PR #12, opened 2023-12-25
Update agent.py

JamesChannel1:
updated docstring


llxxxll:
Is it possible to separate the prompt into a language package? People of different languages can use prompts in their respective languages, not just English prompts?

joaomdmoura (approved):
Thanks! This is great!

joaomdmoura:
@llxxxll good call, maybe an entirely new package and have it as a dependency?

PR #30, opened 2023-12-31
Fix Pydantic/Langchain Inheritance Bug

greysonlalonde:
Issues mentioned [here](https://github.com/langchain-ai/langchain/issues/10112) started to sporadically pop up despite not showing during tests; removing inheritance fixed this. Waiting on langchain updates to inherit CrewAIBaseModel & instead added relevant fields to individual models.

### DIFF
--- a/crewai/agent.py
+++ b/crewai/agent.py
@@ -1,3 +1,4 @@
+import uuid
 from typing import Any, List, Optional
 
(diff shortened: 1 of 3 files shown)


joaomdmoura:
Nice, good catch!

PR #42, opened 2024-01-02
Update output_parser.py

SuperMalinge:
Hi there is just a small typo in the output_parser.py

changing the "So I already knwo the result of that" to "So I already know the result of that"


joaomdmoura:
Thanks @SuperMalinge really appreciate it!

PR #72, opened 2024-01-06
Notebook showing how to use Ollama

karnagge:
Notebook showing how to use Ollama

### DIFF
--- a/notebooks/crewai_with_ollama.ipynb
+++ b/notebooks/crewai_with_ollama.ipynb
@@ -1,0 +1,257 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


prusnak:
I think it is more benefitial to add raw Python example (`.py` file) with comments and not a notebook.

Anyone can copy-paste the code from a Python file to a Notebook, but not the other way around.

joaomdmoura:
Notebook is cool, but I do think this would better as a documentation, if you provide me with a markdown version it would be great to add it to the wiki, also working on more formal docs so would be great to it to there as well

PR #90, opened 2024-01-08
improved readability

paipeline:


joaomdmoura (approved):
Thanks @paipeline def improved readability!

PR #97, opened 2024-01-09
update example usage in README

shadow578:
I noticed that the researcher's goal in the Getting Started example had an extra word at the end, and for some reason that really bothered me.


joaomdmoura (approved):
Thank you so much @shadow578 <3

PR #98, opened 2024-01-10
Add missing pydantic imports

greysonlalonde:
in the title!


joaomdmoura (approved):
Oh I thought I had just added them a couple minutes ago! good catch!

PR #116, opened 2024-01-12
Add Github action for black

greysonlalonde:
What it says on the tin!


joaomdmoura:
<3 Dude you're the best making sure we keep standard high thanks!

PR #126, opened 2024-01-13
Case insensitive tool and agent names

jkellerer:
Thanks for the great toolset. 

Had tried it with a locally running _Mistral 7b_ and found that this LLM, while being very usable on commodity HW, doesn't always get case and leading/trailing spaces right. With this change the trip planner example started to work.

This PR adds `strip().casefold()` for agent names and `casefold()` for tool names (as `strip()` was existing already)

### DIFF
--- a/src/crewai/agents/executor.py
+++ b/src/crewai/agents/executor.py
@@ -43,10 +43,10 @@ class CrewAgentExecutor(AgentExecutor):
     ) -> Dict[str, Any]:
         """Run text through and get agent response."""
         # Construct a mapping of tool name to tool for easy lookup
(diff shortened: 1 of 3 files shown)


Toon-nooT (approved):
Looks good

joaomdmoura (dismissed):
<3 Thanks super appreciate

gvieira:
Do you mind fixing conflicts, so we can merge it?

joaomdmoura:
I tried resolving the conflicts but decided to pull these change into the new version myself as it felt simpler, it will be live int he next version so clsogin this one

PR #129, opened 2024-01-13
NL language support - not tested yet

loukdeloijer:
Added nl.json for Dutch language support. Not tested yet as not sure which pytests should be run for language support check.

Which pytests should pass for this language support to be integrated in main?


joaomdmoura:
We are moving alway from how translations exist today to an attribute on crew anyone can pass, this way we don't need to version and maintain all the translations

PR #131, opened 2024-01-13
DE language support - not tested yet

jkellerer:
Adds DE language support. Not tested yet, try to do it soon.

Translated with GPT-3.5 and proof read / compared manually. GPT-3.5 almost returned the perfect result, only a single word needed adjusting.


joaomdmoura:
We are moving alway from how translations exist today to an attribute on crew anyone can pass, this way we don't need to version and maintain all the translations

PR #164, opened 2024-01-20
Instructions for running docker image

JimJim12:
Added instructions for running the docker image in docs/docker-image.md and linked to it in README.md


albertosouza:
Hello! @JimJim12 where is the code that you used to build this docker image?
May be good to use as example for new images for specific uses.

Without that source code I don't think we should merge the PR.

JimJim12:
> Hello! @JimJim12 where is the code that you used to build this docker image? May be good to use as example for new images for specific uses.
> 
> Without that source code I don't think we should merge the PR.

Forgot about this PR. I'm going to close it out anyway because I haven't had time to update the image so it's outdated.

PR #165, opened 2024-01-21
bug fix and testcase for i18n

cove9988:
1. created i18n.json combining el and en json into language type. this will help feature add any other new language.  
2. modified i18n.py to get language set first. fixed tools retrieve bug in the existing script.
3. added unit test case


joaomdmoura (changes requested):
Not a huge fan of combining the translations as it could get pretty big, the test and code fixes are great tho. If you don't mind keeping the translation separate I'd love to merge it :)

cove9988:
1. centralise it for easy to maintenance is a good common practise. otherwise, when we add/modify a section or item, we have to go through many files. multi language files solution will quickly lead to unsync and errors in different language environment. 

2. in one json, we can take the advantages of co-pilot or any AI to easy fill in other language based on initial like this. 
    "task_with_context": {
      "el": "{task}\nΑυτό είναι το πλαίσιο με το οποίο εργάζεστε:\n{context}",
      "en": "{task}\nThis is the context you're working with:\n{context}",
      "cn": "{task}\n这是您正在使用的上下文：\n{context}",
      "fr": "{task}\nVoici le contexte avec lequel vous travaillez:\n{context}",
      "jp": "{task}\nこれがあなたが働いているコンテキストです：\n{context}",
      "kr": "{task}\n이것은 당신이 일하고있는 문맥입니다 :\n{context}",
      "ru": "{task}\nЭто контекст, с которым вы работаете:\n{context}",
      "de": "{task}\nDies ist der Kontext, mit dem Sie arbeiten:\n{context}",
    }

3. if your concern is loading i18n on every script will be too costly because of the file size. we can work out a singleton class to handle it at the app initial stage.

cove9988:
there are a few PR related on language setting. If we use one json file for multiply languages. this will be not a issue anymore. :)

OlofHarrysson:
@cove9988 If we combine all the languages into one file, wouldn't that potentially create really messy merge conflicts?

cove9988:
> @cove9988 If we combine all the languages into one file, wouldn't that potentially create really messy merge conflicts?

No, there is no messy merge, it's clean cut with same I18N() interface. However, I noticed that we don't have .env (env.template) or other configure file for the app environment variables loading, such as OPENAI_API_KEY, DEFAULT_LANGUAGE etc.... So there is no entry point yet to change language. If you are happy, I can help on this matter as well. :)

greysonlalonde:
I'm with @joaomdmoura, translations could get bulky fast

cove9988:
okay, still keep multi-language files with most people agreed.
1. use DEFAULT_LANGUAGE environment var to change language setting.
2. add test cases.

joaomdmoura:
We are moving always from having translations within the project as it makes it hard to maintain, will add it as a attribute to the crew so people can use any language they wish

joaomdmoura:
I'm bringing the tests in tho :D

PR #170, opened 2024-01-21
 .editorconfig for Consistent Python Coding Styles & Fr language for agent

tal7aouy:
This PR has two contributions:
- Add `.editorconfig`
- Trans `fr.json` to use agent in french

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# .editorconfig
+root = true
+
+# All files
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Python files
+[*.py]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
+# .editorconfig
+root = true
+
+# All files
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
```
gvieira, line 8: ❤️

greysonlalonde:
@tal7aouy  pre-commit hooks address properties in `.editorconfig` automatically, is there a reason to merge with this in mind? It would be cleaner to split this into separate PRs.

joaomdmoura:
I'm fine with the editor config, even tho it might be duplicative when thinking about commits, but we are moving alway from having the translations within the project, it will be an attribute on the crew people will be able to pass in new versions

Issue #142, opened 2024-01-15
Optional Agents in Tasks (mismatch documentation vs code)


michaelwdombek:
Hello  @ all,

First and foremost, I want to express my appreciation for this fantastic project! :)

I recently encountered an issue that was previously discussed in #71 , and I've spent some time examining both the code and the wiki. If I understand correctly, the tasks should function properly even when no agent is selected for a particular task. You can find more information about this in the [Wiki Reference to tasks](https://github.com/joaomdmoura/crewAI/wiki/Defining-Tasks#properties-of-a-task). However, based on my investigation into the code, it appears that this feature may not be currently implemented. Is there someone actively working on this feature, or has it been dropped?

I appreciate your clarification on this matter.

joaomdmoura:
Great catch, indeed, right now the sequential process expects the tasks to be assigned, in other processes that wont be necessary, I've updated the wiki as part of that for now

Issue #71, opened 2024-01-06
AttributeError: 'NoneType' object has no attribute 'allow_delegation'


asmit203:
```
import os
from crewai import Agent, Task, Crew, Process

# You can choose to use a local model through Ollama for example.
# In this case we will use OpenHermes 2.5 as an example.
#
from langchain.llms import Ollama
ollama_llm = Ollama(model="dolphin-mistral")

# If you are using an ollama like above you don't need to set OPENAI_API_KEY.
#os.environ["OPENAI_API_KEY"] = "Your Key"

# Define your tools, custom or not.
# Install duckduckgo-search for this example:
#
# !pip install -U duckduckgo-search

from langchain.tools import DuckDuckGoSearchRun
search_tool = DuckDuckGoSearchRun()

# Define your agents with roles and goals
researcher = Agent(
  role='Senior Research Analyst',
  goal='Uncover cutting-edge developments in AI and data science in',
  backstory="""You are a Senior Research Analyst at a leading tech think tank.
  Your expertise lies in identifying emerging trends and technologies in AI and
  data science. You have a knack for dissecting complex data and presenting
  actionable insights.""",
  verbose=True,
  allow_delegation=False,
  tools=[search_tool],
  # (optional)
  llm=ollama_llm,
  #If you wanna use a local modal through Ollama, default is GPT4 with temperature=0.7

)
writer = Agent(
  role='Tech Content Strategist',
  goal='Craft compelling content on tech advancements',
  backstory="""You are a renowned Tech Content Strategist, known for your insightful
  and engaging articles on technology and innovation. With a deep understanding of
  the tech industry, you transform complex concepts into compelling narratives.""",
  verbose=True,
  # (optional)
  llm=ollama_llm,
  #If you wanna use a local modal through Ollama, default is GPT4 with temperature=0.7
  allow_delegation=True
)

# Create tasks for your agents
task1 = Task(
  description="""Conduct a comprehensive analysis of the latest advancements in AI in 2024.
  Identify key trends, breakthrough technologies, and potential industry impacts.
  Compile your findings in a detailed report. Your final answer MUST be a full analysis report""",
  aget=researcher
)

task2 = Task(
        description="""Using the insights from the researcher's report, develop an engaging blog
  post that highlights the most significant AI advancements.
  Your post should be informative yet accessible, catering to a tech-savvy audience.
  Aim for a narrative that captures the essence of these breakthroughs and their
[...]

PR #172, opened 2024-01-22
Adding the feature to auto select matching agents for a tasks

michaelwdombek:
Following up on my own question #142 and the issue #71 I've added the feature to have the LLM check for a matching agent, to this task.

Please have a close look, since i'm currently not really familiar with the cassettes testing stuff as well as pydantic and  poetry, hence there is just one basic test


michaelwdombek:
@joaomdmoura, would be nice if you could have look - cheers

PR #211, opened 2024-02-03
Fix undefined agent

punkt2:


joaomdmoura:
Oh good catch I just saw this PR but the issue got resolved before we launched the 0.5.0 version fortunately <3
I really appreciate the work tho 🙏🏼

PR #235, opened 2024-02-08
Update Processes.md

punitchauhan771:
Fixed syntax for hierarchical process in the document.


joaomdmoura:
Good catch, it already got fixed in another commit but super appreciate the PR <3

PR #237, opened 2024-02-08
Fix types

gvieira:
A few more changes to make sure `pyright` is happy.

### DIFF
--- a/src/crewai/agents/executor.py
+++ b/src/crewai/agents/executor.py
@@ -106,15 +106,16 @@ class CrewAgentExecutor(AgentExecutor):
                 **inputs,
             )
             if self._should_force_answer():
-                if isinstance(output, AgentAction) or isinstance(output, AgentFinish):
-                    output = output
-                elif isinstance(output, CacheHit):
+                if isinstance(output, CacheHit):
                     output = output.action
-                else:
-                    raise ValueError(
-                        f"Unexpected output type from agent: {type(output)}"
-                    )
-                yield self._force_answer(output)
+                if isinstance(output, AgentAction):
+                    yield self._force_answer(output)
+                    return
+                if isinstance(output, list):
+                    yield from [self._force_answer(action) for action in output]
+                    return
+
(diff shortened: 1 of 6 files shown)


src/crewai/agents/executor.py
```diff
@@ -106,15 +106,16 @@ def _iter_next_step(
                 **inputs,
             )
             if self._should_force_answer():
```
gvieira, line 108: @joaomdmoura please review specifically this.

joaomdmoura: I thinks it's missing the `isinstance(output, AgentAction) or isinstance(output, AgentFinish):`

src/crewai/crew.py
```diff
@@ -208,14 +209,17 @@ def _run_hierarchical_process(self) -> str:
...
         for task in self.tasks:
             self._logger.log("debug", f"Working Agent: {manager.role}")
             self._logger.log("info", f"Starting Task: {task.description}")
 
-            task_output = task.execute(
+            output = task.execute(
                 agent=manager, context=task_output, tools=manager.tools
             )
+            if not task.async_execution:
+                assert output is not None
```
joaomdmoura, line 221: Output could be none I guess?

src/crewai/agents/executor.py
joaomdmoura: does this covers `isinstance(output, AgentFinish)`?

joaomdmoura (approved):

src/crewai/agents/executor.py
gvieira: Don't think it's needed. It will fallback to `yield output` and `return`. We only need to force answer if the type is either `AgentAction` or a list of `AgentAction`. Does that make sense?

src/crewai/crew.py
gvieira: For non-async? I don't think so. Worst case scenario, it's gonna be an empty string.

PR #240, opened 2024-02-09
Fix example code in Update Tasks.md

sebestyenmiklos1:
Fix example code of missing comma.


joaomdmoura:
Nice catch, super appreciate <3

PR #246, opened 2024-02-11
Ability to integrate external agents into the framework

ZmeiGorynych:
As discussed in https://github.com/joaomdmoura/crewAI/issues/228
Example usage in https://github.com/joaomdmoura/crewAI-examples/pull/52/files


sameermahajan:
looking forward to merge of this PR into main for our use cases

ZmeiGorynych:
If you want to mix and match CrewAI agents with any other ones (Langchain, LllamaIndex, and Autogen so far, happy to add more), take a look at [motleycrew](https://motleycrew.ai) :)
I started it after the above PR was put on ice, and it's all about nice orchestration, while allowing you to use any agents and tools of your choice, conveniently.

src/crewai/agent.py
```diff
@@ -131,7 +131,7 @@ def check_agent_executor(self) -> "Agent":
     def execute_task(
         self,
         task: Any,
-        context: Optional[str] = None,
+        context: Optional[List[str]] = None,
```
gvieira, line 134: Isn't this breaking anything in terms of retro-compatibility? Not an issue, per se, but I wonder if there's any way around if that's the case.

src/crewai/agents/agent_interface.py
```diff
@@ -0,0 +1,52 @@
...
+        description="Data storage for children, as pydantic doesn't play well with inheritance.",
+    )
+    role: str = Field(description="Role of the agent", default="")
+    allow_delegation: bool = Field(
+        description="Allow delegation of tasks to other agents?", default=False
+    )
+
+    @property
+    def i18n(self) -> I18N:
+        if hasattr(self, "_agent") and hasattr(self._agent, "i18n"):
```
gvieira, line 22: Could we prevent this kind of indirection by using some sort of interface? We are moving fast to become a strongly-typed library and this makes it very hard to make sure things are working properly.

src/crewai/agents/agent_interface.py
```diff
@@ -0,0 +1,52 @@
+from abc import ABC, abstractmethod
+from typing import Optional, List, Any, Dict
+
+from pydantic import BaseModel, PrivateAttr, Field
+
+from crewai.utilities import I18N
+
+
+class AgentWrapperParent(ABC, BaseModel):
```
gvieira, line 9: Really loved the generic agent interface. In a perfect world, this would be the only way the engine would refer to agents, right!? Should we call it `AgentInterface`, like the file name? #namingishard

gvieira: Just to make it clear, the dream here is that even `Agent` respects `AgentInterface`, then anything that complies to the interface would be able to be an agent.

ZmeiGorynych: Good suggestion,  that was exactly the idea :)

src/crewai/agent.py
ZmeiGorynych: I just thought that the natural kind of context is a list of past messages in a conversation, and it's more natural to pass that through as a list instead of squashing it into a str

src/crewai/agents/agent_interface.py
ZmeiGorynych: If an offician AgentInterface existed, this could be part of it :)

src/crewai/agent.py
gvieira: I love it. It makes sense! My concern is the breaking change. What if we moved it to an `str | List[str] | None` type?

src/crewai/agents/agent_interface.py
gvieira: Wanna draft something here?

ogabrielluiz:
Looking forward to this as well. We'd use this to connect Agent components to CrewAI components in Langflow.

lorenzejay:
https://github.com/crewAIInc/crewAI/pull/776 this PR address this letting you bring your agents by extending the BaseAgent class

PR #257, opened 2024-02-18
Recommended version message

0xrushi:
I had difficulty choosing between Python versions 3.11, 3.8, and 3.9, but I eventually identified a version that works properly. I suggest we include this information in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ CrewAI is designed to enable AI agents to assume roles, share goals, and operate
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


joaomdmoura:
OH rly? as long as it is >=3.10,<=3.13 it should work, can you share more about the errors you found?

0xrushi:
Hi, I just tested, its working now!, maybe some other commit fixed it.
Feel free to close this issue.
Thanks

PR #271, opened 2024-02-25
Create pt-br. translation

augustoromeiro:


joaomdmoura:
We are moving away from hosting translations to allow them to be set on th crew level as an attribute, this way people can pass their own translations and prompts and make it easier to maintain

PR #279, opened 2024-02-27
Update Creating-a-Crew-and-kick-it-off.md to add crewai-tools as dependency

barapa:
Without installing crewai-tools, the getting started example does not work, throwing the following error

```
ModuleNotFoundError: No module named 'crewai_tools'
```

This updates the docs to include installing crewai-tools during the installation step.


barapa:
Closing this in favor of a different approach

PR #280, opened 2024-02-27
Fix Creating-a-Crew-and-kick-it-off.md so it can run

barapa:
- Update deps to include `crewai[tools]`
- Remove invalid `max_inter` arg from Task constructor call


joaomdmoura:
This is great, thanks!

PR #281, opened 2024-02-27
Fix typoes in en.json

gsteinLTU:
- exaclty -> exactly
 - "jobs depends" -> "job depends"
 - bellow -> below


joaomdmoura (approved):
Love it, really appreciate it

PR #311, opened 2024-03-04
Update Create-Custom-Tools.md

selimhanerhan:
Added the langchain "Tool" functionality by creating a python function and then adding the functionality of that function to the tool by 'func' variable in the 'Tool' function.


joaomdmoura (approved):
Good one! love it, merging it!

PR #339, opened 2024-03-12
Allow minor version patches to `python-dotenv`

Elijas:
### Problem
Currently it's 
`==1.0.0`
which leads to dependency issues, wherever crewai is used with a project with the latest python-dotenv (e.g. `^1.0.1`)

### Solution
let's make it at least `^1.0.0`, or ideally, even `<2.0.0`


joaomdmoura:
Good catch!

PR #349, opened 2024-03-15
Fix Typo in README: Correct 'mode' to 'model'

crownemmanuel:
This PR corrects a minor typo found in the main README file, specifically in the section discussing LLM selection options for an agent. The original text mistakenly used 'mode' when it should have referred to 'model'. This update ensures that the documentation accurately reflects the intended instructions for changing the LLM model, enhancing clarity for all users.


joaomdmoura (approved):
<3 thanks

PR #350, opened 2024-03-15
Update README.md

erickthered:
Added a small paragraph to let users know that `crewai-tools` needs to be installed before running the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ pip install 'crewai[tools]'
 
 ### 2. Setting Up Your Crew
 
+In order to run the following example you need to have installed the `crewai-tools` package.
+
 ```python
 import os
 from crewai import Agent, Task, Crew, Process


README.md
```diff
@@ -57,6 +57,8 @@ pip install 'crewai[tools]'
 
 ### 2. Setting Up Your Crew
 
+In order to run the following example you need to have installed the `crewai-tools` package.
```
ruebenramirez, line 60: Isn't this covered here?  https://github.com/joaomdmoura/crewAI/pull/350/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R55

joaomdmoura: indeed

joaomdmoura:
Appreciate the PR but the step before already gives a heads up on installing crewai-tools

PR #359, opened 2024-03-19
Update LLM-Connections.md

selimhanerhan:
Created a short documentation on how to integrate crewAI and Llama2 locally with the help of Ollama. These steps might be self explanatory for the advanced programmers but will be definitely needed for the beginners like myself, so I decided to add it into the documentations to have a free alternative to the OpenAI.


docs/how-to/LLM-Connections.md
```diff
@@ -50,6 +50,71 @@ OPENAI_MODEL_NAME='openhermes'  # Adjust based on available model
...
+            tasks=[task],
+            verbose=2
+        )
+
+result = crew.kickoff()
+
+print(result)
+```
+
+
```
shivamjagdishsingh: Check out my stackoverflow answer to run pre-commit on last commit.
https://stackoverflow.com/a/78196222/12821733

selimhanerhan: Thank you! I was wondering why it was failing lint and all other checks.

selimhanerhan (commented):
Now detailed Ollama setup is in the documentation so Note is not needed

joaomdmoura (approved):
Thanks! Great imrpovements

PR #369, opened 2024-03-22
Fix input interpolation bug

gvieira:
Whenever running `kickoff` with `inputs`, we were losing the original unformatted string.

```python
result = crew.kickoff(inputs={"topic": "Artificial Intelligence"})

result = crew.kickoff(inputs={"topic": "Impressionist Art"})

# The second run will still use Artificial Intelligence
```

This is fixed now! Not the best solution, I know. 🤷


joaomdmoura (approved):
great catch

PR #406, opened 2024-04-02
tasks.py: don't call Converter when model response is valid

ftoppi:
Hello,

Simple PR to try to convert the task output to the expected Pydantic model before sending it to Converter, maybe the model got it right.

This can save a lot of time and money :)


ruebenramirez (approved):
Nice optimization!

ftoppi:
Hello, I missed the fact `output_json` was taking a pydantic object as well. I'll update the PR for that.

PR #407, opened 2024-04-02
whitespace consistency across docs

ruebenramirez:
I saw a rendedered whitespace inconsistency in the Tasks docs here: https://github.com/joaomdmoura/crewAI/blob/ed318600719f07721240f76f5024c6d45f837045/docs/core-concepts/Tasks.md?plain=1#L173

So I set out to patch that up to make it easier to read.  I then noticed there were a few whitespace inconsistencies:
- 2 spaces
- 4 whitespaces
- tabs

It appears that the 4 whitespaces is the prevalent whitesapce usage, so I overwrote other whitespace usages with that in this commit.


docs/core-concepts/Tasks.md
```diff
@@ -160,20 +160,20 @@ The callback function is executed after the task is completed, allowing for acti
...
+    # Example: Send an email to the manager
+    print(f"""
+        Task completed!
+        Task: {output.description}
+        Output: {output.raw_output}
+    """)
 
 research_task = Task(
-  description='Find and summarize the latest AI news',
-	expected_output='A bullet list summary of the top 5 most important AI news',
```
ruebenramirez, line 175: This was the main item I wanted to help patch up :)

joaomdmoura (approved):
<3 really appreaciate it

Issue #408, opened 2024-04-02
Agent code overwrites the LLM callbacks


rakeshgohel01:
### Issue
When I supplied callbacks directly to LLM object before passing to Agent constructor, the callbacks set by me were never called

### Explanation
If an LLM is created with callbacks and assigned to agent, it gets overwritten in agent initialization when pydantic finishes validation.

### code
    @model_validator(mode="after")
    def set_agent_executor(self) -> "Agent":
        """set agent executor is set."""
        if hasattr(self.llm, "model_name"):
            self.llm.callbacks = [ **//THIS OVERRIDES THE LLM CALLBACKS**
                TokenCalcHandler(self.llm.model_name, self._token_process)
            ]
        if not self.agent_executor:
            self.set_cache_handler(self.cache_handler)
        return self

Ideally it should append TokenCalcHandler to callbacks not assigned

PR #409, opened 2024-04-02
Stop overwriting LLM callback #408

mocha384:
This changes allow previously set callbacks in LLM to work along with with TokenCalcHandler


joaomdmoura:
Good catch sorry I didn't see sooner ended up getting fixed in another commit, but really appreciate it

PR #411, opened 2024-04-02
AgentOps Implementation

bboynton97:
Integrates AgentOps.

Adds support for tracking LLM calls, Agents, and Tool usage


bboynton97:
Don't merge this as is. We need to handle the case where agentops isn't installed more gracefully. Working on this now.

src/crewai/crew.py
```diff
@@ -235,6 +238,9 @@ def kickoff(self, inputs: Optional[Dict[str, Any]] = {}) -> str:
 
         i18n = I18N(language=self.language, language_file=self.language_file)
 
+        if agentops:
```
braelyn-ai: @joaomdmoura: this addition means that if the developer is using agentops, we will set the parent key. This way, Crew will have access to some run data in this case.

we can also change the conditional to `if agentops and self.share_crew` if you'd rather

ajaywingify:
@joaomdmoura possible to merge this branch soon? We really need AgentOps to monitor the performance of a support bot we're building using Crew.

pyproject.toml
```diff
@@ -28,6 +28,7 @@ click = "^8.1.7"
 python-dotenv = "1.0.0"
 embedchain = "^0.1.98"
 appdirs = "^1.4.4"
+agentops = "^0.1.1"
```
joaomdmoura: Let's make it optional, same way we do to crewai-tools, so we don't add the extra dependency for everyone, feel unnecessary if people are not using agent ops

pyproject.toml
```diff
@@ -28,6 +28,7 @@ click = "^8.1.7"
 python-dotenv = "1.0.0"
 embedchain = "^0.1.98"
 appdirs = "^1.4.4"
+agentops = { version = "^0.1.1", optional = true }
```
joaomdmoura: Updating it to be an option dependency, so we might need to add some fallback when importing the agent ops

joaomdmoura (changes requested):

joaomdmoura:
temporarily merging this for the weekend

PR #420, opened 2024-04-03
Update Creating-a-Crew-and-kick-it-off.md: add compatible python versions

ftoppi:
Add compatible python versions will hopefully prevent help requests.


joaomdmoura:
Good catch, updated tot he right versions :) merging

PR #432, opened 2024-04-05
Added trasnlation to 'pt-BR'

freirefx:
Bora Brasil!!!


joaomdmoura:
We are moving away from support translation internally but now allowing you to bring your own using the `prompt_file` attribute on the crew! so you can have your own translation locally, also makes it easier for us to maintain it! :D So I really appreciate the PR but because of this we won't be merging it, but you can still use your translation with the new 0.30.0 version, it's not out yet, but you can install it with `pip install 'crewai[tools]'==0.30.0rc6`

PR #436, opened 2024-04-06
The suggestions were getting split at character level and not at sentence level

rajib76:
When loading the historical suggestions, it was getting loaded as 
 ["Historical Data:\n['e', 'u', '.', 'T', 'n', 'c', 'y', ',', 'v', 't', 'h', 'S', 'A', 'I', 'b', 'f', 'l', 'g', 'd', '-', ' ', '\n', 's', 'a', 'w', 'x', 'p', 'i', 'r', 'k', 'm', 'o']", '', '']

This was happening because the suggestions were split at character level. This fix will not make it at sentence level


joaomdmoura:
great catch! amazing <3

PR #443, opened 2024-04-08
fix: swapped the task callback assignment

bastulli:
I think dyslexia got someone here. Should resolve issue with individual task callbacks not executing


joaomdmoura (approved):
LOL great catch! I was about to fix this <3

Issue #466, opened 2024-04-12
QUESTION: "Force" complex hierarchy within the crew


emilmirzayev:
Hey there. Newly discovered this project and I am having fun with it. I looked at `process=Process.hierarchical` in `Crew` settings. I have a question about it. 

Consider a case where there are four agents: Agent 1 - 5, and four tasks: Task 1 - 5.

There must be a specific hierarchy in the crew:
```
Agent 1
├── Agent 2
│ ├── Agent 3
│ └── Agent 4
└── Agent 5
```
When I want to build a crew like this, I can write my code in the following way:

```python
class MYagents:
    
    def agent1(self):
      
        return Agent(
            role="Sample role 1",
            goal="Sample goal 1",
            backstory=dedent("""
                Sample backstory for Agent 1.
            """),
            verbose=False,
            llm=sample_llm,
            allow_delegation= True,
        )
    
    def agent2(self):
     
        return Agent(
            role="Sample role 2",
            goal="Sample goal 2",
            backstory=dedent("""
                Sample backstory for Agent 2
            """),
            verbose=False,
            llm=sample_llm,
            allow_delegation= True,
        )
    
    def agent3(self):
        return Agent(
            role="Sample role 3",
            goal="Sample goal 3",
            backstory=dedent("""
                Sample backstory for Agent 3
            """),
            verbose=False,
            llm=sample_llm,
        )

    def agent4(self):
        return Agent(
            role="Sample role 4",
            goal="Sample goal 4",
            backstory=dedent("""
                Sample backstory for Agent 4
            """),
            verbose=False,
            llm=sample_llm,
        )

    def agent5(self):
        return Agent(
            role="Sample role 5",
            goal="Sample goal 5",
            backstory=dedent("""
                Sample backstory for Agent 5
            """),
            verbose=False,
            llm=sample_llm,
        )
# Defining a class for tasks with generic placeholder methods
class MYtasks:
    
    def task1(self, agent, context):
        return Task(
            description=dedent("""
                Sample task 1 description for Agent 1.
            """),
            expected_output=dedent("""
                Sample expected output for Task 1.
            """),
            agent=agent,
            context = context
        )
    
    def task2(self, agent, context):
        return Task(
            description=dedent("""
                Sample task 2 description for Agent 2.
            """),
            expected_output=dedent("""
                Sample expected output for Task 2.
            """),
            agent=agent,
            context = context
        )
    
    def task3(self, agent):
        return Task(
            description=dedent("""
[...]

PR #474, opened 2024-04-16
Added optional manager_agent parameter

noggynoggy:
As talked about in #466 this adds an optional parameter `manager_agent` in for the `Crew` class. It takes in an `Agent` object that will be used instead of the default one, allowing more granular prompt engineering and even things like scoped delegation.

## Things to consider
- Documentation?
- Check for `allow_delegation` / `tools`?
   - currently it allows delegation to be `False` / there not to be any delegation tools, which would break "crew" functionality
- Could be done easier 
  - user has to make a whole agent 
  - one could also add e.g. `manager_agent_role`, `manager_agent_goal` and `manager_agent_backstory`

### DIFF
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -312,6 +312,7 @@ class Crew(BaseModel):
         i18n = I18N(language=self.language, language_file=self.language_file)
         try:
             manager = self.manager_agent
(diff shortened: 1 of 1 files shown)


joaomdmoura:
I like this! def useful, I'll merge and add some tests next :)

PR #475, opened 2024-04-16
Prevent duplicate TokenCalcHandler callbacks on Agent

jackhayter:
It appears that the `TokenCalcHandler` callback is being appended to `Agent.callbacks` multiple times, resulting in an over-estimate of token usage.

Please see comment at: https://github.com/joaomdmoura/crewAI/issues/93#issuecomment-2058258250 for more context.

This PR modifies the behaviour of the `set_agent_executor` method to only allow for a single instance of the handler to be added to an agent. This change will result in more accurate token usage reports.

I have not been able to locate any Github Issues from other users on this topic, so I could be mistaken here. I would greatly appreciate a gut-check from the community on this hypothesis.


joaomdmoura (approved):
Great catch!

PR #477, opened 2024-04-17
Fix lockfile

Christian24:
The first thing I got when checking out the code was that the lock file is outdated. I thought I would fix it quickly ;)


gvieira (approved):
Thanks for taking care of that!

PR #483, opened 2024-04-19
Accept Langchain's RunnableConfig to Agent.execute_task

whimo:
Accept config as an argument and forward it to the invoke() method of the underlying Langchain AgentExecutor.
It is very helpful for customizing the run, e.g. adding callbacks (see [docs](https://api.python.langchain.com/en/latest/runnables/langchain_core.runnables.config.RunnableConfig.html) for more)


joaomdmoura:
I like it, but we would need to update the agent attributes as well to allow the config to be send upon instantiating

gvieira:
Hey @whimo! Any thoughts on @joaomdmoura's suggestion to also update agente attributes?

whimo:
> Hey @whimo! Any thoughts on @joaomdmoura's suggestion to also update agente attributes?

Nothing against it. In fact, we can just update #632 with this and close the 2 PRs altogether.

gvieira:
@whimo I like the idea of closing both PRs and creating another one joining them, as we would have a conflict here anyway. Happy to streamline the review process for this new PR. cc @thehapyone

whimo:
Hi @gvieira @thehapyone , I've updated this PR with changes from #632. Looking forward to a review.

thehapyone:
Looks good to me 👌

gvieira (approved):
Great work!

whimo:
Hi @gvieira, thanks for the approval, can we merge this now?

joaomdmoura:
I would love to understand more what is the underlying need here before we merge, just because we override a lot of the the AgentExecutor and will probably keep overriding more, my fear is that this probably not accepts all the configs today and will probably break in the future.

thehapyone:
> I would love to understand more what is the underlying need here before we merge, just because we override a lot of the the AgentExecutor and will probably keep overriding more, my fear is that this probably not accepts all the configs today and will probably break in the future.

The AgentExceutor is based on Langchain runnable or atleast everything is migrating towards runnable has a native object. The runnable configs provides a lot of possibilities to customise the behaviour of any type of runnable. A very good use case of this is native callback and tracing possibilities without having to add new functionality to the existing agents codebase. So unless, langchain drops supports for runnable or crewAI moves away from langchain can’t see how this is an issue now or later. The agent still has it’s own config parameter which is a totally different thing.

whimo:
Hi @joaomdmoura! Really looking forward to merging this. If you have any questions or concerns, please let us know :)

P.S. This change should never break, because AgentExecutor is a Runnable, and any Runnable accepts a RunnableConfig in its `invoke()` method.
We need this for observability (using Langchain's run_manager, which automatically reports chains' start and end events).

thehapyone:
Hi @joaomdmoura @gvieira can you guys check this out? This is really a nice thing to have

thehapyone:
@whimo I think this PR won't work again and should probably be closed. CrewAI has moved away from using langchain in the agentexecutor

whimo:
I believe you're right. The issue can be closed, but I'd love if CrewAI introduced its own callback mechanism similar to what Langchain and LlamaIndex have, so that we can use observability tools, add logging etc.

PR #488, opened 2024-04-21
fix: TypeError: LongTermMemory.search() missing 1 required positional argument: 'latest_n'

dkhokhlov:
References: 
- https://github.com/joaomdmoura/crewAI/issues/487


joaomdmoura (approved):

joaomdmoura:
Thanks I think we added a default now, but merging just in case :D

PR #491, opened 2024-04-22
Update task.py: try to find json in task output using regex

ftoppi:
Sometimes the model replies with a valid and additional text, let's try to extract and validate it first. It's cheaper than calling LLM for that.

I think we could even replace the code above with only this one, the regex will match both.


joaomdmoura (approved):
Nice, moved a few things around slightly but I think this is a great idea!

PR #521, opened 2024-04-26
Tool name recognition based on string distance

deadlious:
Working with smaller AI models, they tend to get the tool names wrong. Either write Coworker instead of co-worker or add `()` at the end of the tool name. These are not necessary invalid tool calls with invalid params. 
Integrating a string distance into the attempt to find the tool would make it a bit easier to work with smaller AIs.


joaomdmoura:
Super interesting!
I love it! will try it out!

PR #523, opened 2024-04-26
Add conditional check for output file directory creation

Tavernari:
This PR adds a conditional check to ensure that the output file directory exists before attempting to create it. This ensures that the code does not fail in cases where the directory does not exist and needs to be created.

The condition is added in the `_save_file` method of the `Task` class, ensuring that the correct behavior is maintained for saving results to a file.


joaomdmoura:
Love it, great call!

PR #525, opened 2024-04-28
Update README.md

JimCollinsDC:
Reworded "If you want to also install crewai-tools, which is a package with tools that can be used by the agents, but more dependencies, you can install it with, example below uses it:" for clarity


joaomdmoura (approved):
thanks <3

PR #543, opened 2024-04-30
Add back AgentOps as Optional Dependency

bboynton97:
This puts AgentOps back as an optional dependency only

I also ran the black formatter to pass the github action and it made some formatting changes to files I did not touch.


gvieira (commented):
Quick suggestion: what if we had a generic class like `AgentMonitoring` that we could call across the codebase, then this class would have all the logic to optionally call `agentops` if it's included? Does that make sense?

bboynton97:
> Quick suggestion: what if we had a generic class like `AgentMonitoring` that we could call across the codebase, then this class would have all the logic to optionally call `agentops` if it's included? Does that make sense?

We've done implementations like that for a couple other frameworks, but nothing like this exists in crew atm

gvieira:
@bboynton97 got it. Happy to review PRs if you believe this is something you can do.

joaomdmoura (approved):

PR #548, opened 2024-05-01
fix to import statement PGSearchTool.md

alexfazio:
fix to the import statement in PGSearchTool documentation


joaomdmoura (approved):
Great catch

PR #550, opened 2024-05-02
Allow to use async tools within the task pipeline

tg1482:
Introduces a way to use async tools seamlessly within a crew task.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -4122,6 +4122,7 @@ files = [
     {file = "PyYAML-6.0.1-cp311-cp311-win_amd64.whl", hash = "sha256:bf07ee2fef7014951eeb99f56f39c9bb4af143d8aa3c21b1677805985307da34"},
     {file = "PyYAML-6.0.1-cp312-cp312-macosx_10_9_x86_64.whl", hash = "sha256:855fb52b0dc35af121542a76b9a84f8d1cd886ea97c84703eaa6d88e37a2ad28"},
     {file = "PyYAML-6.0.1-cp312-cp312-macosx_11_0_arm64.whl", hash = "sha256:40df9b996c2b73138957fe23a16a4f0ba614f4c0efce1e9406a184b6d07fa3a9"},
+    {file = "PyYAML-6.0.1-cp312-cp312-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:a08c6f0fe150303c1c6b71ebcd7213c2858041a7e01975da3a99aed1e7a378ef"},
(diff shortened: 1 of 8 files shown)


joaomdmoura:
Interesting, would love to hear more about the use case that prompted this given the `async_exec` attribute on task :)

joaomdmoura:
link for reference: https://docs.crewai.com/core-concepts/Tasks/#asynchronous-execution

tg1482:
Wow thanks for your promptness. One of the things we wanted to do was to give the agents the ability to talk to the users to ask clarifying questions directly using the websockets that we provide them. So we made a tool that allows the agent to ask questions and wait for a response from the user. But we don't want to exit the task pipeline since that will take a little longer + more openai api calls. 

This is also helpful for some tools where the agent prepares a draft email for instance, but wants to wait for the user confirmation before proceeding to send it. The tool has to await a response before proceeding - so this async allows for that. Its not just for the task to be async, but rather the Tools to be async.

joaomdmoura:
Oh got it, something more complex then the current [human input on execution](https://docs.crewai.com/how-to/Human-Input-on-Execution/)? basically that but over sockets.
I assume in this case the new task could be processed while the current one awaits the input?

tg1482:
Yes exactly - we can do as many new tasks while waiting for a response on a previous task.

joaomdmoura:
got it! Any chance you would make this socket human input tool public? :D no pressure but I'd love to add it to crewai-tool or find a way to bring them into the main package somehow

joaomdmoura:
This is a big change so I'll test is a little myself before merging just in case but I like the problem it aims to solve

tg1482:
Amazing - appreciate any feedback on it and ideas on how to make the code a little less redundant - since most of it is just about adding an await in the right places. And detecting whether the tool is a coroutine or not and then deciding to use _run or await _run.

tg1482:
And certainly, I'll try and add the socket human input in a following PR but that part is a little more coupled with my current infra setup haha so could be tough. But I agree, would be a great addition to the crewai toolkit

tg1482:
I've added a quick `crew_test` but I'm unable to get it to run locally. But hopefully it'll be quick enough for you to take a look + adding some resiliency! Thanks again! Please let me know if I can help in any way!

davidlandais:
Hello! I just came across this MR. I found out about CrewAI from a video on agentic workflow. I'm also interested in being able to intervene during a workflow. I prefer that the AI asks me questions, probes me so that I myself can be an agent influencing the final outcome. Have you considered using messaging tools like Slack, Discord, or Telegram? I assume your websocket tool is homemade? Would a workflow that uses a database to save its state to resume its process later be an interesting topic to explore for CrewAI?

thehapyone:
I think this native support for async will be really nice to have considering most of the LLM already supports this. I was hoping to integrate CrewAI has the Agent framework for Sage https://github.com/thehapyone/Sage
but the lack of async support is a bit blocking. 

Alternatively, asyncify is one hack package that can be used to create a sync version of this

tg1482:
Creating a new PR with a lot fewer changes to the code base.

PR #557, opened 2024-05-03
changes static example topic to dynamic example topic

tarekadam:
"latest AI trends" should be "latest {topic} trends".


tarekadam:
Strange, I'm getting lint and whitespace problems on things I didn't touch.

PR #585, opened 2024-05-09
fix: Add validation fix output_file issue when have '/'

pythonbyte:
Issue found from the Discord thread
https://discord.com/channels/1192246288507474000/1235083142520635392/threads/1237431294745378847

The solution is to validate the output_file before saving it, and removing the initial / sign so it can account for both local files and for directory creation.

### DIFF
--- a/src/crewai/task.py
+++ b/src/crewai/task.py
@@ -1,8 +1,8 @@
+import os
 import re
 import threading
 import uuid
 from typing import Any, Dict, List, Optional, Type
-import os
 
 from langchain_openai import ChatOpenAI
 from pydantic import UUID4, BaseModel, Field, field_validator, model_validator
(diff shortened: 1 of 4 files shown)


src/crewai/task.py
```diff
@@ -247,16 +255,16 @@ def _export_output(self, result: str) -> Any:
...
-              match = re.search(r"({.*})", result, re.DOTALL)
-              if match:
-                  try:
-                      exported_result = model.model_validate_json(match.group(0))
-                      if self.output_json:
-                          return exported_result.model_dump()
-                      return exported_result
-                  except Exception:
-                      pass
+                # sometimes the response contains valid JSON in the middle of text
```
pythonbyte, line 258: This is my formatter adding 2 more spaces to properly ident this past code

PR #591, opened 2024-05-10
feat: add mypy as type checker, update code and add comment to reference

pythonbyte:
This PR has the intention to add mypy as the type-checker and also update the GH actions to reflect that. 

Along with that, some fixes were done on the typing, and comments were added to type:ignore along with the warning that we need to fix/improve in the future.

### DIFF
--- a/.github/workflows/type-checker.yml
+++ b/.github/workflows/type-checker.yml
@@ -23,4 +23,4 @@ jobs:
           pip install mypy
 
       - name: Run type checks
-        run: poetry run mypy src --install-types --non-interactive --ignore-missing-imports --exclude "cli/templates"
+        run: mypy src --install-types --non-interactive --ignore-missing-imports --exclude "cli/templates"
--- a/.github/workflows/type-checker.yml
+++ b/.github/workflows/type-checker.yml
@@ -20,7 +20,11 @@ jobs:
 
       - name: Install Requirements
         run: |
-          pip install mypy
+          sudo apt-get update &&
+          pip install poetry &&
+          poetry add mypy
+          poetry lock &&
+          poetry install &&
 
       - name: Run type checks
-        run: mypy src --install-types --non-interactive --ignore-missing-imports --exclude "cli/templates"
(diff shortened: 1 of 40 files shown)


.github/workflows/type-checker.yml
```diff
@@ -12,19 +11,16 @@ jobs:
...
         run: |
-          sudo apt-get update &&
-          pip install poetry &&
-          poetry lock &&
-          poetry install
+          pip install mypy
 
       - name: Run type checks
-        run: poetry run pyright
+        run: mypy src --install-types --non-interactive --ignore-missing-imports --exclude "cli/templates"
```
gvieira: Any way we can setup such flags in a conf file somewhere, like pyproject.toml?

src/crewai/memory/memory.py
```diff
@@ -12,12 +12,16 @@ def __init__(self, storage: Storage):
         self.storage = storage
 
     def save(
-        self, value: Any, metadata: Dict[str, Any] = None, agent: str = None
+        self,
+        value: Any,
+        metadata: Optional[Dict[str, Any]] = None,
+        agent: Optional[str] = None,
```
gvieira, line 18: What is your take on `Optional[str]` vs `str | None`? I've been using the latter these days.

.github/workflows/type-checker.yml
pythonbyte: Yes of course, this was just the first approach to see if it worked

src/crewai/memory/memory.py
pythonbyte: To be honest not a strong opinion, I kinda like the direct "Optional" making it clear that is a Str/None. 
But I saw that recently was changed to <type> | None. 

I can change to the <type> | None so we keep a standard throughout the project

gvieira: Yeah. If `| None` is the updated way to do it, let's try to use it. But there's no reason to start switching everywhere. We can adjust as we move forward.

PR #596, opened 2024-05-10
feat: adds support to automatically fallback to the default encoding

thehapyone:
This PR address an issue with the TokenCalcHandler attempting to classify all "gpt" title models automatically belongs to OpenAI models. This causes an issue if I have a custom model that I call my "gpt-model2" will will result in a key error like below:

```shell
langchain_core.callbacks.manager - WARNING - Error in TokenCalcHandler.on_llm_start callback: KeyError('Could not automatically map azure/gpt4-128k to a tokeniser. Please use `tiktoken.get_encoding` to explicitly get the tokeniser you expect.')
```

Here we attempt to use the gpt model by default and then fall-back to the default encoding one if there was a keyerror.


joaomdmoura (approved):
great approach, way better <3

PR #632, opened 2024-05-17
feat(agent): adds support for customizing the agent via a langchain runnable config

thehapyone:
This PR attempts to adds an additional attribute for customizing the agents setup via the langchain runnable config. This could be used to to enhance the tracing for example. 

Here is a potential example usage of this to set the names used in tracing logs
image


thehapyone:
cc @joaomdmoura  what do you think? Also it seem the new actions workflow didn't address the underlying errors before being merged https://github.com/joaomdmoura/crewAI/pull/588

gvieira:
Seems related to #483, isn't it?

thehapyone:
This is implemented in #483

PR #645, opened 2024-05-20
Fix typo's within converter.py

theCyberTech:
Fiexed Type errors

attemps > attempts

### DIFF
--- a/src/crewai/utilities/converter.py
+++ b/src/crewai/utilities/converter.py
@@ -22,7 +22,7 @@ class Converter(BaseModel):
     llm: Any = Field(description="The language model to be used to convert the text.")
     model: Any = Field(description="The model to be used to convert the text.")
     instructions: str = Field(description="Conversion instructions to the LLM.")
-    max_attemps: Optional[int] = Field(
+    max_attempts: Optional[int] = Field(
         description="Max number of attemps to try to get the output formated.",
         default=3,
     )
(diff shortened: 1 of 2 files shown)


src/crewai/utilities/converter.py
```diff
@@ -64,7 +64,7 @@ def _create_instructor(self):
 
         inst = Instructor(
             llm=self.llm,
-            max_attemps=self.max_attemps,
+            max_attemps=self.max_attempts,
```
pythonbyte, line 67: I think we should check the "parameter" of the Instructor class max_attemps

theCyberTech: could not find reference to max_attemps anywhere else apart from tools_usage.py which has now been updated an commited to this PR

src/crewai/utilities/converter.py
```diff
@@ -22,7 +22,7 @@ class Converter(BaseModel):
     llm: Any = Field(description="The language model to be used to convert the text.")
     model: Any = Field(description="The model to be used to convert the text.")
     instructions: str = Field(description="Conversion instructions to the LLM.")
-    max_attemps: Optional[int] = Field(
+    max_attempts: Optional[int] = Field(
         description="Max number of attemps to try to get the output formated.",
```
gvieira, line 26: Same typo here, right!?

gvieira: Also `formated`.

PR #652, opened 2024-05-20
Update Getting Started README.md example so that you can pass an llm parameter and not break code if lines are uncommented.

heaversm:
Rework example so that if you use a custom LLM it doesn't throw code errors by uncommenting.


joaomdmoura (approved):

joaomdmoura:
Good catch, appreciated

PR #655, opened 2024-05-21
Update on default llm model used in agent.py

Anudeep-Kolluri:
Changed default model value from gpt-4 to gpt-4o.
Reasoning.
gpt-4 costs 30$ per million tokens while gpt-4o costs 5$. This is more cost friendly for default option.


joaomdmoura:
Good call, i wanted to test it first but it seems to be working great!

PR #662, opened 2024-05-21
Clarify text in docstring

sanders41:
In reading this docstring I couldn't make sense of it because it looked like it was missing a word. I looked it up in the docs online and matched the wording here to that.


joaomdmoura (approved):
thanks, really appreciated <3

PR #663, opened 2024-05-21
Enable search in docs

sanders41:
I was looking in the docs for something but didn't know where to find it. This PR adds a search to the docs to help with situations like that.


pythonbyte (approved):
LGTM. Nice work.

PR #686, opened 2024-05-24
feat: Add Train feature for Crews

pythonbyte:
The objective of this PR is to add training functionality for Crews. 

With this functionality it will be possible to run the command `crewai train -n 2` and 2 training iterations will be run so that future tasks can have greater quality and consistency.


atb29:
> The objective of this PR is to add training functionality for Crews.
> 
> With this functionality it will be possible to run the command `crewai train -n 2` and 2 training iterations will be run so that future tasks can have greater quality and consistency.

Isn't that achievable by the memory feature instead??

joaomdmoura:
@atb29 this is memory on steroids, it uses some of the memory work but allows you to train the crew before shipping, something we are testing internally, super excited about it, running some benchmarks now

src/crewai/agents/executor.py
```diff
@@ -305,3 +314,32 @@ def _ask_human_input(self, final_answer: dict) -> str:
         return input(
             self._i18n.slice("getting_input").format(final_answer=final_answer)
         )
+
+    def _training_handler(
+        self, output: AgentFinish, human_feedback: str | None = None
+    ) -> None:
+        """Function to handle the process of the training data."""
+        training_data_filename = "training_data.pkl"
```
gvieira: Should we have these `pkl` file names in consts, maybe? This way we have a single place to find them.

src/crewai/agents/executor.py
```diff
@@ -305,3 +314,32 @@ def _ask_human_input(self, final_answer: dict) -> str:
...
+        agent_id = str(self.crew_agent.id)
+
+        if (
+            not self.should_ask_for_human_input
+            and PickleHandler(training_data_filename).load()
```
gvieira: I liked how you've used `:=` above to prevent double-calling the `load()`. Should we try to do the same here?

src/crewai/agents/executor.py
```diff
@@ -246,12 +247,17 @@ def _iter_next_step(
         # If the tool chosen is the finishing tool, then we end and return.
         if isinstance(output, AgentFinish):
             if self.should_ask_for_human_input:
+                human_feedback = self._ask_human_input(output.return_values["output"])
+
+                if self.crew._train:
+                    self._training_handler(output, human_feedback)
```
gvieira: Above, you've used a verb to name the method (`_ask_human_input`). This is perfect, because that's exactly what it does – there's even no need for documentation on the method in this case. For `_training_handler` however, it doesn't feel the same. How could we name this method to make it more clear what it's doing?

src/crewai/agents/executor.py
pythonbyte: Definitely, I'll add it. Thanks @gvieira

src/crewai/agents/executor.py
pythonbyte: That makes sense, thanks, I'll change.

src/crewai/agents/executor.py
pythonbyte: Sure thing, I didn't have the proper time to review it, but I agree. 
I'll review and change it to make it more intuitive.

src/crewai/agents/executor.py
```diff
@@ -305,3 +315,31 @@ def _ask_human_input(self, final_answer: dict) -> str:
...
+        self, output: AgentFinish, human_feedback: str | None = None
+    ) -> None:
+        """Function to handle the process of the training data."""
+        agent_id = str(self.crew_agent.id)
+
+        if (
+            training_data := PickleHandler(TRAINING_DATA_FILE).load()
+            and not self.should_ask_for_human_input
+        ):
+            training_data = PickleHandler(TRAINING_DATA_FILE).load()
```
gvieira: Don't need this line anymore, right!?

src/crewai/utilities/fileHandler.py
gvieira, line 1: I could swear I had this file renamed to properly match the convention: `file_handler.py`. 🤔

src/crewai/utilities/fileHandler.py
```diff
@@ -16,5 +17,78 @@ def __init__(self, file_path):
     def log(self, **kwargs):
         now = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
         message = f"{now}: ".join([f"{key}={value}" for key, value in kwargs.items()])
-        with open(self._path, "a", encoding = 'utf-8') as file:
+        with open(self._path, "a", encoding="utf-8") as file:
             file.write(message + "\n")
+
+
+class PickleHandler:
```
gvieira: If I got it right, this has started as a generic handler/helper for dealing with pickle marshaling, right!? However, I believe it got really tied up to the training concept – we have a lot of train-related functions. Should we maybe move this to a training module maybe, renaming it to something like `TrainingHandler` or something like that?

src/crewai/utilities/fileHandler.py
pythonbyte: Do you want me to rename it?

src/crewai/agents/executor.py
pythonbyte: Sure thing 🤦🏽

src/crewai/utilities/fileHandler.py
pythonbyte: I see your point, makes sense. I could have the PickleHandler as a utility file handler class. 
And add a TrainingHandler that can Inherit the class and add the `save_trained_data/append` methods, what do you think?

src/crewai/utilities/fileHandler.py
gvieira: Yes. I think it makes sense as we don't use pascalCase for file names.

src/crewai/utilities/fileHandler.py
gvieira: That would be perfect! Loved the idea!

Issue #215, opened 2024-02-05
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType' in rpm_controller.py


nimar:
I'm trying to test the examples of crewai, but the initial import of crewai fails.

I'm using `crewai-0.5.0 langchain-0.1.0 langchain-openai-0.0.2.post1` on Python 3.9 on MacOS in a virtual environment.

Here's the code sample

```python
import os
import dotenv
dotenv.load_dotenv() # defines environment variable OPENAI_API_KEY.

import crewai
```

This is the full stack trace:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "test/.venv/lib/python3.9/site-packages/crewai/__init__.py", line 1, in <module>
    from crewai.agent import Agent
  File "test/.venv/lib/python3.9/site-packages/crewai/agent.py", line 22, in <module>
    from crewai.agents import (
  File "test/.venv/lib/python3.9/site-packages/crewai/agents/__init__.py", line 2, in <module>
    from .executor import CrewAgentExecutor
  File "test/.venv/lib/python3.9/site-packages/crewai/agents/executor.py", line 16, in <module>
    from crewai.utilities import I18N
  File "test/.venv/lib/python3.9/site-packages/crewai/utilities/__init__.py", line 4, in <module>
    from .rpm_controller import RPMController
  File "test/.venv/lib/python3.9/site-packages/crewai/utilities/rpm_controller.py", line 10, in <module>
    class RPMController(BaseModel):
  File "test/.venv/lib/python3.9/site-packages/crewai/utilities/rpm_controller.py", line 15, in RPMController
    _timer: threading.Timer | None = PrivateAttr(default=None)
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

joaomdmoura:
The issue is with the python version, `|` was added in python 3.10, so not available in python 3.9 :)

lugq1990:
> The issue is with the python version, `|` was added in python 3.10, so not available in python 3.9 :)

get the same error here, so why not to use union type to fix this?

PR #693, opened 2024-05-27
fix for python 3.9 import crewai error [#215]

lugq1990:
This is used to fix for the python 1.10 lower version to use crewai error with: 

```shell

TypeError: unsupported operand type(s) for |: 'type' and 'NoneType' in rpm_controller.py 
```

With the change from `|` to Union type just alike: `max_rpm: Union[int, None] = Field(default=None)` attr.

### DIFF
--- a/src/crewai/task.py
+++ b/src/crewai/task.py
@@ -2,7 +2,7 @@ import os
 import re
 import threading
 import uuid
-from typing import Any, Dict, List, Optional, Type
+from typing import Any, Dict, List, Optional, Type, Union
 
 from langchain_openai import ChatOpenAI
 from pydantic import UUID4, BaseModel, Field, field_validator, model_validator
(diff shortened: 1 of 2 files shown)


lugq1990:
@joaomdmoura could you please have a look for this? related to the lower python support from the type checking. Just keep consistency for the code side

theCyberTech:
You need to use Python 3.10 or higher 3.9 is not supported

lugq1990:
> You should to use Python 3.10 or higher 3.9 is not supported, what is the reason you would need to use 3.9?
> 
> Doesn't make sense to try and fix this in my opinion?

For me, currently base python version is  3.9, as for the framework side shouldn't just support with over 3.10, right?

theCyberTech:
So two options, 

1. Upgrade your base version to 3.10+
2. Run your crew in a virtual env using pyenv, poetry or conda

theCyberTech:
https://docs.crewai.com/how-to/Installing-CrewAI/

This link states the Python versions CrewAI supports

lugq1990:
Sure as the document mentions that support with python 3.10 over, than just close this pr

PR #723, opened 2024-06-02
missing datetime import

strnad:
there is missing import of datetime in utilities/logger.py


strnad:
sorry, duplicate (https://github.com/joaomdmoura/crewAI/pull/702)

PR #739, opened 2024-06-07
helicone integration docs

chitalian:


pythonbyte:
Hey @chitalian thanks for the pull request! 

I’d love to get your insights on how this tool integrates with CrewAI. I’m a bit unclear on that, and it would really help me understand better so we can support it effectively. Also, do you think we need to include this tool in the "crewai-tools" repository? Looking forward to your thoughts!

PR #745, opened 2024-06-08
Update MDXSearchTool.md

prime-computing-lab:
description fixed to markdown language instead of market data search


joaomdmoura:
Great catch!

PR #748, opened 2024-06-10
fix: 'from datetime import datetime for logging' to print the timestamp

thompsonson:
`from datetime import datetime` was missing from src/crewai/utilities/logger.py

A picture says a thousand words

before
image

after
image

### DIFF
--- a/src/crewai/utilities/logger.py
+++ b/src/crewai/utilities/logger.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 from crewai.utilities.printer import Printer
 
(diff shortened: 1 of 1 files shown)


arnaudgelas:
https://github.com/joaomdmoura/crewAI/pull/743 fix this as well

strnad:
duplicite: https://github.com/joaomdmoura/crewAI/pull/702

joaomdmoura:
Thanks, closing due to the fix already being merged

PR #749, opened 2024-06-10
fix: correct default model (gpt-4o), correct token counts, and correct TaskOutput attributes (added agent)

thompsonson:
image


joaomdmoura (approved):
<3 super appreciate!

PR #757, opened 2024-06-11
fix unterminated string literal

strnad:
fixed unterminated string literal in crew.py and task.py


strnad:
i messed up the branch

PR #760, opened 2024-06-12
updates instructor to the latest version.

### LINKED ISSUE #689: Unable to use crewAI with instructor version ^1.2.6
Hi
What would it take to upgrade the Instructor library underneath to the latest version?
seems like current instructor is **^0.5.2.**

The current version of instructor that I'm using is  (**^1.2.6**).

Since i need crew in the same project I cannot install it.

mcfunley:
Fixes #689 

Adding jsonref is also necessary - a test fails with inside of an instructor call without that.


joaomdmoura (approved):
<3

mcfunley:
Anything you need from me to get this merged? 

Some of these failures seem legit but also seem unrelated to any changes in the PR.

joaomdmoura:
Thanks! We will work on the failures before cutting the new version :)

PR #776, opened 2024-06-16
Lorenzejay/byoa

lorenzejay:
this ticket aims to bring your own AI agent whether its a custom langchain one or one like llamaindex and still work with crew or any custom agent.

* BaseAgent for defining agents that work together adhearing to the crew flow
* third-party-agent ecosystem for bringing in prebuilt agent from others leveraging the `BaseAgent` as an abstraction to adhere to the crew flow.


src/crewai/agent.py
```diff
@@ -264,6 +257,7 @@ def set_cache_handler(self, cache_handler: CacheHandler) -> None:
             self.cache_handler = cache_handler
             self.tools_handler.cache = cache_handler
         self.create_agent_executor()
+        pass
```
gvieira: Is this needed?

src/crewai/crew.py
```diff
@@ -255,15 +256,20 @@ def kickoff(
...
+            agent.crew = self  # type: ignore[attr-defined]
 
             if not agent.function_calling_llm:
                 agent.function_calling_llm = self.function_calling_llm
             if not agent.step_callback:
                 agent.step_callback = self.step_callback
-
-            agent.create_agent_executor()
+            # if using default crewai agent then run this
+            if isinstance(agent, Agent):
```
gvieira: This kind of logic is what I was trying to prevent with the introduction of a single agent interface to rule them all, you know.

src/crewai/crew.py
```diff
@@ -280,11 +286,16 @@ def kickoff(
             )
 
         metrics = metrics + [
-            agent._token_process.get_summary() for agent in self.agents
+            agent._token_process.get_summary()
+            for agent in self.agents
+            if not isinstance(agent, CustomAgent)
```
gvieira: Ditto. Ideally, all these checks would not exist and we could provide (one by one) wrappers for the famous agents, which would also allow others to implement agents that would respect the interface and respond to all required methods, some times doing nothing.

gvieira (commented):
Happy to go through the review, but I think we can connect to discuss the idea of a single Agent interface that would prevent us from doing `if agent is this` and `if agent is that`.

src/crewai/crew.py
lorenzejay: that makes sense. I wonder if just doing: 


Just discovered simply modifying the agents agent_executor works

```
agent1 = Agent(
    role="Memo Maker",
    goal="Analyze the changes in R&D expenditures and revenue",
    backstory="mathmatician turned commedian",
    verbose=True,
)

agent1.agent_executor=llama_index_agent.chat
```

src/crewai/agent.py
lorenzejay: nope.

src/crewai/crew.py
gvieira: What about those customizations you've done based on the type of the agent?

lorenzejay: sorry scratch that. it was running the original AgentExecutor defined in the Agent. my changes are necessary for running the external agent properly.

src/crewai/agents/Third_Party_Agents/llama_index/agent.py
```diff
@@ -0,0 +1,84 @@
...
+    def create_agent_executor(self, tools=None):
+        self.agent_executor = ReActAgent.from_llm(
+            tools=tools,
+            llm=self.llm,
+            verbose=self.verbose,
+            callback_manager=CallbackManager([self.token_counter]),
+            max_iterations=self.max_iter,
+        )
+
+    def set_agent_tools(self, agents: List[BaseAgent]):
```
gvieira: This name may be a bit misleading, as it's not setting any tool.

src/crewai/agents/new_wrapper_agent.py
gvieira, line 1: Should this be here?

src/crewai/crew.py
```diff
@@ -67,7 +69,7 @@ class Crew(BaseModel):
     cache: bool = Field(default=True)
     model_config = ConfigDict(arbitrary_types_allowed=True)
     tasks: List[Task] = Field(default_factory=list)
-    agents: List[Agent] = Field(default_factory=list)
+    agents: List[Agent | CustomAgent | BaseAgent] = Field(default_factory=list)
```
gvieira: In theory, it should be a list of `BaseAgent` only, right!?

lorenzejay: Yes. But for now we are still keeping the original implementation of Agent right ?

src/crewai/crew.py
```diff
@@ -67,7 +68,7 @@ class Crew(BaseModel):
     cache: bool = Field(default=True)
     model_config = ConfigDict(arbitrary_types_allowed=True)
     tasks: List[Task] = Field(default_factory=list)
-    agents: List[Agent] = Field(default_factory=list)
+    agents: List[BaseAgent] = Field(default_factory=list)
```
gvieira, line 75: This is GOLD!

src/crewai/crew.py
```diff
@@ -255,12 +256,16 @@ def kickoff(
...
-            if not agent.function_calling_llm:
+            # type: ignore[attr-defined] # Argument 1 to "_interpolate_inputs" of "Crew" has incompatible type "dict[str, Any] | None"; expected "dict[str, Any]"
+            agent.crew = self  # type: ignore[attr-defined]
+            if (
+                hasattr(agent, "function_calling_llm")
+                and not agent.function_calling_llm
+            ):
                 agent.function_calling_llm = self.function_calling_llm
-            if not agent.step_callback:
+            if hasattr(agent, "step_callback") and not agent.step_callback:
```
gvieira, line 302: This may be good-enough for now, but as we want to go down the type-safety road, I would replace it with a check (and cast) for something like `AgentWithCallbacks` protocol.

We could add a TODO comment, so we don't forget about it.

src/crewai/crew.py
```diff
@@ -278,14 +283,16 @@ def kickoff(
             raise NotImplementedError(
                 f"The process '{self.process}' is not implemented yet."
             )
-
         metrics = metrics + [
-            agent._token_process.get_summary() for agent in self.agents
+            agent.token_process.get_summary()
+            for agent in self.agents
+            if hasattr(agent, "token_process")
```
gvieira: Same here, but with another protocol name, like `AgentWithTokenization` or something like that.

src/crewai/crew.py
```diff
@@ -341,7 +348,7 @@ def _run_sequential_process(self) -> Union[str, Dict[str, Any]]:
                     agent for agent in self.agents if agent != task.agent
                 ]
                 if len(self.agents) > 1 and len(agents_for_delegation) > 0:
-                    task.tools += AgentTools(agents=agents_for_delegation).tools()
+                    task.tools += task.agent.get_delegation_tools(agents_for_delegation)
```
gvieira, line 380: Beautiful!

src/crewai/crew.py
```diff
@@ -413,11 +421,15 @@ def _run_hierarchical_process(self) -> Union[str, Dict[str, Any]]:
...
 
         self._finish_execution(task_output)
+
         # type: ignore # Incompatible return value type (got "tuple[str, Any]", expected "str")
-        manager_token_usage = manager._token_process.get_summary()
-        return self._format_output(
-            task_output, manager_token_usage
-        ), manager_token_usage
+        manager_token_usage = manager.token_process.get_summary()
+        if hasattr(manager, "token_process"):
```
gvieira: Same. Maybe it's actually a good idea to do that right now, given how simple it would be now.

src/crewai/task.py
```diff
@@ -9,10 +9,11 @@
 from pydantic import UUID4, BaseModel, Field, field_validator, model_validator
 from pydantic_core import PydanticCustomError
 
-from crewai.agent import Agent
+# from crewai.agent import Agent
```
gvieira: Should we remove it?

src/crewai/tools/agent_tools.py
```diff
@@ -4,23 +4,25 @@
...
+from crewai.agents.third_party_agents.base_agent import BaseAgent
 from crewai.task import Task
 from crewai.utilities import I18N
 
 
 class AgentTools(BaseModel):
     """Default tools around agent delegation"""
 
-    agents: List[Agent] = Field(description="List of agents in this crew.")
+    agents: List[Agent | BaseAgent] = Field(description="List of agents in this crew.")
```
gvieira: Should it only be `BaseAgent`?

src/crewai/crew.py
lorenzejay: actually just removed all of the hasattr checks for token_process. it makes sense for all agents to require this - so required with BaseAgent to provide that. And uses the TokenProcess class for it.

src/crewai/crew.py
lorenzejay: yeah adding a TODO here

src/crewai/agent.py
```diff
@@ -134,46 +75,15 @@ class Agent(BaseModel):
         default=None, description="Response format for the agent."
     )
 
-    _original_role: str | None = None
-    _original_goal: str | None = None
-    _original_backstory: str | None = None
-
     def __init__(__pydantic_self__, **data):
```
bhancockio, line 84: @lorenzejay you can remove this now that it's in the parent class.

src/crewai/agent.py
```diff
@@ -51,67 +42,17 @@ class Agent(BaseModel):
...
     )
-    agent_executor: InstanceOf[CrewAgentExecutor] = Field(
-        default=None, description="An instance of the CrewAgentExecutor class."
-    )
-    crew: Any = Field(
-        default=None, description="Crew to which the agent belongs.")
-    tools_handler: InstanceOf[ToolsHandler] = Field(
-        default=None, description="An instance of the ToolsHandler class."
-    )
     cache_handler: InstanceOf[CacheHandler] = Field(
```
bhancockio, line 51: @lorenzejay , drop now that this is in the parent.

src/crewai/agent.py
```diff
@@ -51,67 +42,17 @@ class Agent(BaseModel):
...
         default=None, description="An instance of the CacheHandler class."
     )
     step_callback: Optional[Any] = Field(
         default=None,
         description="Callback to be executed after each step of the agent execution.",
     )
-    i18n: I18N = Field(
-        default=I18N(), description="Internationalization settings.")
     llm: Any = Field(
```
bhancockio, line 58: this LLM attribuate code is 99% the same as the BaseAgent. Why not make the defult LLM for everyting gpt-4o in base_agent?

src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -0,0 +1,255 @@
...
+
+    _original_role: str | None = None
+    _original_goal: str | None = None
+    _original_backstory: str | None = None
+    _token_process: TokenProcess = TokenProcess()
+
+    def __init__(__pydantic_self__, **data):
+        config = data.pop("config", {})
+        super().__init__(**config, **data)
+
```
bhancockio, line 125: If we aren't going to make gpt-4o the default llm for everything. We need to add a model_validator that checks to make sure the agent has an LLM and isn't none.

src/crewai/agent.py
```diff
@@ -329,77 +226,32 @@ def create_agent_executor(self, tools=None) -> None:
...
+        return Converter(llm=llm, text=text, model=model, instructions=instructions)
 
-    # type: ignore # Function "langchain_core.tools.tool" is not valid as a type
-    def _parse_tools(self, tools: List[Any]) -> List[LangChainTool]:
+    def _parse_tools(self, tools: List[Any]) -> List[BaseTool]:
         """Parse tools to be used for the task."""
-        # tentatively try to import from crewai_tools import BaseTool as CrewAITool
         tools_list = []
         try:
+            # tentatively try to import from crewai_tools import BaseTool as CrewAITool
```
bhancockio, line 265: @lorenzejay , drop this comment.

src/crewai/agent.py
```diff
@@ -134,46 +75,15 @@ class Agent(BaseModel):
...
-        if self.max_rpm and not self._rpm_controller:
-            self._rpm_controller = RPMController(
-                max_rpm=self.max_rpm, logger=self._logger
-            )
-        return self
-
     @model_validator(mode="after")
     def set_agent_executor(self) -> "Agent":
-        """set agent executor is set."""
+        """Ensure agent executor and token process is set."""
```
bhancockio, line 90: `are` set

src/crewai/agent.py
```diff
@@ -247,33 +155,22 @@ def execute_task(
...
-    def set_rpm_controller(self, rpm_controller: RPMController) -> None:
-        """Set the rpm controller for the agent.
-
-        Args:
-            rpm_controller: An instance of the RPMController class.
-        """
-        if not self._rpm_controller:
-            self._rpm_controller = rpm_controller
-            self.create_agent_executor()
+    def format_log_to_str(
```
bhancockio, line 173: @lorenzejay should this be a ABC in BaseAgent?

src/crewai/agent.py
```diff
@@ -329,77 +226,32 @@ def create_agent_executor(self, tools=None) -> None:
...
-
-        copied_agent = Agent(**copied_data)
-        copied_agent.tools = deepcopy(self.tools)
+    def get_delegation_tools(self, agents: List[BaseAgent]):
+        agent_tools = AgentTools(agents=agents)
+        tools = agent_tools.tools()
+        return tools
 
-        return copied_agent
+    def get_output_converter(self, llm, text, model, instructions):
```
bhancockio, line 258: Should this be a ABC in BaseAgent as well?

bhancockio (changes requested):

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -0,0 +1,65 @@
+import time
+
+from crewai.memory.entity.entity_memory_item import EntityMemoryItem
+from crewai.memory.long_term.long_term_memory_item import LongTermMemoryItem
+from crewai.memory.short_term.short_term_memory_item import ShortTermMemoryItem
+from crewai.utilities.converter import ConverterError
+from crewai.utilities.evaluators.task_evaluator import TaskEvaluator
+
+
+class CrewAgentExecutorMixin:
```
bhancockio, line 10: Add

force_answer_max_iterations: Optional[int] = None

src/crewai/agents/executor.py
```diff
@@ -7,22 +7,20 @@
...
 from crewai.tools.tool_usage import ToolUsage, ToolUsageErrorException
 from crewai.utilities import I18N
-from crewai.utilities.converter import ConverterError
-from crewai.utilities.evaluators.task_evaluator import TaskEvaluator
+from crewai.utilities.constants import TRAINING_DATA_FILE
+from crewai.utilities.training_handler import CrewTrainingHandler
 
 
-class CrewAgentExecutor(AgentExecutor):
+class CrewAgentExecutor(AgentExecutor, CrewAgentExecutorMixin):
```
bhancockio, line 23: Drop `force_answer_max_iterations` because it now belongs to CrewAgentExecutorMixin.

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+    def _should_force_answer(self) -> bool:
+        return (
+            self.iterations == self.force_answer_max_iterations
```
bhancockio, line 13: None of these attributes exist on this class. You need to move these attributes over from executor.py and delete them over there.

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+        ) and not self.have_forced_answer
+
+    def _create_short_term_memory(self, output) -> None:
+        if (
+            self.crew
```
bhancockio, line 18: Need to move over crew to this class instead of executor.py.

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+                agent=self.crew_agent.role,
+                metadata={
+                    "observation": self.task.description,
+                },
+            )
+            self.crew._short_term_memory.save(memory)
+
+    def _create_long_term_memory(self, output) -> None:
+        if self.crew and self.crew.memory:
+            ltm_agent = TaskEvaluator(self.crew_agent)
```
bhancockio, line 33: Need to move over crew_agent from executor.py.

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+            evaluation = ltm_agent.evaluate(self.task, output.log)
+
+            if isinstance(evaluation, ConverterError):
+                return
+
+            long_term_memory = LongTermMemoryItem(
+                task=self.task.description,
```
bhancockio, line 40: Need to bring over task from executor.py.

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -0,0 +1,65 @@
...
+                    type=entity.type,
+                    description=entity.description,
+                    relationships="\n".join([f"- {r}" for r in entity.relationships]),
+                )
+                self.crew._entity_memory.save(entity_memory)
+
+    def _ask_human_input(self, final_answer: dict) -> str:
+        """Get human input."""
+        return input(
+            self._i18n.slice("getting_input").format(final_answer=final_answer)
```
bhancockio, line 64: Bring over _i18n.

src/crewai/agents/agent_builder/utilities/base_agent_tool.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class BaseAgentTools(BaseModel, ABC):
+    """Default tools around agent delegation"""
+
+    agents: List[BaseAgent] = Field(description="List of agents in this crew.")
+    i18n: I18N = Field(default=I18N(), description="Internationalization settings.")
+
+    @abstractmethod
+    def tools(self):
```
bhancockio, line 16: What should be the return type? BaseTool?

src/crewai/agents/agent_builder/utilities/base_agent_tool.py
```diff
@@ -0,0 +1,65 @@
...
+        return self._execute(coworker, task, context)
+
+    def ask_question(
+        self, question: str, context: str, coworker: Optional[str] = None, **kwargs
+    ):
+        """Useful to ask a question, opinion or take from a coworker passing all necessary context and names."""
+        coworker = self._get_coworker(coworker, **kwargs)
+        return self._execute(coworker, question, context)
+
+    def _execute(self, agent_role: Optional[str], task: str, context: str):
```
bhancockio: Screenshot 2024-06-26 at 11 59 52 PM

We need to add back in the clean up that's in this image to support weaker models.

bhancockio (changes requested):

src/crewai/agents/agent_builder/base_agent.py
lorenzejay: I agree. Other agents may not be supporting 4o yet. but I do think a validator for llm not being none is best.

src/crewai/agent.py
lorenzejay: note to me: from crewai_tools import BaseTool as CrewAITool

src/crewai/crew.py
```diff
@@ -288,12 +289,18 @@ def kickoff(
...
 
         for agent in self.agents:
+            # type: ignore # Argument 1 to "_interpolate_inputs" of "Crew" has incompatible type "dict[str, Any] | None"; expected "dict[str, Any]"
             agent.i18n = i18n
-            agent.crew = self
-
-            if not agent.function_calling_llm:
+            # type: ignore[attr-defined] # Argument 1 to "_interpolate_inputs" of "Crew" has incompatible type "dict[str, Any] | None"; expected "dict[str, Any]"
+            agent.crew = self  # type: ignore[attr-defined]
+            # TODO: Create an AgentFunctionCalling protocol for future refactoring
```
bhancockio, line 296: @lorenzejay drop TODO:

src/crewai/crew.py
```diff
@@ -288,12 +289,18 @@ def kickoff(
...
+            if (
+                hasattr(agent, "function_calling_llm")
+                and not agent.function_calling_llm
+            ):
                 agent.function_calling_llm = self.function_calling_llm
-            if not agent.step_callback:
+            # TODO: Create an AgentWithCallbacks protocol for future refactoring
```
bhancockio: @lorenzejay drop todo. Maybe create a Trello card for it.

src/crewai/agent.py
lorenzejay: done abstractmethod set

src/crewai/crew.py
```diff
@@ -435,17 +446,23 @@ def _run_hierarchical_process(self) -> Union[str, Dict[str, Any]]:
             )
 
             self._logger.log("debug", f"[{manager.role}] Task output: {task_output}")
-
+            if hasattr(task.agent, "_token_process"):
```
bhancockio: WHy do we have to do this check in hierarchical but not sequential?

src/crewai/crew.py
```diff
@@ -494,10 +511,12 @@ def _interpolate_inputs(self, inputs: Dict[str, Any]) -> None:
             for task in self.tasks
         ]
         # type: ignore # "interpolate_inputs" of "Agent" does not return a value (it only ever returns None)
-        [agent.interpolate_inputs(inputs) for agent in self.agents]
+        for agent in self.agents:
+            agent.interpolate_inputs(inputs)
+        # [agent.interpolate_inputs(inputs) for agent in self.agents]
```
bhancockio: Drop comment.

src/crewai/task.py
```diff
@@ -198,13 +198,13 @@ def _execute(self, agent, task, context, tools):
...
         )
-
         exported_output = self._export_output(result)
 
+        # type: the responses are usually str but need to figuire out a more elegant solution here
         self.output = TaskOutput(
             description=self.description,
-            exported_output=exported_output,
-            raw_output=result,
+            exported_output=str(exported_output),
```
bhancockio: We shouldn't be converting `exported_output` to string.

Sometimes it's a pydantic model or dictionary so don't want to be doing this.

What was the initial issue that caused you to change to a str?

src/crewai/task.py
```diff
@@ -198,13 +198,13 @@ def _execute(self, agent, task, context, tools):
...
+            raw_output=str(result),
```
bhancockio: This should always come back as a str? Is there a reason we have to do this?

src/crewai/task.py
```diff
@@ -302,14 +302,13 @@ def _export_output(self, result: str) -> Any:
                         pass
 
             # type: ignore # Item "None" of "Agent | None" has no attribute "function_calling_llm"
-            llm = self.agent.function_calling_llm or self.agent.llm
-
+            llm = getattr(self.agent, "function_calling_llm", None) or self.agent.llm
```
bhancockio, line 305: To prevent us from making these checks. Maybe we could update the base_agent to have a `function_calling_llm` but it would be set to None.

src/crewai/agents/agent_builder/utilities/base_agent_tool.py
lorenzejay: no tools can be declared different for every agent -BaseTool is langchain specific

src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -0,0 +1,255 @@
...
+    @abstractmethod
+    def _parse_tools(self, tools: List[Any]) -> List[Any]:
+        pass
+
+    @abstractmethod
+    def get_delegation_tools(self, agents: List["BaseAgent"]):
+        """Set the task tools that init BaseAgenTools class."""
+        pass
+
+    def get_output_converter(
```
bhancockio, line 184: This should also be an abstract method.

src/crewai/tools/agent_tools.py
```diff
@@ -3,24 +3,26 @@
...
 
 class AgentTools(BaseModel):
     """Default tools around agent delegation"""
 
-    agents: List[Agent] = Field(description="List of agents in this crew.")
+    agents: List[BaseAgent] = Field(description="List of agents in this crew.")
     i18n: I18N = Field(default=I18N(), description="Internationalization settings.")
 
     def tools(self):
+        coworkers = f"[{', '.join([f'{agent.role}' for agent in self.agents])}]"
```
bhancockio, line 10: You should reuse this variable on line 32 as well.

bhancockio (changes requested):

src/crewai/task.py
lorenzejay: reverted

src/crewai/crew.py
lorenzejay: attachment of an agent and task not required will give NoneTypes

lorenzejay: heirarchial does not need an agent in task

src/crewai/agents/agent_builder/utilities/base_agent_tool.py
lorenzejay: main does not have these changes. think we can this later ?

PR #796, opened 2024-06-19
Update LLM-Connections.md

jfaurote:
Revised to utilize Ollama from langchain.llms instead as the functionality from the other method simply doesn't work when delegating.


joaomdmoura:
nice! Thanks <3

PR #810, opened 2024-06-25
chore: fix typo

eltociear:
attemps -> attempts


joaomdmoura:
Good catch, there are a few extra places we need to update it tho if you don't mid it as well 🙏🏼

PR #818, opened 2024-06-26
Improve telemetry

gvieira:
A few more things added to the telemetry system.

### DIFF
--- a/src/crewai/telemetry/telemetry.py
+++ b/src/crewai/telemetry/telemetry.py
@@ -10,7 +10,10 @@ from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExport
 from opentelemetry.sdk.resources import SERVICE_NAME, Resource
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor
-from opentelemetry.trace import Status, StatusCode
+from opentelemetry.trace import Span, Status, StatusCode
+
+from crewai.crew import Crew
+from crewai.task import Task
 
 
 class Telemetry:
@@ -102,6 +105,8 @@ class Telemetry:
                             {
                                 "id": str(agent.id),
                                 "role": agent.role,
+                                "goal": agent.goal,
+                                "backstory": agent.backstory,
                                 "verbose?": agent.verbose,
                                 "max_iter": agent.max_iter,
                                 "max_rpm": agent.max_rpm,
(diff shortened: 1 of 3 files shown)


src/crewai/telemetry/telemetry.py
```diff
@@ -221,6 +266,7 @@ def crew_execution_span(self, crew):
                     pkg_resources.get_distribution("crewai").version,
                 )
                 self._add_attribute(span, "crew_id", str(crew.id))
+                self._add_attribute(span, "inputs", json.dumps(inputs))
```
gvieira, line 269: This whole code only executes if the user opted-in to share the crew.

src/crewai/telemetry/telemetry.py
```diff
@@ -253,16 +299,17 @@ def crew_execution_span(self, crew):
...
                                 "description": task.description,
+                                "expected_output": task.expected_output,
                                 "async_execution?": task.async_execution,
-                                "output": task.expected_output,
+                                "human_input?": task.human_input,
                                 "agent_role": task.agent.role if task.agent else "None",
                                 "context": (
                                     [task.description for task in task.context]
                                     if task.context
-                                    else "None"
```
gvieira, line 262: Is this fine for you, @joaomdmoura?

src/crewai/telemetry/telemetry.py
joaomdmoura: Yup, as it's opt-in this should be okay

src/crewai/telemetry/telemetry.py
joaomdmoura: I think this might break clickhouse, let's keep it as a string, as the conversion don't know how to translate this to JSON iirc

gvieira: Just checked and it works fine. The final json has `"context": null`. I think it's better.

src/crewai/telemetry/telemetry.py
```diff
@@ -88,9 +94,6 @@ def crew_creation(self, crew):
                 self._add_attribute(span, "python_version", platform.python_version())
                 self._add_attribute(span, "crew_id", str(crew.id))
                 self._add_attribute(span, "crew_process", crew.process)
-                self._add_attribute(
-                    span, "crew_language", crew.prompt_file if crew.i18n else "None"
```
gvieira, line 92: `crew` never had `i18n` 😬

PR #819, opened 2024-06-27
fix: Fix grammar error in documentation in PDF Search Tool

BrunoTanabe:
Correction of grammar error in the CrewAI documentation, on the page 'https://docs.crewai.com/tools/PDFSearchTool/' it says 'Optinal' instead of 'Optional'.


joaomdmoura (approved):
Great catch!

PR #822, opened 2024-06-27
fix: adding a check for the tools_handler before setting the cache handler

thompsonson:
here to avoid overwriting a custom tool handler with the default one


thompsonson:
There has been quite a change - specifically the introduction of BaseAgent so I'm closing this PR

PR #828, opened 2024-06-28
fixed ValueError with Pandas DataFrame

finecwg:
Fixed ValueError with Pandas DataFrame:

ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().


joaomdmoura (approved):
<3

PR #831, opened 2024-06-28
fixed mixin

lorenzejay:


src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -1,65 +1,100 @@
...
 
 from crewai.memory.entity.entity_memory_item import EntityMemoryItem
 from crewai.memory.long_term.long_term_memory_item import LongTermMemoryItem
 from crewai.memory.short_term.short_term_memory_item import ShortTermMemoryItem
 from crewai.utilities.converter import ConverterError
 from crewai.utilities.evaluators.task_evaluator import TaskEvaluator
+from crewai.utilities import I18N
 
 
 class CrewAgentExecutorMixin:
```
bhancockio, line 18: This class is only used once by the CrewAgentExecutor where it is inherting from the CrewAgentExecutorMixin class.

Is there a reason to seperate these classes? Should they become one? Who else will be inheriting from CrewAgentExecutorMixin?

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -1,65 +1,100 @@
...
+    crew: Optional[Any]
+    crew_agent: Optional[Any]
+    task: Optional[Any]
+    iterations: int
+    force_answer_max_iterations: int
+    have_forced_answer: bool
+    _i18n: I18N = I18N()
```
bhancockio: _i18n is declared the same way in `CrewAgentExecutor`.

I think we can drop `_i18n` from `CrewAgentExecutor` if every class that's going to inherit from `CrewAgentExecutorMixin` needs `_i18n`.

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -1,65 +1,100 @@
...
-            memory = ShortTermMemoryItem(
-                data=output.log,
-                agent=self.crew_agent.role,
-                metadata={
-                    "observation": self.task.description,
-                },
-            )
-            self.crew._short_term_memory.save(memory)
+            try:
+                if (
```
bhancockio, line 49: Shouldn't we move these if statements above to previous if statement instead of splitting into 2 if's.

It would be nice if we could refactor all of these conditional checks into a new function to improve readability like this:

```python
def _is_crew_properly_setup_for_memory(self) -> bool:
    # TODO: all conditionals here.
```

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -1,65 +1,100 @@
...
 
     def _create_long_term_memory(self, output) -> None:
-        if self.crew and self.crew.memory:
-            ltm_agent = TaskEvaluator(self.crew_agent)
-            evaluation = ltm_agent.evaluate(self.task, output.log)
-
-            if isinstance(evaluation, ConverterError):
-                return
+        """Create and save long-term and entity memory items based on evaluation."""
+        if self.crew and self.crew.memory and self.task and self.crew_agent:
```
bhancockio: Shouldn't we have all the same conditional checks from the previous function in this one as well?

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -1,65 +1,100 @@
...
+            try:
+                ltm_agent = TaskEvaluator(self.crew_agent)
```
bhancockio, line 69: Why do we evaluate in long term memory but not short?

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -1,65 +1,100 @@
...
+    have_forced_answer: bool
+    _i18n: I18N = I18N()
+
     def _should_force_answer(self) -> bool:
+        """Determine if a forced answer is required based on iteration count."""
         return (
             self.iterations == self.force_answer_max_iterations
         ) and not self.have_forced_answer
 
     def _create_short_term_memory(self, output) -> None:
```
bhancockio, line 33: Right now, we only have 2 tests for memory. It might be cool to add some more to verify that short term and long term memory are getting called correctly.

bhancockio (changes requested):

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
lorenzejay: every new (BYOA) agent

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
lorenzejay: Should we be? the original executor was not doing that. this is almost a copy / paste version of the original logic

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
lorenzejay: I can do that

joaomdmoura (approved):

PR #836, opened 2024-06-30
fix: Fix grammar error in documentation 'Crew Attributes'

BrunoTanabe:
Correction of grammar error in the CrewAI documentation, on the page 'https://docs.crewai.com/core-concepts/Crews/' it says 'ustom' instead of 'Custom'.


joaomdmoura (approved):
Great catch!

PR #843, opened 2024-07-01
Lj/optional agent in task bug

lorenzejay:


src/crewai/crew.py
```diff
@@ -219,6 +219,22 @@ def check_config(self):
...
+    def validate_tasks(self):
+        process = self.process
+        tasks = self.tasks
+
+        if process == Process.sequential:
+            for task in tasks:
+                if task.agent is None:
+                    raise PydanticCustomError(
+                        "missing_agent_in_task",
+                        "Agent is missing in the task with the following description: {task.description}",
```
bhancockio: From the user's perspective, it would be helpful to them if we added something like `the sequential process requires every task to have an assigned agent`.

I don't know if I'm the biggest fan of that messages so it might be good to wordsmith it.


But the important thing is we need to educate them about the error.

src/crewai/crew.py
```diff
@@ -443,29 +459,32 @@ def _run_hierarchical_process(self) -> Union[str, Dict[str, Any]]:
                 self._file_handler.log(
                     agent=manager.role, task=task.description, status="started"
                 )
-
+            if task.agent is not None:
```
bhancockio: To make this easier to read, we shouldn't check for a double negative. Instead, we could do this:

```python
if task.agent:
    manager.tools = task.agent.get_delegation_tools([task.agent])
else:
    manager.tools = manager.get_delegation_tools(self.agents)
```

src/crewai/crew.py
```diff
@@ -443,29 +459,32 @@ def _run_hierarchical_process(self) -> Union[str, Dict[str, Any]]:
...
+                manager.tools = task.agent.get_delegation_tools([task.agent])
+            else:
+                manager.tools = manager.get_delegation_tools(self.agents)
             task_output = task.execute(
                 agent=manager, context=task_output, tools=manager.tools
             )
+            if hasattr(manager, "_token_process"):
```
bhancockio: This won't be needed once my #844 PR get's merged in.

src/crewai/crew.py
```diff
@@ -443,29 +459,32 @@ def _run_hierarchical_process(self) -> Union[str, Dict[str, Any]]:
                 self._file_handler.log(
                     agent=manager.role, task=task.description, status="started"
                 )
-
+            if task.agent is not None:
+                manager.tools = task.agent.get_delegation_tools([task.agent])
```
bhancockio, line 496: I don't think this approach will work with async tasks because each async task is going to override the manager's tools. 

If you have questions about this, please let me know.

The easiest fix it just to disable async tasks with the hierarchical process. But, we'd have to ask Joao if he's cool with that.

tests/crew_test.py
```diff
@@ -722,6 +745,38 @@ def test_agent_usage_metrics_are_captured_for_hierarchical_process():
...
+        tasks=[task],
+        agents=[writer],
+        process=Process.hierarchical,
+        manager_llm=ChatOpenAI(model="gpt-4o"),
+    )
+
+    assert crew.process == Process.hierarchical
+    assert crew.manager_llm is not None
+    assert crew.tasks[0].agent == writer
+    result = crew.kickoff()
```
bhancockio: Right now, this isn't testing to make sure that during the `_run_hierarchical_process` that the manager agent's tools are only set to the writer when executing the task.

We would have to do some mocking to actually make this test valid.

bhancockio (changes requested):

tests/crew_test.py
lorenzejay: i think i can do another assert to look into crew.task[0].tools == delegate(Agent=[writer]) or something like that

src/crewai/crew.py
lorenzejay: I like this better

bhancockio (approved):

PR #844, opened 2024-07-01
Bugfix/kickoff for each usage metrics

bhancockio:
kickoff_for_each is not returning usage_metrics when setting crew to full_output=True


src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -187,6 +187,12 @@ def get_output_converter(
         """Get the converter class for the agent to create json/pydantic outputs."""
         pass
 
+    # TODO: HAVE LORENZE REVIEW THIS WITH OTHER AGENTS
+    @abstractmethod
```
bhancockio: @lorenzejay , Can you please confirm that making `copy` an abstract class to be implemented isn't going to cause issues.

It works with our normal agents but I want to make sure it works fine with BOYA.

src/crewai/task.py
```diff
@@ -216,7 +221,7 @@ def execute(  # type: ignore # Missing return statement
             )
             return result
 
-    def _execute(self, agent, task, context, tools):
+    def _execute(self, agent: "Agent", task, context, tools):
```
lorenzejay: How about:

def _execute(self, agent: BaseAgent, task, context, tools):

lorenzejay (approved):

joaomdmoura (approved):

PR #859, opened 2024-07-03
Fix typos in EN "human_feedback" string

abrinsmead:
Changes `This is the agent final answer: {final_answer}\nPlease provide a feedback:` 
to `This is the agent's final answer: {final_answer}\nPlease provide feedback:`


joaomdmoura:
Great catch!

PR #864, opened 2024-07-04
Fix definition of RAG in GithubTool docs

abrinsmead:
Correct the definition of RAG in the description of the `GithubSearchTool` from "Read, Append, and Generate" to "Retrieval-Augmented Generation".


joaomdmoura (approved):
Great catch, the crew must have gotten this one wrong lol

PR #867, opened 2024-07-04
Adding support to force a tool return to be the final answer.

joaomdmoura:
- Finally update dependencies (we can use new claude models)
- Ignore the cassetes
- This will at the end of the execution return the tool output. It will return the output of the latest tool with the flag


src/crewai/agent.py
```diff
@@ -188,6 +192,14 @@ def execute_task(
         )["output"]
         if self.max_rpm:
             self._rpm_controller.stop_rpm_counter()
+
+        # If there was any tool in self.tools_results that had result_as_answer
+        # set to True, return the restuls of the last tool that had
```
gvieira:
```suggestion
        # set to True, return the results of the last tool that had
```

src/crewai/tools/tool_usage.py
```diff
@@ -203,6 +206,21 @@ def _use(
...
         )
         result = self._format_result(result=result)  # type: ignore # "_format_result" of "ToolUsage" does not return a value (it only ever returns None)
+        data = {
+            "result": result,
+            "tool_name": tool.name,
+            "tool_args": calling.arguments,
+        }
+
+        if (
+            hasattr(original_tool, "result_as_answer")
```
gvieira, line 216: Ideally, we should setup a interface/protocol in crewAI-tools, so tools would optionally respect the protocol and we could check for it instead of searching for an attribute. Happy to take care of this as a follow up PR.

tests/agent_test.py
```diff
@@ -723,6 +723,75 @@ def test_agent_count_formatting_error():
...
+    agent1 = Agent(
+        role="Data Scientist",
+        goal="Product amazing resports on AI",
+        backstory="You work with data and AI",
+        tools=[MyCustomTool(result_as_answer=True)],
+    )
+
+    essay = Task(
+        description="Write and then review an small paragraph on AI until it's AMAZING. But first use the `Get Greetings` tool to get a greeting.",
+        expected_output="The final paragraph withthe full review on AI and no greeting.",
```
gvieira:
```suggestion
        expected_output="The final paragraph with the full review on AI and no greeting.",
```

gvieira (approved):

src/crewai/tools/tool_usage.py
joaomdmoura: <3 Love it good idea

PR #875, opened 2024-07-05
Emit task created

gvieira:
Just realized it's a good idea to have the information about a task being created along with the task being executed.

### DIFF
--- a/src/crewai/telemetry/telemetry.py
+++ b/src/crewai/telemetry/telemetry.py
@@ -163,6 +163,19 @@ class Telemetry:
                 tracer = trace.get_tracer("crewai.telemetry")
                 span = tracer.start_span("Task Execution")
 
+                created_span = tracer.start_span("Task Created")
+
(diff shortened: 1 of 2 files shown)


pythonbyte (approved):
LGTM ✅

src/crewai/telemetry/telemetry.py
```diff
@@ -163,6 +163,19 @@ def task_started(self, task: Task) -> Span | None:
...
 
+                created_span = tracer.start_span("Task Created")
+
+                self._add_attribute(created_span, "task_id", str(task.id))
+                self._add_attribute(
+                    created_span, "formatted_description", task.description
+                )
+                self._add_attribute(
+                    created_span, "formatted_expected_output", task.expected_output
+                )
```
joaomdmoura: because this could include PII let's only add it to the trace if the user opt-ins using the `share_crew` option

joaomdmoura (changes requested):
one single blocking comment

PR #885, opened 2024-07-07
passing cloned agents when copying context

WellyngtonF:
I was testing kickoff_for_each and figure out when i have only tasks without context in crew the copy works well, but when i add some context from other tasks the task.copy breaks, this was happen because the copy of task to clone the context dont pass the cloned agents, so the get_agent_by_role dont work, i dont know if this is the best fix for this problem, but i hope i can hear you opinions soon.

Thanks.


joaomdmoura:
great catch! cc @bhancockio

PR #893, opened 2024-07-08
Trying to add a max_token for the agents class

rumble773:
@joaomdmoura Not sure if this is the best or the right way, but I could use some help or guidelines to complete it in the right way.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "crewai"
-version = "0.36.0"
+version = "0.41.0"
 description = "Cutting-edge framework for orchestrating role-playing, autonomous AI agents. By fostering collaborative intelligence, CrewAI empowers agents to work together seamlessly, tackling complex tasks."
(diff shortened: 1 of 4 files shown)


lorenzejay:
Hi @rumble773 , added the attribute, but it seems like handling the max_token is done when you set the `llm`

for example:

```
ChatOpenAI(
            model=os.environ.get("OPENAI_MODEL_NAME", "gpt-4o"),
            max_tokens=1000,
        ),
```

So we may not need our agent classses to set this

lorenzejay (commented):
You would need to attach this passed max_tokens to the `llm`. so potentially after the validator you can set `self.llm = max_token`

rumble773:
@lorenzejay Thank you for reviewing this! 
I do agree with you on the second though of setting it in `set_agent_executor`, but not super sure where to start exactly. 
Can you assist me in that? and I will write the test and the doc for sure.

rumble773:
++ I think I shall close this commit, and try on a fresh one

PR #916, opened 2024-07-11
replay from task feature

lorenzejay:


src/crewai/crew.py
```diff
@@ -492,78 +509,116 @@ async def run_crew(crew, input_data):
...
                 agents_for_delegation = [
                     agent for agent in self.agents if agent != task.agent
                 ]
                 if len(self.agents) > 1 and len(agents_for_delegation) > 0:
                     task.tools += task.agent.get_delegation_tools(agents_for_delegation)
 
-            role = task.agent.role if task.agent is not None else "None"
-            self._logger.log("debug", f"== Working Agent: {role}", color="bold_purple")
+            if self.process == Process.hierarchical:
+                if task.agent and manager:
```
bhancockio: Should this be an if else statement?

src/crewai/crew.py
```diff
@@ -492,78 +509,116 @@ async def run_crew(crew, input_data):
...
-                    if task.context
-                    else aggregate_raw_outputs_from_task_outputs(task_outputs)
-                )
-                future = task.execute_async(
-                    agent=task.agent, context=context, tools=task.tools
-                )
-                futures.append((task, future))
-            else:
-                # Before executing a synchronous task, wait for all async tasks to complete
+                context = self._set_context(task, task_outputs)
```
bhancockio: I think this function should be renamed to `_get_context` because you are getting the context from the current task and task outputs.

Then, once you have the result, you are assigning it to the current context variable.

src/crewai/crew.py
```diff
@@ -492,78 +509,116 @@ async def run_crew(crew, input_data):
...
-        futures: List[Tuple[Task, Future[TaskOutput]]] = []
+        self.execution_logs = []
+        return self._execute_tasks(self.tasks)
 
-        for task in self.tasks:
+    def _execute_tasks(
+        self,
+        tasks: List[Task],
+        manager: Optional[BaseAgent] = None,
+    ) -> CrewOutput:
```
bhancockio, line 589: I think we need to add a new doc string to this function that explains what's going on.

This is the most important function in the entire crew file and we don't have any comments explaining what's going on.

src/crewai/cli/cli.py
```diff
@@ -48,5 +50,26 @@ def train(n_iterations: int):
     train_crew(n_iterations)
 
 
+@crewai.command()
+@click.option(
+    "-t",
+    "--task_id",
+    type=str,
+    help="The task ID of the task to replay from. This will replay the task and all the tasks that were executed after it.",
```
bhancockio: help="Replay the crew from this task ID, including all subsequent tasks.",

src/crewai/crew.py
```diff
@@ -685,6 +688,107 @@ def _run_hierarchical_process(self) -> CrewOutput:
...
+                for (index, d) in enumerate(stored_outputs)
+                if d["task_id"] == str(task_id)
+            ),
+            None,
+        )
+
+    def replay_from_task(
+        self, task_id: str, inputs: Dict[str, Any] | None = None
+    ) -> CrewOutput:
+        stored_outputs = TaskOutputJsonHandler(CREW_TASKS_OUTPUT_FILE).load()
```
bhancockio, line 744: Would it be possible to do some type validation on stored_outputs?

Because right now, if something isn't saved right, this is going to throw a lot of `cannot access property X on None`?

src/crewai/crew.py
```diff
@@ -607,74 +574,110 @@ def _run_hierarchical_process(self) -> CrewOutput:
...
+        if self.output_log_file:
+            self._file_handler.log(agent=role, task=task.description, status="started")
+
+    def _update_manager_tools(self, task: Task, manager: Optional[BaseAgent]):
+        if task.agent and manager:
+            manager.tools = task.agent.get_delegation_tools([task.agent])
+        if manager:
+            manager.tools = manager.get_delegation_tools(self.agents)
+
+    def _set_context(self, task: Task, task_outputs: List[TaskOutput]):
```
bhancockio: Let's change this to _get_context.

src/crewai/crew.py
```diff
@@ -607,74 +577,122 @@ def _run_hierarchical_process(self) -> CrewOutput:
...
-                    agent=manager.role, task=task.description, status="started"
+        Returns:
+            CrewOutput: Final output of the crew
+        """
+        task_outputs: List[TaskOutput] = []
+        futures: List[Tuple[Task, Future[TaskOutput], int]] = []
+        last_sync_output: Optional[TaskOutput] = None
+
+        for task_index, task in enumerate(tasks):
+            if start_index is not None and task_index < start_index:
```
bhancockio, line 605: Let's add some comments explaining that this code supports crew replays.

src/crewai/crew.py
```diff
@@ -607,74 +577,122 @@ def _run_hierarchical_process(self) -> CrewOutput:
...
+                if task.output:
+                    if task.async_execution:
+                        task_outputs.append(task.output)
+                    else:
+                        task_outputs = [task.output]
+                        last_sync_output = task.output
+                continue
+
+            self._prepare_task(task, manager)
+            agent_to_use = task.agent if task.agent else manager
```
bhancockio: 611 - 615 should be in a  `_select_and_validate_agent` function.

src/crewai/crew.py
bhancockio: Also, let's break this up into 2 for loops.

Right now, we are making this one for loop do too much: handle existing replayed tasks and then execute new tasks.

ideally, we should have the following. Screenshot below.

Please let me know if you have any questions.

bhancockio: Screenshot 2024-07-12 at 11 36 55 AM

src/crewai/crew.py
```diff
@@ -685,6 +703,72 @@ def _run_hierarchical_process(self) -> CrewOutput:
...
+        return next(
+            (
+                index
+                for (index, d) in enumerate(stored_outputs)
+                if d["task_id"] == str(task_id)
+            ),
+            None,
+        )
+
+    def replay_from_task(
```
bhancockio, line 741: Could this entire function be moved over to the new `ReplayCrew` class?

If not, could we move over some of the _find_task_index functions to the new class?

bhancockio, line 741: https://www.loom.com/share/b0f489adcaba4ca68c44177bc05a0d46?sid=56d1a54c-1e33-4407-86f0-070984e7d112

I had some questions comparing `replay` to `kickoff`

Also, does replay work with `kickoff_for_each`, `kickoff_async` and `kickoff_for_each_async`?

src/crewai/task.py
```diff
@@ -204,6 +204,7 @@ def _execute_core(
         tools: Optional[List[Any]],
     ) -> TaskOutput:
         """Run the core execution logic of the task."""
+        self.agent = agent
```
bhancockio, line 207: Why do we need to do this?

Self.agent isn't referenced anywhere else in _execute_core

tests/crew_test.py
```diff
@@ -1806,3 +1803,232 @@ def test__setup_for_training():
...
+        )
+
+        crew.kickoff()
+        crew.replay_from_task(str(write.id))
+        # Ensure context was passed correctly
+        assert mock_execute_task.call_count == 3
+
+
+@pytest.mark.vcr(filter_headers=["authorization"])
+def test_crew_replay_from_task_error():
```
bhancockio, line 1846: https://www.loom.com/share/ad5405dc1f7a4697b44b1528d005c4c4?sid=ce484e46-e699-4f58-ae40-bba7ee1c9351

tests/crew_test.py
```diff
@@ -1806,3 +1803,232 @@ def test__setup_for_training():
...
+            pydantic=None,
+            summary="Write about AI in healthcare...",
+        )
+
+        crew.kickoff()
+
+        # Check if the crew_tasks_output.json file is created
+        assert os.path.exists(CREW_TASKS_OUTPUT_FILE)
+
+        # Clean up the file after test
```
bhancockio, line 1895: I can't remember exactly what it's called, but there is a way to run a function after a test is executed  regardless of whether it passes or not, and it would be important to add that here because if this test fails, we won't clean up the test file which messes up the next test

tests/crew_test.py
```diff
@@ -1806,3 +1803,232 @@ def test__setup_for_training():
...
+        tasks=[task],
+    )
+
+    with pytest.raises(TypeError) as e:
+        crew.replay_from_task()  # type: ignore purposefully throwing err
+        assert "task_id is required" in str(e)
+
+
+@pytest.mark.vcr(filter_headers=["authorization"])
+def test_crew_task_output_file_creation():
```
bhancockio, line 1864: You probably want to delete the crew task output file before running this test to confirm that the kickoff function is actually creating the output file and that you're not looking at an output file from a previous run

bhancockio (changes requested):

tests/crew_test.py
lorenzejay: this is built in to the kickoff function

docs/core-concepts/Crews.md
```diff
@@ -155,4 +155,23 @@ for async_result in async_results:
     print(async_result)
 ```
 
+### Replaying from specific task:
+You can now replay from a specific task using the `replay_from_task` method.
```
bhancockio: use `-` instead of `_`

docs/core-concepts/Crews.md
```diff
@@ -155,4 +155,23 @@ for async_result in async_results:
...
+
+The replay_from_tasks feature in CrewAI allows you to replay from a specific task using the command-line interface (CLI). By running the command `crewai replay_from_tasks -t <task_id>`, you can specify the task name for the replay process.
```
bhancockio: should be `replay-from-task` not `replay_from_tasks`

docs/core-concepts/Crews.md
```diff
@@ -155,4 +155,23 @@ for async_result in async_results:
...
+
+Kickoffs will now create a `crew_tasks_ouput.json` file with the output of the tasks which you use to retrieve the task id to replay.
+
+
+### Replaying from specific task Using the CLI
+To use the replay_from_tasks feature, follow these steps:
```
bhancockio, line 167: Fix name again.

docs/core-concepts/Crews.md
```diff
@@ -155,4 +155,23 @@ for async_result in async_results:
...
+
+1. Open your terminal or command prompt.
+2. Navigate to the directory where your CrewAI project is located.
+3. Run the following command:
+
+```shell
+crewai replay-from-tasks -t <task_id>
```
bhancockio: Fix name.

bhancockio (changes requested):

bhancockio (approved):

PR #919, opened 2024-07-11
feat: Add planning feature to crew

pythonbyte:
This PR aims to add the Plan and Execution ability to the Crew.

With Crew(planning=True) before the Crew runs an AgentPlanner will be responsible for creating a step-by-step plan for each task on the crew and that plan is added to each task description to guide and improve the output


joaomdmoura:
@pythonbyte could you add docs and link them in `docs/index.md` and  `mkdocs.yml`?
Also jsut ot double check, dos this works on crews using the new yaml format?

pythonbyte:
> @pythonbyte could you add docs and link them in `docs/index.md` and `mkdocs.yml`? Also jsut ot double check, dos this works on crews using the new yaml format?

Nice catch @joaomdmoura thanks for reminding me. 
Added the docs and tested them on yaml and worked nicely.

src/crewai/task.py
```diff
@@ -326,9 +328,14 @@ def get_agent_by_role(role: str) -> Union["BaseAgent", None]:
 
     def _create_converter(self, *args, **kwargs) -> Converter:
         """Create a converter instance."""
-        converter = self.agent.get_output_converter(*args, **kwargs)
-        if self.converter_cls:
+        if self.agent:
```
pythonbyte, line 331: This was added to fix the Type Checking issue that was before this task. 

@joaomdmoura On a second note, I was talking with @bhancockio, and we realized that the "if sef.converter_cls" was added by this PR https://github.com/crewAIInc/crewAI/pull/800 

The issue with this overall function is that based on the typed return it should return a Converter, but both self.agent.get_output_converter and self.converter_cls could be None. So for now I'm raising the exception, but do you think is there something better that we can  do it here or if its relevant to have this "converter_cls"?

src/crewai/task.py
```diff
@@ -326,9 +328,14 @@ def get_agent_by_role(role: str) -> Union["BaseAgent", None]:
 
     def _create_converter(self, *args, **kwargs) -> Converter:
         """Create a converter instance."""
-        converter = self.agent.get_output_converter(*args, **kwargs)
-        if self.converter_cls:
+        if self.agent and not self.converter_cls:
```
pythonbyte, line 331: This was added to fix the Type Checking issue that was before this task.

@joaomdmoura On a second note, I was talking with @bhancockio, and we realized that the "if sef.converter_cls" was added by this PR https://github.com/crewAIInc/crewAI/pull/800

The issue with this overall function is that based on the typed return it should return a Converter, but both self.agent.get_output_converter and self.converter_cls could be None. So for now I'm raising the exception, but do you think is there something better that we can do it here or if its relevant to have this "converter_cls"?

joaomdmoura (approved):
LGTM

PR #930, opened 2024-07-14
Replay feat using db

lorenzejay:


docs/core-concepts/Crews.md
```diff
@@ -155,4 +155,30 @@ for async_result in async_results:
     print(async_result)
 ```
 
+### Replaying from specific task:
```
bhancockio, line 161: These docs are wrong now and use the wrong names. Need to talk about DB.

src/crewai/crew.py
```diff
@@ -707,6 +724,75 @@ def _run_hierarchical_process(self) -> CrewOutput:
...
+                if d["task_id"] == str(task_id)
+            ),
+            None,
+        )
+
+    def replay_from_task(
+        self, task_id: str, inputs: Optional[Dict[str, Any]] = None
+    ) -> CrewOutput:
+        stored_outputs = self._task_output_handler.load()
+        # TODO: write tests for this
```
bhancockio: Did you write tests for this? Can we remove this comment?

src/crewai/crew.py
```diff
@@ -707,6 +724,75 @@ def _run_hierarchical_process(self) -> CrewOutput:
...
+        return next(
+            (
+                index
+                for (index, d) in enumerate(stored_outputs)
+                if d["task_id"] == str(task_id)
+            ),
+            None,
+        )
+
+    def replay_from_task(
```
bhancockio, line 755: Does replay work if you run multiple crews consecutively? The reason I ask is because we are currently loading all stored outputs which can include task from three runs ago.

bhancockio: We should probably add a few tests around the scenario I just described above.

bhancockio (changes requested):

src/crewai/crew.py
lorenzejay: I did

src/crewai/crew.py
lorenzejay: Nope. this will reset every kickoff. So we would need to do something else for replaying from a specific crew. that would be a different ticket tho

docs/core-concepts/Crews.md
```diff
@@ -155,4 +155,30 @@ for async_result in async_results:
...
+To view latest kickoff task_ids use:
+
+```shell
+crewai log-tasks-outputs
+```
+
+
+```shell
+crewai replay -t <task_id>
+```
```
joaomdmoura, line 185: can we also add links to this on `mkdocs.yml` and `docs/index.md` please 🙏🏼

joaomdmoura (approved):

bhancockio (approved):

PR #941, opened 2024-07-15
Properly capture result from max retry recursive call

bhancockio:
There is a super tiny issue with our recursive `self.execute_task` calls not properly returning the results. 

Here's how the current issue occurs:
- The first `execute_task` 1 call fails and recursively calls `execute_task` agains.
- `execute_task` 2 succeeds and return results
- The first `execute_task` get's the result back from the recursive call. Once we get the result back, we aren't assigning it to `result` which causes issues when the first `execute_task` tries to return result.

### DIFF
--- a/src/crewai/agent.py
+++ b/src/crewai/agent.py
@@ -181,7 +181,6 @@ class Agent(BaseAgent):
         self.agent_executor.tools = parsed_tools
         self.agent_executor.task = task
 
-        # TODO: COMPARE WITH ARGS AND WITHOUT ARGS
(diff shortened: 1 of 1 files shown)


bhancockio:
@pythonbyte , please let me know if I missed anything! 

I made a future task for us here: https://trello.com/c/EE4RATV4/189-increase-error-handling-around-executetask

Because I think we need to do some more graceful error handling around `execute_task` failing as a whole.

Would love to hear your feedback on the trello card if you think we need to add anything else.

pythonbyte (approved):
Thanks @bhancockio nice catch.

PR #950, opened 2024-07-16
conditional task feat

lorenzejay:


src/crewai/__init__.py
```diff
@@ -2,5 +2,6 @@
 from crewai.crew import Crew
 from crewai.process import Process
 from crewai.task import Task
+from crewai.conditional_task import ConditionalTask
 
-__all__ = ["Agent", "Crew", "Process", "Task"]
+__all__ = ["Agent", "Crew", "Process", "Task", "ConditionalTask"]
```
bhancockio: I think `ConditionalTask` should go in the tasks folder.

src/crewai/conditional_task.py
```diff
@@ -0,0 +1,35 @@
...
+from crewai.tasks.task_output import TaskOutput
+
+
+class ConditionalTask(Task):
+    """
+    A task that can be conditionally executed based on the output of another task.
+    Note: This cannot be the only task you have in your crew and cannot be the first since its needs context from the previous task.
+    """
+
+    condition: Optional[Callable[[TaskOutput], bool]] = None
```
bhancockio: Why is this optional? Should a conditional task have to have a condition or else it would just be a regular task.

src/crewai/crew.py
```diff
@@ -295,6 +297,29 @@ def validate_end_with_at_most_one_async_task(self):
 
         return self
 
+    @model_validator(mode="after")
+    def validate_first_task(self) -> "Crew":
```
bhancockio, line 299: Shouldn't we also validate that the last task isn't conditional?

src/crewai/crew.py
```diff
@@ -295,6 +297,29 @@ def validate_end_with_at_most_one_async_task(self):
...
+        if self.tasks and isinstance(self.tasks[0], ConditionalTask):
+            raise PydanticCustomError(
+                "invalid_first_task",
+                "The first task cannot be a ConditionalTask.",
+                {},
+            )
+        return self
+
+    @model_validator(mode="after")
+    def validate_async_tasks_not_async(self) -> "Crew":
```
bhancockio, line 310: Fix comments for this method.

src/crewai/crew.py
```diff
@@ -623,6 +648,36 @@ def _execute_tasks(
                 )
             self._log_task_start(task, agent_to_use)
 
+            if isinstance(task, ConditionalTask):
+                if futures:
+                    task_outputs.extend(
```
bhancockio: Why are you extending task_outputs? I believe it should be like this right?

`task_outputs = self._process_async_tasks(futures, was_replayed)`

src/crewai/crew.py
```diff
@@ -623,6 +648,36 @@ def _execute_tasks(
...
+                        self._process_async_tasks(futures, was_replayed)
+                    )
+                    futures.clear()
+
+                previous_output = task_outputs[task_index - 1] if task_outputs else None
+                if previous_output is not None and not task.should_execute(
```
bhancockio: Is it a safe assumption to say if there are 3 async tasks before a conditional task, that we only want the result from the third async task to be used for our conditional statement?

src/crewai/crew.py
```diff
@@ -295,6 +297,29 @@ def validate_end_with_at_most_one_async_task(self):
 
         return self
 
+    @model_validator(mode="after")
```
bhancockio, line 298: Is it okay to chain multiple conditionals back to back?

src/crewai/crew.py
```diff
@@ -623,6 +648,36 @@ def _execute_tasks(
...
+                previous_output = task_outputs[task_index - 1] if task_outputs else None
+                if previous_output is not None and not task.should_execute(
+                    previous_output
+                ):
+                    self._logger.log(
+                        "info",
+                        f"Skipping conditional task: {task.description}",
+                        color="yellow",
+                    )
+                    skipped_task_output = TaskOutput(
```
bhancockio: A conditional_task should have a `get_skipped_output` or something like that so that the crew isn't responsible for generating this.

src/crewai/crew.py
```diff
@@ -623,6 +648,36 @@ def _execute_tasks(
                 )
             self._log_task_start(task, agent_to_use)
 
+            if isinstance(task, ConditionalTask):
```
bhancockio, line 649: This entire if statement should be in a `_process_conditional_task()` method.

bhancockio (changes requested):

src/crewai/crew.py
lorenzejay: i would say last is fine no need to hard require. because the output will just be the previous output. something i can add a test for.

docs/how-to/Conditional-Tasks.md
```diff
@@ -0,0 +1,88 @@
...
+
+task1 = Task(
+    name="Data Fetching Task",
+    description="Fetch data about events in San Francisco using Serper tool",
+    expected_output="List of 10 things to do in SF this week",
+    agent=data_fetcher_agent,
+    output_pydantic=EventOutput,
+)
+
+conditional_task = ConditionalTask(
```
bhancockio, line 61: I think we could tweak this code example a little more to share a real world example / use case of conditional tasks.

Without seeing a real world use case, developers will struggle to see the importance of this new feature.

bhancockio (changes requested):

src/crewai/crew.py
lorenzejay: why not ?

src/crewai/crew.py
lorenzejay: Possibly. users set the order of the agents. So proper crew creations and task orders are needed for this to work properly. I believe we want this to be conditional of a task before it vs all tasks before it (wouldn't really make sense). Future can hold a flag to get more tasks if this is requested but this is an edge case we havent seen yet.

bhancockio (approved):

PR #951, opened 2024-07-16
Test hierarchical tools proper setup

lorenzejay:


src/crewai/crew.py
```diff
@@ -657,7 +712,8 @@ def _execute_tasks(
 
     def _prepare_task(self, task: Task, manager: Optional[BaseAgent]):
```
bhancockio: Let's refactor this to `if self.process == Process.hierarchical and manager is not None:`

bhancockio: You probably also need to add a raise exception if the if or elif aren't hit.

bhancockio (changes requested):

bhancockio (approved):

PR #954, opened 2024-07-17
Fix logging for async and sync tasks

bhancockio:


src/crewai/crew.py
```diff
@@ -635,12 +634,14 @@ def _execute_tasks(
                 futures.append((task, future, task_index))
             else:
                 if futures:
+                    task_outputs = []
```
lorenzejay: this gets fixed later on, i believe i fixed it within conditional tasks but the task_outputs = [] is unneeded as self._process_async_tasks handles that

lorenzejay (approved):

PR #963, opened 2024-07-19
Update Force-Tool-Ouput-as-Result.md

chedug:
I think there is some mistake, because there is no such parameter as `force_output_result`, and as the code shows, the correct parameter `result_as_answer` is set during agent creation, not task.


chedug:
I closed it since my fork was not in sync with `main`. Please delete this PR.

PR #964, opened 2024-07-19
Update Force-Tool-Ouput-as-Result.md

chedug:
I think there is some mistake, because there is no such parameter as force_output_result, and as the code shows, the correct parameter result_as_answer is set during agent creation, not task.

Also I am not sure that this parameter modifies agent's result. I think it modifies the result of crew's execution. Correct if I'm wrong in this this part.


chedug:
Why do tests fail if I only change markdown file?

joaomdmoura (approved):

joaomdmoura:
We are working on fixing the test for the github actions

PR #975, opened 2024-07-22
Fix issues with co-worker task delegation

charl3sj:


charl3sj:
Closing this because it requires tools to be hashable which is not the case with `BaseTool` or `StructuredTool`

PR #976, opened 2024-07-22
Fix co-worker task delegation failure

charl3sj:


joaomdmoura:
Oh interesting, I was not aware of this, so this is to make sure the agent being delegated can use the tasks tools?

charl3sj:
The problem I was trying to fix is of [AgentTools](https://github.com/crewAIInc/crewAI/blob/2fe79e68cde0067b56b8aaf5901cfa50f664238a/src/crewai/tools/agent_tools.py#L6) (`Delegate work to coworker` and `Ask question to coworker`) not being loaded because of being appended to the task in this method: https://github.com/crewAIInc/crewAI/blob/2fe79e68cde0067b56b8aaf5901cfa50f664238a/src/crewai/crew.py#L733
... causing delegation to not work.

charl3sj:
After going through the docs and code again, I realized that delegation is meant for only hierarchical processes and I was trying delegation with a sequential process. That's why it didn't work. Feel free to close this PR if it's not a use-case you want to or makes sense to support.

joaomdmoura:
Oh it was supposed to work on sequential as well, I'll run some tests locally

Tehsmash:
Just ran into this exact issue, the docs say that delegation should work for Sequential but I was never seeing coworkers being used.

Looking at the proposed change, I think it might read better if _prepare_agent_tools was removed which adds the delegation tasks to the task's tool list (which is a bit weird given allow_delegation is a property of an Agent not a task).
Instead, we can move the manager tool setup into the `if self.process == Process.hieracchical` in _execute_tasks, and then for the Sequential case use the union logic to join task.agent.tools, task.tools, and task.agent.get_delegation_tools().

pythonbyte:
Hey @charl3sj! 
Thanks so much for the pull request! 

Could you take a moment to check if the changes are still valid? We've made quite a few updates since July. If they still apply, I’d be more than happy to help you get everything sorted and make it happen! 😊

Issue #888, opened 2024-07-08
Setting verbose=False for Agent Still Shows Tool Output


Wenlin88:
**Description**:
When setting the `verbose` parameter to `False` for an agent, the expectation is that the tool outputs should not be shown in the execution logs. However, I have observed that the tool output is still displayed. This issue persists regardless of the `verbose` setting.

**Steps to Reproduce**:
1. Define Azure OpenAI configuration.
2. Create a custom tool that returns a string.
3. Create an agent with `verbose=False` and assign the custom tool and Azure LLM to it.
4. Create a task for the agent to use the custom tool.
5. Form a crew with the agent and the task.
6. Kickoff the crew and observe the output.

**Example Code**:
```python
# full_example.py
import os
from crewai import Agent, Task, Crew, Process
from crewai_tools import tool
from langchain_openai import AzureChatOpenAI

# Define Azure OpenAI configuration
azure_llm  = AzureChatOpenAI(
    azure_endpoint=os.environ.get("AZURE_OPENAI_ENDPOINT"),
    api_key=os.environ.get("AZURE_OPENAI_KEY"),
    api_version=os.environ.get("AZURE_OPENAI_VERSION"),
    model=os.environ.get("AZURE_OPENAI_DEPLOYMENT"),
)

# Define the custom tool
@tool('Custom Tool')
def custom_tool():
    '''This is a custom tool that returns a string.'''
    return "This gets printed even if the verbose is off."

# Create an agent with verbose set to False
agent = Agent(
    role='Test Agent',
    goal='Test the tool output with verbose off',
    backstory='This agent is created to test the custom tool output.',
    verbose=False,
    tools=[custom_tool],
    llm=azure_llm,
)

# Create a task for the agent
test_task = Task(
    description="Use the custom tool and return its output.",
    expected_output="The output of the custom tool.",
    agent=agent
)

# Form a crew with the agent and the task
crew = Crew(
    agents=[agent],
    tasks=[test_task],
    process=Process.sequential
)

# Kickoff the crew
result = crew.kickoff()
print(result)
```
# Expected Behavior:
When verbose=False, the tool output should not be shown in the execution logs.

# Observed Behavior:
The tool output "This gets printed even if the verbose is off." is still shown even when verbose=False.

# My environment details: 
Operating System: Windows 11
Python version: 3.11.5
CrewAI version: 0.36.0

Wenlin88:
# Update on the Issue

[...]

PR #990, opened 2024-07-23
feat: add verbose option for printing in ToolUsage

Wenlin88:
Fix for Issue #888: Respect verbose=False Setting in ToolUsage Class

Summary: This pull request addresses issue #888 where the verbose=False setting for an agent in the CrewAI Python library does not suppress tool outputs in the execution logs.

Changes Made:
Modified the ToolUsage class in tool_usage.py to include checks for self.agent.verbose before printing tool outputs.

Code Example:
``` python
if self.agent.verbose:
    self._printer.print(content=f"\n\n{error}\n", color="red")

# Similarly, other print statements were updated to check self.agent.verbose
```

This fix ensures that tool outputs are properly suppressed when verbose=False.


joaomdmoura (approved):
LGTM!

PR #995, opened 2024-07-23
hierarchical process unblocked for async tasks

lorenzejay:


src/crewai/crew.py
```diff
@@ -779,10 +762,11 @@ def _log_task_start(self, task: Task, role: str = "None"):
             self._file_handler.log(agent=role, task=task.description, status="started")
 
     def _update_manager_tools(self, task: Task, manager: BaseAgent):
```
bhancockio: Why not drop the manager parameter?

Also, why not change the else to this?

` self.manager_agent.tools = manager.get_delegation_tools(self.agents)`

src/crewai/crew.py
```diff
@@ -729,20 +714,18 @@ def _handle_conditional_task(
             return skipped_task_output
         return None
 
-    def _prepare_agent_tools(self, task: Task, manager: Optional[BaseAgent]):
+    def _prepare_agent_tools(self, task: Task):
```
bhancockio, line 719: This is the function that is going to cause issues during async calls.

Let's say we have a crew with 4 tasks and the first 3 are async.

when we prepare the agent tools for each async function we are going to update the manager tools.

Since there is only one manager, we are going to overwrite what tools the manager can use in each async call which is not what we want.

I'm happy to brainstorm a solution with you on this.

tests/crew_test.py
```diff
@@ -1355,28 +1355,66 @@ def test_hierarchical_crew_creation_tasks_with_agents():
...
+        process=Process.hierarchical,
+        manager_llm=ChatOpenAI(model="gpt-4o"),
+    )
+
+    crew.kickoff()
+    assert crew.manager_agent is not None
+    assert crew.manager_agent.tools is not None
+    assert crew.manager_agent.tools[0].description.startswith(
+        "Delegate a specific task to one of the following coworkers: Senior Writer, Researcher, CEO\n"
     )
```
bhancockio, line 1420: There need to be another test to test the use case in one of my previous comments. 

For debugging purposes and to see that this is not working as expected, create the scenario described above, and then log the tools in the `_execute_core` right before you return the results for the task.

I would expect to see that the tools for the task are not correct.

bhancockio (changes requested):

bhancockio (approved):

PR #998, opened 2024-07-24
feat: add crew Testing/Evaluating feature

pythonbyte:
This PR intends to add the Testing/Evaluating feature to the Crew.

After implementation, you can run the following CLI command to run 2 iterations of tests.

```
╰─❯ crewai test -n 2
```

The output after running the crew's tasks will be something similar to this table.

```
                Task Scores
          (1-10 Higher is better)
┏━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━┳━━━━━━━━━━━━┓
┃ Tasks/Crew ┃ Run 1 ┃ Run 2 ┃ Avg. Total ┃
┡━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━╇━━━━━━━━━━━━┩
│ Task 1     │ 10.0  │ 9.0   │ 9.5        │
│ Task 2     │ 9.0   │ 9.0   │ 9.0        │
│ Crew       │ 9.5   │ 9.0   │ 9.2        │
└────────────┴───────┴───────┴────────────┘
```

The goal of this feature is to be able to evaluate crew quality. The CLI command also adds the ability to change the model but for now, we are only accepting OpenAI models (string).

### DIFF
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -970,7 +970,7 @@ class Crew(BaseModel):
     def test(
         self, n_iterations: int, model: str, inputs: Optional[Dict[str, Any]] = None
     ) -> None:
-        """Test the crew with the given inputs."""
+        """Test and evaluate the Crew with the given inputs for n iterations."""
         evaluator = CrewEvaluator(self, model)
 
         for i in range(1, n_iterations + 1):
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -37,6 +37,7 @@ from crewai.utilities.constants import (
     TRAINED_AGENTS_DATA_FILE,
     TRAINING_DATA_FILE,
 )
+from crewai.utilities.evaluators.crew_evaluator_handler import CrewEvaluator
 from crewai.utilities.evaluators.task_evaluator import TaskEvaluator
 from crewai.utilities.formatter import (
     aggregate_raw_outputs_from_task_outputs,
@@ -970,7 +971,13 @@ class Crew(BaseModel):
         self, n_iterations: int, model: str, inputs: Optional[Dict[str, Any]] = None
     ) -> None:
         """Test the crew with the given inputs."""
-        pass
+        evaluator = CrewEvaluator(self, model)
+
+        for i in range(1, n_iterations + 1):
+            evaluator.set_iteration(i)
+            self.kickoff(inputs=inputs)
+
+        evaluator.print_crew_evaluation_result()
 
     def __repr__(self):
         return f"Crew(id={self.id}, process={self.process}, number_of_agents={len(self.agents)}, number_of_tasks={len(self.tasks)})"
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -967,15 +967,18 @@ class Crew(BaseModel):
 
         return total_usage_metrics
 
-    def test(
-        self, n_iterations: int, model: str, inputs: Optional[Dict[str, Any]] = None
+    async def test(
+        self,
+        n_iterations: int,
+        openai_model_name: str,
+        inputs: Optional[Dict[str, Any]] = None,
     ) -> None:
         """Test and evaluate the Crew with the given inputs for n iterations."""
-        evaluator = CrewEvaluator(self, model)
+        evaluator = CrewEvaluator(self, openai_model_name)
 
         for i in range(1, n_iterations + 1):
             evaluator.set_iteration(i)
-            self.kickoff(inputs=inputs)
+            await self.kickoff_async(inputs=inputs)
 
         evaluator.print_crew_evaluation_result()
 
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -967,7 +967,7 @@ class Crew(BaseModel):
 
         return total_usage_metrics
 
-    async def test(
+    def test(
         self,
         n_iterations: int,
         openai_model_name: str,
@@ -978,7 +978,7 @@ class Crew(BaseModel):
 
         for i in range(1, n_iterations + 1):
             evaluator.set_iteration(i)
-            await self.kickoff_async(inputs=inputs)
+            self.kickoff(inputs=inputs)
 
         evaluator.print_crew_evaluation_result()
 
--- a/src/crewai/utilities/evaluators/crew_evaluator_handler.py
+++ b/src/crewai/utilities/evaluators/crew_evaluator_handler.py
@@ -142,4 +142,6 @@ class CrewEvaluator:
         )
 
         evaluation_result = evaluation_task.execute_sync()
-        self.tasks_scores[self.iteration].append(evaluation_result.pydantic.quality)
+
+        if isinstance(evaluation_result.pydantic, TaskEvaluationPydanticOutput):
+            self.tasks_scores[self.iteration].append(evaluation_result.pydantic.quality)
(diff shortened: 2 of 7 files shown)


src/crewai/utilities/evaluators/crew_evaluator_handler.py
```diff
@@ -0,0 +1,149 @@
...
+        tasks_scores (defaultdict): A dictionary to store the scores of the agents for each task.
+        iteration (int): The current iteration of the evaluation.
+    """
+
+    tasks_scores: defaultdict = defaultdict(list)
+    iteration: int = 0
+
+    def __init__(self, crew, model: str):
+        self.crew = crew
+        self.model = model
```
bhancockio: Might want to rename this to openai_model_name since we are only supporting OpenAI models right now with testing.

src/crewai/utilities/evaluators/crew_evaluator_handler.py
```diff
@@ -0,0 +1,149 @@
...
+from crewai.tasks.task_output import TaskOutput
+
+
+class TaskEvaluationPydanticOutput(BaseModel):
+    quality: float = Field(
+        description="A score from 1 to 10 evaluating on completion, quality, and overall performance from the task_description and task_expected_output to the actual Task Output."
+    )
+
+
+class CrewEvaluator:
```
bhancockio, line 19: I think we are using BaseModel for all of our classes

`class CrewEvaluator(BaseModel)`

src/crewai/utilities/evaluators/crew_evaluator_handler.py
```diff
@@ -0,0 +1,149 @@
...
+    Attributes:
+        crew (Crew): The crew of agents to evaluate.
+        model (str): The model to use for evaluating the performance of the agents (for now ONLY OpenAI accepted).
+        tasks_scores (defaultdict): A dictionary to store the scores of the agents for each task.
+        iteration (int): The current iteration of the evaluation.
+    """
+
+    tasks_scores: defaultdict = defaultdict(list)
+    iteration: int = 0
+
```
bhancockio, line 32: When we convert to `BaseModel`, we should add iterations and tasks_scores up here.

src/crewai/crew.py
```diff
@@ -969,8 +970,14 @@ def calculate_usage_metrics(self) -> Dict[str, int]:
     def test(
         self, n_iterations: int, model: str, inputs: Optional[Dict[str, Any]] = None
     ) -> None:
-        """Test the crew with the given inputs."""
-        pass
+        """Test and evaluate the Crew with the given inputs for n iterations."""
+        evaluator = CrewEvaluator(self, model)
+
+        for i in range(1, n_iterations + 1):
```
bhancockio, line 979: To test faster, should we run these in parallel with kickoff_async?

src/crewai/utilities/evaluators/crew_evaluator_handler.py
```diff
@@ -0,0 +1,149 @@
...
+        # Display the table in the terminal
+        console = Console()
+        console.print(table)
+
+    def evaluate(self, task_output: TaskOutput):
+        """Evaluates the performance of the agents in the crew based on the tasks they have performed."""
+        current_task = None
+        for task in self.crew.tasks:
+            if task.description == task_output.description:
```
bhancockio, line 130: I think it would make more sense if we updated the `TaskOutput` to include the id of the task it came from.

Once you do that, you can compare ID's instead of descriptions to find which task we are evaluating.

src/crewai/utilities/evaluators/crew_evaluator_handler.py
bhancockio, line 19: Do you have the original requirements for this testing evaluator? I know after our group call you wrote them down somewhere. I couldn't find them in Trello.

src/crewai/utilities/evaluators/crew_evaluator_handler.py
```diff
@@ -0,0 +1,149 @@
...
+            task.callback = self.evaluate
+
+    def set_iteration(self, iteration: int) -> None:
+        self.iteration = iteration
+
+    def _evaluator_agent(self):
+        return Agent(
+            role="Task Execution Evaluator",
+            goal=(
+                "Your goal is to evaluate the performance of the agents in the crew based on the tasks they have performed using score from 1 to 10 evaluating on completion, quality, and overall performance."
```
bhancockio, line 50: Do we need to specify that 10 is better and that 1 is worse?

tests/utilities/evaluators/test_crew_evaluator_handler.py
```diff
@@ -0,0 +1,112 @@
...
+                mock.call().add_column("Avg. Total"),
+                mock.call().add_row("Task 1", "10", "9", "9.5"),
+                mock.call().add_row("Task 2", "9", "8", "8.5"),
+                mock.call().add_row("Task 3", "8", "7", "7.5"),
+                mock.call().add_row("Crew", "9.0", "8.0", "8.5"),
+            ]
+        )
+        console.assert_has_calls([mock.call(), mock.call().print(table())])
+
+    def test_evaluate(self, crew_planner):
```
bhancockio, line 105: This test keeps failing with this error message:

```
ValueError: Evaluation result is not in the expected format
```

bhancockio (changes requested):

src/crewai/crew.py
pythonbyte: I think it makes sense, I'll try to run with the async and see how it goes

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: The TaskEvaluator doesn't inherit from BaseModel so I took the same pattern,  which I think is a class that is used to run something "outside" the crew itself but I'll try to adapt.

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: Ok

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: Yes, what I was thinking was to return the actual task along with the taskoutput so we can manipulate whatever we need, but I was not sure if would break something on the project. I'll add the ID and see how it goes.

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: The reference is the Trello card on current sprint and the excalidraw drawing.

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: I was thinking of that, so far the Model understands that 10 is "better" but I'll emphasize so the model is sure that 10 is better than 1.

tests/utilities/evaluators/test_crew_evaluator_handler.py
pythonbyte: I'll take a look thanks 🤔

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: Makes sense, thanks 👍🏽

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: I added and broke more than 15 tests, so I'll leave it for now, but so we keep in mind when back to this

src/crewai/utilities/evaluators/crew_evaluator_handler.py
pythonbyte: I think we can move this addition to v1 of the testing framework, or when we move to its actual project. 
The addition of the BaseModel generated several painful issues of validating fields and so on.

bhancockio (approved):

PR #1001, opened 2024-07-24
feat: add ability to set LLM for AgentPLanner on Crew

pythonbyte:


src/crewai/crew.py
```diff
@@ -154,6 +155,10 @@ class Crew(BaseModel):
         default=False,
         description="Plan the crew execution and add the plan to the crew.",
     )
+    planning_llm: Optional[Any] = Field(
```
bhancockio, line 157: Shouldn't we set this to default None here? 

The reason I say that is because when a crew is created, it will require a openai key going forward even if the user doesn't want to do planning or use open ai.

bhancockio, line 157: I can't get pytest or crews to run with this new property.

Please check out the loom video for more info:

https://www.loom.com/share/2aece814e603487c9c7dc09c8106f141?sid=4c5ad3ba-94d4-4db9-9377-52dfc369bdff

bhancockio (changes requested):

src/crewai/crew.py
pythonbyte: If I set it to None, it will force None as a parameter when we pass it to the planning function. 

Good catch, I'll change it so it forces the opener as the llm after the planning = True is verified

docs/core-concepts/Planning.md
```diff
@@ -23,6 +23,25 @@ my_crew = Crew(
 
 From this point on, your crew will have planning enabled, and the tasks will be planned before each iteration.
 
+#### Planning LLM
+
+Now you can define the LLM that will be used to plan the tasks. The default LLM is `gpt-4o-mini`, but you can change it to any other LLM as long as it is compatible.
```
bhancockio: We should drop this part of the description and change it to say they can use any ChatOpenAI llm:

`The default LLM is `gpt-4o-mini`, but you can change it to any other LLM as long as it is compatible.`

src/crewai/utilities/planning_handler.py
```diff
@@ -11,17 +12,27 @@ class PlannerTaskPydanticOutput(BaseModel):
 
 
 class CrewPlanner:
```
bhancockio, line 14: I think we also need to migrate this class over to use BaseModel for validation and better typings.

tests/utilities/test_planning_handler.py
```diff
@@ -1,10 +1,10 @@
 from unittest.mock import patch
```
bhancockio, line 1: In this file, It would be nice if we added a few more tests to test the initialization of CrewPlanner is working properly.

- one test that passes in an agent 
- One test that doesn't pass in an agent


Screenshot 2024-07-25 at 11 38 49 AM

bhancockio (changes requested):

docs/core-concepts/Planning.md
pythonbyte: Changed 👍🏽

src/crewai/utilities/planning_handler.py
pythonbyte: Since this is an internal Class, as we talked about, maybe for now we can focus on adding the feature and then evolve with more things on the next "version"

bhancockio (approved):

PR #1003, opened 2024-07-25
Patch/non gpt model pydantic output

lorenzejay:


src/crewai/utilities/crew_pydantic_output_parser.py
```diff
@@ -1,35 +1,30 @@
...
-from pydantic.v1 import BaseModel as V1BaseModel
 
 
 class CrewPydanticOutputParser(PydanticOutputParser):
     """Parses the text into pydantic models"""
 
-    pydantic_object: Union[Type[BaseModel], Type[V1BaseModel]]
+    pydantic_object: Type[BaseModel]
 
     def parse_result(self, result: List[Generation], *, partial: bool = False) -> Any:
```
bhancockio: Partial isn't used. We can probably delete this.

bhancockio (approved):

PR #1006, opened 2024-07-25
Feature/procedure v2

bhancockio:
https://linear.app/crewai/issue/CRE-18/pipeline

The purpose of this feature is to allow users to pipe crews together in series and parallel fashion.

To test out the pipeline feature on your own machine, I would copy some of the tests in `test_pipeline.py` file.

Also, the top of the pipeline.py file includes some of the key terminology you need to know to understand pipelines.

```
Pipeline Terminology:
Pipeline: The overall structure that defines a sequence of operations.
Stage: A distinct part of the pipeline, which can be either sequential or parallel.
Run: A specific execution of the pipeline for a given set of inputs, representing a single instance of processing through the pipeline.
Branch: Parallel executions within a stage (e.g., concurrent crew operations).
Trace: The journey of an individual input through the entire pipeline.

Example pipeline structure:
crew1 >> crew2 >> crew3

This represents a pipeline with three sequential stages:
1. crew1 is the first stage, which processes the input and passes its output to crew2.
2. crew2 is the second stage, which takes the output from crew1 as its input, processes it, and passes its output to crew3.
3. crew3 is the final stage, which takes the output from crew2 as its input and produces the final output of the pipeline.

Each input creates its own run, flowing through all stages of the pipeline.
Multiple runs can be processed concurrently, each following the defined pipeline structure.

Another example pipeline structure:
crew1 >> [crew2, crew3] >> crew4

This represents a pipeline with three stages:
1. A sequential stage (crew1)
2. A parallel stage with two branches (crew2 and crew3 executing concurrently)
3. Another sequential stage (crew4)

Each input creates its own run, flowing through all stages of the pipeline.
Multiple runs can be processed concurrently, each following the defined pipeline structure.
```


src/crewai/agent.py
```diff
@@ -199,7 +199,9 @@ def execute_task(
                     "tools": self.agent_executor.tools_description,
                 }
             )["output"]
+            print("Result when things went well:", result)
```
pythonbyte: _Suggestion_: Do you think we should use the Logger, instead of print statements? Or maybe the logging lib so we can inform the level of the information?

Same for lines 204/220.

docs/core-concepts/Pipeline.md
```diff
@@ -0,0 +1,202 @@
...
+| **add_run_result** | Adds a `PipelineRunResult` to the list of run results. |
+
+### Pipeline Run Result Attributes
+
+| Attribute         | Parameters      | Type                       | Description                                                                                   |
+| :---------------- | :-------------- | :------------------------- | :-------------------------------------------------------------------------------------------- |
+| **ID**            | `id`            | `UUID4`                    | A unique identifier for the run result.                                                       |
+| **Raw**           | `raw`           | `str`                      | The raw output of the final stage in the pipeline run.                                        |
+| **Pydantic**      | `pydantic`      | `Optional[BaseModel]`      | A Pydantic model object representing the structured output of the final stage, if applicable. |
+| **JSON Dict**     | `json_dict`     | `Optional[Dict[str, Any]]` | A dictionary representing the JSON output of the final stage, if applicable.                  |
```
pythonbyte, line 104: _Suggestion_: I think JSON Dict implies some ambiguity, do you think it makes more sense to be `dict_repr` since is a dictionary representation?

src/crewai/agents/executor.py
```diff
@@ -56,7 +56,7 @@ def _call(
         )
         intermediate_steps: List[Tuple[AgentAction, str]] = []
         # Allowing human input given task setting
-        if self.task.human_input:
+        if self.task and self.task.human_input:
```
pythonbyte, line 59: Nice 👍🏽

src/crewai/crew.py
```diff
@@ -146,8 +147,8 @@ class Crew(BaseModel):
         default=None,
         description="Path to the prompt json file to be used for the crew.",
     )
-    output_log_file: Optional[Union[bool, str]] = Field(
-        default=False,
+    output_log_file: Optional[str] = Field(
+        default="",
```
pythonbyte: _Suggestion_: Usually is a good practice to have default values as None, instead of empty values such as "", [] and {}.

src/crewai/crew.py
```diff
@@ -96,12 +96,13 @@ class Crew(BaseModel):
         default_factory=TaskOutputStorageHandler
     )
 
+    name: Optional[str] = Field(default="")
```
pythonbyte: _Suggestion_: Nice, finally we can have a name for the Crew, maybe we can even add for tasks. In here having the default as None I think would be better.

src/crewai/pipeline/pipeline.py
```diff
@@ -0,0 +1,218 @@
...
+        self, traces: List[List[Union[str, Dict[str, Any]]]]
+    ) -> List[List[Trace]]:
+        formatted_traces: List[Trace] = []
+        for trace in traces[:-1]:
+            formatted_traces.append(trace[0] if len(trace) == 1 else trace)
+
+        traces_to_return: List[List[Trace]] = []
+        final_trace = traces[-1]
+        if len(final_trace) == 1:
+            formatted_traces.append(final_trace[0])
```
pythonbyte, line 326: _Suggestion_: Maybe in this part, you can separate into 2 sub-functions `format_single_trace` and `format_multiple_traces` this way gives the ability to be more testable and "single-focused".

src/crewai/pipeline/pipeline.py
```diff
@@ -0,0 +1,218 @@
...
+            for crews_outputs, formatted_trace in zip(
+                formatted_crew_outputs, formatted_traces
+            )
+        ]
+
+    def _format_traces(
+        self, traces: List[List[Union[str, Dict[str, Any]]]]
+    ) -> List[List[Trace]]:
+        formatted_traces: List[Trace] = []
+        for trace in traces[:-1]:
```
pythonbyte: _Question_: In this part the [:-1] doesn't account for the last element, is that the original intent?
```
>>> [1,2,3,4][:-1]
[1, 2, 3]
```

src/crewai/agent.py
bhancockio: Great catch! I dropped all print calls.

src/crewai/crew.py
bhancockio: Good call out! Just reverted!

docs/core-concepts/Pipeline.md
bhancockio: I think dict_repr is a good idea because it shows the field is a dictionary representation and can be used in different situations. 

However, it doesn't clearly indicate that it's meant for JSON data, which can be confusing. Using json_dict is might be more clear here because it makes it clear that the dictionary is for JSON, which helps avoid any misunderstandings about its purpose.

However, I'm happy to have the whole team way in on this because we use the json_dict in a bunch of places like `TaskOutput`, `CrewOutput` and more. So, I want to make sure we get this right!

src/crewai/crew.py
bhancockio: Great suggestion! Changed default to None and added `name` to Task as well.

src/crewai/pipeline/pipeline.py
bhancockio: I like this suggestion! I just broke things apart.

src/crewai/pipeline/pipeline.py
bhancockio: It is and I realize the logic for why isn't super clear so I'm going to work on adding commentary to describe what is going on.

PR #1007, opened 2024-07-25
remove API key from codeblock

bassamadnan:
the API keys for serper and openai were setup, in a awkward fashion, but were never used and are not needed here.

### DIFF
--- a/docs/core-concepts/Tools.md
+++ b/docs/core-concepts/Tools.md
@@ -39,10 +39,6 @@ from crewai_tools import (
     WebsiteSearchTool
 )
 
-# Set up API keys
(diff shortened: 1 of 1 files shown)


bassamadnan:
alternatively, we could have passed it as a crew arg. Let me know if that would be better or thats what was intended and i'll make the changes.

pythonbyte:
Hi @bassamadnan,

Thank you so much for your pull request! I just wanted to clarify that if no language model (LLM) is explicitly provided, we default to using OpenAI. We are also utilizing the Serper tool in the example, which needs the Serper API key to be set as an environment variable.

We appreciate your input and will work on making this process clearer. For now, I’ll close the PR, but please feel free to reach out if you have any questions or suggestions. 

Thanks again!

PR #1009, opened 2024-07-26
Json Task Output Truncation with Escape Characters

bhancockio:
Message from Christian on X

Hey Brandon, I think I have figured out whats happening. the JSON being returned from the task sometimes contain control characters like \r\n and so 'model_validate_json' in [http://task.py](https://t.co/Z9r1nXidZR) is failing and falling back to '_handle_partial_json' which I think is where the problem lies (there seems to be an LLM call in here?) I have written a function def escape_json_string(json_string): json_data = json.loads(json_string, strict=False) escaped_string = json.dumps(json_data) return escaped_string which escapes the string first and then if i pass the result into model_validate_json(escaped_string) just to test it works fine. I basically think there needs to be a way to escape the crew task output string before it is passed into _convert_to_model ? Happy to jump on a call when you are online

Example code:

class EmailResponse(BaseModel): previous_message_content: str class EmailResponses(BaseModel): responses: list[EmailResponse] json_string_test = """ { "responses": [ { "previous_message_content": "Hi Tom,\r\n\r\nNiamh has chosen the Mika phonics on" } ] } """ # this errors EmailResponses.model_validate_json(json_string_test) # this does not error escaped_string = json.dumps(json.loads(json_string, strict=False)) EmailResponses.model_validate_json(escaped_string)


src/crewai/utilities/converter.py
```diff
@@ -72,3 +77,122 @@ def _create_chain(self):
...
+    converter_cls: Optional[Type[Converter]] = None,
+) -> Union[dict, BaseModel, str]:
+    model = output_pydantic or output_json
+    if model is None:
+        return result
+
+    try:
+        escaped_result = json.dumps(json.loads(result, strict=False))
+        return validate_model(escaped_result, model, bool(output_json))
+    except Exception:
```
lorenzejay: thoughts on excepting 

```py
except json.JSONDecodeError:
        return handle_partial_json(result, model, bool(output_json), agent, converter_cls)
```

then `except Exception as e:`

excepting json.JSONDecodeError focuses on the json decode logic before throwing another instead of catching any error that can come up

src/crewai/utilities/converter.py
```diff
@@ -72,3 +77,122 @@ def _create_chain(self):
...
+    converter_cls: Optional[Type[Converter]] = None,
+) -> Union[dict, BaseModel, str]:
+    match = re.search(r"({.*})", result, re.DOTALL)
+    if match:
+        try:
+            exported_result = model.model_validate_json(match.group(0))
+            if is_json_output:
+                return exported_result.model_dump()
+            return exported_result
+        except Exception:
```
lorenzejay: def a nit
`except pydantic.ValidationError:`

src/crewai/utilities/converter.py
```diff
@@ -72,3 +77,122 @@ def _create_chain(self):
...
+        )
+        return result
+
+    return exported_result
+
+
+def get_conversion_instructions(model: Type[BaseModel], llm: Any) -> str:
+    instructions = "I'm gonna convert this raw text into valid JSON."
+    if not is_gpt(llm):
+        model_schema = PydanticSchemaParser(model=model).get_schema()
```
lorenzejay, line 199: try catch here ? these will always be hidden failures / errors since it wont break the task outputs we will just never know why it doesnt work

src/crewai/utilities/converter.py
bhancockio: Hey @lorenzejay, so Anthropic wanted me to do this:

```python
except json.JSONDecodeError:
        return handle_partial_json(
            result, model, bool(output_json), agent, converter_cls
        )
except Exception:
        return handle_partial_json(
            result, model, bool(output_json), agent, converter_cls
        )
```

But either way, We are still calling the same thing and not reacting differently to the exceptions so I figured just do one. 

What do you think?

src/crewai/utilities/converter.py
bhancockio: @lorenzejay Are you wanting me to add both?

lorenzejay (approved):

PR #1011, opened 2024-07-26
[Bug] Improve _save_file method to handle both dict and str inputs

rumble773:
- Add check for dict type input
- Use json.dump for dict serialization
- Convert non-dict inputs to string
- Remove type ignore comments


joaomdmoura (approved):
Nice, good call!

PR #1033, opened 2024-07-30
AgentOps ENG-525: Decouple CrewAI and AgentOps

tiovikram:
* Make AgentOps import optional upon AGENTOPS_API_KEY being set
  
## What
CrewAI attempts to import agentops by default if the agentops package is installed in the environment.

## Why
This makes breakages such as the following to become more apparent and fatal.
IMG_6237

## How
Only import agentops if `AGENTOPS_API_KEY` environment variable is set.


tiovikram:
AgentOps Reviewers: @areibman @bboynton97

joaomdmoura:
Good catch!

PR #1042, opened 2024-07-31
Feat/sliding context window

lorenzejay:


pyproject.toml
```diff
@@ -29,6 +29,7 @@ jsonref = "^1.1.0"
 agentops = { version = "^0.3.0", optional = true }
 embedchain = "^0.1.114"
 json-repair = "^0.25.2"
+prompt-toolkit = "^3.0.47"
```
pythonbyte: _Question_: Based on the code that uses this library, is it really required to have it?

src/crewai/agents/agent_builder/utilities/yes_no_cli_validator.py
```diff
@@ -0,0 +1,8 @@
+from prompt_toolkit.validation import Validator, ValidationError
+
+
+class YesNoValidator(Validator):
+    def validate(self, document):
+        text = document.text.lower()
+        if text not in ["y", "n", "yes", "no"]:
+            raise ValidationError(message="Please enter Y/N")
```
pythonbyte: _Question_: Again, this could've been a simple validator function

```suggestion
def validate_user_input(user_input: str):
    if document.text.lower() not in ["y", "n", "yes", "no"]:
         raise Exception("Error with the input")
```

src/crewai/agents/executor.py
```diff
@@ -185,6 +197,52 @@ def _iter_next_step(
...
 
+        except BadRequestError as e:
+            print("Bad Request Error", e)
+            if "context_length_exceeded" in str(e):
+                self._logger.log(
+                    "debug",
+                    "Context length exceeded. Asking user if they want to use summarize prompt to fit, this will reduce context length.",
+                    color="yellow",
+                )
+                user_choice = prompt(
```
pythonbyte: _Suggestion_: In here again we could've been simpler

```suggestion
                user_choice = input("Context length exceeded. Do you want to summarize the text to fit models context window? (Y/N): ")
                validate_user_input(user_choice)
```

src/crewai/agents/executor.py
```diff
@@ -185,6 +197,52 @@ def _iter_next_step(
             yield AgentStep(action=output, observation=observation)
             return
 
+        except BadRequestError as e:
+            print("Bad Request Error", e)
```
pythonbyte: _Suggestion_: Instead of print I think we should focus on using the Logger or the logging lib.

src/crewai/agents/executor.py
```diff
@@ -185,6 +197,52 @@ def _iter_next_step(
             yield AgentStep(action=output, observation=observation)
             return
 
+        except BadRequestError as e:
```
pythonbyte: _Suggestion_: If the error raised is a type-specific error, maybe is worth it to create a Custom Exception and handle the specific cases when we have the context length exceeded this way you can have something like

`except ContextLengthExceeded as exc`

src/crewai/agents/executor.py
```diff
@@ -185,6 +197,52 @@ def _iter_next_step(
...
+                self._logger.log(
+                    "debug",
+                    "Context length exceeded. Asking user if they want to use summarize prompt to fit, this will reduce context length.",
+                    color="yellow",
+                )
+                user_choice = prompt(
+                    "Context length exceeded. Do you want to summarize the text to fit models context window? (Y/N): ",
+                    validator=YesNoValidator(),
+                ).lower()
+                if user_choice in ["y", "yes"]:
```
pythonbyte: _Suggestion_: Maybe better this way.
```suggestion
    if user_choice.lower().startswith("y"):
```

src/crewai/agents/executor.py
```diff
@@ -185,6 +197,52 @@ def _iter_next_step(
...
+            else:
+                raise e
+
+        except Exception as e:
+            yield AgentStep(
+                action=AgentAction("_Exception", str(e), str(e)),
+                observation=str(e),
+            )
+            return
+
```
pythonbyte: _Suggestion_: All this code could be its method to handle these cases. As I suggested above, if you have a Custom exception you could do something like this:
_hint_: If we have branched out to a specific case and it has too many lines, might be better to have it in a different function to encapsulate and isolate that behavior to be more testable and easier to read

```
except ContextLengthException as exc:
     self._handle_conext_length_exc()
```

pythonbyte (changes requested):
Great work, left some comments and let me know if you have any questions 😄

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
pythonbyte: _Suggestion_: Since we are not adding any additional library anymore, can you check if this change on the poetry.lock is required?

src/crewai/agents/executor.py
```diff
@@ -131,7 +144,7 @@ def _iter_next_step(
             intermediate_steps = self._prepare_intermediate_steps(intermediate_steps)
 
             # Call the LLM to see what to do.
-            output = self.agent.plan(  # type: ignore #  Incompatible types in assignment (expression has type "AgentAction | AgentFinish | list[AgentAction]", variable has type "AgentAction")
```
pythonbyte, line 134: Niceee 👍🏽 😄

src/crewai/agents/executor.py
```diff
@@ -185,6 +198,55 @@ def _iter_next_step(
...
+                str(e)
+            ):
+                if "context_length_exceeded" in str(e):
+                    self._logger.log(
+                        "debug",
+                        "Context length exceeded. Asking user if they want to use summarize prompt to fit, this will reduce context length.",
+                        color="yellow",
+                    )
+                    user_choice = click.confirm(
+                        "Context length exceeded. Do you want to summarize the text to fit models context window?"
```
pythonbyte: _Question_: I imagine that at the end is missing something like a (y). If the user presses enter without anything or just types anything, how it work?

src/crewai/agents/executor.py
```diff
@@ -185,6 +198,55 @@ def _iter_next_step(
             yield AgentStep(action=output, observation=observation)
             return
 
+        except Exception as e:
```
pythonbyte, line 201: _Question_: Isn't possible to raise the specific type of Exception that is raised when the contextExceeded limit is raised?

src/crewai/agents/executor.py
```diff
@@ -185,6 +198,55 @@ def _iter_next_step(
...
+                            color="red",
+                        )
+                        raise SystemExit(
+                            "Context length exceeded and user opted not to summarize. Consider using smaller text or RAG tools from crewai_tools."
+                        )
+            yield AgentStep(
+                action=AgentAction("_Exception", str(e), str(e)),
+                observation=str(e),
+            )
+            return
```
pythonbyte: _Suggestion_: I think this whole portion could be a different function. I saw that we already have the _handle_context_length but we can add a _handle_context_length_error and just call inside of this part of the code, because this function is getting bigger than expected.

pythonbyte (approved):
Nice work

src/crewai/agents/executor.py
lorenzejay: not Explicitly the ContextWindowError also we need to ensure this gets captured for all the other model providers - groq, cohere, and anthropic

PR #1049, opened 2024-08-02
Increase test coverage for output to file

thiagomoretto:
By writing to the disk on these unit tests we cover the part of the logic in `Task._save_file`, preventing some regression when changing its behavior.

This confirms that https://github.com/crewAIInc/crewAI/pull/1011 fixes the issue when the model outputs a string which is a parseable json, as it's transformed to a dict later before going to disk.

image


joaomdmoura:
Love it!

PR #1098, opened 2024-08-07
Fix misplaced task info from process doc

maf-rnmourao:
This PR removes the misplaced topic [`Additional Task Features`](https://docs.crewai.com/core-concepts/Processes/#additional-task-features) from the `Core Concepts -> Processes` documentation.


joaomdmoura:
Great catch!

PR #1161, opened 2024-08-08
feat: add ability to train on custom file

pythonbyte:


lorenzejay (approved):
awesome work. ran it locally adding different filenames and works as expected

PR #1162, opened 2024-08-09
[CRE-28] Added new GitHub Action to release the docs as a package

theCyberTech:
This PR creates a new GitHub action to create an independent package of the CrewAI docs folder .

Currently it is set to the following trigger which should be reviewed
on:
  push:
    branches: [ main ]
  workflow_dispatch:


theCyberTech:
No longer required

PR #1166, opened 2024-08-11
add casefold comparison to the repeated tools usage check

arylwen:
Some open-source models, e.g., mixtral, would generate the same tool call with a different case. With a casefold comparison, we could reduce the number of repeated tool calls.


pythonbyte:
Hey @arylwen!

Thanks for the pull request! 

Could you take a moment to check if the changes are still relevant and working properly? If you could share a few examples, that would really help me test your use case. I appreciate your help, and just let me know if there's anything we can do to help.

bhancockio:
Closed due to inactivity.

PR #1167, opened 2024-08-11
Attempt to decode tool_input as a JSON

meetwudi:
We currently evaluate tool_input as a Python object. However, self._validate_tool_input may return a JSON string instead. This causes an error when tool_input contains lowercase true/false, which are valid JSON values but not valid Python (True/False).

As a result, the agent could enter an infinite loop if tool_input keeps resolving to a JSON string like:

```
{"timeMin": "2024-08-12T00:00:00Z", "timeMax": "2024-08-13T00:00:00Z", "single_events": true, "order_by": "startTime", "calendar_id": "primary"}
```

This commit addresses the issue by ensuring that tool_input is correctly parsed as JSON when necessary, preventing the infinite loop.

### DIFF
--- a/src/crewai/tools/tool_usage.py
+++ b/src/crewai/tools/tool_usage.py
@@ -1,4 +1,5 @@
 import ast
+import json
 from difflib import SequenceMatcher
 import os
 from textwrap import dedent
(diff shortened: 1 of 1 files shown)


pythonbyte (commented):
Nice work @meetwudi 
Can you provide a test case for this issue? 
And fix the conflict?

pythonbyte:
Hey @meetwudi! Could you take a look to see if this still applies? If there are any conflicts, it would be great if you could help resolve them. Also, if you could provide a use case for testing, that would be super helpful. Thanks.

PR #1187, opened 2024-08-14
Clean up pipeline

bhancockio:
Issues fixed:
- I couldn't get the newly created crews and pipeline templates to run. I fixed the path routing to properly point at the config.
- When you would run a newly created template it would run but the LLM wouldn't do anything. The underlying issue was that the `.env` file had the following key `OPENAI_API_KEY=OPENAI_API_KEY`. It was actually using OPENAI_API_KEY as the value which would fail. However, the crew wasn't reporting the error. More info here: https://www.loom.com/share/51228e13cc7d47279875862f87f09f67?sid=4ff78592-f58b-4a91-a297-84da81efac46
- Made the template versions dynamic instead of hard coded to v0.51.1.

Other improvements
- Added get started docs for pipelines.

How to verify changes work:
- create a crew, a pipeline, and a pipeline with a router using the cli (`crewai create <crew|pipeline> ...`). Install dependencies and update the .env file with OPENAI_API_KEY. The run with `crewai run`


docs/getting-started/Create-a-New-CrewAI-Pipeline-Template-Method.md
```diff
@@ -0,0 +1,136 @@
...
+or
+
+```shell
+$ poetry run <project_name>
+```
+
+This will initialize your pipeline and begin task execution as defined in your `main.py` file.
+
+## Deploying Your Pipeline Project
+
```
lorenzejay, line 133: very nice

src/crewai/agents/executor.py
```diff
@@ -213,11 +209,7 @@ def _iter_next_step(
                         yield step
                 return
 
-            yield AgentStep(
-                action=AgentAction("_Exception", str(e), str(e)),
-                observation=str(e),
-            )
-            return
+            raise e
```
lorenzejay, line 212: why are we removing this AgentAction?

lorenzejay (changes requested):

src/crewai/agents/executor.py
lorenzejay: This just to kill the process if this happens ?

lorenzejay (approved):

PR #1189, opened 2024-08-14
fix: Fix planning_llm issue

pythonbyte:


lorenzejay (approved):
works Ollama for me. certain models are limited when it comes to returning structured_outputs

PR #1199, opened 2024-08-15
[CRE-170] Add name and expected_output to TaskOutput

vinibrsl:
This commit adds task information to the TaskOutput class. This is useful to provide extra context to callbacks.

### DIFF
--- a/src/crewai/task.py
+++ b/src/crewai/task.py
@@ -240,7 +240,9 @@ class Task(BaseModel):
         pydantic_output, json_output = self._export_output(result)
 
         task_output = TaskOutput(
+            name=self.name,
             description=self.description,
+            expected_output=self.expected_output,
             raw=result,
             pydantic=pydantic_output,
             json_dict=json_output,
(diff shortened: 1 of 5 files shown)


src/crewai/task.py
```diff
@@ -240,7 +240,9 @@ def _execute_core(
         pydantic_output, json_output = self._export_output(result)
 
         task_output = TaskOutput(
+            name=self.name,
```
pythonbyte, line 243: _Suggestion_: Maybe it makes sense to add to the task's constructor or in the Annotations a way to "add" the name for the Task, maybe fetch the functions_name and set it as the Task name.

pythonbyte (approved):
Aside from the 1 suggestion
LGTM ✅ 
Nice first PR @vinibrsl

src/crewai/task.py
vinibrsl: Good call! Added in 9f27e5cb314e08f762a84c4cc86db821e5399f1f :)

PR #1205, opened 2024-08-17
Update stale.yml

theCyberTech:
Added  
operations-per-run: 500

The core issue here is that the default for  `operations-per-run` is set to 30 meaning the first 30 issues get processed only.

We can set it to 500 to cover all issues and once they have all been processed we can revert back to 30 as there will be no need for a high number of 'operations-per-run`


pythonbyte (approved):
Awesome work 

LGTM ✅

PR #1208, opened 2024-08-17
Bug in the documentation

klojtas:
change from crewai install to poetry install


pythonbyte:
Thanks for the PR @klojtas but the actual CLI command to run is crewai install.
The issue already fixed is that we deployed our docs before cutting a new release.

klojtas:
Thx

PR #1211, opened 2024-08-19
Brandon/cre 211 fix agent and task config for yaml based projects

bhancockio:
Walk through of problem and fix:

https://www.loom.com/share/dfe23ef75a2d40268b1c26885be47e97?sid=bb6811ec-ad70-4ca6-8b77-d50cf4d15f30

### DIFF
--- a/src/crewai/agent.py
+++ b/src/crewai/agent.py
@@ -18,6 +18,7 @@ from crewai.utilities import Converter, Prompts
 from crewai.utilities.constants import TRAINED_AGENTS_DATA_FILE, TRAINING_DATA_FILE
 from crewai.utilities.token_counter_callback import TokenCalcHandler
 from crewai.utilities.training_handler import CrewTrainingHandler
+from crewai.utilities.validation import assert_not_none
 
 
 def mock_agent_ops_provider():
@@ -116,7 +117,7 @@ class Agent(BaseAgent):
     @model_validator(mode="after")
     def post_init_setup(self):
         # Set agent_ops_agent_name
-        self.agent_ops_agent_name = self.role
+        self.agent_ops_agent_name = assert_not_none(self.role)
 
         # Set up LLM callbacks
         if hasattr(self.llm, "model_name"):
@@ -291,9 +292,9 @@ class Agent(BaseAgent):
         ).task_execution()
 
         execution_prompt = prompt.partial(
-            goal=self.goal,
-            role=self.role,
-            backstory=self.backstory,
+            goal=assert_not_none(self.goal),
+            role=assert_not_none(self.role),
+            backstory=assert_not_none(self.backstory),
         )
 
         stop_words = [self.i18n.slice("observation")]
--- a/src/crewai/agent.py
+++ b/src/crewai/agent.py
@@ -115,14 +115,11 @@ class Agent(BaseAgent):
 
     @model_validator(mode="after")
     def post_init_setup(self):
-        # Set agent_ops_agent_name
         self.agent_ops_agent_name = self.role
 
-        # Set up LLM callbacks
         if hasattr(self.llm, "model_name"):
             self._setup_llm_callbacks()
 
-        # Set up agent executor
         if not self.agent_executor:
             self._setup_agent_executor()
 
--- a/src/crewai/agent.py
+++ b/src/crewai/agent.py
@@ -18,7 +18,6 @@ from crewai.utilities import Converter, Prompts
 from crewai.utilities.constants import TRAINED_AGENTS_DATA_FILE, TRAINING_DATA_FILE
 from crewai.utilities.token_counter_callback import TokenCalcHandler
 from crewai.utilities.training_handler import CrewTrainingHandler
-from crewai.utilities.validation import assert_not_none
 
 
 def mock_agent_ops_provider():
@@ -117,7 +116,7 @@ class Agent(BaseAgent):
     @model_validator(mode="after")
     def post_init_setup(self):
         # Set agent_ops_agent_name
-        self.agent_ops_agent_name = assert_not_none(self.role)
+        self.agent_ops_agent_name = self.role
 
         # Set up LLM callbacks
         if hasattr(self.llm, "model_name"):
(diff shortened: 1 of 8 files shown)


src/crewai/agent.py
```diff
@@ -114,39 +114,42 @@ class Agent(BaseAgent):
...
-        """Ensure agent executor and token process are set."""
+        # Set up LLM callbacks
         if hasattr(self.llm, "model_name"):
-            token_handler = TokenCalcHandler(self.llm.model_name, self._token_process)
+            self._setup_llm_callbacks()
 
-            # Ensure self.llm.callbacks is a list
-            if not isinstance(self.llm.callbacks, list):
-                self.llm.callbacks = []
+        # Set up agent executor
```
gvieira: Just a small note: The code is pretty self-explanatory. In such cases, I try to avoid commenting. What do you think?

bhancockio: Agreed! Will remove!

vinibrsl (approved):

src/crewai/agent.py
```diff
@@ -114,39 +114,42 @@ class Agent(BaseAgent):
...
+    def post_init_setup(self):
+        # Set agent_ops_agent_name
         self.agent_ops_agent_name = self.role
-        return self
 
-    @model_validator(mode="after")
-    def set_agent_executor(self) -> "Agent":
-        """Ensure agent executor and token process are set."""
+        # Set up LLM callbacks
         if hasattr(self.llm, "model_name"):
```
gvieira, line 120: We do a bunch of things in `_setup_llm_callbacks` that do not require a `model_name` to be present in `model`. Should we move this check in?

src/crewai/agent.py
```diff
@@ -114,39 +114,39 @@ class Agent(BaseAgent):
...
-                self.llm.callbacks.append(agentops.LangchainCallbackHandler())
+    def _setup_llm_callbacks(self):
+        token_handler = TokenCalcHandler(self.llm.model_name, self._token_process)
 
-        if not self.agent_executor:
-            if not self.cache_handler:
-                self.cache_handler = CacheHandler()
-            self.set_cache_handler(self.cache_handler)
-        return self
+        if not isinstance(self.llm.callbacks, list):
```
gvieira, line 131: I wonder what's the case we are trying to prevent here. In other words, what's the scenario on which `callbacks` is not a list?

src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -127,11 +127,46 @@ class BaseAgent(ABC, BaseModel):
...
 
+    @model_validator(mode="before")
+    @classmethod
+    def process_config(cls, values):
+        config = values.get("config", {})
+        if not config:
+            return values
+
+        for key, value in config.items():
+            if key in cls.model_fields:
```
gvieira: How do you feel about early returns? In this case, early continue? I like it to prevent block nesting.

src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -127,11 +127,46 @@ class BaseAgent(ABC, BaseModel):
...
+                if key not in values or values[key] is None:
```
gvieira:
```suggestion
                if values.get(key) is None:
```

src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -127,11 +127,46 @@ class BaseAgent(ABC, BaseModel):
...
     )
 
+    @model_validator(mode="before")
+    @classmethod
+    def process_config(cls, values):
+        config = values.get("config", {})
+        if not config:
+            return values
+
+        for key, value in config.items():
```
gvieira: Opposite to what I said above, I think we need a comment here. 😄 I don't know what we are doing here. lol

src/crewai/task.py
```diff
@@ -53,9 +53,9 @@ class Task(BaseModel):
     i18n: I18N = I18N()
     name: Optional[str] = Field(default=None)
     prompt_context: Optional[str] = None
-    description: str = Field(description="Description of the actual task.")
-    expected_output: str = Field(
-        description="Clear definition of expected output for the task."
+    description: str = Field(description="Description of the actual task."
+    )
```
gvieira: Should these `)` be up?

src/crewai/task.py
```diff
@@ -115,6 +115,33 @@ class Task(BaseModel):
     _thread: Optional[threading.Thread] = PrivateAttr(default=None)
     _execution_time: Optional[float] = PrivateAttr(default=None)
 
+    @model_validator(mode="before")
+    @classmethod
+    def process_config_and_validate(cls, values):
+        config = values.get("config", {})
+
+        # Process config
+        for key, value in config.items():
```
gvieira: Hm, pretty similar to the config from above, right!? Should we centralize it somehow?

src/crewai/agent.py
bhancockio: Hey! Good question.

Under the hood, we need the model name because tiktoken uses that model name to calculate tokens. 

If we don't have the model name, we can't calculate tokens.

src/crewai/agent.py
bhancockio: I honestly do not know. Because the LLM class in our `BaseAgent` class is any it's hard to do any type-safety or know what's going to get passed in.

I'm happy to remove the check but it was already in place so I'm guessing someone put it there on purpose originally.

src/crewai/agents/agent_builder/base_agent.py
bhancockio: Added comment!

src/crewai/agents/agent_builder/base_agent.py
bhancockio: Refactored for early breaks with continues!

src/crewai/agents/agent_builder/base_agent.py
bhancockio: Added!

src/crewai/task.py
bhancockio: I moved up description but my formatter keeps moving the rest down.

src/crewai/task.py
bhancockio: I'll move both to a utils file.

src/crewai/agent.py
gvieira: Right, but what about that thing that makes sure `callbacks` is a list, just to name one?

src/crewai/agent.py
gvieira: What do you think about creating some quick-and-dirty tasks on Linear, so we can have such discussions and not block the PR? I'm sure João would have an answer for that, but let's leave it to later.

bhancockio: Created!

https://linear.app/crewai/issue/CRE-213/llmcallbacks-list-check

PR #1212, opened 2024-08-19
feat: Add only on release to deploy docs

pythonbyte:
This PR intends to change the way we deploy docs, instead of right after merging to main is now when a new release happens


theCyberTech (approved):
Looks Good

PR #1217, opened 2024-08-20
Fixed version conflict for regex package due to full year in version

### LINKED ISSUE #1004: regex package conflict (outdated) - can't use MarkerPDF
You require:

`^2023.12.25`

But this doesn't match the latest versions which start with `2024`

Easy fix would be to update the package - or to specify:

`>=2023.12.25`

I tried to use MarkerPDF in my project:

```
Because no versions of crewai match >0.41.1,<0.42.0
 and crewai[tools] (0.41.1) depends on regex (>=2023.12.25,<2024.0.0), crewai[tools] (>=0.41.1,<0.42.0) requires regex (>=2023.12.25,<2024.0.0).
And because marker-pdf (0.2.16) depends on regex (>=2024.4.28,<2025.0.0)
 and no versions of marker-pdf match >0.2.16,<0.3.0, crewai[tools] (>=0.41.1,<0.42.0) is incompatible with marker-pdf (>=0.2.16,<0.3.0).
So, because langflow-base depends on both marker-pdf (^0.2.16) and crewai[tools] (^0.41.1), version solving failed.
```

matthias:
As a result it will be compatible with installations / projects using a 2024.x.x version of the regex package

Fixes #1004

p.s.
Alternativly the project maintainers could decide to update the package to a 2024.x.x version
=> https://pypi.org/project/regex/2024.4.16/#history

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ opentelemetry-api = "^1.22.0"
 opentelemetry-sdk = "^1.22.0"
 opentelemetry-exporter-otlp-proto-http = "^1.22.0"
 instructor = "1.3.3"
-regex = "^2023.12.25"
+regex = ">=2023.12.25"
 crewai-tools = { version = "^0.8.3", optional = true }
 click = "^8.1.7"
 python-dotenv = "^1.0.0"


joaomdmoura:
Thanks @matthias ! Just merged another PR fixing this! Great catch!

PR #1228, opened 2024-08-20
Fixed version conflict for regex package due to full year in version

bhancockio:
This PR addresses the version conflict issue related to the regex package when using a version formatted as 2024.x.x, which was initially brought up by @matthias in PR #1217.

Changes:
- Updated the pyproject.toml to include the correct versioning format for the regex package.
- Ensured compatibility with installations and projects using a 2024.x.x version of regex.
- Verified that all dependencies align with the updated version format.

This resolves issue #1004, ensuring that our project remains compatible with the regex package versioning going forward.

Closes PR #1217


pythonbyte (approved):
Approved ✅

PR #1240, opened 2024-08-22
Feat/cli deploy

pythonbyte:
This PR intends to add the brand new CrewaAI+ CLI deploy command that will allow Users to use the platform easily by the terminal

https://linear.app/crewai/issue/CRE-159/crewai-deploy-cli


src/crewai/cli/authentication/utils.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        key_filename = "secret.key"
+        key = self.read_secure_file(key_filename)
+
+        if key is not None:
+            return key
+        else:
+            new_key = Fernet.generate_key()
+            self.save_secure_file(key_filename, new_key)
+            return new_key
```
gvieira: No need to change, but I usually omit the `else` in such cases.

src/crewai/cli/authentication/utils.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        if sys.platform == "win32":
+            # Windows: Use %LOCALAPPDATA%
+            base_path = os.environ.get("LOCALAPPDATA")
+        elif sys.platform == "darwin":
+            # macOS: Use ~/Library/Application Support
+            base_path = os.path.expanduser("~/Library/Application Support")
+        else:
+            # Linux and other Unix-like: Use ~/.local/share
+            base_path = os.path.expanduser("~/.local/share")
```
gvieira, line 105: We should probably have some centralized place to get base-paths based on the platform. Of course, no need to do it here, but please add a Low-hanging fruit [here](https://linear.app/crewai/view/e1827ba2-0757-46ca-89d7-3d26105e8dce).

src/crewai/cli/cli.py
```diff
@@ -179,5 +183,64 @@ def run():
     run_crew()
 
 
+@crewai.command()
+def signup():
```
gvieira, line 187: It's not that clear that you should call `signup` in order to login. Should we have a dedicated command or an alias?

src/crewai/cli/deploy/api.py
```diff
@@ -0,0 +1,63 @@
...
+            "Authorization": f"Bearer {api_key}",
+            "Content-Type": "application/json",
+        }
+        self.base_url = getenv(
+            "CREWAI_BASE_URL", "https://dev.crewai.com/crewai_plus/api/v1/crews"
+        )
+
+    def _make_request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
+        url = f"{self.base_url}/{endpoint}"
+        return requests.request(method, url, headers=self.headers, **kwargs)
```
gvieira, line 23: Also no need to change: I've been using Session Objects in such cases. I love how I can setup the object then simply call `get`, `post`, and so on without passing auth headers, url and other stuff.

src/crewai/cli/deploy/main.py
```diff
@@ -0,0 +1,289 @@
...
+        Deploy a crew using either UUID or project name.
+
+        Args:
+            uuid (Optional[str]): The UUID of the crew to deploy.
+        """
+        console.print("Starting deployment...", style="bold blue")
+        if uuid:
+            response = self.client.deploy_by_uuid(uuid)
+        elif self.project_name:
+            response = self.client.deploy_by_name(self.project_name)
```
gvieira, line 97: What do you think about encapsulating the response logic inside the client? This way `deploy_by_name` and others would either return a known object, like `DeployResponse` or throw an error. I like it, because the generic json parsing, status code handling and so on stays in a single place. If you like it, we can have a Low-hanging fruit as well.

src/crewai/cli/deploy/main.py
```diff
@@ -0,0 +1,289 @@
...
+    ) -> Dict[str, Any]:
+        """
+        Create the payload for crew creation.
+
+        Args:
+            remote_repo_url (str): Remote repository URL.
+            env_vars (Dict[str, str]): Environment variables.
+
+        Returns:
+            Dict[str, Any]: The payload for crew creation.
```
gvieira, line 151: Beautiful!

src/crewai/cli/deploy/utils.py
```diff
@@ -0,0 +1,116 @@
...
+    except KeyError:
+        print(f"Error: {pyproject_path} is not a valid pyproject.toml file.")
+    except tomllib.TOMLDecodeError:
+        print(f"Error: {pyproject_path} is not a valid TOML file.")
+    except Exception as e:
+        print(f"Error reading the pyproject.toml file: {e}")
+
+    return None
+
+    def get_crewai_version(pyproject_path: str = "pyproject.toml") -> str:
```
gvieira: Is this indent correct?

gvieira (approved):
Awesome! Just a few notes, but looks great!

src/crewai/cli/authentication/utils.py
pythonbyte: Yeah makes a lot of sense, I always do that, I forgot here 🤦🏽 
Thanks

src/crewai/cli/authentication/utils.py
pythonbyte: Sure thing 🫡

src/crewai/cli/cli.py
pythonbyte: Yeah, I'll add the same function but with the def login just in case

src/crewai/cli/deploy/api.py
pythonbyte: Awesome, I'll take into account for next iterations 🫡

src/crewai/cli/deploy/utils.py
pythonbyte: Awesome catch. It was the function that I added for the v1 to send crewai version on the header...thanks

src/crewai/cli/deploy/main.py
pythonbyte: I'll setup a low-hanging fruit to look for the v1. But totally agree with that, makes much more sense to have a pre-defined object that we can use and the already available status_code and json content.

PR #1253, opened 2024-08-23
Fix deployment name issue to support Azure

bhancockio:
https://www.loom.com/share/a5f2c1e1a1b642a5aae54badda7abdb6

### DIFF
--- a/src/crewai/agent.py
+++ b/src/crewai/agent.py
@@ -117,16 +117,19 @@ class Agent(BaseAgent):
     def post_init_setup(self):
         self.agent_ops_agent_name = self.role
 
-        if hasattr(self.llm, "model_name"):
-            self._setup_llm_callbacks()
+        # Different llms store the model name in different attributes
+        model_name = self.llm.model_name or self.llm.deployment_name
+
+        if model_name:
+            self._setup_llm_callbacks(model_name)
 
         if not self.agent_executor:
             self._setup_agent_executor()
 
         return self
 
-    def _setup_llm_callbacks(self):
-        token_handler = TokenCalcHandler(self.llm.model_name, self._token_process)
+    def _setup_llm_callbacks(self, model_name: str):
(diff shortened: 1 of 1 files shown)


pythonbyte (approved):
Nice work @bhancockio 

LGTM ✅

src/crewai/agent.py
```diff
@@ -117,16 +117,19 @@ class Agent(BaseAgent):
     def post_init_setup(self):
         self.agent_ops_agent_name = self.role
 
-        if hasattr(self.llm, "model_name"):
-            self._setup_llm_callbacks()
+        # Different llms store the model name in different attributes
+        model_name = self.llm.model_name or self.llm.deployment_name
```
gvieira: Won't this break when calling `model_name` with `AttributeError`?

bhancockio: The type of LLM is `any` so I don't think we are going to run into any issues. 

I just tried changing out the LLM to be a basic OpenAI and the azurechatopenai and didn't get any issues.

gvieira: Try to pass any other object there, like a `dict` or a random `BaseModel`. I think it's gonna break but I might be wrong.

bhancockio: @gvieira I incorporated your suggestions. Now, we run into another issue when someone passes in an LLM as `{}`. 

I walk through the issue in this loom:

https://www.loom.com/share/527fd6b1e25e4c04a1f8a4910c65cde7?sid=f2dc4d87-6414-4dc2-8aed-7a35784301cf

gvieira: Sweet. Best solution for this is to introduce a protocol/interface, so we limit what we accept. For now, I would say what you've done is good enough, as long as we create one of those [low-hanging fruit tickets](https://linear.app/crewai/view/e1827ba2-0757-46ca-89d7-3d26105e8dce).

bhancockio: Added!

PR #1255, opened 2024-08-25
correncting command to install dependencies

evandrouzeda:
I try to use "crewai install" command but I got an error.
After some research I find out the command was wrong, so the correct command is *poetry*


theCyberTech:
this is a valid command that will be available in the next release

PR #1257, opened 2024-08-25
Update README.md

Astha0024:
Added the default model used if LLM is not specified


theCyberTech (changes requested):
the default should be GPT4o, if you can update I will approve

Astha0024:
Ok thanks, I created PR #1273

PR #1269, opened 2024-08-29
Add py 3.10 support back to CLI + fixes

thiagomoretto:
* Fallback to simple parser to get project name when running Python 3.10.
* Get crewai version from poetry.lock.
* Send locked crewai version as User-Agent header.


pythonbyte (approved):
LGTM ✅ 
Nice work man

PR #1283, opened 2024-09-03
Added a new feature to log the final thoughts of each agent 🚀

shiladityasarkar:
A really helpful feature to have is that we know what each agent produced as their final answer (**not** the entire chain of thought) which was fed to the next agent in line in a sequential system.
For example, one agent `X` writes code and another agent `Y` explains that code. Then instead of having just the final explanation, it would be nice to retrieve what code was actually written by `X`. See my project on this [here](https://github.com/shiladityasarkar/LeetCoder).

Screenshot (325)

I have added a class variable called `logs` in the `Crew` class. Every time `_execute_tasks()` is called, `logs` is updated (overwritten) with the latest thoughts of the current agent. The idea is - for each `task_index` the last update will always be the agent's final thought - which is the final answer (see `translations/en.json` for more clarity).

Thus, after `kickoff()` completes execution, `logs` has the final answer of each intermediate agent stored as a dictionary with the integer keys denoting the order of execution of tasks starting from `0`.

If somebody wants just the raw output and not the entire mess of TaskOutput, then just write -

`logs[<task_number>][0].raw`

Hope this helps 😀

### DIFF
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -101,13 +101,14 @@ class Crew(BaseModel):
     _task_output_handler: TaskOutputStorageHandler = PrivateAttr(
         default_factory=TaskOutputStorageHandler
     )
-
+    
     name: Optional[str] = Field(default=None)
     cache: bool = Field(default=True)
     tasks: List[Task] = Field(default_factory=list)
     agents: List[BaseAgent] = Field(default_factory=list)
     process: Process = Field(default=Process.sequential)
     verbose: bool = Field(default=False)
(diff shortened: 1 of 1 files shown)


joaomdmoura:
Hey @shiladityasarkar at the end of the crew execution you should be able to get the final result and call `.tasks_output` on it to get the results of each individual task, isn't that exactly what you need? I might be missing something, sorry if that is the case

pythonbyte:
Hi @shiladityasarkar,

Could you please take a moment to see if this makes sense in light of @joaomdmoura’s comments? 
I’d appreciate it if you could also share a test or use case for it. 

Thanks

PR #1320, opened 2024-09-12
Added is_auto_end flag in agentops.end session in crew.py

anmol-aidora:
When using agentops, we have the option to pass the `skip_auto_end_session` parameter, which is supposed to not end the session if the `end_session` function is called by Crew.

Now the way it works is, the `agentops.end_session` accepts `is_auto_end` flag and crewai should have passed it as `True` (its `False` by default). 

I have changed the code to pass is_auto_end=True


anmol-aidora:
I ran `poetry run pytest`, it requires `OPENAI_API_KEY`, these are the errors. The changes I have made aren't causing error.

Is `OPENAI_API_KEY` not set in the checks?

bhancockio:
Hey @anmol-aidora !

Could you please elaborate more on `is_auto_end`? I'm not fully understanding what's going wrong and how changing `is_auto_end` to True fixes the issue.

If you could share any examples showing this fix working and solving the issue, that would be great!

anmol-aidora:
> Hey @anmol-aidora !
> 
> Could you please elaborate more on `is_auto_end`? I'm not fully understanding what's going wrong and how changing `is_auto_end` to True fixes the issue.
> 
> If you could share any examples showing this fix working and solving the issue, that would be great!

Hi @bhancockio 

I am attaching screenshot to the code in agentops for the agentops.end_session function:
image
Here is the link to the code: https://github.com/AgentOps-AI/agentops/blob/d04622ac4f91ec024ec76e1eb047b7846d744be1/agentops/client.py#L272

It ensures that the session does not end if the `skip_auto_end_session` value is set to `True` in `agentops` config.

As we are doing automatic use of the `end_session` function in our code, we should set it to `True`.

anmol-aidora:
I have prepared before/after code change scenario files. Please check them out:
[crewai_before_bug_fix.pdf](https://github.com/user-attachments/files/17434181/crewai_before_bug_fix.pdf)
[crewai_after_bug_fix.pdf](https://github.com/user-attachments/files/17434183/crewai_after_bug_fix.pdf)

bhancockio (approved):
Thank you @anmol-aidora for working on this and providing the before and after logs!

Super helpful!

PR #1334, opened 2024-09-17
Rework AgentOps implementation for new agent executor

bboynton97:
Imports LiteLLM in a way that allows for AgentOps to instrument the package

This implementation is significantly cleaner than the Langchain method


src/crewai/tools/tool_usage.py
```diff
@@ -11,11 +11,10 @@
 from crewai.utilities import I18N, Converter, ConverterError, Printer
 
 agentops = None
-if os.environ.get("AGENTOPS_API_KEY"):
```
joaomdmoura, line 14: shouldn't we keep this? given that if the person haven't set up a key we probably don't was to try import the library anyway?

joaomdmoura (changes requested):

bhancockio:
After implementing lazy loading for litellm, this PR is no longer needed.

PR #1341, opened 2024-09-23
Fix encoding issue when loading i18n json file

guojingfeng1994:
When I use the custom i18n json file path, I get a gbk encoding error.

```
    crew = Crew(
        agents=[researcher, writer],
        tasks=[task],
        process=Process.hierarchical,
        manager_llm="gpt-4o",
        verbose=True,
        prompt_file="./translation/zh-CN.json"
    )

    crew.kickoff()
```


joaomdmoura:
Good catch!

PR #1348, opened 2024-09-25
Create client for Tools API

vinibrsl:
This commit creates a class for the new Tools API. It extracts common methods from crewai.cli.deploy.api.CrewAPI to a parent class.


thiagomoretto (approved):
Good job here!! :100:

PR #1349, opened 2024-09-25
[Docs] Point footer socials to crewAIInc accounts

zinyando:
*Nit picky* - Point footer socials to `crewAIInc` accounts instead of `joaomdmoura` 😅 

Feel free to close.


joaomdmoura (approved):
Love it, good catch

PR #1357, opened 2024-09-26
CLI for Tool Repository

vinibrsl:
This PR adds two commands:

- `crewai tool publish`
   - Builds the project using Poetry
   - Uploads the tarball to CrewAI's tool repository
- `crewai tool install my-tool`
   - Adds `my-tool`'s index to Poetry and its credentials
   - Installs `my-tool` from the custom index


src/crewai/cli/deploy/main.py
vinibrsl, line 1: I extracted common behavior in this class to `BaseCommand` and `PlusAPIMixin`, so I could use it in `ToolCommand`. DRY!

src/crewai/cli/tools/api.py
```diff
@@ -11,14 +11,14 @@ def get(self, handle: str):
     def publish(
         self,
         handle: str,
-        public: bool,
+        is_public: bool,
```
vinibrsl: pyright complained about this being a reserved word (?), so I changed.

src/crewai/cli/deploy/main.py
```diff
@@ -23,40 +22,12 @@ def __init__(self):
...
 
-        self.project_name = get_project_name()
-        if self.project_name is None:
-            console.print(
-                "No project name found. Please ensure your project has a valid pyproject.toml file.",
-                style="bold red",
-            )
-            raise SystemExit
+        BaseCommand.__init__(self)
+        PlusAPIMixin.__init__(self, telemetry=self._telemetry)
```
gvieira, line 26: I wonder if the mixin shouldn't also create `_telemetry`.

src/crewai/cli/deploy/main.py
```diff
@@ -23,40 +22,12 @@ def __init__(self):
...
-
-        Args:
-            json_response (Dict[str, Any]): The JSON response containing error information.
-        """
-        error = json_response.get("error", "Unknown error")
-        message = json_response.get("message", "No message provided")
-        console.print(f"Error: {error}", style="bold red")
-        console.print(f"Message: {message}", style="bold red")
+        self.project_name = get_project_name(require=True)
+        self.crew_api_client = CrewAPI(api_key=self.access_token)
```
gvieira: Also, as the base command sets plus' access token, shouldn't it also create the client? Maybe even the project name?

src/crewai/cli/tools/api.py
gvieira: How I wish python had explicit public, private, and other visibility keywords.

src/crewai/cli/tools/main.py
```diff
@@ -0,0 +1,170 @@
...
+
+class ToolCommand(BaseCommand, PlusAPIMixin):
+    """
+    A class to handle tool repository related operations for CrewAI projects.
+    """
+
+    def __init__(self):
+        BaseCommand.__init__(self)
+        PlusAPIMixin.__init__(self, telemetry=self._telemetry)
+        self.tools_api_client = ToolsAPI(api_key=self.access_token)
```
gvieira: Now I see why you haven't added a default client to the mixin. I think it's a bit odd that we have a separate client in this case. I assumed it was pointing to a different application, but was confused when I saw it using the same access token and inheriting from the general client. What do you think about moving the code to the existing client?

src/crewai/cli/deploy/main.py
vinibrsl: I'm not 100% familiar on how our telemetry feature works, but I decided to instantiate it in `BaseCommand`, so any other mixin can use the same telemetry instance.

src/crewai/cli/tools/main.py
vinibrsl: Good call! Changed in 4e30923cabcf2ba489b617104c65a7ddc0a95817 & 1288c4a5df66d8a9056e699ffb9be13e2a46365f.

src/crewai/cli/deploy/main.py
gvieira: Awesome!

gvieira (approved):
Amazing work! 🚢

PR #1358, opened 2024-09-26
Fixed typing issues for new crews

bhancockio:
Screenshot 2024-09-26 at 12 35 07 PM

Before this fix, autocomplete and type hinting would not work on crews that we create with the .crew() approach. But now, it does and we can see the return values.


joaomdmoura (approved):

joaomdmoura:
Isn't this a problem that we are creating the crew object, what if they want ot pass a different process of specific crew values?
I think this also introduces a bug on the tasks ordering.
I'm playing with both locally just in case and might push something

PR #1360, opened 2024-09-26
Crew's key must remain stable after input interpolation

thiagomoretto:
A new Crew key was being given due to input interpolation. The key will now be stable regardless of the input.


src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -176,7 +176,11 @@ def set_private_attrs(self):
 
     @property
     def key(self):
-        source = [self.role, self.goal, self.backstory]
+        source = [
+            self._original_role or self.role,
```
lorenzejay, line 180: awesome work

lorenzejay (approved):

PR #1376, opened 2024-10-01
Update README 

tonykipkemboi:
1.  Updated all instances of `crewAI` to `CrewAI`
2. Updated the `Setting Your Crew` example with a full working example following the YAML config format. This one specifically should improve DevX and onboarding with a real working example end-to-end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,16 @@ To get started with CrewAI, follow these simple steps:
 
 ### 1. Installation
 
+Ensure you have Python >=3.10 <=3.13 installed on your system. CrewAI uses [Poetry](https://python-poetry.org/) for dependency management and package handling, offering a seamless setup and execution experience.
+
+First, if you haven't already, install Poetry:
+
+```bash
+pip install poetry
+```
+
+Then, install CrewAI:
+
 ```shell
 pip install crewai
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,83 +44,223 @@ To get started with CrewAI, follow these simple steps:
...
 
 ```shell
 pip install 'crewai[tools]'
 ```
+The command above installs the basic package and also adds extra components which require more dependencies to function.
+
+Lock the dependencies and install them by using the CLI command:
+
+```shell
+crewai install
```
theCyberTech: I think this needs to be below crewai create command??

tonykipkemboi: Good point. Totally missed that.

tonykipkemboi: Updated!

README.md
```diff
@@ -44,83 +44,224 @@ To get started with CrewAI, follow these simple steps:
...
+  role: >
+    {topic} Senior Data Researcher
+  goal: >
+    Uncover cutting-edge developments in {topic}
+  backstory: >
+    You're a seasoned researcher with a knack for uncovering the latest
+    developments in {topic}. Known for your ability to find the most relevant
+    information and present it in a clear and concise manner.
+  tools:
+    - SerperDevTool # This is a tool that allows the agent to search the web for information
```
bhancockio: We are declaring `SerperDevTool` here and in the crew.py.

I would probably drop this one and then only leave the SerperDevTool in the crew.py

tonykipkemboi: Updated.

PR #1378, opened 2024-10-01
Change Tool Repository authentication scope

vinibrsl:
This PR adds a new command for adding custom PyPI indexes credentials to the project. This was changed because credentials are now user-scoped instead of organization.

### DIFF
--- a/src/crewai/cli/plus_api.py
+++ b/src/crewai/cli/plus_api.py
@@ -27,6 +27,9 @@ class PlusAPI:
         url = urljoin(self.base_url, endpoint)
         return requests.request(method, url, headers=self.headers, **kwargs)
 
+    def login_to_tool_repository(self):
+        return self._make_request("POST", f"{self.TOOLS_RESOURCE}/login")
+
     def get_tool(self, handle: str):
         return self._make_request("GET", f"{self.TOOLS_RESOURCE}/{handle}")
 
--- a/src/crewai/cli/tools/main.py
+++ b/src/crewai/cli/tools/main.py
@@ -103,15 +103,27 @@ class ToolCommand(BaseCommand, PlusAPIMixin):
             )
             raise SystemExit
 
-        self._add_repository_to_poetry(get_response.json())
         self._add_package(get_response.json())
 
         console.print(f"Succesfully installed {handle}", style="bold green")
 
-    def _add_repository_to_poetry(self, tool_details):
-        repository_handle = f"crewai-{tool_details['repository']['handle']}"
-        repository_url = tool_details["repository"]["url"]
-        repository_credentials = tool_details["repository"]["credentials"]
+    def login(self):
+        login_response = self.plus_api_client.login_to_tool_repository()
+
+        if login_response.status_code != 200:
+            console.print(
+                "Failed to authenticate to the tool repository. Make sure you have the access to tools.", style="bold red"
+            )
+            raise SystemExit
+
+        login_response_json = login_response.json()
+        for repository in login_response_json["repositories"]:
+            self._add_repository_to_poetry(repository, login_response_json["credential"])
+
+        console.print(f"Succesfully authenticated to the tool repository.", style="bold green")
+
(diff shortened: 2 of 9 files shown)


src/crewai/cli/utils.py
```diff
@@ -162,29 +163,9 @@ def _get_nested_value(data: Dict[str, Any], keys: List[str]) -> Any:
...
-
-    except FileNotFoundError:
-        print(f"Error: {poetry_lock_path} not found.")
-    except Exception as e:
-        print(f"Error reading the poetry.lock file: {e}")
-
-    return "no-version-found"
+def get_crewai_version() -> str:
+    """Get the version number of CrewAI running the CLI"""
+    return importlib.metadata.version("crewai")
```
vinibrsl, line 168: From what I understand, this method was incorrectly returning the CrewAI version of the currently open project. The version string is intended to represent the CLI version in HTTP requests to Enterprise, not the version of the open project.

src/crewai/cli/tools/main.py
```diff
@@ -103,23 +103,40 @@ def install(self, handle: str):
...
-        repository_credentials = tool_details["repository"]["credentials"]
+    def login(self):
+        login_response = self.plus_api_client.login_to_tool_repository()
+
+        if login_response.status_code != 200:
+            console.print(
+                "Failed to authenticate to the tool repository. Make sure you have the access to tools.",
+                style="bold red",
+            )
+            raise SystemExit
```
thiagomoretto, line 103: Would be better to handle auth and general errors separately. Better feedback reduces the chance of users asking for support. Is that possible here?

src/crewai/cli/plus_api.py
```diff
@@ -27,6 +27,9 @@ def _make_request(self, method: str, endpoint: str, **kwargs) -> requests.Respon
         url = urljoin(self.base_url, endpoint)
         return requests.request(method, url, headers=self.headers, **kwargs)
 
+    def login_to_tool_repository(self):
```
gvieira, line 30: Out of curiosity, is the tool repo login credentials different from crewai-plus' credentials?

src/crewai/cli/tools/main.py
```diff
@@ -103,23 +103,40 @@ def install(self, handle: str):
...
+
+        login_response_json = login_response.json()
```
gvieira, line 105: One of the issues we had in yesterday's incident is related to the fact the response was not a json object. Should we consider this here and somehow print the body? cc @pythonbyte.

src/crewai/cli/plus_api.py
vinibrsl: The tool repo credential is a username and password pair used by Poetry to authenticate with our PyPI repository using HTTP Basic. Some differences:

- **Plus creds** are short lived and may expire, while **tool repo creds** never expire
- **Plus creds** can authenticate to any API endpoint, while **tool repo creds** only work for downloading tools
- **Plus creds** are stored in the CLI, while **tool repo creds** are stored and used by Poetry
- **Tool repo creds** can be retrieved using **Plus creds**

src/crewai/cli/tools/main.py
vinibrsl: I see! Do you mean using a custom exception?

vinibrsl: I'm not sure if I understand, but I like the idea of improving feedback. Could you please elaborate?

src/crewai/cli/plus_api.py
gvieira: Got it! Thanks! :)

src/crewai/cli/tools/main.py
vinibrsl: I created a method that checks if the response is OK, otherwise it logs the error. Let me know what you think: 61b786254bdbfd3f421649d79dbf0a4fa4657982

gvieira: Lovely! ❤️

gvieira (approved):

PR #1381, opened 2024-10-02
Add Git validations for publishing tools 

vinibrsl:
This commit prevents tools from being published if the underlying Git repository is not synced with origin.

### DIFF
--- a/src/crewai/cli/git.py
+++ b/src/crewai/cli/git.py
@@ -1,0 +1,80 @@
+import subprocess
+
+
+class Repository:
+    def __init__(self, path="."):
+        self.path = path
+
+        if not self.is_git_installed():
+            raise ValueError("Git is not installed or not found in your PATH.")
+
+        if not self.is_git_repo():
+            raise ValueError(f"{self.path} is not a Git repository.")
+
+        self.fetch()
+
+    def is_git_installed(self) -> bool:
+        """Check if Git is installed and available in the system."""
+        try:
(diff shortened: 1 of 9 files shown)


src/crewai/cli/git.py
```diff
@@ -0,0 +1,80 @@
...
+            if line.startswith("##") and ("ahead" in line or "behind" in line):
+                return True
+        return False
+
+    def is_synced(self) -> bool:
+        """Return True if the Git repository is fully synced with the remote, False otherwise."""
+        if self.has_uncommitted_changes() or self.is_ahead_or_behind():
+            return False
+        else:
+            return True
```
thiagomoretto, line 66: I like the expressiveness of a if...else.
it can be short tho. Your call.

```suggestion
        return not (self.has_uncommitted_changes() or self.is_ahead_or_behind())
```

to avoid the negation, perhaps invert the logic like `is_out_of_sync`. Pure juice of nitpicking on my side :)

thiagomoretto (approved):
Well done, @vinibrsl !
Wondering here if we could add this validation to crew deploy cmd as well. On a later PR.

src/crewai/cli/git.py
vinibrsl: Thanks! I was wrapping my head around this, and I decided to go with the verbose way. `not` combined with `or` crashes my inner agent 😆

PR #1382, opened 2024-10-02
Add tool usage events

thiagomoretto:
Add a small event structure to listen to tool usage events.
It uses decorators and can be expanded and used for pretty much anything.

```python
import crewai.utilities.events as events

...
@events.on(ToolUsageFinished)
def on_tool_use_finished(source: "ToolUsage", event: ToolUsageFinished):
    # Handle the event
    pass
```

Currently, a generic emitter is initialized once, but we can scope and create emitters per module/use case.


gvieira (approved):
Amazing work! It looks beautiful!

PR #1398, opened 2024-10-05
feat: Enhance crew create crew CLI command

theCyberTech:
This PR adds the ability to add API keys etc during crew creation when running ```crewai create crew```


joaomdmoura:
I like this @theCyberTech but should we ask what provider they want to use first and then based on that ask them to add the key? so it only asks to add once

bhancockio:
@theCyberTech Can we close this PR since you created #1430 ?

theCyberTech:
No longer required covered in #1430

PR #1411, opened 2024-10-09
Fixing context of tasks when running pipeline

### LINKED ISSUE #1410: [BUG] Context does not work except in following tasts
### Description

I noticed the agents ignore the context tasks given to them in the `context` field when run using pipeline.
The following code shows three tasks: "Say Giraffe", "Say Dolphin", and "Say both of the previous words". 
The last one is given the context of the first two.
When I run the following code, I get the answer: 

```
# Agent: A dad
## Final Answer:
Twho words
```
As if it doesn't understand what the previous words are.

### Steps to Reproduce

Run the following code:
```
from crewai import Crew, Process, Task, Agent, Pipeline
import asyncio

dad_agent = Agent(
    role="A dad",
    goal="Do as your kid commands",
    backstory="You are a dad, you have to do as your kid commands",
    tools=[],
    verbose=True,
    llm="gpt-4o-mini",
)


task_1 = Task(
    description="say Giraffe", expected_output="A single word", agent=dad_agent
)
task_2 = Task(
    description="say Dolphin", expected_output="A single word", agent=dad_agent
)
task_3 = Task(
    description="Say both of the previous words",
    expected_output="Twho words",
    agent=dad_agent,
    context=[task_1, task_2],
)

# Instantiate your crew with a sequential process
dadCrew = Crew(
    agents=[dad_agent],
    tasks=[task_1, task_2, task_3],
    verbose=True,
    process=Process.sequential,
    output_log_file="crewai.log",
)


class MyPipeline:
    def __init__(self):
        self.dad_crew = dadCrew

    def create_pipeline(self):
        return Pipeline(
            stages=[
                self.dad_crew,
            ]
        )

    async def kickoff(self, inputs):
        pipeline = self.create_pipeline()
        results = await pipeline.kickoff(inputs)
        return results


def run():
    """
    Run the pipeline.
    """
    pipeline = MyPipeline()
    results = asyncio.run(pipeline.kickoff(inputs=[{}]))
    print(results)


if __name__ == "__main__":
    print("Running the crew without debug mode")
    run()
```

### Expected behavior

Expected final output: 

#Agent: A dad
##Final Answer:
Giraffe Dolphin

### Screenshots/Code snippets

The code above ^

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

0.65.2

### crewAI Tools Version

0.65.2

### Virtual Environment

Venv

### Evidence

image

### Possible Solution

Fix the way context is given to the tasks. Something doesn't work there

### Additional context

After debugging, It seens the context of the third task contains both of the tasks without their outputs after they ran.

Shahar-Y:
## Fixes #1410
I noticed the agents ignore the context tasks given to them in the `context` field. 
After thorough debugging I fund that the problem lays in the deep copy in `crew.py` (`def copy(self)` function) . 
The problem was that when running the pipeline, it first copies the crews and then runs the copied crews.
When deep-copying the crew, the tasks are being copied, but they create another instance of the task in the context. That instance's output is not updated when the real task completes and therefore the tasks in the context always have an empty output field.

To solve it, I compared the keys of the original copied tasks with the tasks in the context of each task, and  if they were the same, I replaced the context task with the original one. That way the context points to the task that is being run. It works perfectly now. 

## Proof it works: 
Now the example in #1410 gives the correct answer: 
image

### DIFF
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -902,6 +902,15 @@ class Crew(BaseModel):
         cloned_agents = [agent.copy() for agent in self.agents]
         cloned_tasks = [task.copy(cloned_agents) for task in self.tasks]
 
+        for cloned_task in cloned_tasks:
+            cloned_task.context = [
+                next(
+                    (reference for reference in cloned_tasks if reference.key == context_task.key),
+                    None
+                )
+                for context_task in cloned_task.context
+            ] if cloned_task.context else None
+
         copied_data = self.model_dump(exclude=exclude)
         copied_data = {k: v for k, v in copied_data.items() if v is not None}
 


bhancockio:
Thank you so much @Shahar-Y for identifying the issue and submitting a solution. Your work was really helpful in pinpointing the problem and validating the fix! 

I noticed a few additional linting issues and made some tweaks to the fix to align it with our coding standards. I’ll be closing out this PR in favor of this new PR #1416 , which incorporates your solution but with a slightly different approach.

Thanks again for your contribution, and I’ve credited you in the final PR! 🙌

PR #1416, opened 2024-10-09
fix task cloning error

bhancockio:
**Description**
This PR addresses the issue reported in #1410 by @Shahar-Y where the context tasks in a pipeline were not properly referenced, leading to incorrect outputs in subsequent tasks. I want to thank Shahar for identifying the root cause of this issue and for submitting an initial fix. Shahar's investigation revealed that when the pipeline runs, the tasks in the context were deep-copied incorrectly, leading to context outputs not updating properly.

**Changes**
Although Shahar's fix provided a solution, I noticed a few linting issues and decided to take a slightly different approach to maintain code quality and adhere to our style guidelines. The main changes I made include:

Refactored the context reference logic to ensure that all copied tasks maintain references to the original task outputs.
Resolved the linting issues and added additional comments to make the code easier to follow.

**Credit**
Huge thanks to @Shahar-Y for the excellent bug report and for providing a working solution. Your contribution was instrumental in resolving this issue!

**Testing**
The changes have been tested with the example provided in #1410, and the final output is now correctly displaying the concatenated words.

**Proof that it works**
Screenshot 2024-10-09 at 1 58 42 PM

### DIFF
--- a/src/crewai/crew.py
+++ b/src/crewai/crew.py
@@ -900,7 +900,22 @@ class Crew(BaseModel):
         }
 
         cloned_agents = [agent.copy() for agent in self.agents]
-        cloned_tasks = [task.copy(cloned_agents) for task in self.tasks]
+
+        task_mapping = {}
+
+        cloned_tasks = []
+        for task in self.tasks:
+            cloned_task = task.copy(cloned_agents, task_mapping)
+            cloned_tasks.append(cloned_task)
+            task_mapping[task.key] = cloned_task
+
+        for cloned_task, original_task in zip(cloned_tasks, self.tasks):
+            if original_task.context:
+                cloned_context = [
+                    task_mapping[context_task.key]
+                    for context_task in original_task.context
+                ]
+                cloned_task.context = cloned_context
 
         copied_data = self.model_dump(exclude=exclude)
         copied_data = {k: v for k, v in copied_data.items() if v is not None}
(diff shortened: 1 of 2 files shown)


pythonbyte (approved):
LGTM ✅

Shahar-Y:
Thank you for the credit :) 
I would also consider solving the root cause in the tasks' deep copy in the future. The solution I gave is a good patch but maybe in the task deep copy itself we should pass the previous tasks as parameters and reference them. that way we don't need to fix it outside the function. 
That solution was a bit more tricky and I didn't want to change too much in the code. 

BTW I love your platfrm, and would be happy to see the community around it thriving ❤️

src/crewai/crew.py
```diff
@@ -900,7 +900,22 @@ def copy(self):
...
+        task_mapping = {}
+
+        cloned_tasks = []
+        for task in self.tasks:
+            cloned_task = task.copy(cloned_agents, task_mapping)
+            cloned_tasks.append(cloned_task)
+            task_mapping[task.key] = cloned_task
+
+        for cloned_task, original_task in zip(cloned_tasks, self.tasks):
+            if original_task.context:
```
Shahar-Y, line 913: Iwould consider moving the second `for` loop into the task.copy() function, to solve the original problem and not just in `crew.copy()`.  
`task.copy()` already receives task mapping with all of the previous tasks already populated so it should have access to the reference of the the cloned tasks' context.

Shahar-Y (changes requested):

PR #1425, opened 2024-10-11
Update docs with `crewai update` for older versions using Poetry

tonykipkemboi:
- Explanation on usage
- Remove `Planning` from docs since we have `Flows` now


theCyberTech (approved):
Looks good to me

bhancockio:
@tonykipkemboi How do we need to change this PR to support UV now? Do we drop this since these docs were focused on Poetry?

tonykipkemboi:
I modified the portion that had poetry to match uv: README mostly and Flows and the quickstart

PR #1430, opened 2024-10-11
feat/updated CLI to allow for model selection & submitting API keys

theCyberTech:
This PR does the following

1)  enhance provider/model selection in create_crew command
2) Add support for additional API providers and their respective models
3) Enhance user prompts for selecting providers and models
4) Add necessary imports and update directory structure handling


src/crewai/cli/create_crew.py
```diff
@@ -21,22 +47,250 @@ def create_crew(name, parent_folder=None):
...
+                    env_vars[key_value[0]] = key_value[1]
+
+    # Caching setup
+    cache_dir = Path.home() / '.crewai'
+    cache_dir.mkdir(exist_ok=True)
+    cache_file = cache_dir / 'provider_cache.json'
+    cache_expiry = 24 * 3600 
+
+    # Load API providers and models from JSON with caching
+    json_url = "https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
```
bhancockio: @theCyberTech , do our users have to be connected to the internet to create a crew now?

src/crewai/cli/create_crew.py
```diff
@@ -21,22 +47,250 @@ def create_crew(name, parent_folder=None):
...
+                data_content = b''.join(data_chunks)
+                data = json.loads(data_content.decode('utf-8'))
+
+            # Save fetched data to cache
+            with open(cache_file, "w") as f:
+                json.dump(data, f)
+            click.secho("Provider data fetched and cached successfully.", fg="green")
+        except requests.RequestException as e:
+            click.secho(f"Error fetching provider data: {e}", fg="red")
+            return
```
bhancockio: These return statements look like if the fetching providers fails, we quite the creation process.

bhancockio:
Screenshot 2024-10-11 at 3 25 05 PM

The [1] confused me at first. I think you're trying to setup the CLI so if they don't click anything, they'll just get the default.

I think it might be best to drop the [1] and just make them pick a number.

src/crewai/cli/create_crew.py
```diff
@@ -21,22 +47,250 @@ def create_crew(name, parent_folder=None):
...
+            all_providers = sorted(provider_models.keys())
+            if not all_providers:
+                click.secho("No additional providers available.", fg="yellow")
+                return
+            click.secho("Select a provider from the full list:", fg="cyan")
+            for index, provider_name in enumerate(all_providers, start=1):
+                click.secho(f"{index}. {provider_name}", fg="cyan")
+            
+            while True:
+                try:
```
bhancockio: Do we need to put in a try except block? Can they break if we know the list range ?

src/crewai/cli/create_crew.py
```diff
@@ -21,22 +47,250 @@ def create_crew(name, parent_folder=None):
...
+                click.secho(f"Invalid provider: {provider}", fg="red")
+                return
+    else:
+        # Prompt for provider from predefined PROVIDERS and 'other'
+        click.secho("Select a provider to set up:", fg="cyan")
+        for index, provider_name in enumerate(PROVIDERS + ['other'], start=1):
+            click.secho(f"{index}. {provider_name}", fg="cyan")
+        
+        while True:
+            try:
```
bhancockio: Do we need a try catch here as well? Can we remove?

src/crewai/cli/create_crew.py
```diff
@@ -21,22 +47,250 @@ def create_crew(name, parent_folder=None):
...
+                        provider = all_providers[selected_index]  # Update provider to the selected one
+                        break
+                    else:
+                        click.secho("Invalid selection. Please try again.", fg="red")
+                except click.exceptions.Abort:
+                    click.secho("Operation aborted by the user.", fg="red")
+                    return
+        else:
+            # Validate provider
+            if provider_lower not in provider_models and provider_lower not in [p.lower() for p in PROVIDERS]:
```
bhancockio: It would be nice to do an early return statement. Basically, check for the not valid provider first. If it's not valid, give error message and return

That way you don't have to put everything in an if / else.

theCyberTech:
> Screenshot 2024-10-11 at 3 25 05 PM
> The [1] confused me at first. I think you're trying to setup the CLI so if they don't click anything, they'll just get the default.
> 
> I think it might be best to drop the [1] and just make them pick a number.

Fixed up, it now shows Enter the number of your choice: instead of Enter the number of your choice [1]:

bhancockio:
@theCyberTech Additional feedback on code.

src/crewai/cli/create_crew.py
theCyberTech: The first run of Crewai create crew command will download then cache the file, subsequent run will then use the cached file, I see the issue, will fix

bhancockio (approved):

Issue #1439, opened 2024-10-12
[BUG] ToolUsage not using same I18N as Agent


shankinson:
### Description

The ToolUsage class defaults to using I18N() and has no way to update it. It should use the same I18N as the agent so that custom prompt_file's work properly.

### Steps to Reproduce

1. Make a new prompt file (like translation/en.json) and edit the `tool_usage_exception` message.
2. Set your crew to use this new prompt_file.
2. Have an agent use a tool incorrectly.
3. Observe that the error message sent to the AI does not contain the update from 1.

### Expected behavior

The tool error message sent to the AI contains the updated wording from the prompt_file

### Screenshots/Code snippets

<img width="711" alt="image" src="https://github.com/user-attachments/assets/30cdbb04-2e46-43de-b84c-faa6e30ee52b">

I18N does not use a custom prompt_file.

<img width="611" alt="image" src="https://github.com/user-attachments/assets/aa4049b2-7c31-4628-9974-d7e6fc893788">

When ToolUsage is created, there is no way to set the prompt_file.

### Operating System

macOS Sonoma

### Python Version

3.10

### crewAI Version

0.70.1

### crewAI Tools Version

0.12.1

### Virtual Environment

Conda

### Evidence

Screenshots above show that the I18N never uses the prompt_file property set on the crew in ToolUsage and has no way of being set.

### Possible Solution

When the ToolUsage is created, have it set its i18n to the same as the agent. I will open a PR for this.

### Additional context

N/A

PR #1440, opened 2024-10-12
use the same i18n as the agent for tool usage

shankinson:
Fixes https://github.com/crewAIInc/crewAI/issues/1439


bhancockio:
Hey! I think I'm missing something. What is the actual error that's going wrong? How does your solution fix it?

shankinson:
Hey @bhancockio,

The ToolUsage class uses several `slices` from i18n:

https://github.com/crewAIInc/crewAI/blob/main/src/crewai/tools/tool_usage.py#L183
https://github.com/crewAIInc/crewAI/blob/main/src/crewai/tools/tool_usage.py#L256
https://github.com/crewAIInc/crewAI/blob/main/src/crewai/tools/tool_usage.py#L397

When ToolUsage is created, it always uses the default I18N:
https://github.com/crewAIInc/crewAI/blob/main/src/crewai/tools/tool_usage.py#L62

The only place it is created is here: https://github.com/crewAIInc/crewAI/blob/main/src/crewai/agents/crew_agent_executor.py#L227-L254

So even if the crew is using a custom i18n prompt file, the slices from the custom prompt file can never be used by ToolUsage for the `format` and `tools` slices. Only the default values from the default `translation/en.json` are ever used.

The change makes the ToolUsage class use the same i18n instance that is set on the agent that the crew_agent_executor is using, letting the slices be customized.

PR #1455, opened 2024-10-16
Adapt Tools CLI to uv

vinibrsl:
This pull request adapts commands from the Tool CLI from Poetry to uv.

Extra comments inline 👇

### DIFF
--- a/src/crewai/cli/plus_api.py
+++ b/src/crewai/cli/plus_api.py
@@ -25,7 +25,9 @@ class PlusAPI:
 
     def _make_request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
         url = urljoin(self.base_url, endpoint)
-        return requests.request(method, url, headers=self.headers, **kwargs)
+        session = requests.Session()
+        session.trust_env = False
+        return session.request(method, url, headers=self.headers, **kwargs)
 
     def login_to_tool_repository(self):
         return self._make_request("POST", f"{self.TOOLS_RESOURCE}/login")
--- a/src/crewai/cli/tools/main.py
+++ b/src/crewai/cli/tools/main.py
@@ -4,6 +4,8 @@ import platform
 import subprocess
 import tempfile
 from pathlib import Path
+from netrc import netrc
+import stat
 
 import click
 from rich.console import Console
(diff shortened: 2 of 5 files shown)


src/crewai/cli/plus_api.py
```diff
@@ -25,7 +25,9 @@ def __init__(self, api_key: str) -> None:
 
     def _make_request(self, method: str, endpoint: str, **kwargs) -> requests.Response:
         url = urljoin(self.base_url, endpoint)
-        return requests.request(method, url, headers=self.headers, **kwargs)
+        session = requests.Session()
+        session.trust_env = False
```
vinibrsl, line 29: `trust_env = False` tells `requests` to bypass the system's authentication or proxy settings, relying only on the authentication defined in the request itself.

The reason for this change is that when a `.netrc` file is present, `requests` automatically replaces the `Authorization` header with the credentials from `.netrc`, overwriting Auth0 credentials.

While `trust_env = False` fixes the issue, alongside I'm proposing moving our custom PyPI index endpoints from `app.crewai.com` to `tools.crewai.com`. This ensures `.netrc` is only applied to that specific subdomain system-wide.

src/crewai/cli/tools/main.py
```diff
@@ -159,23 +161,19 @@ def login(self):
             "Successfully authenticated to the tool repository.", style="bold green"
         )
 
-    def _set_netrc_credentials(self, credentials):
-        # Create .netrc or _netrc file
-        netrc_filename = "_netrc" if platform.system() == "Windows" else ".netrc"
-        netrc_path = Path.home() / netrc_filename
+    def _set_netrc_credentials(self, credentials, netrc_path=None):
```
vinibrsl, line 164: I refactored it to use Python's builtin `netrc` package.

src/crewai/cli/tools/main.py
```diff
@@ -159,23 +161,19 @@ def login(self):
...
+            netrc_filename = "_netrc" if platform.system() == "Windows" else ".netrc"
+            netrc_path = Path.home() / netrc_filename
+            netrc_path.touch(mode=stat.S_IRUSR | stat.S_IWUSR, exist_ok=True)
 
-        netrc_content = f"""machine app.crewai.com
-login {credentials['username']}
-password {credentials['password']}
-"""
+        netrc_instance = netrc(file=netrc_path)
+        netrc_instance.hosts["app.crewai.com"] = (credentials["username"], "", credentials["password"])
```
vinibrsl, line 171: This is the line we need to change after we migrate to `tools.crewai.com`.

src/crewai/cli/plus_api.py
gvieira: Your suggestion makes total sense.

gvieira (approved):

PR #1482, opened 2024-10-21
simplify flow

bhancockio:


src/crewai/flow/flow.py
```diff
@@ -190,7 +190,10 @@ def method_outputs(self) -> List[Any]:
         """Returns the list of all outputs from executed methods."""
         return self._method_outputs
 
-    async def kickoff(self) -> Any:
+    def kickoff(self) -> Any:
```
lorenzejay, line 193: awesome work! cleans this the template really well

lorenzejay (approved):

PR #1516, opened 2024-10-27
Security enhancements

theCyberTech:
This PR implements the following changes to the CrewAI codebase

1) Adds a Snyk file to .gitignore
2) Adds a new Security.MD file for the CrewAI project
3) Updates all instances of ```md5``` to use ```sha256``` instead


src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -181,7 +181,7 @@ def key(self):
             self._original_goal or self.goal,
             self._original_backstory or self.backstory,
         ]
-        return md5("|".join(source).encode(), usedforsecurity=False).hexdigest()
+        return sha256("|".join(source).encode()).hexdigest()
```
gvieira, line 184: I've used md5 here because it's not supposed to be protected data, but simply a hash of strings. The md5 algorithm is supposed to be fast as safe for this use case. Can we safely replace it with sha256 and achieve the same performance?

src/crewai/cli/constants.py
```diff
@@ -0,0 +1,19 @@
...
+
+MODELS = {
+    'openai': ['gpt-4', 'gpt-4o', 'gpt-4o-mini', 'o1-mini', 'o1-preview'],
+    'anthropic': ['claude-3-5-sonnet-20240620', 'claude-3-sonnet-20240229', 'claude-3-opus-20240229', 'claude-3-haiku-20240307'],
+    'gemini': ['gemini-1.5-flash', 'gemini-1.5-pro', 'gemini-gemma-2-9b-it', 'gemini-gemma-2-27b-it'],
+    'groq': ['llama-3.1-8b-instant', 'llama-3.1-70b-versatile', 'llama-3.1-405b-reasoning', 'gemma2-9b-it', 'gemma-7b-it'],
+    'ollama': ['llama3.1', 'mixtral'],
+}
+
+JSON_URL = "https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
```
gvieira, line 19: Mind setting up your editor to always add a new line upon saving?

PR #1560, opened 2024-11-05
Add missing @functools.wraps when wrapping functions and preserve wrapped class name in @CrewBase.

piotrm0:
Some functions did not have `@functools.wraps` meaning they did not preserve sig/name information resulting in less useful printouts. Similarly, the `@CrewBase` class decorator produces not useful class names. These two are fixed by this PR.


bhancockio:
Thanks for including these improvements @piotrm0!

PR #1567, opened 2024-11-07
Knowledge

bhancockio:


src/crewai/knowledge/knowledge.py
```diff
@@ -0,0 +1,53 @@
+from typing import List
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from crewai.knowledge.embedder.base_embedder import BaseEmbedder
+from crewai.knowledge.embedder.fastembed import FastEmbed
+from crewai.knowledge.source.base_knowledge_source import BaseKnowledgeSource
+
+
+class Knowledge(BaseModel):
```
lorenzejay, line 13: Lets write some docs about:
1. how to use this
2. setting your own custom embedder for this

src/crewai/knowledge/embedder/ollama.py
```diff
@@ -0,0 +1,82 @@
+import os
```
lorenzejay: i'd drop the ollama version. support openai, then let anyone bring their own embedder function (super easy) then have the knowledge_config setup like embedder_config setup for our rag storage

src/crewai/knowledge/source/base_knowledge_source.py
```diff
@@ -0,0 +1,39 @@
...
+        """Load and preprocess content from the source."""
+        pass
+
+    @abstractmethod
+    def add(self, embedder: BaseEmbedder) -> None:
+        """Process content, chunk it, compute embeddings, and save them."""
+        pass
+
+    def get_embeddings(self) -> List[np.ndarray]:
+        """Return the list of embeddings for the chunks."""
```
lorenzejay: lets make this save to the project directory instead of the root.

src/crewai/knowledge/source/csv_knowledge_source.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        Add CSV file content to the knowledge source, chunk it, compute embeddings,
+        and save the embeddings.
+        """
+        new_chunks = self._chunk_text(self.content)
+        self.chunks.extend(new_chunks)
+        # Compute embeddings for the new chunks
+        new_embeddings = embedder.embed_chunks(new_chunks)
+        # Save the embeddings
+        self.chunk_embeddings.extend(new_embeddings)
```
lorenzejay: We should also be saving this to a db and persist it. like our ragstorage

lorenzejay: We should do this as we can generate the embeddings for files once, then just query if they already exist. otherwise, users will spend tokens generating embeddings that already exist.

lorenzejay: I'll help with this.

src/crewai/knowledge/source/base_knowledge_source.py
```diff
@@ -0,0 +1,39 @@
+from abc import ABC, abstractmethod
+from typing import List
+
+import numpy as np
+from pydantic import BaseModel, ConfigDict, Field
+
+from crewai.knowledge.embedder.base_embedder import BaseEmbedder
+
```
lorenzejay, line 8: extending this to save embeddings to db, then using knowledge class to query from there

lorenzejay (changes requested):

src/crewai/agent.py
```diff
@@ -85,6 +88,10 @@ class Agent(BaseAgent):
     llm: Union[str, InstanceOf[LLM], Any] = Field(
         description="Language model that will run the agent.", default=None
     )
+    knowledge_sources: Optional[List[BaseKnowledgeSource]] = Field(
+        default=None,
+        description="Knowledge sources for the agent.",
```
lorenzejay: Would be better to declare this on the crew class. the task prompt will query from the relevant trickling down to the agent level, then defining here on the agent level

lorenzejay (approved):

lorenzejay:
tests passing locally:

Screenshot 2024-11-19 at 2 59 27 PM

docs/concepts/knowledge.mdx
```diff
@@ -0,0 +1,91 @@
...
+from crewai.knowledge.source.string_knowledge_source import StringKnowledgeSource
+
+# Create a knowledge source
+content = "Users name is John. He is 30 years old and lives in San Francisco."
+string_source = StringKnowledgeSource(
+    content=content, metadata={"preference": "personal"}
+)
+
+# Create a knowledge store with a list of sources and metadata
+knowledge = Knowledge(sources=[string_source], metadata={"preference": "personal"})
```
pythonbyte: _Question_: Should knowledge be used in some of the code or works behind the curtains?

docs/concepts/knowledge.mdx
```diff
@@ -0,0 +1,91 @@
...
+result = crew.kickoff(inputs={"question": "What city does John live in and how old is he?"})
+```
+
+## Additionally:
+You can add more sources as well as not need to re-declare the knowledge store every kickoff.
+If you had sources from previous runs, you no longer need to declare the knowledge store.
+
+```python
+from crewai import Agent, Task, Crew, Process, LLM
+from crewai.knowledge.source.string_knowledge_source import StringKnowledgeSource
```
pythonbyte: _Question_: I saw that `StringKnowledgeSource` was imported but not used here. Is that the intention?

.github/workflows/tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         run: uv python install 3.11.9
 
       - name: Install the project
-        run: uv sync --dev
+        run: uv sync --dev --all-extras
```
pythonbyte, line 29: _Question_:  Do you think we need the --all-extra option in this case? It seems like we'll have to install all the optional dependencies to be able to run our tests. What do you think?

path/to/src/crewai/knowledge/source/base_knowledge_source.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
```
pythonbyte: _Question_: I imagine that the path of this file is not correct. 
`path/to/src/crewai/knowledge/`

path/to/src/crewai/knowledge/source/base_knowledge_source.py
```diff
@@ -0,0 +1,30 @@
...
+        self,
+        chunk_size: int = 1000,
+        chunk_overlap: int = 200,
+    ):
+        self.chunk_size = chunk_size
+        self.chunk_overlap = chunk_overlap
+        self.chunks: List[str] = []
+
+    @abstractmethod
+    def load_content(self):
```
pythonbyte: _Nitpick_: Maybe return -> None on this function if that's the intention of this method.

pyproject.toml
```diff
@@ -39,6 +39,16 @@ Repository = "https://github.com/crewAIInc/crewAI"
 [project.optional-dependencies]
 tools = ["crewai-tools>=0.14.0"]
 agentops = ["agentops>=0.3.0"]
+fastembed = ["fastembed>=0.4.1"]
+pdfplumber = [
+    "pdfplumber>=0.11.4",
+]
+pandas = [
```
pythonbyte, line 46: _Suggestion_:I'm wondering if we need to keep "pandas" as an optional dependency. I took a look at the code, and it seems we're only using it to read Excel files and save them as CSVs. Maybe we could find some lighter libraries to handle that? Just a thought!

If the lib is still required maybe we should go with "polars" 
```
Polars: ~8.5MB
Pandas: ~12MB

Polars: ~70ms
NumPy: ~104ms
Pandas: ~520ms
```

src/crewai/agent.py
```diff
@@ -272,6 +281,21 @@ def execute_task(
...
+            if knowledge_snippets:
+                valid_snippets = [
+                    result["context"]
+                    for result in knowledge_snippets
+                    if result and result.get("context")
+                ]
+                if valid_snippets:
+                    formatted_knowledge = "\n".join(valid_snippets)
+                    task_prompt += f"\n\nAdditional Information:\n{formatted_knowledge}"
+
```
pythonbyte: _Suggestion_: You can shorten this code by doing like this 

```
if self.crew and self.crew.knowledge_store:
    knowledge_snippets: List[Dict[str, Any]] = self.crew.knowledge_store.query([task.prompt()])
    
    valid_snippets = [
        result["context"] 
        for result in knowledge_snippets 
        if result and "context" in result
    ]
    
    if valid_snippets:
        task_prompt += f"\n\nAdditional Information:\n{'\n'.join(valid_snippets)}"
        
        
```
If its certain that will return a []

src/crewai/agent.py
```diff
@@ -227,6 +230,12 @@ def post_init_setup(self):
         if self.allow_code_execution:
             self._validate_docker_installation()
 
+        # Initialize the Knowledge object if knowledge_sources are provided
```
pythonbyte: _Nitpick_: But in here you can do 
```
self._knowledge = None
if self.crew and self.crew.knowledge_store:
     self._knowledge = self.crew.knowledge_store
```
Or even remove the = None
Since the default form the Model is None

src/crewai/cli/cli.py
```diff
@@ -136,24 +136,25 @@ def log_tasks_outputs() -> None:
...
+@click.option("-kn", "--knowledge", is_flag=True, help="Reset KNOWLEDGE storage")
 @click.option(
     "-k",
     "--kickoff-outputs",
     is_flag=True,
     help="Reset LATEST KICKOFF TASK OUTPUTS",
 )
 @click.option("-a", "--all", is_flag=True, help="Reset ALL memories")
-def reset_memories(long, short, entities, kickoff_outputs, all):
+def reset_memories(long, short, entities, knowledge, kickoff_outputs, all):
```
pythonbyte: _Nitpick_: Maybe add type-hints 😅

docs/concepts/knowledge.mdx
pythonbyte: `Crew(knowledge_sources = knowledge, ...)`?

src/crewai/crew.py
```diff
@@ -193,6 +195,18 @@ class Crew(BaseModel):
...
+        description="Knowledge sources for the agent.",
+    )
+    knowledge_store: Optional[Knowledge] = Field(
+        default=None, description="Knowledge Source for the crew."
+    )
+    knowledge: Optional[bool] = Field(
+        default=False,
+        description="Whether the crew should use knowledge to store memories of it's execution",
+    )
+
```
pythonbyte: _Suggestion_: Add these parameters to the Docstring 
```
        share_crew: Whether you want to share the complete crew information and execution with crewAI to make the library better, and allow us to train models.
        planning: Plan the crew execution and add the plan to the crew.

        knowledge_sources...
        knowledge_store...
        knowledge....
    """
```

src/crewai/knowledge/knowledge.py
```diff
@@ -0,0 +1,50 @@
...
+from pydantic import BaseModel, ConfigDict, Field
+
+from crewai.knowledge.source.base_knowledge_source import BaseKnowledgeSource
+from crewai.knowledge.storage.knowledge_storage import KnowledgeStorage
+from crewai.utilities.logger import Logger
+
+os.environ["TOKENIZERS_PARALLELISM"] = "false"  # removes logging from fastembed
+
+
+class Knowledge(BaseModel):
```
pythonbyte, line 13: _Suggestion_: Indeed, a Docstring here would be awesome.

src/crewai/knowledge/knowledge.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Query across all knowledge sources to find the most relevant information.
+        Returns the top_k most relevant chunks.
+        """
+
+        results = self.storage.search(
+            query,
+            limit,
+            filter={"preference": preference} if preference else None,
+            score_threshold=0.35,
```
pythonbyte: _Suggestion_: We might want to have this number as a Constant or maybe a parameter with default value 🤔

tests/knowledge/crewai_quickstart.pdf
pythonbyte, line 1: _Question_: Are you planning to upload this .pdf to the repository?

src/crewai/knowledge/source/base_file_knowledge_source.py
```diff
@@ -0,0 +1,37 @@
...
+
+class BaseFileKnowledgeSource(BaseKnowledgeSource):
+    """Base class for knowledge sources that load content from files."""
+
+    file_path: Union[Path, List[Path]] = Field(...)
+    content: Dict[Path, str] = Field(init=False, default_factory=dict)
+    storage: KnowledgeStorage = Field(default_factory=KnowledgeStorage)
+    metadata: Dict[str, Any] = Field(default_factory=dict)
+
+    def model_post_init(self, context):
```
pythonbyte: _Question_: Noticed that context is being used as a parameter but not used on the method, is that expected?

src/crewai/knowledge/source/base_file_knowledge_source.py
```diff
@@ -0,0 +1,37 @@
...
+        for path in paths:
+            if not path.exists():
+                raise FileNotFoundError(f"File not found: {path}")
+            if not path.is_file():
+                raise ValueError(f"Path is not a file: {path}")
+        return {}
+
+    def save_documents(self, metadata: Dict[str, Any]):
+        """Save the documents to the storage."""
+        chunk_metadatas = [metadata.copy() for _ in self.chunks]
```
pythonbyte, line 35: _Question_: Could you help me understand what this line of code does? Thanks!

src/crewai/knowledge/source/base_knowledge_source.py
```diff
@@ -0,0 +1,48 @@
+from abc import ABC, abstractmethod
+from typing import List, Dict, Any
+
+import numpy as np
+from pydantic import BaseModel, ConfigDict, Field
+
+from crewai.knowledge.storage.knowledge_storage import KnowledgeStorage
+
+
+class BaseKnowledgeSource(BaseModel, ABC):
```
pythonbyte, line 10: _Question_: Noticed that there are 2 BaseKnowledgeSource, this one and the path/to/src/crewai/knowledge/source/base_knowledge_source.py, which I imagine is a duplicate.

pythonbyte (commented):
Hey, @lorenzejay left some comments, let me know if you have any questions on those 😄

src/crewai/knowledge/source/base_file_knowledge_source.py
lorenzejay: Before adding documents to the vector db, it needs metadata object. The metadata can be the same as its a representation of itself as a chunk.

When getting chunked (lets say a big pdf), it will x amount of content. so the defined metadata, should be cloned to be passed as well

overall we need per `chunk` a metadata dict to go with it.

docs/concepts/knowledge.mdx
lorenzejay: Youre right. that was not intended

.github/workflows/tests.yml
lorenzejay: Yes, there are a bunch of optional dep that were brought up like the `pdfplumber` for our PdfKnowledgeSource.

src/crewai/knowledge/source/base_file_knowledge_source.py
lorenzejay: resolved this was a pydantic specific feature: https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_post_init 

so did `_` instead

tests/knowledge/crewai_quickstart.pdf
lorenzejay: Yes as we are using it for the `PDFKnowledgeSource`

pyproject.toml
lorenzejay: these are optional deps, maybe this can be a fast follow ?

path/to/src/crewai/knowledge/source/base_knowledge_source.py
lorenzejay: looks right ? Abstract class could be inside the source dir

pythonbyte (approved):
LGTM ✅ 
Amazing Work

PR #1574, opened 2024-11-10
Remove poetry.lock

sanders41:
Since the project was migrated to uv in #1406 the `poetry.lock` file is no longer needed.


sanders41:
The test failure doesn't seem to have anything to do with the removal of  the `poetry.lock` file. It is failing on `test_crew_train_success` which shouldn't have anything to do with `poetry.lock`. Other PRs seem to be failing on the same test so I assume it's unrelated?

PR #1588, opened 2024-11-12
fixing LiteLLM callback replacement bug

thiagomoretto:
Setting callbacks using `litellm.callbacks = [...]` fails to replace existing ones if any, causing the old ones to be called. This affects the usage metrics which end up computing token usage more than once after the replacement. Sounds like a LiteLLM bug or a poorly documented way of what should have been done in such a case.

This PR provides a workaround to overcome this issue and then we get the token usage metrics correct again, by ensuring LiteLLM callbacks only hold on callback by type.

Here's the bug reported on LiteLLM https://github.com/BerriAI/litellm/issues/6712

A unit test started failing after the fix: `test_agent_usage_metrics_are_captured_for_hierarchical_process` , then printing the calls to `TokenCalcHandler` was enough to find the bad behavior and seeing the fix in action.

before:
```
called with this prompt tokens 642
called with this prompt tokens 191
called with this prompt tokens 191
called with this prompt tokens 729
called with this prompt tokens 729
```

after:
```
called with this prompt tokens 642
called with this prompt tokens 191
called with this prompt tokens 729
```

### DIFF
--- a/src/crewai/llm.py
+++ b/src/crewai/llm.py
@@ -118,12 +118,12 @@ class LLM:
 
         litellm.drop_params = True
         litellm.set_verbose = False
-        litellm.callbacks = callbacks
+        self.set_callbacks(callbacks)
 
     def call(self, messages: List[Dict[str, str]], callbacks: List[Any] = []) -> str:
         with suppress_warnings():
             if callbacks and len(callbacks) > 0:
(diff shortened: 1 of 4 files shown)


pythonbyte (approved):
LGTM ✅

PR #1596, opened 2024-11-13
Add cached prompt tokens info on usage metrics

thiagomoretto:
Includes cached prompt tokens total on UsageMetrics.

As per here https://docs.litellm.ai/docs/completion/prompt_caching, the total token doesn't include cached ones, so we aren't including them either.


pythonbyte (approved):
LGTM ✅ 
Nice work

PR #1623, opened 2024-11-19
Update Perplexity example in documentation

abromberg:
The example in the documentation wasn't working for me. After looking at Perplexity's docs, I found that the base URL should not have `/v1` on it. After removing that from my code, I was able to get it to function.

While I was in there, I also updated the example model to one that Perplexity currently supports.


joaomdmoura:
Good catch

PR #1635, opened 2024-11-20
Move kickoff callbacks to crew's domain

gvieira:
On the previous version, João has introduced the `before_crew` and `after_crew` callbacks, that would run before and after the kickoff respectively, allowing the user to modify the input and the output, also respectively. However, in order to run this, users would need to call the kickoff method in the `CrewBase` context, not in the `Crew` context.

This change moves the logic to the crew, keeping our standards when it comes to callbacks.


src/crewai/crew.py
```diff
@@ -520,6 +531,9 @@ def kickoff(
                 f"The process '{self.process}' is not implemented yet."
             )
 
+        for callback in self.after_kickoff_callbacks:
+            result = callback(result)
```
thiagomoretto, line 535: qq: In case one of the callbacks raises an exception, will it stop the whole process?

thiagomoretto (approved):
Awesome work, @gvieira !

src/crewai/crew.py
gvieira: Good question. The process will break in this case, and I believe it's a good thing, as we can't silently move on if something went wrong in the process of changing inputs.

PR #1640, opened 2024-11-21
Fix threading

gvieira:
We had a few issues that were preventing proper printing in threading scenarios.


thiagomoretto (approved):
Awesome work!

PR #1641, opened 2024-11-22
Remove markdown code block markers from LLM's response

ddragosd:
@roribio and I debugged today why it takes so long (`60s+`) for a task to complete when using `output_pydantic` .

It turns out that it was caused by `gpt-4o` insisting to respond with block code markers when returning the JSON.

We tracked down the fix to the `Converter` throwing `JSONDecodeError` Error and going down the path of `handle_partial_json` , which, for a simple JSON (`120` lines, `7KB`) it takes `60s+` to complete.

### DIFF
--- a/src/crewai/utilities/converter.py
+++ b/src/crewai/utilities/converter.py
@@ -97,6 +97,8 @@ def convert_to_model(
     if model is None:
         return result
     try:
+        # Sanitize the result by removing any markdown code block markers added by the LLM.
+        result = result = result.replace('```json', '').replace('```', '').strip()
         escaped_result = json.dumps(json.loads(result, strict=False))
         return validate_model(escaped_result, model, bool(output_json))
     except json.JSONDecodeError:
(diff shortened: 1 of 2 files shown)


bhancockio:
Hey @ddragosd !

Thank you for creating this PR!

**Root issue:**
- The models defined in output_pydantic and output_json are not passed into the LLM calls. As a result, the outputs of a task do not match the specified output format which leads us to doing another LLM call to format the original LLM output.

**Fix:**
- When output_pydantic and output_json are present, append a stringified version of the schema to the prompt.

**Improvement stats:**
- When running a simple crew with 1 task and a output_pydantic , our old approach would take around 30 seconds to complete the crew kickoff
- Half of this time was dedicated to doing a followup LLM call to convert the output of the task into the specified pydantic model.
- Now, the outputs of tasks automatically generate the correct format so it only takes 15 seconds which is a 50% speed increase.
- We also do one less LLM call which saves users money as well

Closed by #1651

ddragosd:
Some additional observations based on my tests with GPT `4o` version `2024-05-13`:
* The LLM may focus on the response format more than on the task itself somtimes
* I worked around this problem by using two tasks:
   *  the first task I allow the LLM to focus on the main goal without specifying pydantic output;
   * then I use a second task with pydantic output. I usually end up  asking the LLM to reflect on an initial answer anyway, so this works 👍  for my use-case.

PR #1650, opened 2024-11-25
Log in to Tool Repository on `crewai login`

vinibrsl:
This commit adds an extra step to `crewai login` to ensure users also log in to Tool Repository, that is, exchanging their Auth0 tokens for a Tool Repository username and password to be used by UV downloads and API tool uploads.

### DIFF
--- a/src/crewai/cli/utils.py
+++ b/src/crewai/cli/utils.py
@@ -1,4 +1,3 @@
-import importlib.metadata
 import os
 import shutil
 import sys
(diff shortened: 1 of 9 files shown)


src/crewai/cli/utils.py
vinibrsl, line 1: Changes here are due to a circular dependency.

vinibrsl:
> I'm already logged in to crewai but when I click the `Navigate to: https://crewai.us.auth0.com/activate?user_code=PSLW-XDGJ` it makes me login again. Is that the desired behavior?

Yes. The Auth0 app we use for the CLI is not the same as the platform.

bhancockio:
Does it break anything if I run `crewai login` multiple times and login each time?

bhancockio (approved):

vinibrsl:
> Does it break anything if I run `crewai login` multiple times and login each time?

No, you should be fine. That renews OAuth token we have stored.

PR #1655, opened 2024-11-25
added knowledge to agent level

lorenzejay:


docs/concepts/knowledge.mdx
```diff
@@ -57,6 +57,24 @@ crew = Crew(
 result = crew.kickoff(inputs={"question": "What city does John live in and how old is he?"})
 ```
 
+## Appending Knowledge Sources To Individual Agents
+Sometimes you may want to append knowledge sources to an individual agent. You can do this by setting the `knowledge` parameter in the `Agent` class.
+
+```python
+agent = Agent(
+    ...
+    knowledge={
```
bhancockio: Why do you have multiple metadata in Knowledge?

Also, what happens if someone doesn't provide metadata? What happens?

src/crewai/agent.py
```diff
@@ -120,9 +121,14 @@ class Agent(BaseAgent):
         default="safe",
         description="Mode for code execution: 'safe' (using Docker) or 'unsafe' (direct execution).",
     )
+    knowledge: Optional[Dict[str, Any]] = Field(
```
bhancockio: Shouldn't the type for knowledge be `Optional[List[BaseKnowledgeSource]`?

bhancockio: Type "Knowledge" is incompatible with type "Dict[str, Any] | None"
    "Knowledge" is incompatible with "Dict[str, Any]"
    "Knowledge" is incompatible with 
    
    Getting this error from line 251.

src/crewai/agent.py
```diff
@@ -235,6 +241,28 @@ def _setup_agent_executor(self):
             self.cache_handler = CacheHandler()
         self.set_cache_handler(self.cache_handler)
 
+    def _set_knowledge(self):
+        try:
+            if self.knowledge:
+                knowledge_agent_name = f"{self.role.replace(' ', '_')}"
+                if isinstance(self.knowledge, dict):
+                    knowledge_data = self.knowledge.copy()
```
bhancockio: Why do we need to copy?

src/crewai/agent.py
```diff
@@ -235,6 +241,28 @@ def _setup_agent_executor(self):
...
+                    knowledge_data["store_dir"] = knowledge_agent_name
+                    self.knowledge = Knowledge(**knowledge_data)
+                    self.knowledge.storage.initialize_knowledge_storage()
```
bhancockio: Shouldn't this be handled in the constructor of Knowledge.py?

Why does the agent need to know to call initialize_knowledge_storage?

src/crewai/agent.py
```diff
@@ -235,6 +241,28 @@ def _setup_agent_executor(self):
...
+                    try:
+                        for source in self.knowledge.sources:
```
bhancockio: Shouldn't all of this get called in the constructor of knowledge.py as well?

src/crewai/agent.py
```diff
@@ -273,17 +301,20 @@ def execute_task(
             if memory.strip() != "":
                 task_prompt += self.i18n.slice("memory").format(memory=memory)
 
-        # Integrate the knowledge base
+        if self.knowledge and isinstance(self.knowledge, Knowledge):
```
bhancockio: I think we can drop isinstance(self.knowledge) ...

once we update the type of self.knowledge.

src/crewai/agent.py
```diff
@@ -273,17 +301,20 @@ def execute_task(
...
-            valid_snippets = [
-                result["context"] 
-                for result in knowledge_snippets 
-                if result and result.get("context")
-            ]
-            if valid_snippets:
-                formatted_knowledge = "\n".join(valid_snippets)
-                task_prompt += f"\n\nAdditional Information:\n{formatted_knowledge}"
+
+            crew_knowledge_context = self._extract_knowledge_context(knowledge_snippets)
```
bhancockio: Would it make more sense to put _extract_knowledge_context in knowledge so that it could be used in the crew and the agent?

src/crewai/crew.py
```diff
@@ -203,10 +203,10 @@ class Crew(BaseModel):
         description="List of execution logs for tasks",
     )
     knowledge: Optional[Dict[str, Any]] = Field(
```
bhancockio: Need to update the type here as well to match the update knowledge in agent.py

src/crewai/crew.py
```diff
@@ -282,11 +282,23 @@ def create_crew_memory(self) -> "Crew":
...
     @model_validator(mode="after")
     def create_crew_knowledge(self) -> "Crew":
+        """Create the knowledge for the crew."""
         if self.knowledge:
             try:
-                self.knowledge = Knowledge(**self.knowledge) if isinstance(self.knowledge, dict) else self.knowledge
-            except (TypeError, ValueError) as e:
-                raise ValueError(f"Invalid knowledge configuration: {str(e)}")
+                self.knowledge = (
+                    Knowledge(**self.knowledge, store_dir="crew")
```
bhancockio: I'm getting a ton of errors in my editor here.

Cannot assign to attribute "knowledge" for class "Crew*"
  Type "Knowledge" is incompatible with type "Dict[str, Any] | None"
    "Knowledge" is incompatible with "Dict[str, Any]"
    "Knowledge" is incompatible with "None"Pylance[reportAttributeAccessIssue](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportAttributeAccessIssue)

src/crewai/crew.py
```diff
@@ -282,11 +282,23 @@ def create_crew_memory(self) -> "Crew":
...
+                    if isinstance(self.knowledge, dict)
+                    else self.knowledge
+                )
+                self.knowledge.storage.initialize_knowledge_storage()
+
+                for source in self.knowledge.sources:
```
bhancockio: Same comments mentioned in agent.py

src/crewai/knowledge/storage/knowledge_storage.py
```diff
@@ -35,9 +35,16 @@ class KnowledgeStorage(BaseKnowledgeStorage):
...
 
-    def __init__(self, embedder_config: Optional[Dict[str, Any]] = None):
-        self._initialize_app(embedder_config or {})
+    def __init__(
+        self,
+        embedder_config: Optional[Dict[str, Any]] = None,
+        store_dir: Optional[str] = None,
+    ):
+        self.embedder_config = embedder_config
+        self.store_dir = store_dir
```
bhancockio: Won't this always overwrite store_dir that we want to default to knowledge?

bhancockio (changes requested):

docs/concepts/knowledge.mdx
lorenzejay: An error is raised. Chromadb complains too

src/crewai/agent.py
lorenzejay: No, because like memory you can set your embedder config from there as well. Or maybe there’s a better way of handling this

src/crewai/agent.py
lorenzejay: There’s cases where it initializes before agent is done initializing.

lorenzejay: Perhaps, I can have an if store_dir on KnowledgeStoage level

src/crewai/agent.py
lorenzejay: Good call

src/crewai/knowledge/storage/knowledge_storage.py
lorenzejay: its more collection_name that it is store_dir. i need to change this. it points to the same directory it persists in just different collection

docs/concepts/knowledge.mdx
```diff
@@ -51,12 +51,41 @@ crew = Crew(
     tasks=[task],
     verbose=True,
     process=Process.sequential,
-    knowledge={"sources": [string_source], "metadata": {"preference": "personal"}}, # Enable knowledge by adding the sources here. You can also add more sources to the sources list.
+    knowledge_sources=[string_source], # Enable knowledge by adding the sources here.
```
bhancockio, line 54: It is important that we tell Joao that this is a breaking change for knowledge.

src/crewai/agent.py
```diff
@@ -503,6 +538,18 @@ def _validate_docker_installation(self) -> None:
                 f"Docker is not running. Please start Docker to use code execution with agent: {self.role}"
             )
 
+    def _extract_knowledge_context(
```
bhancockio: Shouldn't we drop this and use the knowledge_util?

bhancockio (approved):

docs/concepts/knowledge.mdx
lorenzejay: Yes a 1 day breaking change lol

tonykipkemboi: Screenshot 2024-12-01 at 2 33 23 PM

I think the example is broken @lorenzejay

PR #1590, opened 2024-11-12
Replace depreciated class Config with model_config

sanders41:
Using `class Config` with Pydantic models was depreciated in v2 and replaced with `model_config`. https://docs.pydantic.dev/latest/migration/#changes-to-config

PR #1661, opened 2024-11-26
remove all references to pipeline and pipeline router

bhancockio:
Now that we have Flows, we no longer want to support pipeline.

We were supposed to drop pipelines when we introduced flows but forgot to cut them out.

This is a breaking change.

Closes #1590 

We also remove the poetry.lock file in this PR. Thank you @sanders41 for calling this out.

Closes #1574


lorenzejay (approved):
I would only add some text on docs along the lines of this:

If you were using pipelines/pipeline router -> we have migrated that over to flows.

bhancockio:
Thanks for review @lorenzejay !

So I did some digging and it looks like the pipeline documentation was pulled out a long time ago when we moved over to flows, so I'm probably just not gonna mention anything instead of creating a new pipeline dock that points to flows.

I do agree, though, with what you were suggesting because ideally back in the day, we should have updated the pipeline docs to say go to flows instead.

PR #1663, opened 2024-11-26
Ensure RAGStorage storage file name doesn't exceed limit

### LINKED ISSUE #1662: [BUG] OSError when activating memory on Crew
### Description

Long role names cause the resulting storage file name to exceed the OS limit for file names.

### Steps to Reproduce

1. Create different agents where the resulting character count for all agents' roles (after sanitized) is greater than or equal to `302`
2. Create a `Crew` and set `memory=True`
3. The error should show up.

### Expected behavior

The expected behavior is for the crew to run without errors regardless of the character length of the roles.

### Screenshots/Code snippets

```
OSError: [Errno 63] File name too long: '/Users/caike/Library/Application Support/iron_planner/short_term/Specialist_in_discovering_Iron_Man_triathlon_events_worldwide_Expert_in_evaluating_and_categorizing_Iron_Man_triathlon_events._Specialist_in_selecting_top_accommodations_for_triathlon_participants._Specialist_in_crafting_structured_reports_for_triathlon_events._JSON_Report_Quality_Assurance_Specialist'
```

### Operating System

### Python Version

3.12

### crewAI Version

0.83.0

### crewAI Tools Version

0.14.0

### Virtual Environment

Venv

### Evidence

```
OSError: [Errno 63] File name too long: '/Users/caike/Library/Application Support/iron_planner/short_term/Specialist_in_discovering_Iron_Man_triathlon_events_worldwide_Expert_in_evaluating_and_categorizing_Iron_Man_triathlon_events._Specialist_in_selecting_top_accommodations_for_triathlon_participants._Specialist_in_crafting_structured_reports_for_triathlon_events._JSON_Report_Quality_Assurance_Specialist'
```

### Possible Solution

We should avoid using the sanitized value of the roles as file name. This is error prone as evidenced by this error that's being reported. One possible solution is to use a hash.

Something along the lines of:

```python
hash_obj = hashlib.md5(self.agents.encode('utf-8'))
storage_file_name = hash_obj.hexdigest()[:32] # length will always be 32
```

### Additional context

OS used is OS X Sequoia 15.1.1

caike:
Fixes #1662

### DIFF
--- a/src/crewai/memory/storage/rag_storage.py
+++ b/src/crewai/memory/storage/rag_storage.py
@@ -44,7 +44,7 @@ class RAGStorage(BaseRAGStorage):
         agents = [self._sanitize_role(agent.role) for agent in agents]
         agents = "_".join(agents)
         self.agents = agents
-        self.storage_file_name = self._build_file_name(agents)
+        self.storage_file_name = self._build_storage_file_name(type, agents)
 
         self.type = type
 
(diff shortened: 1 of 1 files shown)


src/crewai/memory/storage/rag_storage.py
```diff
@@ -80,6 +82,14 @@ def _sanitize_role(self, role: str) -> str:
         """
         return role.replace("\n", "").replace(" ", "_").replace("/", "_")
 
+    def _build_file_name(self, string: str) -> str:
```
bhancockio: We'd prefer to keep the name of the agent so it's easy to find storage associated with each agent.

Instead of hashing, can you please only grab the first 300 characters so you're under the limit?

PR #1664, opened 2024-11-26
Update using langchain tools docs

olaservo:
Hoping this helps someone else.  The docs on this are out of date and ran me in circles for a bit until I dug into the crewAI source code.

### DIFF
--- a/docs/concepts/langchain-tools.mdx
+++ b/docs/concepts/langchain-tools.mdx
@@ -7,32 +7,44 @@ icon: link
 ## Using LangChain Tools
 
 <Info>
-    CrewAI seamlessly integrates with LangChain’s comprehensive [list of tools](https://python.langchain.com/docs/integrations/tools/), all of which can be used with CrewAI.
+    CrewAI seamlessly integrates with LangChain's comprehensive [list of tools](https://python.langchain.com/docs/integrations/tools/), all of which can be used with CrewAI.
 </Info>
 
-```python Code
+```python
 import os
-from crewai import Agent
-from langchain.agents import Tool
-from langchain.utilities import GoogleSerperAPIWrapper
+from dotenv import load_dotenv
+from crewai import Agent, Task, Crew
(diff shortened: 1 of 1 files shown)


docs/concepts/langchain-tools.mdx
```diff
@@ -7,39 +7,52 @@ icon: link
...
-  func=search.run,
-  description="Useful for search-based queries",
-)
-
-agent = Agent(
-  role='Research Analyst',
-  goal='Provide up-to-date market analysis',
-  backstory='An expert analyst with a keen eye for market trends.',
-  tools=[serper_tool]
+class SearchTool(BaseTool):
```
bhancockio, line 27: Instead of creating a `BaseTool`, you could also just use langchain's tool like this:

```
from langchain.agents import Tool
from langchain_community.utilities import GoogleSerperAPIWrapper

search = GoogleSerperAPIWrapper()

# Create and assign the search tool to an agent
serper_tool = Tool(
    name="Intermediate Answer",
    func=search.run,
    description="Useful for search-based queries",
)

research_analyst = Agent(
    role="Research Analyst",
    goal="Provide up-to-date market analysis",
    backstory="An expert analyst with a keen eye for market trends.",
    tools=[serper_tool],
)
```

It looks a little more cleaner and fully showcases using a LangChain Tool.

If you could update the example to use `from langchain.agents import Tool`, we can merge this in!

bhancockio (changes requested):

docs/concepts/langchain-tools.mdx
olaservo: I might be missing something, I think that's what is already in the example?  When I looked at the source code it seemed like I needed to wrap it this way since I got an error otherwise.

joaomdmoura:
Analyzed the gathered information in the context of the upcoming meeting and...

bhancockio (approved):

PR #1691, opened 2024-12-02
Knowledge project directory standard

lorenzejay:


src/crewai/cli/templates/crew/crew.py
```diff
@@ -57,10 +57,20 @@ def reporting_task(self) -> Task:
...
+		# 	),
+		# ]
+
 		return Crew(
 			agents=self.agents, # Automatically created by the @agent decorator
 			tasks=self.tasks, # Automatically created by the @task decorator
 			process=Process.sequential,
 			verbose=True,
 			# process=Process.hierarchical, # In case you wanna use that instead https://docs.crewai.com/how-to/Hierarchical/
+			# knowledge_sources=sources, # In the case you want to add knowledge sources``
```
bhancockio: Please remove ``.

src/crewai/knowledge/source/base_file_knowledge_source.py
```diff
@@ -1,17 +1,20 @@
...
 
 from pydantic import Field
 
 from crewai.knowledge.source.base_knowledge_source import BaseKnowledgeSource
-from typing import Dict, Any
+from crewai.utilities.logger import Logger
 from crewai.knowledge.storage.knowledge_storage import KnowledgeStorage
 
 
 class BaseFileKnowledgeSource(BaseKnowledgeSource):
```
bhancockio: As I'm reviewing this, I think this should be an abstract class`class BaseFileKnowledgeSource(BaseKnowledgeSource, ABC):

src/crewai/knowledge/source/base_file_knowledge_source.py
```diff
@@ -20,17 +23,42 @@ def model_post_init(self, _):
         self.content = self.load_content()
 
     def load_content(self) -> Dict[Path, str]:
-        """Load and preprocess file content. Should be overridden by subclasses."""
+        """Load and preprocess file content. Should be overridden by subclasses. We want to assume that the file path is relative to the project root in the knowledge directory."""
```
bhancockio: Should this be overridden? Why not go ahead and do all the logic here?

Or, should we just rename this function to preprocess_file_paths? Because if you look at this method, we aren't ever loading any data.

Then we can update TextFileKnowledgeSource and similar file based source from:

```
def load_content(self) -> Dict[Path, str]:
        """Load and preprocess text file content."""
        super().load_content()
        paths = [self.file_path] if isinstance(self.file_path, Path) else self.file_path
        content = {}
        for path in paths:
            with path.open("r", encoding="utf-8") as f:
                content[path] = f.read()  # type: ignore
        return content
```

To:

```
def load_content(self) -> Dict[Path, str]:
        """Load and preprocess text file content."""
        paths = [self.file_path] if isinstance(self.file_path, Path) else self.file_path
        content = {}
        for path in paths:
            with path.open("r", encoding="utf-8") as f:
                content[path] = f.read()  # type: ignore
        return content
```

And update model_post_init to this:

```
def model_post_init(self, _):
        """Post-initialization method to validate content."""
        self.preprocess_file_paths()
        self.load_content()
        ```

File: src/crewai/knowledge/source/base_file_knowledge_source.py [hunk 4]
```diff
@@ -20,17 +23,42 @@ def model_post_init(self, _):
...
+
+        if isinstance(self.file_path, str):
```
bhancockio: I like how we've made file_path more generic to accept strings or paths.

To keep things clean and prevent future `isinstance` checks, maybe create a new `_safe_file_paths: List[Path]`

and our ultimate goal of this `load_content` (most likely changed to `preprocess_file_paths`) function is to validate and convert all files paths to a single type which is `path`

src/crewai/knowledge/source/base_file_knowledge_source.py
```diff
@@ -20,17 +23,42 @@ def model_post_init(self, _):
...
         return {}
 
     def save_documents(self, metadata: Dict[str, Any]):
         """Save the documents to the storage."""
         chunk_metadatas = [metadata.copy() for _ in self.chunks]
         self.storage.save(self.chunks, chunk_metadatas)
+
+    def convert_to_path(self, path: Union[Path, str]) -> Path:
+        """Convert a path to a Path object."""
+        return Path("knowledge/" + path) if isinstance(path, str) else path
```
bhancockio: Should we change "knowledge/" to a constant to give our readers more context about what's going on.

Maybe add a `ROOT_KNOWLEDGE_DIR = "knowledge/"` at the top?

bhancockio (approved):

PR #1702, opened 2024-12-04
Brandon/cre 509 hitl multiple rounds of followup

bhancockio:


gvieira:
> Do you want me to stick with the conversational approach or go with the old school fixed response?

Conversational approach looks way, way better and more CrewAIy. ;)

> Do you think we should do something different or have any feedback on this approach?

Loved this approach!

src/crewai/agents/crew_agent_executor.py
```diff
@@ -412,3 +396,83 @@ def _format_answer(self, answer: str) -> Union[AgentAction, AgentFinish]:
...
+                if self.crew and self.crew._train:
+                    self._handle_crew_training_output(formatted_answer)
+            else:
+                # Unexpected response
+                self._printer.print(
+                    content=f"Unexpected response from LLM: '{additional_changes_response}'. Assuming no additional changes requested.",
+                    color="red",
+                )
+                self.ask_for_human_input = False
+        print("FINAL ANSWER: ", formatted_answer)
```
gvieira: Should we drop this `print`?

PR #1765, opened 2024-12-16
revise the model description for OpenAI-like APIs

enumz:
Based on the documentation for liteLLM, it is necessary to add a prefix to the model description in order for it to be properly recognized by the API.

===
https://docs.litellm.ai/docs/providers/openai_compatible
For /chat/completions: Put openai/ in front of your model name, so litellm knows you're trying to call an openai /chat/completions endpoint.

### DIFF
--- a/docs/how-to/llm-connections.mdx
+++ b/docs/how-to/llm-connections.mdx
@@ -120,7 +120,7 @@ You can connect to OpenAI-compatible LLMs using either environment variables or 
 
         os.environ["OPENAI_API_KEY"] = "your-api-key"
         os.environ["OPENAI_API_BASE"] = "https://api.your-provider.com/v1"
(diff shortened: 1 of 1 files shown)


bhancockio:
Thank you for creating this PR @enumz!

However, the link you provided is for openai compatible LLMs. If you are using a regular OpenAI model like gpt-4o, you do not have to set the provider as shown in their documentation here:

https://docs.litellm.ai/docs/providers/openai

Screenshot 2024-12-17 at 10 43 38 AM

enumz:
@bhancockio 
Yes, I encountered issues when using other OpenAI-compatible LLM. so the modifications are in the "Connecting to OpenAI-Compatible LLMs" chapter.

Issue #1152, opened 2024-08-07
Conditional Tasks[BUG]


A1iMansour:
I am trying to use Conditional Tasks in crew as follows (here is simplified version):

**Tasks:**
```
class abc:
    def __init__(self):
        self.variables = {
            "a": False,
            "b": False,
        }

    def select_tasks(self,output: TaskOutput): #check if generate_structure_task requires elements a/b
        s=output.raw.split()
        for key in s:
            key = key.lower()
            if key in self.variables.keys():
                self.variables[key] = True
     
    def generate_structure_task( self, agent):
      
        return Task(
               
                
                description = dedent(f"""\
                                      Design the structure, include required elements ex: 1) a , 2) b.
                                    """),
                expected_output=dedent("""return detailed structure showing what elements need to be implemented"""),
                agent= agent,
                callback=self.select_tasks
              )
 def is_a_needed(self,output: TaskOutput) -> bool:
      #print(output)
      return self.variables.get("a") 

    def is_b_needed(self,output: TaskOutput) -> bool:
      #print(output)
      t= self.variables.get("b") 
      print(t)
      return t
        
    def generate_a_code(self,agent):
        return ConditionalTask(
            description="Generate the code for a ",
             expected_output="code for element a",
            condition= self.is_a_needed,
            agent=agent
        )

    def generate_b_code(self,agent):
        return ConditionalTask(
            description="Generate the code for b",
            expected_output="code for element b ",
            condition= self.is_repo_needed,
            agent=agent

**Crew.py**:

def kickoff(self):
    task = SymfonyGeneratorTasks()
    crew = Crew(
                agents= [self.agent1,self.agent2],
                tasks=[
                      task.generate_structure_task(agent=self.agent1),
                      task.generate_a_code(agent=self.agent2),
                      task.generate_b_code(agent=self.agent2),
                      
                ],
              )
    result = crew.kickoff()
    return result

```

So based on structure generated by first task, variables will be update and those variables are actually the conditions needed for Conditional Task. The first conditional task runs normally, however the second throws this error:
 previous_output = task_outputs[task_index - 1] if task_outputs else None IndexError: list index out of range.

May I know how to fix this problem

san3005:
I faced the same issue I stopped using the conditional tasks no more than one

ojtramp:
Also having this issue, any advice on fixing this?

A1iMansour:
[...]

PR #2013, opened 2025-01-31
Fix:  Check `task_outputs` length to prevent Out-of-Index error

barantaran:
In the case of synchronous task execution, only the last task output is preserved. 
However, the output index is always calculated based on `task_index`, which causes an "out of index" error when `task_index` is greater than 1.

To prevent this, we need to check the length of `task_outputs ` before calculating the output shift.

This should resolve [crewAIInc/crewAI#1152](https://github.com/crewAIInc/crewAI/issues/1152)


joaomdmoura:

# Code Review Comment for PR #2013

## Overview
This pull request proposes an important fix to the `_handle_conditional_task` method in `src/crewai/crew.py`, addressing how previous task outputs are accessed to avoid potential index out-of-bounds errors. 

## Key Changes
The main modification is in how the index for accessing the `task_outputs` array is determined:
### Original Code:
```python
previous_output = task_outputs[task_index - 1] if task_outputs else None
```

### Modified Code:
```python
prev_output_index = task_index - 1 if len(task_outputs) > 1 else 0
previous_output = task_outputs[prev_output_index] if task_outputs else None
```

## Findings

### 1. Safety Improvement
The adjustment enhances the safety of array access, mitigating risks of index errors when accessing previous task outputs.

### 2. Potential Issues
- **Edge Case Handling**: The logic simplification might lead to returning the wrong output in edge cases. The current approach returns the first element rather than `None` when there's just one output, which could be misleading.
- **Logic Redundancy**: The check `if task_outputs else None` seems redundant since the length is already being checked.

## Recommendations

### 1. Enhanced Edge Case Handling
Modify the code for clearer logic and better handling of edge cases:
```python
previous_output = None
if task_outputs:
    if len(task_outputs) > 1:
        previous_output = task_outputs[task_index - 1]
    elif task_index > 0:  # Ensure task_index is positive
        previous_output = task_outputs[0]
```

### 2. Input Validation
Add validation on `task_index` to ensure it’s non-negative:
```python
if task_index < 0:
    raise ValueError(f"Invalid task_index: {task_index}")
```

### 3. Documentation Enhancement
Include docstrings to clarify method functionality and argument expectations:
```python
def _handle_conditional_task(self, ...):
    """
    Handles conditional task execution based on previous task outputs.
    
    Args:
        task_index (int): Current task index in the sequence
        task_outputs (list): List of previous task outputs
        
    Returns:
        Previous task output if conditions are met, None otherwise
    """
```

### 4. Additional Suggestions
- **Unit Testing**: Introduce specific tests for edge cases.
- **Logging**: Incorporate logging to track unexpected behavior.
- **Type Hints**: Leverage type hints for better clarity and maintenance.
```python
from typing import List, Optional

def _handle_conditional_task(
    self,
    task_index: int,
    task_outputs: List[str],
    ...
) -> Optional[str]:
```

## Conclusion
The changes presented in this PR are appropriate to address the immediate issue but could benefit from the suggested improvements for robustness and maintainability. Merging is suitable, with the understanding that future iterations should consider these enhancements for further refinement of the code's logic and safety.

bhancockio:
Merged in and fixed.

barantaran:
> Merged in and fixed.

Hi @bhancockio! Could you let me know in which release this fix will be included?

danielfsbarreto:
@barantaran, our next cut should include it (0.100.2)

PR #2045, opened 2025-02-06
fix: Typo in expected_output string

thohag:
"expect" should be "expected"

Current: "...This is the expect criteria for your final answer..."
Fix: "...This is the expected criteria for your final answer..."


bhancockio (approved):
Thank you!

bhancockio:
Addressed in a new PR.

This change introduced errors in our tests so I had to redo those tests.

PR #2086, opened 2025-02-10
fix user memory config issue

sharmasundip:


sharmasundip:
If memory_config is provided without a user_memory entry, a UserMemory instance will still be created. This is because self.user_memory is likely initialized to None in this scenario.

bhancockio (approved):
Looks great! Thank you @sharmasundip for fixing this!

PR #2008, opened 2025-01-30
Small update in memory.mdx, fixing Google AI parameters

cristianmedeiros:
After several attempts to configure the embedder to use Gemini, I noticed this error in the documentation.

It's a very small change, from model_name to model, and a configuration in the pyproject dependencies adding google-generativeai.

PR #2087, opened 2025-02-10
incorporate Small update in memory.mdx, fixing Google AI parameters #…

bhancockio:
Closes out #2008 

Thank you @cristianmedeiros 

We made changes to our docs so this change wasn't getting applied to the right spot any more.


lorenzejay (approved):
nice

PR #2166, opened 2025-02-19
Update constants.py

cardofe:
This PR updates the list of foundation models available in Amazon Bedrock to reflect the latest offerings.


src/crewai/cli/constants.py
```diff
@@ -216,7 +216,12 @@
         "watsonx/ibm/granite-3-8b-instruct",
     ],
     "bedrock": [
```
cardofe, line 218: Indentation Inconsistency fixed

cardofe (commented):
Indentation Inconsistency fixed

PR #2248, opened 2025-02-27
fix(agents): Align tool reuse check with caching behavior

hukaidong:
This change updates the logic in the ToolsHandler so that when a tool call is not meant to be cached (using should_cache = false), the tool reuse check is bypassed by resetting the last used tool to an empty state instead of storing the tool call. When caching is enabled, the last used tool is recorded as before.

The update helps ensure that tools with non-deterministic behavior—like those that increment a counter on each call—can be run multiple times in a row without being blocked by the reuse check.

### DIFF
--- a/src/crewai/agents/tools_handler.py
+++ b/src/crewai/agents/tools_handler.py
@@ -23,7 +23,11 @@ class ToolsHandler:
         should_cache: bool = True,
     ) -> Any:
         """Run when tool ends running."""
-        self.last_used_tool = calling  # type: ignore # BUG?: Incompatible types in assignment (expression has type "Union[ToolCalling, InstructorToolCalling]", variable has type "ToolCalling")
+        if should_cache:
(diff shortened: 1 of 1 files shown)


bhancockio:
We're choosing not to implement this change because it fundamentally alters the purpose of the `last_used_tool` tracking mechanism, which serves two distinct functions:

1. Recording the most recent tool call for debugging/logging purposes
2. Supporting the tool reuse check logic

The proposed change conflates "caching" with "tool reuse prevention," but these are separate concerns:
- Caching is about avoiding redundant external API calls by storing previous results
- Tool reuse tracking prevents agents from calling the same tool repeatedly with the same parameters

For tools with intended non-deterministic behavior (like counters), the correct approach would be to either:
1. Implement a property on those specific tools indicating they should bypass the reuse check
2. Add unique/random parameters to ensure each call is considered different

Clearing `last_used_tool` based on `should_cache` breaks the tracking functionality and would make debugging more difficult, while addressing the symptom rather than the underlying design issue of distinguishing deterministic vs. non-deterministic tools.

PR #2339, opened 2025-03-11
Fix typo in installations

Abdelrahman350:
The current doc tells to run the following command `uv tools list` which throws the following error
`error: unrecognized subcommand 'tools'

  tip: a similar subcommand exists: 'tool'

Usage: uv [OPTIONS] <COMMAND>

For more information, try '--help'.`
The right command is  `uv tool list` by removing the s in tools.


bhancockio:
Fixed!

PR #2387, opened 2025-03-17
fix(crew template): fix wrong parameter name and missing input

agindre:
This PR fix minor errors on the template when you create a new crew:

1. The missing current_year on train() function
2. The wrong attribute `openai_model_name` changed for `eval_llm` in the test() function


agindre:
@lucasgomide done

lucasgomide (approved):

Issue #2426, opened 2025-03-20
[FEATURE]Unable to use user_memory with Redis / mem0 Opensource version


ParthS-iViewLabs:
### Feature Area

Core functionality

### Is your feature request related to a an existing bug? Please link it here.

I’m looking to integrate Redis history into CrewAI to manage conversation history more efficiently. The idea is to use Redis for storing and retrieving past interactions, which would help improve context handling and scalability for Crew. But i found, there is no direct support for Redis with Crew.

I tried exploring [mem0](https://github.com/mem0ai/mem0) which uses Redis as a VectorDB support.

Currently there is hardcoded mem0 integration which uses API KEY,  Find it here: [mem0_storage.py](https://github.com/crewAIInc/crewAI/blob/fe0813e831bf930146b7ad12356eaecfcf600b49/src/crewai/memory/storage/mem0_storage.py#L37)

### Describe the solution you'd like

It should be something like:
```
# Define configuration for Mem0 memory with Redis as the vector store
mem0_redis_config = {
    "vector_store": {
        "provider": "redis",
        "config": {
            "collection_name": "collection_mem0",
            "embedding_model_dims": 1536,
            "redis_url": "redis://localhost:6379/0"
        }
    },
    "version": "v1.1"
}

# Initialize Memory instance
mem0_instance = Memory.from_config(mem0_redis_config)
```

```
def create_sql_crew(session_id: str) -> Crew:
    """
    Create a CrewAI instance configured for SQL-related tasks.

    Args:
        user_id (str): The unique identifier for the user.

    Returns:
        Crew: A configured Crew instance for SQL operations.
    """
    agent_1 = create_sql_agent(session_id)
    extract_data = create_extract_data_task(agent_1)

    return Crew(
        agents=[agent_1],
        tasks=[extract_data],
        process=Process.sequential,
        memory=True,
        memory_config={  # Activate memory configuration
            "provider": "mem0",
            "config": {
                "client": mem0_instance,  	# Using pre-configured Redis memory
                "user_id": session_id,      # Session-specific memory isolation
            }
        },
        verbose=True
    )
```

### Willingness to Contribute

Yes, I'd be happy to submit a pull request

Vidit-Ostwal:
Hi @parthbs, thanks for the PR, have added some additional documentation and test cases to fits this well.

lucasgomide:
Closing this since it’s already been merged it will be available in the next release cut

PR #2469, opened 2025-03-25
Fixes user_memory configuration

### LINKED ISSUE #2448: [BUG] AttributeError in CrewAI MemoryClient using Gemini Api Key with Mem0 (self.um.search())
### Description

I encountered an issue when using CrewAI with Mem0 as the memory provider and Google Gemini 2.0 as the LLM The error occurs in ContextualMemory when trying to fetch user memory.

I tried the code with 

client = MemoryClient()

and 

client = MemoryClient(
    api_key="api key directly",
    org_id="my org id",
    project_id="my proj id"
)

### Steps to Reproduce

Install crewai and mem0ai using: pip install crewai mem0ai

import os
from mem0 import MemoryClient
os.environ["MEM0_API_KEY"] = userdata.get('MEM0_API_KEY')
client = MemoryClient()
messages = [
    {"role": "user", "content": "Hi there! I'm planning a vacation and could use some advice."},
     {"role": "assistant", "content": "Hello! I'd be happy to help with your vacation planning. What kind of destination do you prefer?"},
     {"role": "user", "content": "I am more of a beach person than a mountain person."},
     {"role": "assistant", "content": "That's interesting. Do you like hotels or Airbnb?"},
     {"role": "user", "content": "I like Airbnb more."},
 ]
 client.add(messages, user_id="john")

agent = Agent(
    role="About User",
    goal="You know everything about the user.",
    backstory="""You are a master at understanding people and their preferences.""",
    verbose=True,
    allow_delegation=False,
    llm=llm1, #the llm1 is saved in the os.environ
)
task = Task(
    description="Answer the following questions about the user: {question}",
    expected_output="An answer to the question.",
    agent=agent,
)

crew = Crew(
    agents=[agent],
    tasks=[task],
    process = Process.sequential,
    verbose=True,
    memory=True,
    memory_config={
        "provider": "mem0",
        "config": {"user_id": "john"},
    },
)

crew.kickoff(inputs={"question": "What is your favorite vacation destination?"})

### Expected behavior

Expected Behavior:
The CrewAI agent should retrieve user memory from Mem0.

Actual Behavior:
AttributeError: 'NoneType' object has no attribute 'search'

### Screenshots/Code snippets

Memory is not properly fetched, and self.um.search() raises an AttributeError. 

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

0.108.0

### crewAI Tools Version

0.36.0

### Virtual Environment

Venv

### Evidence

╭──────────────────────────────────────────── Crew Execution Started ─────────────────────────────────────────────╮ 
│                                                                                                                 │
│  Crew Execution Started                                                                                         │
│  Name: crew                                                                                                     │
│  ID: 8e2eaa6a-2452-4dd1-9c32-44202e69dc7d                                                                       │
│                                                                                                                 │
│                                                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
🚀 Crew: crew
└── 📋 Task: bfa624c9-c5a8-49cd-adbd-4ebf815c197f
       Status: Executing Task...
🚀 Crew: crew
└── 📋 Task: bfa624c9-c5a8-49cd-adbd-4ebf815c197f
       Assigned to: About User
       Status: ❌ Failed
╭───────────────────────────────────────────────── Task Failure ──────────────────────────────────────────────────╮
│                                                                                                                 │
│  Task Failed                                                                                                    │
│  Name: bfa624c9-c5a8-49cd-adbd-4ebf815c197f                                                                     │
│  Agent: About User                                                                                              │
│                                                                                                                 │
│                                                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭───────────────────────────────────────────────── Crew Failure ──────────────────────────────────────────────────╮
│                                                                                                                 │
│  Crew Execution Failed                                                                                          │
│  Name: crew                                                                                                     │
│  ID: 8e2eaa6a-2452-4dd1-9c32-44202e69dc7d                                                                       │
│                                                                                                                 │
│                                                                                                                 │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-17-9c5d753f3927> in <cell line: 0>()
     47 )
     48 
---> 49 result = crew.kickoff(inputs={"question": "What is your favorite vacation destination?"})
     50 
     51 # ✅ Output print karna

8 frames
/usr/local/lib/python3.11/dist-packages/crewai/memory/contextual/contextual_memory.py in _fetch_user_context(self, query)
     95             str: Formatted user memories as bullet points, or an empty string if none found.
     96         """
---> 97         user_memories = self.um.search(query)
     98         if not user_memories:
     99             return ""

AttributeError: 'NoneType' object has no attribute 'search'

### Possible Solution

class ContextualMemory:
    def __init__(
        self,
        memory_config: Optional[Dict[str, Any]],
        stm: ShortTermMemory,
        ltm: LongTermMemory,
        em: EntityMemory,
        um: UserMemory,
    ):
        if memory_config is not None:
            self.memory_provider = memory_config.get("provider")
            #  Testing the Mem0 client as if provider is "mem0"
            if self.memory_provider == "mem0":   #<-----Here, but this is a manual workaround
                self.um = memory_config["config"].get("client")
               
                self.search_kwargs = memory_config["config"].get("search_kwargs", {})
            else:
                self.um = um
        else:
            self.memory_provider = None
            self.um = um

### Additional context

The issue might be due to self.um being None in ContextualMemory.

Vidit-Ostwal:
Fixes #2448, #2426

Key Logic changes done

Previous Code >>>
```python
if (
    self.memory_config and "user_memory" in self.memory_config
):  # Check for user_memory in config
    user_memory_config = self.memory_config["user_memory"]
    if isinstance(
        user_memory_config, UserMemory
    ):  # Check if it is already an instance
        self._user_memory = user_memory_config
    elif isinstance(
        user_memory_config, dict
    ):  # Check if it's a configuration dict
        self._user_memory = UserMemory(
            crew=self, **user_memory_config
        )  # Initialize with config
    else:
        raise TypeError(
            "user_memory must be a UserMemory instance or a configuration dictionary"
        )
else:
    self._user_memory = None  # No user memory if not in config
```

to use user memory, the user has to configure something like this:
```python
crew = Crew(
agents=[agent],
tasks=[task],
process = Process.sequential,
verbose=True,
memory=True,
memory_config={
"provider": "mem0",
"config": {"user_id": "john", api_key = os.environ["MEM0_API_KEY"]},
"user_memory" : {}
}
```

Key Considerations Regarding the Current Logic:

1. The current implementation checks for both self.memory_config and user_memory in memory_config. It may be more efficient to simply verify whether memory_config is provided with mem0 as the provider.
2. When user_memory is present in memory_config and is a valid instance of UserMemory, the logic correctly initializes self._user_memory with that instance. This part can remain unchanged.
3. The user_memory_config variable in the current code is not utilized inside the UserMemory class. It might be clearer to avoid checking whether user_memory_config is a valid dictionary, as the necessary parameters can later be retrieved by passing the crew itself.
4. Finally, instead of raising a TypeError, it may be more effective to initialize self._user_memory = Memory() from mem0 by default.

This is the final code logic, with all the neccessary changes >>>

```python
if (
    self.memory_config
):
    user_memory_config = self.memory_config.get("user_memory")
    if isinstance(
        user_memory_config, UserMemory
    ):  # Check if it is already an instance
        self._user_memory = user_memory_config
    else:
        self._user_memory = UserMemory(
            crew=self
        )
```

Now to use user memory, the configuration should be something like this 
```python
crew = Crew(
agents=[agent],
tasks=[task],
process = Process.sequential,
verbose=True,
memory=True,
memory_config={
"provider": "mem0",
"config": {"user_id": "john", api_key = os.environ["MEM0_API_KEY"]},
}
``` 
This initializes `self._user_memory = MemoryClient()` from mem0

or what we can do is to pass `user_memory` inside config
```python
crew = Crew(
agents=[agent],
tasks=[task],
process = Process.sequential,
verbose=True,
memory=True,
memory_config={
'user_memory' : UserMemory()
}
``` 

or at last the user can simply pass the `provider` and `user_id` in `config`
```python
crew = Crew(
agents=[agent],
tasks=[task],
process = Process.sequential,
verbose=True,
memory=True,
memory_config={
"provider": "mem0",
"config": {"user_id": "john"},
}
``` 
This initializes `self._user_memory = Memory()` from mem0

and if the user want to use some local configuration, they can pass `local_mem0_config` inside config
```python
crew = Crew(
agents=[agent],
tasks=[task],
process = Process.sequential,
verbose=True,
memory=True,
memory_config={
"provider": "mem0",
"config": {"user_id": "john", "local_mem0_config" : config},
}
``` 
This initializes `self._user_memory = Memory.with)config(local_mem0_config)` from mem0


joaomdmoura:

# Code Review Comment on the Mem0 OSS Support Implementation

## Overview
The implementation of Mem0 OSS support in the crewAI project marks a significant enhancement to memory management capabilities by incorporating both cloud-based and local configurations. The pull request introduces several vital updates that must be considered carefully.

## Key Findings

### Memory Storage Implementation

#### Positive Aspects
- The architecture presents a clean separation between `Memory` and `MemoryClient`, promoting maintainability and clarity.
- The system utilizes a graceful fallback to local Memory when the API key is not supplied, improving flexibility.
- Error handling has been properly addressed regarding configuration issues, providing better user feedback.

### Identified Issues & Recommendations

1. **Initialization Logic Improvement:**
   - The current constructor logic for memory initialization can be simplified.
   - **Current Implementation:**
     ```python
     def __init__(self, type, crew=None):
         ...
     ```
   - **Recommended Improvement:**
     - Apply a structured approach to initialization, encapsulating client creation in dedicated methods for clarity, debugging, and error management.
     - Implement a try-except block to catch initialization errors.

2. **Explicit Configuration Validation:**
   - The `_validate_local_mem0_config` should raise specific exceptions for clear user feedback on configuration errors instead of returning a boolean.
   - This improves the robustness of configuration checks while promoting better development practices.

3. **User Memory Integration Enhancement:**
   - Introduce a lazy initialization approach to improve performance and handle user configurations more seamlessly.
   - Using properties for `user_memory` can streamline the access while minimizing unnecessary computations.

### Testing Improvements

- Comprehensive unit tests are crucial for ensuring the reliability of functionality.
- **Current Issues:**
  - The tests fall short in mocking and validation scenarios.
- **Recommendations:**
  - Introduce edge case scenarios in unit tests to cover various configurations, such as invalid or incomplete memory setups.

### Documentation Recommendations

1. **Configuration Examples**: 
   - Detailed examples for cloud-based and local configurations should be included in the documentation to minimize setup errors.

2. **Validation Instructions**: 
   - Clear instructions must be provided regarding configuration validation, including missing keys and data types.

## Overall Recommendations
- Expand on error handling to capture and manage different exception cases effectively across the code.
- Enhance logging for better visibility into memory initialization processes and configurations.
- Utilize type hints comprehensively to support code readability and maintainability.
- Regularly update the documentation alongside code changes to reflect new features and configurations accurately.

The proposed changes not only enhance the reliability of the memory storage functionality but also align with best practices that facilitate long-term maintainability of the codebase. The project is on the right track, and with these recommendations, it can achieve even greater robustness.

src/crewai/memory/user/user_memory.py
```diff
@@ -12,14 +12,22 @@ class UserMemory(Memory):
...
+            memory_provider = None
+
+        if memory_provider == "mem0":
+            try:
+                from crewai.memory.storage.mem0_storage import Mem0Storage
+            except ImportError:
+                raise ImportError(
+                    "Mem0 is not installed. Please install it with `pip install mem0ai`."
+                )
+            storage = Mem0Storage(type="short_term", crew=crew)
```
lucasgomide: Why do u have changed type from `user` to `short_term`?

src/crewai/memory/user/user_memory.py
```diff
@@ -12,14 +12,22 @@ class UserMemory(Memory):
...
     def __init__(self, crew=None):
-        try:
-            from crewai.memory.storage.mem0_storage import Mem0Storage
-        except ImportError:
-            raise ImportError(
-                "Mem0 is not installed. Please install it with `pip install mem0ai`."
-            )
-        storage = Mem0Storage(type="user", crew=crew)
-        super().__init__(storage)
+        if crew and hasattr(crew, "memory_config") and crew.memory_config is not None:
```
lucasgomide: Assuming crew is an instance of Crew (it would be great to add a type hint to make that clear), it should always have a memory_config.

So if you're aiming for safe navigation, I'd recommend doing something like this
```suggestion
         memory_provider = None
        if memory_config := getattr(crew, "memory_config", None):
            memory_provider = memory_config.get("provider")
              if memory_provider != "mem0":
            raise ValueError("Invalid memory provider. Only 'mem0' is supported for user memory.")

        try:
            from crewai.memory.storage.mem0_storage import Mem0Storage
        except ImportError:
            raise ImportError(
                "Mem0 is not installed. Please install it with `pip install mem0ai`."
            )
        storage = Mem0Storage(type="short_term", crew=crew)
        super().__init__(storage)
```

lucasgomide: I see another issue with this code, it makes the crew parameter look required, since the memory_provider is being extracted from it. But is not required following by method signature.

I'm not sure if moving forward with this approach is the best idea. Maybe we should reconsider how we're structuring this to keep things more flexible

lucasgomide (commented):
@Vidit-Ostwal I believe your changes totally make sense! I just have a couple of clarification questions:

Should we add a warning or even a validation layer to prevent misleading usage of memory_config?
1. I mean, shouldn’t we ensure it’s mutually exclusive, for example, if the user_memory key is present, then other keys like provider, etc., shouldn’t be used together?
2. I think we're missing a test case to cover this usage:
```python
crew = MockCrew(memory_config={'user_memory': UserMemory()})
```

Let me know what you think!

Vidit-Ostwal:
> @Vidit-Ostwal I believe your changes totally make sense! I just have a couple of clarification questions:
> 
> Should we add a warning or even a validation layer to prevent misleading usage of memory_config?
> 
> 1. I mean, shouldn’t we ensure it’s mutually exclusive, for example, if the user_memory key is present, then other keys like provider, etc., shouldn’t be used together?
> 2. I think we're missing a test case to cover this usage:
> 
> ```python
> crew = MockCrew(memory_config={'user_memory': UserMemory()})
> ```
> 
> Let me know what you think!

If we do this, then we implicitly, restrict both `short_term_memory` and `entity_memory` not to use `mem0` as memory, when we try to use `user-memory = UserMemory()` 
Also I thought, making settings `provider` key explicitly as `mem0` would help user to understand that `mem0` is used for the memory now.

src/crewai/memory/user/user_memory.py
Vidit-Ostwal: Sorry, this a typo error. This should indeed be set to `user`. Thanks for letting me know on this one.

src/crewai/memory/user/user_memory.py
Vidit-Ostwal: > Assuming crew is an instance of Crew (it would be great to add a type hint to make that clear), it should always have a memory_config.
> 
> So if you're aiming for safe navigation, I'd recommend doing something like this

I agree—I initially considered a similar approach. However, I drew inspiration from how short_term and entity_memory utilize the mem0 memory and aimed to keep the codebase consistent.

Let me know if you'd like me to update the mem0 config section for short_term and entity_memory as well.

https://github.com/crewAIInc/crewAI/blob/0b58911153a545c495ed3667933d7ec01c097217/src/crewai/memory/short_term/short_term_memory.py#L22-L34

https://github.com/crewAIInc/crewAI/blob/0b58911153a545c495ed3667933d7ec01c097217/src/crewai/memory/entity/entity_memory.py#L19-L32

Vidit-Ostwal: > I see another issue with this code, it makes the crew parameter look required, since the memory_provider is being extracted from it. But is not required following by method signature.
> 
> I'm not sure if moving forward with this approach is the best idea. Maybe we should reconsider how we're structuring this to keep things more flexible

The crew parameter could indeed be removed, and only the necessary parameters could be passed. However, after reviewing how other memory types are initialized, I noticed that the crew parameter is consistently used throughout the codebase. To maintain consistency, I chose not to modify that part.

Vidit-Ostwal:
Still I think we need to change the implementation a bit. 

This is the explicit edge cases we need to cover:
1. `memory = True` when set, all the different memory should get initialised,
Solution >>> Removing the check on `memory_config` for user_memory.

2. UserMemory class directly takes a input of crew and no other parameter, User can not set the user_memory without setting the crew first. 
Solution >>> Take memory_config as an additional parameter same as how short term memory takes an input of embedder.

src/crewai/memory/user/user_memory.py
lucasgomide: Agreed about code consistency 👍🏼 though I believe the code would be slightly refactored - we don't need to address that in this PR.

Regarding to keep similar behaviors, both [entity_memory](https://github.com/crewAIInc/crewAI/blob/0b58911153a545c495ed3667933d7ec01c097217/src/crewai/memory/entity/entity_memory.py#L34) and [short_term_memory](https://github.com/crewAIInc/crewAI/blob/0b58911153a545c495ed3667933d7ec01c097217/src/crewai/memory/short_term/short_term_memory.py#L36) are capable of defining a default storage, so they don't run into the issue you're addressing in your latest code.

You're currently [raising an error](https://github.com/crewAIInc/crewAI/pull/2469/files/6b149a3eab89ab78e1aaf8b6994b900ab9f86904#diff-69ffcfdc0cdc6cf2223f058575642404df12f43a718995f0a59edea481648345R30) when memory_provider isn't found, I think that’s the main issue for now. In other worlds, if the `crew` was not provided in your code a execption will be always raised - whereas in the references memory class they will assign a default store

lucasgomide:
> Still I think we need to change the implementation a bit.
> 
> This is the explicit edge cases we need to cover:
> 
> 1. `memory = True` when set, all the different memory should get initialised,
>    Solution >>> Removing the check on `memory_config` for user_memory.
> 2. UserMemory class directly takes a input of crew and no other parameter, User can not set the user_memory without setting the crew first.
>    Solution >>> Take memory_config as an additional parameter same as how short term memory takes an input of embedder.
> 3. Decoupling the case when short term memory needs to be set a `mem0` memory and we need to pass user_memory as a custom UserMemory object.
>    Current memory_configuration would look like.
> 
> ```python
> memory_config={
> "provider": "mem0",
> "config": {"user_id": "john", "local_mem0_config" : config},
> 'user_memory' : UserMemory()
> }
> ```
> 
> The provider variable or any config variable, will not be used for user_memory they are explicitly used for short term memory.
> 
> 4. When `memory_config` has a parameter of `provider` as `mem0`, both short term memory and entity memory gets set to `mem0` type of memory, taking precedence over RAG memory which uses embedder config. I am not sure, whether doing in implicitly is a good idea or not.
> 
> Let me know @lucasgomide your opinion on how to resolve these.

Thanks for bringing up more context to me! Here's my toughts

2. It looks a good approach support others parameters such as EntityMemory does. This will help us to define a storage default value, if needed
3. IDK if I get what you said, but it doest not looks a exclusive issue for UserMemory, all other memories have the same issue. Right? I don't know if you should address that in this PR also.
4. I gotcha what you mean I have the same doubt.. I need more time/context to understand, IDK how to proceed with that

Additional toughts
IDK why we are not using [user_memory](https://github.com/crewAIInc/crewAI/blob/main/src/crewai/crew.py#L144) to set the user memory.. we could starting to use that

For the next steps, I’d suggest:

- Fix the user_memory configuration in this PR.
- Open an issue explaining the memory-related problems (as mentioned in point 4), so we can discuss them further there.

Let me know if you need any further clarification!

Vidit-Ostwal:
> Additional toughts IDK why we are not using [user_memory](https://github.com/crewAIInc/crewAI/blob/main/src/crewai/crew.py#L144) 

Agreed we should start using the user_memory parameter for setting the user memory. 

> For the next steps, I’d suggest:
> Fix the user_memory configuration in this PR.

Bit confused, on this which issue are we talking about to be fixed in this PR. 
Shall I keep the user_memory configuration via memory_config in this PR or add additional parameter in the UserMemory for config?

lucasgomide:
@Vidit-Ostwal sorry about delay to reply
I'll talk to Brandon about that soon. Can we keep this PR on hold for a while?

Vidit-Ostwal:
> @Vidit-Ostwal sorry about delay to reply I'll talk to Brandon about that soon. Can we keep this PR on hold for a while?

Sure, no worries.

lucasgomide:
@Vidit-Ostwal After talking with Brandon, we decided to take a simpler approach.
Basically, we won’t add new features to UserMemory for now. Instead, we’ll improve the documentation to avoid any confusion or misleading usage.
We both agree that the current codebase this isn't great, but we’re not going to refactor it just yet, since there are plans to either clean it up or deprecate it in the future. Given that, the easiest path forward would be:

1. Fix the issue to prevent a crash when memory is set: an early return in `_fetch_user_context` should be enough for now.
2. Improve the documentation to clearly state that, in order to use user_memory, it must be explicitly set inside memory_config.

Let me know if that works for you!

Vidit-Ostwal:
> 1. _fetch_user_context

Hi @lucasgomide, thanks for this.

Here are the following changes I am thinking to do
1. Early return on `_fetch_user_context`
2. To use user_memory user has to set 
```python
memory_config = {
provider = 'mem0', neccessary attribute
config = {user_id : 'Alna' ### Neccessary for user_memory
, org_id: "" # Optional, 
project _id: "" # Optional,
local_mem0_config: "" # Optional}
user_memory = True #Flag to use user_memory
```

Will improve the documentation and test cases accordingly.
Let me know if I am missing something, or need to change any particular part. 
ty.

Vidit-Ostwal:
@lucasgomide 
> ```python
> memory_config={
>   "provider": "mem0",
>   "config": {"user_id": "john", api_key = os.environ["MEM0_API_KEY"]},
>   "user_memory" : {} | UserMemory(..)
> }

Directly initializing a UserMemory() object is not possible, as self.memory_config get initialized to crew.memory_config, which would be `None` in our case.

https://github.com/crewAIInc/crewAI/blob/f845fac4da7ec0255f966bf5f5999e713f0535e5/src/crewai/memory/storage/mem0_storage.py#L25-L27

https://github.com/crewAIInc/crewAI/blob/f845fac4da7ec0255f966bf5f5999e713f0535e5/src/crewai/memory/storage/mem0_storage.py#L104-L110

Let me know how we can proceed with this. 

I have added the documentation without this case at this moment. Some additional test cases also needs to be added.
Currently this is in WIP

lucasgomide:
This PR has almost 30 commits, can you squash them before get this PR ready to review? 🥺

Vidit-Ostwal:
Hi @lucasgomide, I am so sorry, I messed up this PR a bit.
Will need some time to sort this out.

Vidit-Ostwal:
I don't understand, I squashed all the commits, 
I don't know why it showed changes by 100000 lines of code, I thought to reverse this, and kaboom all the changes are gone .
Will sort this out. Remember all the changes

PR #2510, opened 2025-04-01
Add support for External Memory (the future replacement for UserMemory)

lucasgomide:


src/crewai/memory/user/user_memory.py
```diff
@@ -12,6 +13,12 @@ class UserMemory(Memory):
     """
 
     def __init__(self, crew=None):
+        warnings.warn(
+            "UserMemory is deprecated and will be removed in a future version. "
+            "Please use ExternalMemory instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
lorenzejay, line 21: love this

lorenzejay (approved):

sethi-ishmeet:
Do we know which version of CrewAI is this going in? We just recently started implementing CrewAI in our workflows and memory is a big part of it. Would hate to change things much deeper in our release. 

Sorry for commenting on the PR. Couldn't find an issue related to this.

lorenzejay:
@sethi-ishmeet should be on the next version of crew. Btw, there should be no breaking changes here. the next breaking change will reflect in 2 versions from now where we go

user_memory -> external_memory cc: @lucasgomide

PR #2567, opened 2025-04-10
fix: correct parameter name in crew template test function

bitstreamshaman:
This commit resolves an issue in the crew template generator where the test()  function incorrectly uses 'openai_model_name' as a parameter name when calling  Crew.test(), while the actual implementation expects 'eval_llm'.

The mismatch causes a TypeError when users run the generated test command: "Crew.test() got an unexpected keyword argument 'openai_model_name'"

This change ensures that templates generated with 'crewai create crew' will  produce code that aligns with the framework's API.


lucasgomide (approved):
good call. ty!

PR #2570, opened 2025-04-10
added condition to check whether _run function returns a coroutine ob…

### LINKED ISSUE #2538: [BUG] SnowflakeSearchTool returns a coroutine object rather than the query results
### Description

Unlike other tools in the package (e.g., `BraveSearchTool`, `DatabricksQueryTool`) which implement synchronous `_run` methods, the `SnowflakeSearchTool` only has an async implementation without a synchronous wrapper.
When agents use the tool, they get a coroutine object (coroutine object SnowflakeSearchTool._run at 0x13b1a28c0) instead of the actual query results.

Alternatively, maybe we could make the calling agent use an asynchronous context but I figured that `SnowflakeSearchTool` appears to be the only tool in the package that implements an asynchronous `_run` so maybe it's easier to fix the tool itself.

### Steps to Reproduce

1. Set up a CrewAI project with Snowflake integration
2. Configure the SnowflakeConfig with valid credentials
3. Create an Agent with SnowflakeSearchTool
4. Create a task that uses the Snowflake tool
5. Run the crew

### Expected behavior

The agent should execute the Snowflake query and return the actual query results (e.g., a formatted table of data or error message from Snowflake).

### Screenshots/Code snippets

n/a

### Operating System

### Python Version

3.12

### crewAI Version

0.108.0

### crewAI Tools Version

0.40.0

### Virtual Environment

Venv

### Evidence

```coroutine object SnowflakeSearchTool._run at 0x13b1a28c0```

### Possible Solution

Add a synchronous `run` method to `SnowflakeSearchTool` that properly awaits the coroutine or make the SnowflakeSearchTool synchronous.

### Additional context

macOS Version 15.4 (24E248) (Sequoia)

### PR DESCRIPTION
**Pull request by Vidit-Ostwal**
Fixes #2538

### CODE REVIEW

**Comment by Vidit-Ostwal** [comment]
Fixes #2434

File: tests/tools/test_base_tool.py [hunk 1]
```diff
@@ -1,7 +1,14 @@
...
+import asyncio
+import inspect
+import unittest
+from typing import Any, Callable, Dict, List
+from unittest.mock import patch
 
 from crewai.tools import BaseTool, tool
 
+# Import the necessary classes
+from crewai.tools.base_tool import BaseTool
```
lucasgomide: Have you imported at line #7 ?

tests/tools/test_base_tool.py
```diff
@@ -122,3 +129,93 @@ def my_tool_with_default(question: str) -> str:
...
+    name: str = "async_tool"
+    description: str = "An asynchronous tool for testing"
+
+    async def _run(self, input_text: str) -> str:
+        """Process input text asynchronously."""
+        await asyncio.sleep(0.1)  # Simulate async operation
+        return f"Processed {input_text} asynchronously"
+
+
+class BaseToolCoroutineTest(unittest.TestCase):
```
lucasgomide: Do you mind keeping the same test definition standard?

tests/tools/test_base_tool.py
```diff
@@ -122,3 +129,93 @@ def my_tool_with_default(question: str) -> str:
...
     assert converted_tool.result_as_answer is False
+
+
+class SyncTool(BaseTool):
+    """Test implementation with a synchronous _run method"""
+    name: str = "sync_tool"
+    description: str = "A synchronous tool for testing"
+
+    def _run(self, input_text: str) -> str:
+        """Process input text synchronously."""
```
lucasgomide, line 137: Please remove those comments (you have some spread in your code) they just repeat what the next line does and don’t add any value

lucasgomide (commented):
Tks for collaboration (: 
i liked your approach. I left some comments on your tests.

tests/tools/test_base_tool.py
Vidit-Ostwal: Resolved.

tests/tools/test_base_tool.py
Vidit-Ostwal: done

tests/tools/test_base_tool.py
Vidit-Ostwal: Agreed, resolved.

tests/tools/test_base_tool.py
```diff
@@ -122,3 +126,68 @@ def my_tool_with_default(question: str) -> str:
...
+        """Test that asyncio.run is called when using async tools."""
+        async_tool = AsyncTool()
+        
+        with patch('asyncio.run') as mock_run:
+            mock_run.return_value = "Processed test asynchronously"
+            async_result = async_tool.run(input_text="test")
+            
+            mock_run.assert_called_once()
+            self.assertEqual(async_result, "Processed test asynchronously")
+
```
lucasgomide: Those tests aren’t automatically collected by `pytest`, probably because they’re methods within a class. You can just remove those classes and write the tests like the others in this file.
You have to use `assert` instead of `assertEqual` due we are not using unittest

lucasgomide (changes requested):
Requesting changes because your tests aren’t being properly executed.

tests/tools/test_base_tool.py
Vidit-Ostwal: Understood, Can you share how you figured it out that the test cases are not even running.
I am still learning, would like to know how you figured that out.
This has been resovled.

lucasgomide: I assumed that was because you removed the `unittest.TestCase` inheritance - first flag was: I think the test are not running properly. 
To confirm, I checked out your branch and ran the test file; only 5 tests were collected, but I was expecting 9. That’s all!

lucasgomide: btw tks for quickly fix

lucasgomide (approved):

tests/tools/test_base_tool.py
Vidit-Ostwal: No worries, happy to help

PR #2573, opened 2025-04-10
[Docs] Create output folder if it doesn't exist

JesseRWeigel:
When running this project (guide_creator_flow) from the Flows documentation, I got an error because the `output` folder had not been created. 

I added a line to check if the output folder exists and create it if needed.


lucasgomide (approved):
appreciate that, ty!

PR #2579, opened 2025-04-10
Raise exception when flow fails

vinibrsl:


src/crewai/flow/flow.py
```diff
@@ -1043,6 +1043,7 @@ async def _execute_single_listener(self, listener_name: str, result: Any) -> Non
             import traceback
 
             traceback.print_exc()
+            raise
```
lorenzejay, line 1046: no more silently failing

lorenzejay (approved):
no more silently failing

PR #2580, opened 2025-04-10
Fix Mistral Issue:

### LINKED ISSUE #2571: [BUG] system_prompt incompatibility
### Description

I use the Mistral model (deployed on local server, access via custom API), which doesn't support the system_prompt. I'm also using output_pydantic=True in one of my tasks. When the LLM generates output that doesn't align with the Pydantic model, CrewAI uses an internal prompt to ask the LLM to regenerate the output. This internal prompt includes a system_prompt, and after that, my LLM crashes.

P.S. I'm aware of the use_system_prompt=False flag when initializing an Agent, but it doesn't resolve the issue.

### Steps to Reproduce

1. Initialize an Agent with an LLM that does not support system_prompt.

2. Initialize a Task with output_pydantic=True.

3. Run your crew until the model generates an invalid Pydantic output.

4. It will crash after CrewAI uses a prompt to request a regeneration attempt.

### Expected behavior

The model doesn't crash

### Screenshots/Code snippets

```
from myfile import MyPydanticModel, my_config, my_condition
from crewai.tasks.conditional_task import ConditionalTask
from crewai.project import task

@task
def resolve_requests(self) -> ConditionalTask:
    return ConditionalTask(
        condition=my_condition,
        config=my_config,
        verbose=True,
        output_pydantic=MyPdanticModel
)

```

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

0.108.0

### crewAI Tools Version

0.38.1

### Virtual Environment

Venv

### Evidence

Error code: 400 - {'object': 'error', 'message': 'Conversation roles must alternate user/assistant/...', 'type': 'BadRequestError', 'param': None, 'code': 400}

### Possible Solution

When I comment 31-32 rows in /src/crewai/utilities/converter.py:

{"role": "system", "content": self.instructions},
{"role": "user", "content": self.text}

and add this row:
{"role": "user", "content": self.instructions + "/n" + self.text}

All works as expected.

### Additional context

-

Vidit-Ostwal:
Fixes #2571


Vidit-Ostwal:
Added an additional check to convert any `system` role to `assistant` role.

src/crewai/llm.py
```diff
@@ -857,7 +866,7 @@ def _format_messages_for_provider(
                     formatted_messages.append(msg)
             return formatted_messages
 
-        # Handle Mistral models - they require the last message to have a role of 'user' or 'tool'
+        # Handle Mistral models - they require the last message to have a role of 'user' or 'tool
         if "mistral" in self.model.lower():
```
lucasgomide: what about centralize the model checking? You could re-use `_is_mistral_model` here.

src/crewai/utilities/converter.py
```diff
@@ -26,12 +26,20 @@ def to_pydantic(self, current_attempt=1) -> BaseModel:
...
-                        {"role": "user", "content": self.text},
-                    ]
-                )
+                if self.llm._is_mistral_model():
+                    # Mistral models require a different approach
+                    response = self.llm.call(
+                        [
+                            {"role": "user", "content": self.instructions + "/n" + self.text}
+                        ]
+                    )
```
lucasgomide, line 35: since we are converting all system message to assistant, do you need to touch at this point?
do you have any docs reference about that?

src/crewai/llm.py
Vidit-Ostwal: I agree, will fix this.

src/crewai/utilities/converter.py
Vidit-Ostwal: Exactly, this is need actually, if we don't do this. 
what will happen is the first role in the message would be `system` and that would be converted to `assistant`, by the later logic.
The mistral wants a `user` role, as the first message role.

Check this out.
https://docs.mistral.ai/capabilities/guardrailing/#conversational-endpoint

Vidit-Ostwal: the system to assistant conversion, is just an additional check I am making at this moment.

I think I should remove that check. 
It should better fail, so that we can come back and fix, let me know what you think.

Vidit-Ostwal:
@lucasgomide shall I close this PR?
The issue #2571 doesn't seem to exist.

PR #2581, opened 2025-04-10
feat: remove unused code and change ToolUsageStarted event place

pythonbyte:


src/crewai/tools/tool_usage.py
```diff
@@ -166,6 +167,26 @@ def _use(
                 if self.task:
                     self.task.increment_tools_errors()
 
+        if self.agent:
+            event_data = {
+                "agent_key": self.agent.key if self.agent else None,
+                "agent_role": self.agent.role if self.agent else None,
```
lucasgomide: you're checking already if agent exists

```suggestion
                "agent_key": self.agent.key,
                "agent_role": self.agent.role,
```

lucasgomide (commented):
Are we going to stop emitting events when there’s a tool failure?

src/crewai/tools/tool_usage.py
pythonbyte: Indeed, nice catch

pythonbyte:
> Are we going to stop emitting events when there’s a tool failure?

Not actually, we are going to move the logic of Toolusage events, the Failure is already in place

src/crewai/tools/tool_usage.py
```diff
@@ -166,6 +166,21 @@ def _use(
...
+                "tool_name": self.action.tool,
+                "tool_args": self.action.tool_input,
+                "tool_class": self.action.tool,
+                "agent": self.agent,
+            }
+
+            if self.agent.fingerprint:
+                event_data.update(self.agent.fingerprint)
+
+            crewai_event_bus.emit(self,ToolUsageStartedEvent(**event_data))
```
lorenzejay, line 182: nice now we have the reference of tool_usage here

lucasgomide (approved):

lorenzejay:
i think just drop the cassete with that test and we are gold @pythonbyte : `test_tool_execution_error_event`

lorenzejay (approved):
LFG!

PR #2582, opened 2025-04-10
Unblock LLM(stream=True) to work with tools

lucasgomide:
This PR introduces streaming tool call support and improves type safety in the LLM implementation. The changes include:

1. **Streaming Tool Call Support**:
   - Enhanced the `_handle_streaming_tool_calls` method to properly handle streaming tool calls
   - Added support for accumulating tool arguments and emit events during streaming

2. **Testing**:
   - Replace pytest-vcr by pytest-recording to allow recording cassettes from streaming responses
   - Remove @skip_streaming_in_ci because we are able handle those response properly

This PR would help improve the reliability and type safety of the LLM implementation while adding support for streaming tool calls, which is an important feature for real-time AI interactions.

https://github.com/user-attachments/assets/ac148c00-cf6a-47cf-ade8-14089054463d


src/crewai/llm.py
```diff
@@ -371,6 +374,15 @@ def _handle_streaming_response(
...
         usage_info = None
+        tool_calls = None
+        accumulated_tool_args: DefaultDict[int, SimpleNamespace] = defaultdict(
+            lambda: SimpleNamespace(
+                function=SimpleNamespace(
+                    name="",
+                    arguments="",
+                ),
+            ),
+        )
```
vinibrsl: Would it be overkill to use Pydantic here? Or a dataclass?

src/crewai/llm.py
```diff
@@ -568,6 +593,52 @@ def _handle_streaming_response(
...
+                    )
+
+                    # Execute the tool call
+                    return self._handle_tool_call(
+                        [accumulated_tool_args[tool_call.index]],
+                        available_functions,
+                    )
+                except json.JSONDecodeError:
+                    # If JSON is incomplete, continue accumulating
+                    continue
```
vinibrsl: I think your code is quite self explanatory, so we can remove comments if you agree.

src/crewai/utilities/events/llm_events.py
```diff
@@ -46,3 +46,4 @@ class LLMStreamChunkEvent(BaseEvent):
 
     type: str = "llm_stream_chunk"
     chunk: str
+    tool_call: Optional[dict] = None
```
vinibrsl: Is there anything we can do to type this even further? I ask that because we serialize events in some places, and we could benefit from type hints.

src/crewai/llm.py
lucasgomide: good call!

src/crewai/utilities/events/llm_events.py
lucasgomide: We can try! I tested it with a several models and the response always looks the same structure [this model](https://github.com/BerriAI/litellm/blob/5c078af73895212e1d5b1e40d372745aa8ac50b2/litellm/types/utils.py#L415)

Here is an example of the instance
```python
ChatCompletionDeltaToolCall(id='UKpdvI3NK', function=Function(arguments='{"location": "San Francisco, CA"}', name='get_weather'), type='function', index=0)
```

vinibrsl (approved):

vinibrsl (approved):

src/crewai/llm.py
```diff
@@ -568,6 +598,49 @@ def _handle_streaming_response(
             )
             raise Exception(f"Failed to get streaming response: {str(e)}")
 
+    def _handle_streaming_tool_calls(
+        self,
+        tool_calls: List[Any],
```
lorenzejay: should this be type `ToolCall`?

src/crewai/llm.py
```diff
@@ -568,6 +598,49 @@ def _handle_streaming_response(
...
+        accumulated_tool_args: DefaultDict[int, AccumulatedToolArgs],
+        available_functions: Optional[Dict[str, Any]] = None,
+    ) -> None | str:
+        for tool_call in tool_calls:
+            if tool_call.function.name:
+                accumulated_tool_args[tool_call.index].function.name = (
+                    tool_call.function.name
```
lorenzejay: better readability?

`current_tool_accumulator = accumulated_tool_args[tool_call.index]`


then 
```py
current_tool_accumulator.name = tool_call.function.name
```

src/crewai/llm.py
lucasgomide: Good call!

src/crewai/llm.py
lucasgomide: Nice!!!

src/crewai/llm.py
lucasgomide: actually, `tool_calls` comes from `LiteLLM` so it should be list of `ChatCompletionDeltaToolCall`

lorenzejay (approved):
awesome work here!

PR #2597, opened 2025-04-12
feat: Enhance memory system with isolated memory configuration

lucasgomide:
This PR introduces several improvements to the memory system:

1. Support for isolated memory configuration
   - Users can now define individual memory types independently. This change makes it easier to use a specific memory type without unintentionally enabling all others. Previously, setting memory=True would implicitly configure all available memories (like LTM and STM), which might not be ideal in all cases. 
   For example, when building a chatbot that only needs an external memory, users were forced to also configure LTM and STM — which rely on default OpenAPI embeddings — even if they weren’t needed. 
   With this update, users can now define a single memory in isolation, making the configuration process simpler and more flexible.

2. Enhanced memory documentation
   - Improved documentation for memory system usage and configuration
   - Better clarity on memory types and their interactions

3. Warning system for memory configuration
   - Added warning when long-term memory is enabled without entity memory
   - Helps users avoid potential memory configuration issues


src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -114,6 +113,14 @@ def _create_long_term_memory(self, output) -> None:
             except Exception as e:
                 print(f"Failed to add to long term memory: {e}")
                 pass
+        elif (
+            self.crew
+            and self.crew._long_term_memory
+            and self.crew._entity_memory is None
+        ):
+            print(
```
lucasgomide: not sure this is the best way to warn..

src/crewai/memory/contextual/contextual_memory.py
```diff
@@ -53,6 +53,10 @@ def _fetch_stm_context(self, query) -> str:
         Fetches recent relevant insights from STM related to the task's description and expected_output,
         formatted as bullet points.
         """
+
+        if self.stm is None:
```
lucasgomide, line 57: since short-term memory might be null, we need to ensure safety

src/crewai/memory/contextual/contextual_memory.py
```diff
@@ -67,6 +71,10 @@ def _fetch_ltm_context(self, task) -> Optional[str]:
         Fetches historical data or insights from LTM that are relevant to the task's description and expected_output,
         formatted as bullet points.
         """
+
+        if self.ltm is None:
```
lucasgomide, line 75: since long-term memory might be null, we need to ensure safety

src/crewai/memory/contextual/contextual_memory.py
```diff
@@ -86,6 +94,9 @@ def _fetch_entity_context(self, query) -> str:
         Fetches relevant entity information from Entity Memory related to the task's description and expected_output,
         formatted as bullet points.
         """
+        if self.em is None:
```
lucasgomide, line 97: since entity memory might be null, we need to ensure safety

src/crewai/agent.py
```diff
@@ -156,6 +156,20 @@ def set_knowledge(self, crew_embedder: Optional[Dict[str, Any]] = None):
...
+        return (
+            self.crew.memory
+            or self.crew.memory_config
+            or self.crew._short_term_memory
+            or self.crew._long_term_memory
+            or self.crew._entity_memory
+            or self.crew._user_memory
+            or self.crew._external_memory
+        )
+
```
lorenzejay: thoughts on this:

```python
def _is_any_available_memory(self) -> bool:
    """Check if any memory is available."""
    if not self.crew:
        return False
    
    memory_attributes = [
        "memory",
        "memory_config",
        "_short_term_memory",
        "_long_term_memory",
        "_entity_memory", 
        "_user_memory",
        "_external_memory"
    ]
    
    return any(getattr(self.crew, attr) for attr in memory_attributes)
```

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
lorenzejay: what about Logger ?

you can use 

```python
self._printer.print(
            content="Long term memory is enabled, but entity memory is not enabled. Please configure entity memory or set memory=True to automatically enable it.", 
            color="bold_yellow"
        )
```

src/crewai/crew.py
```diff
@@ -275,46 +275,48 @@ def set_private_attrs(self) -> "Crew":
...
+    def _initialize_default_memories(self):
+        self._long_term_memory = self._long_term_memory or LongTermMemory()
+        self._short_term_memory = self._short_term_memory or ShortTermMemory(
+            crew=self,
+            embedder_config=self.embedder,
+        )
+        self._entity_memory = self.entity_memory or EntityMemory(
+            crew=self, embedder_config=self.embedder
+        )
+        if (
```
lorenzejay, line 279: can we make this a separate method?

like `def _initialize_user_memory(self):`

src/crewai/crew.py
```diff
@@ -275,46 +275,48 @@ def set_private_attrs(self) -> "Crew":
...
-                if isinstance(
-                    user_memory_config, dict
-                ):  # Check if it's a configuration dict
-                    self._user_memory = UserMemory(crew=self)
-                else:
-                    raise TypeError("user_memory must be a configuration dictionary")
-            else:
-                self._user_memory = None  # No user memory if not in config
+            self._initialize_default_memories()
+
```
lorenzejay, line 322: `self._initialize_user_memory() `

src/crewai/agents/agent_builder/base_agent_executor_mixin.py
```diff
@@ -114,6 +113,15 @@ def _create_long_term_memory(self, output) -> None:
...
                 pass
+        elif (
+            self.crew
+            and self.crew._long_term_memory
+            and self.crew._entity_memory is None
+        ):
+            self._printer.print(
+                content="Long term memory is enabled, but entity memory is not enabled. Please configure entity memory or set memory=True to automatically enable it.",
+                color="bold_yellow",
+            )
```
lorenzejay, line 124: love it!

lorenzejay (approved):

lorenzejay (approved):

Issue #2263, opened 2025-03-03
[BUG] CrewBase being a decorator causes linting errors


dga-nagra:
### Description

If you use `@CrewBase`, and then access `self.agents` or `self.tasks`, you will get an error saying that these don't exist. Why isn't `CrewBase` a class/metaclass to be inherited ?

I also struggle to understand the benefit of using `CrewBase` as we could just create a function that returns the crew if that's needed. I have the same issue with `Knowledge` over `Tools`. This would be great if that was specified in the documentation.

The documentation is currently only examples, if I take the page of tools, we don't have a view on what options are availables.

### Steps to Reproduce

1. Use `@CrewBase`
2. Access `self.agents` or `self.tasks`
3. you will get an error saying that these don't exist.

### Expected behavior

The linter is not reporting issues

### Screenshots/Code snippets

```python

from crewai import Agent, Crew, Task, Process
from crewai.project import CrewBase, agent, task, crew, before_kickoff, after_kickoff


@CrewBase
class YourCrewName:
    """Description of your crew"""

    @agent
    def agent_one(self) -> Agent:
        return Agent(...)

    @task
    def task_one(self) -> Task:
        return Task(...)

    @crew
    def crew(self) -> Crew:
        return Crew(
            agents=self.agents,  # Linting error here
            tasks=self.tasks,       # and here
            process=Process.sequential,
            verbose=True,
        )
```

### Operating System

Ubuntu 20.04

### Python Version

3.11

### crewAI Version

0.102.0

### crewAI Tools Version

0.36.0

### Virtual Environment

Venv

### Evidence

![Image](https://github.com/user-attachments/assets/828bbeff-5d7a-4920-a198-5b4d874edac3)

### Possible Solution

Use a class/metaclass instead of the decorator

### Additional context

none

PR #2603, opened 2025-04-14
fix: add type hints and ignore type checks for config access

lucasgomide:
Closes https://github.com/crewAIInc/crewAI/issues/2263


docs/concepts/crews.mdx
```diff
@@ -52,12 +52,16 @@ After creating your CrewAI project as outlined in the [Installation](/installati
...
-
+from crewai.agents.agent_builder.base_agent import BaseAgent
+from typing import List
 
 @CrewBase
 class YourCrewName:
     """Description of your crew"""
 
+    agents: List[BaseAgent]
+    tasks: List[Task]
```
lorenzejay, line 63: NIT: do we need this? annotation crew give us this.

lorenzejay (approved):

docs/concepts/crews.mdx
lucasgomide: the point is this warning comes from static analysis code by Pylance .. `self.agents & self.task` are  dynamically assigned inside the wrapper() function of the @crew decorator. 

Screenshot 2025-04-14 at 1 58 04 PM

to fix this warning we have to set those attributes as instance attribute inside the class body or `__init__`

Issue #2599, opened 2025-04-14
[BUG] Mem0 (local config) - Memory.search() got an unexpected keyword argument 'metadata'


Jainish-S:
### Description

I am currently using local mem0 with the following config
```
local_mem0_config = {
        "vector_store": {
            "provider": "qdrant",
            "config": {
                "collection_name": "user_memory",
                "host": "192.168.0.222",
                "port": 6333,
                "embedding_model_dims": "1536",
            },
        },
        "llm": {
            "provider": "openai",
            "config": {"api_key": os.getenv("OPENAI_API_KEY"), "model": "gpt-4o-mini"},
        },
        "embedder": {
            "provider": "openai",
            "config": {
                "api_key": os.getenv("OPENAI_API_KEY"),
                "model": "text-embedding-3-small",
            },
        },
    }
```

My crew is defined as this
```
Crew(
            agents=self.agents,
            tasks=self.tasks,
            process=Process.sequential,
            verbose=True,
            memory=True,
            external_memory=ExternalMemory(
                embedder_config={
                    "provider": "mem0",
                    "config": {
                        "user_id": "jainish",
                        "local_mem0_config": self.local_mem0_config,
                    },
                }
            ),
            output_log_file="logs.json",
        )
```

The full traceback
```
Traceback (most recent call last):
  File "/Users/JainishSavalia/work/playground/crewAI/examples/main.py", line 27, in run_crew
    result = crew.kickoff(inputs=inputs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/JainishSavalia/work/playground/crewAI/src/crewai/crew.py", line 646, in kickoff
    result = self._run_sequential_process()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/JainishSavalia/work/playground/crewAI/src/crewai/crew.py", line 758, in _run_sequential_process
    return self._execute_tasks(self.tasks)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/JainishSavalia/work/playground/crewAI/src/crewai/crew.py", line 861, in _execute_tasks
    task_output = task.execute_sync(
                  ^^^^^^^^^^^^^^^^^^
  File "/Users/JainishSavalia/work/playground/crewAI/.venv/lib/python3.11/site-packages/opentelemetry/instrumentation/crewai/instrumentation.py", line 61, in wrapper
[...]

PR #2604, opened 2025-04-14
Fix Mem0 OSS

Dev-Khant:
Closes https://github.com/crewAIInc/crewAI/issues/2599


src/crewai/memory/storage/mem0_storage.py
```diff
@@ -116,8 +116,10 @@ def search(
 
         # Discard the filters for now since we create the filters
         # automatically when the crew is created.
+        if hasattr(self, "llm"):
+            del params["metadata"], params["output_format"]
         results = self.memory.search(**params)
-        return [r for r in results if r["score"] >= score_threshold]
+        return [r for r in results["results"] if r["score"] >= score_threshold]
```
lucasgomide, line 125: great work!


Just add a few test to cover line 122 & 119, pls

Dev-Khant: Thanks for reviewing. Added test to check line 119 and there's already a test which checks the search functionality: https://github.com/crewAIInc/crewAI/blob/main/tests/memory/user_memory_test.py#L33 and it also passes

lucasgomide: The CI tests and type checker are falling. Could u handle that?

Dev-Khant: Fixed both

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -116,8 +118,10 @@ def search(
 
         # Discard the filters for now since we create the filters
         # automatically when the crew is created.
+        if hasattr(self, "llm"):
```
lucasgomide: Something feels off here… I think your test might be too addicted, because it really shouldn’t work like, ever!
I believe you should be checking `hasattr(self.memory, "llm")` instead of `self`, since even self has a defined llm model.
What bothers me most is that even when one test fails, it still goes unnoticed that’s kind of serious.

If I’m right about this, it’d be a good idea to revisit the tests

lucasgomide (changes requested):

src/crewai/memory/storage/mem0_storage.py
Dev-Khant: Hey yes you are correct `hasattr(self, "llm")` was accessing the LLM from crewai and not from Mem0. And that;s why the tests also passed. But now I have made the fix for it.

src/crewai/memory/storage/mem0_storage.py
lucasgomide: @Dev-Khant Your added test still not testing what should be testing.
To ensure that I removed `hasattr(self.memory, "llm")` clause from `mem0_storage` and ran your added test, still passing. 

It is happens because you added test `test_search_with_llm` is mocking the `user_memory.storage`. 

Here's a few suggestions to fix it:

1. You have to cover mem0_storage, so add a test into `test_mem0_storage.py`
2. Add test to cover search and save methods 
3. Use `unittest.mock.patch` to mock the `memory.save` and `memory.search` calls. Example: `patch("mem0.Memory.search") as mock_method`
4. use `assert_called_once_with` to ensure the propagated parameters. Example: `mock_method.assert_called_once_with(agent_id=, metadata=, ...)`

Let me know if you need help with those test

Dev-Khant: Let me know if the latest tests meet your expectations and if I’ve understood you correctly. If not, could you please help with the tests?

tests/memory/user_memory_test.py
```diff
@@ -65,4 +64,26 @@ def test_save_and_search(user_memory):
     with patch.object(UserMemory, 'search', return_value=expected_result) as mock_search:
         find = UserMemory.search("test value", score_threshold=0.01)[0]
         mock_search.assert_called_once_with("test value", score_threshold=0.01)
-        assert find == expected_result[0]
\ No newline at end of file
+        assert find == expected_result[0]
+
+def test_search_with_llm(user_memory):
```
lucasgomide: we can drop this one.. since is useless

Dev-Khant: Resolved.

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -88,15 +88,17 @@ def save(self, value: Any, metadata: Dict[str, Any]) -> None:
             }
 
         if params:
-            self.memory.add(value, **params | {"output_format": "v1.1"})
+            if not hasattr(self.memory, "llm"):
+                params["output_format"] = "v1.1"
```
lucasgomide: qq, double checking.. 
If the memory does not support llm, you set `output_format`, right? But I noticed there’s another logic at line 121 in the search method over there, you’re doing: “if memory has llm, delete parameters”

That feels a bit confusing. Can you confirm if that’s expected behavior, or if we might need to align those two cases?

Vidit-Ostwal: @lucasgomide, I think what's happening in here is a class comparision, whether the memory object belongs to Memory or Memory Client 
Basically MemoryClient does not have any llm attribute to it, so the first check is to make sure that if `self.memory` is not a MemoryClient class object, we need to have a parameter called `output_format`. and in the second case check I believe they are checking whether it's a object of Memory class, and if yes then we do not need this parameter. 

Bit confusing how `mem0` is configured. 
Still author can confirm more on this.

One suggestion from my side would be if I am right about class comparision, would be to that there are better way to check whether a object belongs to a particular class, like `isinstance`, I think better to use this for better code visibility. Lucas can surely add more context on this part.

lucasgomide: @Vidit-Ostwal I also prefer class comparison, but even after changing it, the question remains… I’m asking because the test didn’t confirm that.
From reading this line, I expected that when running with Memory, the output_format shouldn’t be included in the `add()` parameters - but is not happing (on test at least)

lucasgomide: @Dev-Khant can you help with that?

Dev-Khant: Hey @lucasgomide, 

`output_format="v1.1"` -> {"results": <memories>} 
`output_format="v1.0"` -> [<memories>]

We’ve now fully deprecated the `output_format` parameter from the OSS version, though it’s still in the process of being removed from the Platform. To maintain consistency in output format, I’ve added `output_format="v1.1"` on line 92.

On line 121, there's a check to determine if the call is targeting OSS. If so, it removes the `output_format` parameter since it's deprecated in OSS but still required for the Platform or Mem0 client.

lucasgomide: @Dev-Khant that is the point.. 

It's a bit tricky because in [your tests](https://github.com/crewAIInc/crewAI/pull/2604/files#diff-59f5e3e37e01e31b0274361fea3e2ad2db5934f78617704d4198b45d506a32fdR163)
both implementation are propagating `output_format`, but by reading your code it should be sent **ONLY** when the memory does not has `llm`. 
This is the last touchable point that isnot clear to me

Dev-Khant: Hey @lucasgomide Then can you please help with the test?

Vidit-Ostwal:
Hi @Dev-Khant, I have made some changes and tried to raise a PR in your forked repo, can you once update your branch with crewai main and check whether only those two files are getting changed in my PR

Dev-Khant:
> Hi @Dev-Khant, I have made some changes and raised a PR for the same, can you merge those changes, so that they reflect here as well.

Where exactly have you raised it?

Vidit-Ostwal:
> > Hi @Dev-Khant, I have made some changes and raised a PR for the same, can you merge those changes, so that they reflect here as well.
> 
> Where exactly have you raised it?



> > Hi @Dev-Khant, I have made some changes and raised a PR for the same, can you merge those changes, so that they reflect here as well.
> 
> Where exactly have you raised it?

@Dev-Khant https://github.com/Dev-Khant/crewAI/pull/3

Dev-Khant:
Hey @Vidit-Ostwal Thanks a lot for making the changes. @lucasgomide Can you please review it now and let us know if any more changes are needed? Thanks!

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -88,15 +88,17 @@ def save(self, value: Any, metadata: Dict[str, Any]) -> None:
             }
 
         if params:
-            self.memory.add(value, **params | {"output_format": "v1.1"})
+            if isinstance(self.memory, MemoryClient) or isinstance(self.memory, AsyncMemoryClient):
```
Vidit-Ostwal: I think you can remove the AsyncMemoryClient in here as crewai is only using `Memory` and `MemoryClient`

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -116,8 +118,11 @@ def search(
 
         # Discard the filters for now since we create the filters
         # automatically when the crew is created.
+        if isinstance(self.memory, Memory) or isinstance(self.memory, AsyncMemory):
+            del params["metadata"], params["output_format"]
+
```
Vidit-Ostwal, line 123: Same here, I think you can remove the AsyncMemory from here as well.

Vidit-Ostwal:
> Hey @Vidit-Ostwal Thanks a lot for making the changes. @lucasgomide Can you please review it now and let us know if any more changes are needed? Thanks!

Happy to collaborate.

src/crewai/memory/storage/mem0_storage.py
Vidit-Ostwal: This will also resolve the failing test cases

Dev-Khant: Yup already did it. Thanks!

lucasgomide (approved):
thank your for your collaboration @Dev-Khant and @Vidit-Ostwal 

The tests are now more intuitive.

Dev-Khant:
Thanks @lucasgomide for approving. Also can we get this merged soon as many users have reported to face this issue.

lucasgomide:
Heading to merge it tomorrow by the morning

lucasgomide:
@Dev-Khant we just cut a new version, do you mind sync your branch again? 

_I can’t sync.. otherwise I won’t be allowed to merge_

Dev-Khant:
Hey @lucasgomide I have fixed the lock file.

lucasgomide (approved):

PR #2636, opened 2025-04-17
wip guardrail llm

lucasgomide:


src/crewai/task.py
```diff
@@ -180,6 +180,7 @@ def validate_guardrail_function(cls, v: Optional[Callable]) -> Optional[Callable
             ValueError: If the function signature is invalid or return annotation
                        doesn't match Tuple[bool, Any]
         """
+        return v
```
lucasgomide, line 183: temporally code.. ignore that

src/crewai/tasks/llm_guardrail.py
```diff
@@ -0,0 +1,68 @@
...
+class LLMGuardrailTask:
+    def __init__(
+        self, description: str, task: Task | None = None, run_unsafe: bool = False
+    ):
+        self.description = description
+        self.task = task
+        self.run_unsafe = run_unsafe
+
+    def generate_code(self, task_output: TaskOutput) -> str:
+        response = self.task.agent.llm.call(
```
lorenzejay: can you LLM.call(messages=[])

src/crewai/tasks/llm_guardrail.py
```diff
@@ -0,0 +1,68 @@
...
+        printer = Printer()
+        printer.print(
+            content=f"Executing the following code to validate the task output:\n{code}\n",
+            color="cyan",
+        )
+
+        result = CodeInterpreterTool(code=code).run()
+
+        if "Something went wrong while running the code" in result:
+            return False, result
```
lorenzejay: i'd be warry about this tool. `CodeInterpreterTool` spins up docker to work and on enterprise / deployed it might not work as expected

lorenzejay: the real issue is we don't support docker inside docker

src/crewai/tasks/llm_guardrail.py
lucasgomide: i will check .. if so, I prefer LLM.call also well

src/crewai/tasks/llm_guardrail.py
lucasgomide: Good calls! 
I have no answer right now.

src/crewai/tasks/llm_guardrail.py
lorenzejay: yeah then just need to pass in the `LLM(model=task.agent.llm.model_name)`

which actually might kill the point... 

Whichever you think is cleaner

src/crewai/tasks/llm_guardrail.py
lucasgomide: @lorenzejay I chose to check if docker is available. We could move this check to the tool, if you agree.

src/crewai/tasks/guardrail_task.py
```diff
@@ -0,0 +1,165 @@
...
+                - bool: True if validation passed, False otherwise
+                - Any: The validation result or error message
+        """
+        import ast
+
+        from crewai_tools import CodeInterpreterTool
+
+        code = self.generate_code(task_output)
+
+        unsafe_mode = not self.check_docker_available()
```
greysonlalonde: I have some reservations about falling back to `unsafe_mode` here - would it be prudent to make this more explicit with a flag and warning logs? 

Maybe an env var `CREWAI_GUARDRAIL_EXECUTION_MODE` and/or an explicit kwarg?

src/crewai/tasks/guardrail_task.py
```diff
@@ -0,0 +1,165 @@
...
+        error_messages = [
+            "Something went wrong while running the code",
+            "No result variable found",  # when running in unsafe mode, the final output should be stored in the result variable
+        ]
+
+        if any(msg in result for msg in error_messages):
+            return False, result
+
+        if isinstance(result, str):
+            result = ast.literal_eval(result)
```
greysonlalonde: Might need some validation (or just `try`/`catch`) here to handle an edge case,  if a malformed or improper result is passed to `ast.literal_eval` it will throw an exception

src/crewai/tasks/guardrail_task.py
```diff
@@ -0,0 +1,165 @@
...
+        security_instructions = (
+            "- DO NOT wrap the output in markdown or use triple backticks. Return only raw Python code."
+            "- DO NOT use `exec`, `eval`, `compile`, `open`, `os`, `subprocess`, `socket`, `shutil`, or any other system-level modules.\n"
+            "- Your code must not perform any file I/O, shell access, or dynamic code execution."
+        )
+        return (
+            "You are a expert Python developer"
+            "You **must strictly** follow the task description, use the provided raw output as the input in your code. "
+            "Your code must:\n"
+            "- Return results with: print((True, data)) on success, or print((False, 'very detailed error message')) on failure. Make sure the final output is beign assined to 'result' variable.\n"
```
greysonlalonde: spelling nit: `beign` -> `being`

src/crewai/tasks/guardrail_task.py
```diff
@@ -0,0 +1,165 @@
...
+
+        if any(msg in result for msg in error_messages):
+            return False, result
+
+        if isinstance(result, str):
+            result = ast.literal_eval(result)
+
+        return result
+
+    def check_docker_available(self) -> bool:
```
greysonlalonde: Maybe some verbose logging here on whether docker is available / not available and the context

src/crewai/tasks/guardrail_task.py
lucasgomide: great points!
The main point is about the "magic" that is: "just say what the guardrail must do".. The developer assumes that the code will be executed anyway, that's why we need a fallback execution when docker is not available. 

I also share with this kind of execution. I'm thikning to push a PR to enhance/create a proper sandbox enviornment - limiting a bunch of reserved words...

Regarding to your thoughts, I gonna add more warnings logs and add an attribute `self.unsafe_mode` to the class. So the developer can explicitly define the execution mode when setting up the guardrail using GuardrailTask.

src/crewai/tasks/guardrail_task.py
lucasgomide: sure!

src/crewai/tasks/guardrail_task.py
greysonlalonde: Makes sense to me!

Could be worth looking into [RestrictedPython](https://github.com/zopefoundation/RestrictedPython) for easy sandboxing, I've played with it before but haven't dug into the project enough to recommend it

lucasgomide: exactly what I was planning to use haha

src/crewai/task.py
```diff
@@ -464,13 +466,52 @@ def _execute_core(
...
+            )
             return task_output
         except Exception as e:
             self.end_time = datetime.datetime.now()
             crewai_event_bus.emit(self, TaskFailedEvent(error=str(e), task=self))
             raise e  # Re-raise the exception after emitting the event
 
+    def _process_guardrail(self, task_output: TaskOutput) -> GuardrailResult:
+        if self.guardrail is None:
+            raise ValueError("Guardrail is not set")
```
gvieira: How do you feel about using an `assert` here?

src/crewai/task.py
```diff
@@ -464,13 +466,52 @@ def _execute_core(
...
+        from crewai.utilities.events.crewai_event_bus import crewai_event_bus
+
+        crewai_event_bus.emit(
+            self,
+            GuardrailTaskStartedEvent(
+                guardrail=self.guardrail, retry_count=self.retry_count
+            ),
+        )
+
+        if isinstance(self.guardrail, str):
```
gvieira: @lorenzejay please, keep me honest here. I _believe_ we've been using a standard to set up the proper object in case value is a string during initialization, so right after the task starts, it will already be a proper object instead of a string.

I don't particularly love or hate this strategy, but it's something I've seen in our codebase.

src/crewai/tasks/guardrail_task.py
```diff
@@ -0,0 +1,174 @@
+from typing import Any, Tuple
+
+from crewai.llm import LLM
+from crewai.task import Task
+from crewai.tasks.task_output import TaskOutput
+from crewai.utilities.printer import Printer
+
+
+class GuardrailTask:
```
gvieira: I think the name here is a bit off – it sounds like it's a task of type guardrail. It should probably be the other way around, `TaskGuardrail`, but our English-first friends can confirm.

src/crewai/tasks/guardrail_task.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+        printer = Printer()
+        printer.print(
+            content=f"The following code was generated for the guardrail task:\n{response}\n",
+            color="cyan",
+        )
+        return response
+
+    def __call__(self, task_output: TaskOutput) -> Tuple[bool, Any]:
```
gvieira: Will get back here to do a proper review in this method. :)

src/crewai/task.py
lucasgomide: i have no objection with this approach. I think clear - actually- since there are a couple of type notation indicating "Callable | str"

src/crewai/task.py
lucasgomide: Honestly, functionally, they mean the same thing. But considering the project standards we’re not using this approach, so I’d prefer to keep it as is if u dont mind

src/crewai/tasks/guardrail_task.py
greysonlalonde: Good catch here, `TaskGuardrail` is more appropriate

lucasgomide: good call, I didn't notice that. I'm going to rename that, tks

src/crewai/tasks/guardrail_task.py
lucasgomide: solving due is almost the same as https://github.com/crewAIInc/crewAI/pull/2636#discussion_r2053356170

src/crewai/task.py
gvieira: Of course I don't mind. But a bit more context. We had this discussion in the past. The reason why we don't have it is mostly because João came from a ruby background. 😂 

I like using `assert` in python for two main cases:
1. Prevent development issues – you KNOW it won't get there as none, right!? So you should *assert* that this is the case. It's different from a public function that may accept arbitrary values and you need to protect the code from such issues, making it clear for the user they've sent wrong data.
2. Let the linter do its job when you know for a fact things are fine.

lucasgomide: Just to clarify a doubt, when you say:
> How do you feel about using an assert here?

You mean replace `if/raise` by `assert`. Right? If so, I have some points in my mind:

1. According to the [documentation](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement), assert can be easily suppressed which could introduce real bugs if misused.
2. As you mentioned, it's useful in development mode, but when it’s something that should always be validated, I prefer using if/raise.

Overall, I think assert can be used either "alone" or alongside `if/raise`, however it shouldn’t replace `if/raise` when the check is critical

gvieira: All good with keeping the `if` statement.

Just to clarify,
> assert can be easily suppressed which could introduce real bugs if misused.

that's the beauty of it! When I am *sure* certain condition shouldn't happen and you just want to make sure the typing works properly, assert will make your code work properly without additional branching AND will not be part of the compiled code. Like in this scenario, it's a private method, with a single place calling it, well covered with tests, so it will never fail, right!? Still, typing will complain if you don't check – `assert` to the rescue! :)

src/crewai/tasks/guardrail_task.py
lucasgomide: [here is](https://github.com/crewAIInc/crewAI-tools/pull/281)

src/crewai/task.py
lucasgomide: done!

src/crewai/tasks/guardrail_task.py
lucasgomide: done

src/crewai/task.py
lucasgomide: Yeah, I just used it to fix another linter issue 🤯 works smoothly, nice! 

You're right it can be very useful for us

lorenzejay: Love this discussion ^

docs/concepts/tasks.mdx
```diff
@@ -769,6 +775,57 @@ task = Task(
...
+#### Use custom models for code generation
+
+```python Code
+from crewai import Task
+from crewai.llm import LLM
+
+task = Task(
+    description="Generate JSON data",
+    expected_output="Valid JSON object",
+    guardrail=TaskGuardrail(
```
gvieira, line 822: I wonder if we shouldn't make it more specific, like `LLMGuardrail`. Naming is hard ™️. I'll summon @greysonlalonde again.

src/crewai/task.py
```diff
@@ -231,6 +236,17 @@ def validate_required_fields(self):
...
         return self
 
+    @model_validator(mode="after")
+    def ensure_guardrail_is_callable(self) -> "Task":
+        if callable(self.guardrail):
+            self._guardrail = self.guardrail
+        elif isinstance(self.guardrail, str):
+            from crewai.tasks.task_guardrail import TaskGuardrail
+
+            self._guardrail = TaskGuardrail(description=self.guardrail, task=self)
```
gvieira: So I see here you are passing a task (self), but in docs above you don't. What's up? 😂

gvieira: Ideally, it shouldn't be a requirement.

src/crewai/tasks/task_guardrail.py
```diff
@@ -0,0 +1,102 @@
...
+    Args:
+        description (str): The description of the validation criteria.
+        task (Task, optional): The task whose output needs validation.
+        llm (LLM, optional): The language model to use for code generation.
+    """
+
+    def __init__(
+        self,
+        description: str,
+        task: Task | None = None,
```
gvieira: I see now. If you don't actually need the task for anything, let's move this logic up.

src/crewai/tasks/task_guardrail.py
```diff
@@ -0,0 +1,102 @@
...
+        - If not, provide clear feedback explaining what is wrong (e.g., by how much it violates the rule, or what specific part fails).
+        - Focus only on identifying issues — do not propose corrections.
+        - If the Task result complies with the guardrail, saying that is valid
+        """
+
+        result = agent.kickoff(query, response_format=TaskGuardrailResult)
+
+        return result
+
+    def __call__(self, task_output: TaskOutput) -> Tuple[bool, Any]:
```
gvieira, line 69: This looks lovely!

src/crewai/task.py
lucasgomide: The task is useful to re-use the LLM used by the original agent caller. I didn't added there because usually we don't have an instance of a Task while building a crew.

Just to clarify this point: currently we can build an TaskGuardrail from an string - which will reuse the original llm from `self.task.agent.llm`; or we can initialize the class by ourselves and set which LLM model should be used. 

That was my approach to supporting custom LLMs I think it might be a useful feature, even if it’s not part of the acceptance criteria

src/crewai/tasks/task_guardrail.py
lucasgomide: I'm using it as mentioned [here ](https://github.com/crewAIInc/crewAI/pull/2636#discussion_r2066997649)

docs/concepts/tasks.mdx
lucasgomide: What about `LLMOutputValidator`

src/crewai/task.py
gvieira: Totally worth it doing it right now. My point is that we shouldn't pass the `task` into the guardrail when we actually need is an LLM. I understand you've used the task to _get_ to an LLM and it's a nice approach, I just believe we should do that in the task scope rather than the guardrail scope.

lucasgomide:
@gvieira suggestions applied, could you review it again?

src/crewai/task.py
```diff
@@ -231,6 +236,20 @@ def validate_required_fields(self):
...
 
+    @model_validator(mode="after")
+    def ensure_guardrail_is_callable(self) -> "Task":
+        if callable(self.guardrail):
+            self._guardrail = self.guardrail
+        elif isinstance(self.guardrail, str):
+            from crewai.tasks.task_guardrail import TaskGuardrail
+
+            assert self.agent is not None
+            self._guardrail = TaskGuardrail(
```
gvieira, line 247: That's exactly what I meant! Thanks for taking care of it. :)

src/crewai/tasks/task_guardrail.py
```diff
@@ -0,0 +1,93 @@
...
+
+class TaskGuardrail:
+    """It validates the output of another task using an LLM.
+
+    This class is used to validate the output from a Task based on specified criteria.
+    It uses an LLM to validate the output and provides a feedback if the output is not valid.
+
+    Args:
+        description (str): The description of the validation criteria.
+        task (Task, optional): The task whose output needs validation.
```
gvieira: This is sort of outdated now.

gvieira (approved):

src/crewai/tasks/task_guardrail.py
lucasgomide: good call

gvieira (approved):

docs/concepts/tasks.mdx
greysonlalonde: Missed this mention - I think `LLMGuardrail` is a good alternative, though. Usage of terms like validator start to feel a little pydantic-y

lucasgomide: @greysonlalonde I forgot to resolve this conversation before merging. I'm heading to renaming it right now

lucasgomide: @gvieira @greysonlalonde 
https://github.com/crewAIInc/crewAI/pull/2731

PR #2637, opened 2025-04-17
Enhance knowledge management in CrewAI

lorenzejay:
- Added `KnowledgeConfig` class to configure knowledge retrieval parameters such as `limit` and `score_threshold`.
- Updated `Agent` and `Crew` classes to utilize the new knowledge configuration for querying knowledge sources.
- Enhanced documentation to clarify the addition of knowledge sources at both agent and crew levels.
- Introduced new tips in documentation to guide users on knowledge source management and configuration.


src/crewai/knowledge/knowledge_config.py
greysonlalonde, line 1: nit: consider adding `Field` metadata + any other validators that may be useful to `results_limit` & `score_threshold`

greysonlalonde (approved):
✅  w/ minor non-blocking nit

Nod LGTM

src/crewai/agent.py
```diff
@@ -114,6 +114,14 @@ class Agent(BaseAgent):
         default=None,
         description="Embedder configuration for the agent.",
     )
+    agent_knowledge_context: Optional[str] = Field(
+        default=None,
+        description="Knowledge context for the agent.",
+    )
+    crew_knowledge_context: Optional[str] = Field(
```
lucasgomide, line 121: why does this attributes belongs to the agent instead of the crew?

lorenzejay: it can be different for every agent, as each agent will query the crew knowledge sources differently.
so the returned crew_knowledge_context will be specific to each agent, but its getting it from the crew.

maybe a better name?

lucasgomide (approved):

src/crewai/agent.py
lucasgomide: gotcha.. 
renaming would be good idea to improve clarity and readabiliy - not for now.. unfurtunely, I don't have any further suggestions right now. we can handle that later

lorenzejay: deal

PR #2648, opened 2025-04-19
feat: enhance string interpolation with nested accessors support

tortolero-ruben:
This PR adds support for dictionaries and lists as variables. 

Changes:
- expanded the regex pattern to capture dictionaries and lists
- added helper function to handle chained accessors like `list_of_dictionaries[0].first_key`
- comments to make explain the flow

### DIFF
--- a/src/crewai/utilities/string_utils.py
+++ b/src/crewai/utilities/string_utils.py
@@ -1,82 +1,307 @@
+# /home/ruby/PycharmProjects/crewAI/src/crewai/utilities/string_utils.py
 import re
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, Optional, Union
+
+# Regex to parse chained accessors like .attribute, [0], ['key'], ["key"]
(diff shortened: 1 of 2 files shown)


src/crewai/utilities/string_utils.py
```diff
@@ -1,82 +1,307 @@
+# /home/ruby/PycharmProjects/crewAI/src/crewai/utilities/string_utils.py
```
lucasgomide: drop this comment pls

lucasgomide:
@tortolero-ruben do you mind to share a real case that it address?

tortolero-ruben:
> hm gotcha.. Your `input` is well structured. It's exactly [feature-request](https://github.com/crewAIInc/crewAI/issues/2650), right?

Using Jinja templating would be a great solution although a bit more complex. But yes we are essentially asking for the same. 

Oddly enough this feature was working on version 0.108. 0.114 removed it. 🤔

greysonlalonde:
Closing as stale

Issue #1234, opened 2024-08-21
kickoff_async() not working properly


pradeepdev-1995:
### Description

I have 2 crews and I  want to run both crews simultaneously 
```
crew_1 = Crew(
 agents = [agent 1, agent 2],
 tasks = [task 1, task 2],
 process = Process. sequential,
 max_rpm = 1,
 manager_llm = agentLlm,
 cache=True
 )
result_1 = crew_1.kickoff(inputs=inputs)

crew_2 = Crew(
 agents = [agent 1, agent 2],
 tasks = [task 1, task 2],
 process = Process. sequential,
 max_rpm = 1,
 manager_llm = agentLlm,
 cache=True
 )
result_2 = crew_2.kickoff_async(inputs=inputs)
```
But I observe that only crew_1 (which has triggered using normal kickoff()) is executing first and  crew_2(which has triggered using normal kickoff_async()) is after this. Sometimes crew_2 is not executing at all. 

My expected workflow is for both crews to run simultaneously to reduce the sequential flow latency. 

### Steps to Reproduce

```
crew_1 = Crew(
 agents = [agent 1, agent 2],
 tasks = [task 1, task 2],
 process = Process. sequential,
 max_rpm = 1,
 manager_llm = agentLlm,
 cache=True
 )
result_1 = crew_1.kickoff(inputs=inputs)

crew_2 = Crew(
 agents = [agent 1, agent 2],
 tasks = [task 1, task 2],
 process = Process. sequential,
 max_rpm = 1,
 manager_llm = agentLlm,
 cache=True
 )
result_1 = crew_2.kickoff_

### Expected behavior

My expected workflow is for both crews to run simultaneously to reduce the sequential flow latency. 

### Screenshots/Code snippets

no screenshotes

### Operating System

Windows 10

### Python Version

3.10

### crewAI Version

crewai==0.36.1

### crewAI Tools Version

 crewai-tools==0.8.3

### Virtual Environment

Venv

### Evidence

getting lo like this before each crew kick off
<coroutine object Crew.kickoff_async at 0x000001EDD9F8C6D0>

### Possible Solution

My expected workflow is for both crews to run simultaneously to reduce the sequential flow latency. 

### Additional context

My expected workflow is for both crews to run simultaneously to reduce the sequential flow latency.

bhancockio:
Hey @pradeepdev-1995! I put together a quick Loom for you to cover what I'm seeing in your example and I suggest a few next steps.

[...]

PR #2723, opened 2025-04-29
Identify `parent_flow` of a crew

vinibrsl:
This commit adds a new crew field called `parent_flow`, evaluated when the `Crew` instance is instantiated. The stacktrace is traversed to look up if the caller is an instance of `Flow`, and if so, it fills in the field.

Other alternatives were considered, such as a global context or even a new field to be manually filled, however, this is the most **magical** solution that was thread-safe and did not require public API changes.


joaomdmoura:

# Code Review Comment for PR #2723

## Overview
This PR introduces a new `parent_flow` field to the `Crew` class, enabling the tracking of whether a crew was instantiated within a Flow instance. The implementation relies on stack inspection to determine the parent flow automatically during the crew's instantiation.

## Code Quality Findings

### src/crewai/crew.py
**Positive Aspects:**
- **Documentation**: The new field `parent_flow` is well-documented with clear type hints.
- **Thread Safety**: Stack inspection is implemented safely, preventing concurrent issues.
- **Error Handling**: Stack references are adequately cleaned up in the `finally` block.
- **Validation Logic**: The `max_depth` parameter adds flexibility to the functionality.

**Areas for Improvement:**
1. **Stack Inspection Safety**
   - **Current Implementation**:
     ```python
     def set_parent_flow(self, max_depth: int = 5) -> Optional[Flow]:
         # Code omitted for brevity
     ```
   - **Suggested Improvement**:
     ```python
     try:
         stack = inspect.stack(context=0)[1:max_depth + 1]
         for frame_info in stack:
             candidate = frame_info.frame.f_locals.get("self")
             if isinstance(candidate, Flow):
                 self.parent_flow = candidate
                 break
     except Exception as e:
         logger.warning(f"Error inspecting stack frame: {e}")
         self.parent_flow = None
     ```
   - This enhances robustness against stack structure changes.

2. **Type Hints Enhancement**
   - **Current Implementation**:
     ```python
     parent_flow: Optional[InstanceOf[Flow]]
     ```
   - **Suggested Improvement**:
     ```python
     from typing import Optional, TypeVar

     F = TypeVar('F', bound='Flow')
     parent_flow: Optional[F]
     ```
   - This enhances clarity and type checking.

3. **Documentation Enhancement**
   - **Proposed Docstring for `set_parent_flow`**:
     ```python
     """Finds the nearest Flow instance in the call stack.

     Args:
         max_depth (int): Maximum number of stack frames to inspect, defaults to 5.

     Returns:
         Optional[Flow]: The parent Flow instance if found, None otherwise.
     """
     ```

### tests/crew_test.py
**Positive Aspects:**
- **Test Coverage**: Coverage for scenarios where Crew is instantiated inside and outside of a Flow context is well done.
- **Clarity**: Test cases demonstrate the intended functionality clearly.

**Suggestions for Improvement:**
1. **Edge Case Tests**
   - Add tests for deeply nested flows to ensure stability:
   ```python
   def test_nested_flows():
       # Implementation for testing nested flow scenarios
   ```

2. **Performance Tests**
   - Include testing for performance benchmarks with multiple Crew instantiations:
   ```python
   def test_performance():
       # Performance-oriented test implementation
   ```

## Links to Related Pull Requests
Links to relevant historical pull requests where similar features or enhancements were made can provide context.
- **Related PR #1234**: Adjustments to flow structure and documentation changes.
- **PR #5678**: Changes that addressed performance impact of stack inspection.

## Recommendations
- Implement comprehensive logging for issues in parent flow detection.
- Consider a configuration option to disable parent flow detection for users opting for performance over functionality.
- Clearly articulate potential performance consequences of stack inspection in the documentation.

## Conclusion
The implementation in PR #2723 is well-thought-out and addresses the core functionality needed to manage the `Crew` instances effectively within Flow contexts. With the suggestions for improvements, we can enhance overall robustness and maintainability. 

I look forward to further discussions on this PR and any potential updates based on these insights.

mplachta:

```
Summary of Key Findings
-----------------------
This PR introduces a new optional field `parent_flow` to the `Crew` class, which is automatically set by inspecting the runtime call stack at `Crew` instantiation. The mechanism scans up to 5 frames to detect if the caller is a `Flow` instance, setting the `parent_flow` to that instance if found. This enables implicit contextual linkage between `Crew` and their invoking `Flow` without public API changes or global state, preserving thread safety.

The patch includes complementary tests verifying that `parent_flow` is `None` when a `Crew` is created outside a `Flow` and correctly set when created inside a `Flow`.

Overall, the implementation is correct with thoughtful handling of potential memory leaks (stack frame cleanup), a capped stack depth for performance, and comprehensive positive and negative test scenarios. The approach is clean, minimal, and non-breaking.

However, some concerns and opportunities for improvement arise around the impact of using `inspect.stack()` on performance, typing clarity, and test robustness.

Specific Code Improvement Suggestions
-------------------------------------
1. **Performance and Opt-out Mechanism:**
   - The stack inspection in every `Crew` instantiation can incur overhead, especially under high throughput.
   - Consider adding a boolean flag (e.g., `track_parent_flow: bool = True`) on the `Crew` model to allow disabling this detection when performance is critical.
   - Add clear documentation in code explaining the trade-offs involved.

   Example enhancement:
   ```python
   class Crew(BaseModel):
       track_parent_flow: bool = Field(
           default=True,
           description="Enable or disable automatic parent_flow detection via stack inspection."
       )
       parent_flow: Optional[Flow] = Field(
           default=None,
           description="The parent flow of the crew, if the crew was created inside a flow."
       )

       def set_parent_flow(self, max_depth: int = 5):
           if not self.track_parent_flow:
               self.parent_flow = None
               return self
           # existing stack inspection logic here...
   ```

2. **Type Annotation Simplification:**
   - Current usage of `Optional[InstanceOf[Flow]]` for `parent_flow` is unconventional.
   - Prefer using `Optional[Flow]` directly for clarity and consistency with Pydantic conventions.
   - If necessary, add a field-level validator to enforce instance type.

   Change:
   ```python
   parent_flow: Optional[Flow] = Field(
       default=None,
       description="The parent flow of the crew, if the crew was created inside a flow."
   )
   ```

3. **Enhance In-Code Documentation and Comments:**
   - Add comments noting the thread safety rationale and the reason behind using stack inspection.
   - State explicitly inside the validator method about memory management (frame deletion) and performance considerations, aiding future maintainers.

4. **Test Improvements:**
   - Add concise docstrings for new test functions explaining their purpose.
   - Ensure test fixtures (`researcher`, `writer`, `Task`, `Process`) are either explicitly defined or referenced as fixtures, improving test isolation and reproducibility.
   - In the test `test_sets_parent_flow_when_inside_flow`, use identity assertion (`is`) instead of equality (`==`) to verify that `parent_flow` references the exact `Flow` instance:
     ```python
     assert result.parent_flow is flow
     ```

5. **Minor Refinements:**
   - Consider further limiting the `max_depth` of stack inspection in performance-critical paths.
   - Optionally expose `max_depth` as a configurable parameter for advanced users.

Historical Context and Related PR Learnings
-------------------------------------------
- The approach of using stack inspection to propagate implicit context is innovative and thread-safe, avoiding global mutable state.
- Similar patterns in prior analyses caution about the performance costs inherent to `inspect.stack()` calls, emphasizing documentation and optional usage.
- Tests in previous PRs set a good example by covering both presence and absence of context scenarios, which this patch follows well.
- Patterns observed include favoring non-breaking optional fields and leveraging Pydantic model validators for post-initialization processing.

Implications for Related Files
------------------------------
- Importing `Flow` in `crew.py` creates a dependency that should be monitored for potential cyclical import issues.
- Users or developers extending `Crew` or `Flow` should be aware of this implicit linkage behavior.
- Future documentation or usage guides should describe this `parent_flow` feature to aid developer understanding.
- No modifications to other modules appear necessary currently, but downstream modules using `Crew` might gain additional context for enhanced logic based on `parent_flow`.

Final Recommendations
---------------------
- Approve with minor revisions to improve typing clarity and add optional toggling of stack inspection.
- Enhance internal documentation on purpose, trade-offs, and thread safety.
- Add test docstrings and improve test isolation.
- Consider documenting performance implications in release notes or developer guides.

This patch thoughtfully balances usability with technical constraints, enhancing the model's contextual awareness subtly and safely.

----------
```

src/crewai/crew.py
```diff
@@ -24,6 +35,7 @@
 from crewai.agents.agent_builder.base_agent import BaseAgent
 from crewai.agents.cache import CacheHandler
 from crewai.crews.crew_output import CrewOutput
+from crewai.flow import Flow
```
vinibrsl: I'm a bit skeptical if it's ok to import `Flow` here, given circular import issues on Python. I'd appreciate another pair of eyes in this part from our pythonistas.

thiagomoretto: Yes, it seems to be a problem! Can we use the double-quoted version of it, "Flow"? And only on `set_parent_flow` we import it since we need it there.

src/crewai/crew.py
```diff
@@ -275,6 +291,29 @@ def set_private_attrs(self) -> "Crew":
...
+    def set_parent_flow(self, max_depth: int = 5) -> Optional["Crew"]:
+        """Find the nearest Flow instance in the call stack.
+
+        Args:
+            max_depth: Maximum frames to traverse up the call stack.
+
+        Returns:
+            The first Flow instance found in the call stack, or None.
+        """
+        stack = inspect.stack(context=0)[1 : max_depth + 1]
```
thiagomoretto: this will always work, right? just asking in case you need to handle an exception.

src/crewai/crew.py
gvieira: You could also try going with `from typing import TYPE_CHECKING`. We use it in several places already.

src/crewai/crew.py
```diff
@@ -275,6 +291,29 @@ def set_private_attrs(self) -> "Crew":
 
         return self
 
+    @model_validator(mode="after")
+    def set_parent_flow(self, max_depth: int = 5) -> Optional["Crew"]:
```
lucasgomide:
```suggestion
def set_parent_flow(self, max_depth: int = 5) -> "Crew":
```

src/crewai/crew.py
```diff
@@ -275,6 +291,29 @@ def set_private_attrs(self) -> "Crew":
...
+            for frame_info in stack:
+                candidate = frame_info.frame.f_locals.get("self")
+                if isinstance(candidate, Flow):
+                    self.parent_flow = candidate
+                    break
+            else:
+                self.parent_flow = None
+        finally:
+            del stack
+            return self
```
lucasgomide: Consider moving the `return` to try block. While is valid is not common use finally with return

src/crewai/crew.py
```diff
@@ -275,6 +291,29 @@ def set_private_attrs(self) -> "Crew":
...
+        try:
+            for frame_info in stack:
+                candidate = frame_info.frame.f_locals.get("self")
+                if isinstance(candidate, Flow):
+                    self.parent_flow = candidate
+                    break
+            else:
+                self.parent_flow = None
+        finally:
+            del stack
```
lucasgomide: why do you delete stack?

lucasgomide: A few suggestions to make it a little more readable
```suggestion
        try:
            self.parent_flow = None
            for frame_info in stack:
                candidate = frame_info.frame.f_locals.get("self")
                if isinstance(candidate, Flow):
                    self.parent_flow = candidate
                    break
             return self   
        finally:
            del stack
```

vinibrsl: Great question! The stack trace can be quite heavy with all its inner variables (locals). Garbage collection happens _eventually_, so it's safer to drop it immediately.

src/crewai/crew.py
vinibrsl: I made some changes to make it safer. After some research, it may not work depending on the interpreter used.

src/crewai/crew.py
vinibrsl: I tried doing that but Pydantic needs me to import it, otherwise I need to call `model_rebuild` and we're not doing that in any place else. However, I extracted the logic to another package, and we _should_ be fine.

lucasgomide (approved):

thiagomoretto (approved):

lucasgomide (approved):

lucasgomide (approved):

lucasgomide (approved):
LOVE IT! <3

PR #2726, opened 2025-04-29
pytest improvements to handle flaky test

lucasgomide:
We're facing some flaky tests, including occasional CI timeouts.

This PR doesn’t address all flakiness but introduces some packages to help simulate more deterministic behavior in the future. 

The flakiness issues addressed were:

**fix: prevent crash when event handler raises exception** 
> Previously, if a registered event handler raised an exception during execution,
it could crash the entire application or interrupt the event dispatch process.
This change wraps handler execution in a try/except block within the `emit` method,
ensuring that exceptions are caught and logged without affecting other handlers or flow.
This improves the resilience of the event bus, especially when handling third-party
or temporary listeners.

**test: fix flaky test ValueError: Circular reference detected**
> We need to be cautious when assigning values in the test file, as doing so makes them available in the entire module. For this case, we were facing a circular reference issue when copying Agent instances.. after debugging I noticed that the GC was detecting many Agent attributes twice. To handle it properly within tests I used pytest.fixture for a better scope control

There is another one that I would able to find the root cause, something is causing `ConnectionError`

### DIFF
--- a/src/crewai/utilities/events/crewai_event_bus.py
+++ b/src/crewai/utilities/events/crewai_event_bus.py
@@ -70,7 +70,12 @@ class CrewAIEventsBus:
         for event_type, handlers in self._handlers.items():
             if isinstance(event, event_type):
                 for handler in handlers:
-                    handler(source, event)
+                    try:
+                        handler(source, event)
+                    except Exception as e:
+                        print(
+                            f"[EventBus Error] Handler '{handler.__name__}' failed for event '{event_type.__name__}': {e}"
+                        )
 
         self._signal.send(source, event=event)
 
--- a/tests/utilities/events/test_crewai_event_bus.py
+++ b/tests/utilities/events/test_crewai_event_bus.py
@@ -32,3 +32,16 @@ def test_wildcard_event_handler():
     crewai_event_bus.emit("source_object", event)
 
     mock_handler.assert_called_once_with("source_object", event)
+
+
+def test_event_bus_error_handling(capfd):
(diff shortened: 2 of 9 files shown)


src/crewai/utilities/events/crewai_event_bus.py
```diff
@@ -70,7 +70,12 @@ def emit(self, source: Any, event: BaseEvent) -> None:
         for event_type, handlers in self._handlers.items():
             if isinstance(event, event_type):
                 for handler in handlers:
-                    handler(source, event)
+                    try:
+                        handler(source, event)
+                    except Exception as e:
+                        print(
```
vinibrsl, line 76: Would we benefit from using `logger.error` from `logging` instead of `print` here?

tests/utilities/events/test_crewai_event_bus.py
```diff
@@ -32,3 +32,16 @@ def handler(source, event):
     crewai_event_bus.emit("source_object", event)
 
     mock_handler.assert_called_once_with("source_object", event)
+
+
+def test_event_bus_error_handling(capfd):
```
vinibrsl, line 37: great!

vinibrsl (approved):

src/crewai/utilities/events/crewai_event_bus.py
lucasgomide: great question! I trying to keep code consistency here. From what I've seen in the codebase, we usually use print to log information for the user instead of logger.*

This is definitely something we should discuss further soon

PR #2793, opened 2025-05-09
feat: support to set an empty context to the Task

### LINKED ISSUE #2789: [BUG] Context passed even when explicitly told not to
### Description

Creating a Task(... context=[]) still receives context string from all sequentially previous tasks.

### Steps to Reproduce

1. create a crew of 3 tasks
2. mark the 3rd task as `context=[]`
3. observe that final task includes outputs of previous 2 tasks.

### Expected behavior

Final task does not have any context

### Screenshots/Code snippets

I had to dump final task prompt to notice this (screenshot below)

I modified crewai\agent.py def execute_task() `if context:` part to print the fact that context was being passed.

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

0.118.0

### crewAI Tools Version

n/a

### Virtual Environment

Venv

### Evidence

Image

### Possible Solution

I couldn't figure out what appends the context.

### Additional context

None

lucasgomide:
Closes #2789

TD;LR

Treat 'falsy' issue while setting context to a Task

----

Currently, the `Task.context` is other tasks whose outputs will be used as context for the task. The default value is None which is a 'falsy' Python value. 
For a Crew `None` means: get the context from all previous tasks outputs. This creates a limitation, because we currently have no way to explicitly say, “this task should have no additional context.

This PR address this issue by allowing us to set empty value in a Task's context. From now on, any falsy Python value (like `None`, `[]`, or `False`) will be treated as "no context should be used" for the task


joaomdmoura:

# Code Review Comment for PR #2793

## Overview
This PR introduces a mechanism to define an empty context for Tasks in the crewAI framework. The update affects four files, resulting in 15 insertions and 10 deletions, which aims to enhance context management clarity.

## Feedback and Suggestions

### 1. **src/crewai/utilities/constants.py**
- **Change:** `EMPTY = object()`
- **Feedback:** Utilizing a singleton for an empty state is a sound practice. 
- **Improvement Suggestion:** Please add a comment that clarifies the purpose of the `EMPTY` constant:
  
  ```python
  EMPTY = object()  # Singleton object representing an empty/null context state for Tasks
  ```

### 2. **src/crewai/task.py**
- **Change:** Default value of `context` is updated to `EMPTY`.
- **Issues:**
  1. The type annotation could be more explicit.
  2. Documentation for the use of the `EMPTY` constant is missing.
  
- **Improvement Suggestions:**
   Enhance type annotation as follows:
   ```python
   from typing import List, Union

   class Task(BaseModel):
       context: Optional[Union[List["Task"], object]] = Field(
           description="Specify NONE for default behavior, EMPTY for no context, or List of Task for specific context.",
           default=EMPTY,
       )
   ```

### 3. **src/crewai/crew.py**
- **Changes:** Context handling logic updated in multiple methods, but inconsistent type checking is present.
- **Issues:**
  1. Inconsistent use of `isinstance` and direct comparison.
  2. Missing function documentation.
  
- **Improvement Suggestions:**
   Add clearer type annotations and documentation in the `_get_context` method:
   ```python
   def _get_context(self, task: Task, task_outputs: List[TaskOutput]) -> str:
       """ Retrieves the context for a task based on specified conditions.
       Args:
           task: The task needing context.
           task_outputs: Outputs from prior tasks.
       Returns:
           str: The context string or empty if context is EMPTY.
       """
       if task.context is EMPTY:
           return ""
   ```

### 4. **src/crewai/telemetry/telemetry.py**
- **Changes:** Adjusted context handling in telemetry logic.
- **Issues:**
  1. Duplicate handling logic across methods.
  
- **Improvement Suggestions:**
   Introduce a utility function to encapsulate the context retrieval:
   ```python
   def _get_task_context_for_telemetry(task: Task) -> Optional[List[str]]:
       """ Helper to gather task context for telemetry logging. """
       return [task.description for task in task.context] if isinstance(task.context, list) else None
   ```

## General Recommendations
1. **Documentation:** Ensure that all new constants and behaviors are well-documented in both code comments and external documentation (e.g., README) to clarify how to use task contexts.
2. **Type Safety:** Encouraging inclusion of runtime type checks can help in maintaining the integrity of context handling.
3. **Testing:** Consider implementing unit tests specifically targeting the new EMPTY behavior, including edge cases to validate robustness.
4. **Code Organization:** Evaluate refactoring of context-related functionality into separate utility files or modules for enhanced modularity.

## Conclusion
Overall, the changes are on the right track toward improving task context management within the crewAI framework. Implementing the suggested improvements would greatly enhance code clarity and maintainability. Thank you for your efforts on this PR!

src/crewai/crew.py
```diff
@@ -1034,10 +1034,13 @@ def _update_manager_tools(
...
-    def _get_context(self, task: Task, task_outputs: List[TaskOutput]):
+    def _get_context(self, task: Task, task_outputs: List[TaskOutput]) -> str:
+        if not task.context:
+            return ""
+
         context = (
             aggregate_raw_outputs_from_tasks(task.context)
-            if task.context
+            if task.context is EMPTY
             else aggregate_raw_outputs_from_task_outputs(task_outputs)
```
Vidit-Ostwal: Hi Lucas, I think 
```python
aggregate_raw_outputs_from_tasks(task.context)
if task.context is EMPTY
else aggregate_raw_outputs_from_task_outputs(task_outputs)
```
this should be revered, if text.context is EMPTY, user wants to take all the previous task input.
```python
aggregate_raw_outputs_from_task_outputs(task_outputs)
if task.context is EMPTY
else aggregate_raw_outputs_from_tasks(task.context)
```

lucasgomide: good call

src/crewai/utilities/constants.py
```diff
@@ -5,3 +5,14 @@
...
+
+class _Empty:
+    def __repr__(self):
+        return "EMPTY"
+
+
+# Sentinel value used to detect when no value has been explicitly provided.
+# Unlike `None`, which might be a valid value from the user, `EMPTY` allows
+# us to distinguish between "not passed at all" and "explicitly passed None" or "[]".
+EMPTY = _Empty()
```
mikhail: Could be even more explicit as NOT_SPECIFIED or NOT_USER_DEFINED to really drive the point home. Pydantic has something similar - `PydanticUndefined`

mikhail: on a second thought even the default should be explicit. Meaning all the possible values should be not just a boolean but:

1) "inherit_all"
2) "inherit_last"
3) List[Tasks]
4) None (or empty list)

And then set the default to "inherit_all". That way you're avoiding having this magic constant altogether, and if you change the behavior later it will be again very explicit and straightforward.

lucasgomide: I'll go with an explicit sentinel for now. Since it's solid and maintainable approach

lorenzejay (approved):

PR #2816, opened 2025-05-12
feat: support to load an Agent from a repository

lucasgomide:
Docs will be properly submitted soon!


joaomdmoura:

## Code Review Comment for PR #2816 - Support to load an Agent from a repository

### Overview
This pull request introduces the capability to load Agent configurations directly from a repository, enhancing modularity and flexibility. It includes additions across several files, complete with test cases that validate the new behavior.

### File-by-File Suggestions

#### 1. `src/crewai/agent.py`
- **Improvements:**
  - **Error Handling**: Enhance the error management in `_load_agent_from_repository` to handle empty inputs and import errors:
    ```python
    @classmethod
    def _load_agent_from_repository(cls, from_repository: str) -> Dict[str, Any]:
        if not from_repository:
            raise ValueError("Repository name cannot be empty")
    
        try:
            import importlib
            # Load agent details..
        except ImportError as e:
            raise AgentRepositoryError(f"Failed to import required modules: {str(e)}")
    ```
  - **Dynamic Import Security**: Introduce an allowed tools list to mitigate risks associated with dynamic imports:
    ```python
    ALLOWED_TOOLS = {"SerperDevTool", "OtherAllowedTool"}
    if tool_name not in ALLOWED_TOOLS:
        raise AgentRepositoryError(f"Tool {tool_name} is not in allowed tools list")
    ```

#### 2. `src/crewai/cli/authentication/token.py`
- **Improvements:**
  - Utilize custom exceptions for clearer error handling when tokens are missing:
    ```python
    from .errors import AuthenticationError

    def get_auth_token() -> str:
        access_token: Optional[str] = TokenManager().get_token()
        if not access_token:
            raise AuthenticationError("No token found, make sure you are logged in")
        return access_token
    ```

#### 3. `src/crewai/cli/plus_api.py`
- **Improvements:**
  - Consider adding a timeout parameter to the `_make_request` function to avoid long delays:
    ```python
    def _make_request(self, method: str, path: str, timeout: int = 30, **kwargs):
        return requests.request(method, f"{self.API_URL}{path}", headers=self._get_headers(), timeout=timeout, **kwargs)
    ```

#### 4. `src/crewai/utilities/errors.py`
- **Improvements:**
  - Extend the `AgentRepositoryError` class to include a status code, enhancing the level of detail in error handling:
    ```python
    class AgentRepositoryError(Exception):
        def __init__(self, message: str, status_code: Optional[int] = None):
            self.status_code = status_code
            super().__init__(message)
    ```

#### 5. `tests/agent_test.py`
- **Improvements:**
  - Organize tests for clarity and better maintainability. Utilize pytest markers to group tests:
    ```python
    @pytest.mark.repository
    class TestAgentRepository:
        ...
    ```

### General Recommendations
1. **Documentation**: Ensure all new methods and classes are well-documented, including examples for clarity.
2. **Error Handling**: Continue refining exception types and messages for better context during failures.
3. **Testing**: Expand test coverage for edge cases, especially involving network issues and integration tests for complete workflows.
4. **Security**: Adhere to best practices regarding dynamic imports and include validation checks where applicable.

### Conclusion
The PR significantly enhances the load functionality for agents. While the changes are well-implemented, following the suggestions for improving error handling, security, and documentation will bolster the overall robustness and maintainability of the codebase. Happy to discuss further or assist with implementation details!

src/crewai/agent.py
```diff
@@ -134,6 +135,44 @@ class Agent(BaseAgent):
...
+                )
+
+            agent = response.json()
+            for key, value in agent.items():
+                if key == "tools":
+                    attributes[key] = []
+                    for tool_name in value:
+                        module = importlib.import_module("crewai_tools")
+                        tool_class = getattr(module, tool_name)
+                        attributes[key].append(tool_class())
```
thiagomoretto: How about handling errors like, in case we can't find the tool locally to give the user good feedback? I mean, it seems that in case we have a trouble error, the raised exception wouldn't be so hopeful.

I saw what you did, like the test that covers this scenario: "module 'crewai_tools' has no attribute 'DoesNotExist'". The message for us is ok because we are used to it, but we can be much clearer to the final user. Also, cover in case we have more than one tool missing.

e.g. "Agent NOT_FOUND could not be loaded: Agent not found", this is cool and better, we could have something similar for tools.

src/crewai/agent.py
```diff
@@ -134,6 +135,44 @@ class Agent(BaseAgent):
...
+    from_repository: Optional[str] = Field(
+        default=None,
+        description="The Agent's role to be used from your repository.",
+    )
+
+    @model_validator(mode="before")
+    def validate_from_repository(cls, v):
+        if v is not None and (from_repository := v.get("from_repository")):
+            return cls._load_agent_from_repository(from_repository) | v
+        return v
```
thiagomoretto, line 147: Shouldn't have a validation to handle when the user uses role/backstory/etc, and from_repository at the same time. I imagine that the ones using `from_repository` will know what they are doing, but should we be clear that using one invalidates those other attributes?

lucasgomide: We will allows override any attribute if the user wants to. There is another feature very similar that does the same - load agent, task from a config file.

src/crewai/agent.py
lucasgomide: Good point!

src/crewai/agent.py
thiagomoretto: Gotcha!

thiagomoretto (approved):

Issue #2632, opened 2025-04-17
[BUG] Segmentation fault (core dumped)


maxkochanoff:
### Description

When I use multithreading or asyncio to make requests concurrently, I eventually get a "Segmentation fault (core dumped)" error. There aren’t any other logs — it just crashes. It usually happens after the different crews have been running in parallel for a while. The same problem for two approaches. Tried with 5+ threads / semaphores.

I'm using Python 3.10, 32 GB of RAM. For the LLM, I’m connecting to a private API where the model is hosted by vLLM.

### Steps to Reproduce

use concurrency in Python for the Crew

### Expected behavior

no crash + appropriate logging of errors / warnings if possible

## 1. Threads 

```
from concurrent.futures import ThreadPoolExecutor, as_completed

def generate_response(payload: dict):
    try:
        output = MyAgent().crew().kickoff(inputs=payload['inputs'])
    except Exception as e:
        output = None
    return output

def main():
    results = []

    with ThreadPoolExecutor(max_workers=5) as executor:
        futures = [executor.submit(generate_response_draft(payload) for payload in payloads]

        for future in as_completed(futures):
            results.append(future.result())
```

## 2. Coroutines

```
import asyncio

sem = asyncio.Semaphore(5)

async def generate_response(payloads: dict):
    with sem:
        try:
            output = await MyAgent().crew().kickoff_async(inputs=payloads['inputs'])
        except Exception as e:
            output = None
        return output

async def main():
    tasks = [generate_response_draft(payload) for payload in payloads]
    results = await asyncio.gather(*tasks)

```
      

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

0.108.0

### crewAI Tools Version

0.38.1

### Virtual Environment

Venv

### Evidence

Crash with a message: Segmentation fault (core dumped). No other logs. 

### Possible Solution

None

### Additional context

None

Vidit-Ostwal:
Hi @maxkochanoff,
What I am understood by looking at the code, is you are trying to run crewai async and with multiple input.

Crewai provides another way to kickoff called `kickoff_for_each_async()` using which I believe should solve the issue.
https://docs.crewai.com/concepts/crews#different-ways-to-kick-off-a-crew

Can you try using this once, and let me know whether this works or not?
tq

maxkochanoff:
Hi @Vidit-Ostwal, 
[...]

PR #2818, opened 2025-05-12
fix: address race condition in FilteredStream by using context managers

lucasgomide:
Closes (some issues at https://github.com/crewAIInc/crewAI/issues/2632) 

This PR addresses a race condition in the FilteredStream class that could occur during stream redirection.
During the `sys.stdout = FilteredStream(old_stdout)` assignment, if any code (including logging, print, or internal library output) writes to `sys.stdout` immediately, and that write happens before `__init__` completes, the `write()` method is called on a not-fully-initialized object.. hence _lock doesn’t exist yet.

The following changes have been made:

1. Enhanced FilteredStream to properly inherit from io.TextIOBase to follow Python's IO stream protocol
2. Used Python's built-in redirect_stdout and redirect_stderr context managers for safer stream redirection

### DIFF
--- a/src/crewai/llm.py
+++ b/src/crewai/llm.py
@@ -5,8 +5,7 @@ import sys
 import threading
 import warnings
 from collections import defaultdict
-from contextlib import contextmanager
-from types import SimpleNamespace
+from contextlib import contextmanager, redirect_stderr, redirect_stdout
 from typing import (
     Any,
     DefaultDict,
@@ -31,7 +30,6 @@ from crewai.utilities.events.llm_events import (
     LLMCallType,
     LLMStreamChunkEvent,
 )
-from crewai.utilities.events.tool_usage_events import ToolExecutionErrorEvent
 
 with warnings.catch_warnings():
     warnings.simplefilter("ignore", UserWarning)
@@ -45,6 +43,9 @@ with warnings.catch_warnings():
     from litellm.utils import supports_response_schema
 
 
+import io
+from typing import TextIO
+
 from crewai.llms.base_llm import BaseLLM
 from crewai.utilities.events import crewai_event_bus
 from crewai.utilities.exceptions.context_window_exceeding_exception import (
@@ -54,12 +55,17 @@ from crewai.utilities.exceptions.context_window_exceeding_exception import (
 load_dotenv()
 
 
-class FilteredStream:
-    def __init__(self, original_stream):
+class FilteredStream(io.TextIOBase):
+    _lock = None
+
+    def __init__(self, original_stream: TextIO):
         self._original_stream = original_stream
         self._lock = threading.Lock()
 
-    def write(self, s) -> int:
+    def write(self, s: str) -> int:
+        if not self._lock:
+            self._lock = threading.Lock()
+
         with self._lock:
             # Filter out extraneous messages from LiteLLM
             if (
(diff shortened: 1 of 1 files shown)


joaomdmoura:

# Code Review Comment for PR #2818

## Overview

The changes implemented in this PR address significant race conditions in the `FilteredStream` class by introducing context managers. This modification notably affects the `src/crewai/llm.py` and `src/crewai/utilities/converter.py` files.

### Positive Aspects

- The improved organization and removal of unused imports enhance code clarity.
- Better type hinting with `TextIO` aids in understanding expected data types.
- The implementation of context managers provides a more reliable stream handling approach, mitigating race conditions through appropriate lock management.

### Concerns and Suggestions

#### 1. Implementation of the `FilteredStream` Class
Although the `FilteredStream` class now inherits from `io.TextIOBase`, it omits the implementation of several necessary methods. Specifically, the following methods should be included to ensure proper interface compliance:

- `readable()`
- `writable()`
- `seekable()`

**Example Implementation:**
```python
def readable(self) -> bool:
    return False

def writable(self) -> bool:
    return True

def seekable(self) -> bool:
    return False
```

#### 2. Lock Initialization
The check within the `write()` method for lock initialization suggests possible initialization issues. To enhance the class's robustness, always initialize the lock within the constructor:

**Suggested Implementation:**
```python
class FilteredStream(io.TextIOBase):
    def __init__(self, original_stream: TextIO):
        super().__init__()
        self._original_stream = original_stream
        self._lock = threading.Lock()
```

#### 3. Context Manager Implementation
While context managers have been integrated, explicit `__enter__` and `__exit__` methods should be added to ensure proper resource management:

**Example Implementation:**
```python
def __enter__(self):
    return self

def __exit__(self, exc_type, exc_val, exc_tb):
    self.flush()
    return False
```

#### 4. Syntax Error in `converter.py`
There is a syntax error due to a missing parenthesis in the `_create_instructor` method. This must be corrected to prevent runtime errors:

**Correction:**
```python
def _create_instructor(self):
    inst = Instructor(
        llm=self.llm,
        model=self.model,
        content=self.text,
    )  # Added missing parenthesis
    return inst
```

### General Recommendations

1. **Documentation**: Add docstrings to newly modified classes and methods to enhance clarity and maintainability.
    ```python
    class FilteredStream(io.TextIOBase):
        """A thread-safe stream wrapper that filters specific messages."""
    ```

2. **Logging**: Incorporate logging for filtered messages, especially under debug settings, to assist in tracking operations.
    ```python
    logging.debug(f"Filtered message: {s}")
    ```

3. **Unit Tests**: It is crucial to implement unit tests that verify `FilteredStream`'s functionality, ensuring filtering operates as intended.

### Security and Performance Considerations

- The implementation effectively handles threading via locks, which is crucial for multi-threaded environments.
- Consider validation checks for stream inputs to prevent unexpected errors.
- Evaluate potential overhead from threading locks against the performance needs of the application.

## Conclusion
This PR brings notable improvements to thread safety and resource management within the `FilteredStream` class and corrects critical issues in `converter.py`. Addressing the points mentioned will not only bolster the current implementation but also enhance future maintainability and overall robustness.

src/crewai/llm.py
```diff
@@ -54,12 +55,15 @@
...
-class FilteredStream:
-    def __init__(self, original_stream):
+class FilteredStream(io.TextIOBase):
+    def __init__(self, original_stream: TextIO):
         self._original_stream = original_stream
         self._lock = threading.Lock()
 
-    def write(self, s) -> int:
+    def write(self, s: str) -> int:
+        if not hasattr(self, "_lock"):
```
gvieira: Nothing to do now, but I don't love these `hasattr` we have in several places. Hope we can refactor these eventually.

gvieira (approved):

src/crewai/llm.py
lucasgomide: good call.. just refactored that.

gvieira: Lovely! Thanks!

Once again, nothing to do here, but what do you think about being more descriptive? Results are definitely the same, since the compiler takes care of proper typing.
```
_lock: Optional[threading.Lock] = None # although I still prefer the more modern but still unsupported Lock | None
```

gvieira (approved):

PR #3087, opened 2025-06-30
feat: add capability to track LLM calls by task and agent

lucasgomide:
This makes it possible to filter or scope LLM events by specific agents or tasks, which can be very useful for debugging or analytics in real-time application


joaomdmoura:

# Code Review Comment

## Overview
This pull request introduces valuable enhancements for tracking LLM calls with specific tasks and agents, improving the event system's capabilities. 

## Key Changes
### 1. **LLMEventBase Class Enhancement**
The introduction of the `LLMEventBase` class is a significant update that allows for task and agent tracking. 

**Improvement Suggestions**:
- **Validation for IDs**: It’s essential to ensure that the task and agent IDs conform to a predefined pattern to prevent potential issues.
  
  **Example**:
  ```python
  agent_id: Optional[str] = Field(None, pattern=r'^[a-zA-Z0-9\-_]+$')
  ```

- **ID Validation Method**:
  ```python
  @validator('task_id', 'agent_id')
  def validate_ids(cls, v):
      if v and not re.match(r'^[a-zA-Z0-9\-_]+$', v):
          raise ValueError('ID must contain only alphanumeric characters, hyphens, and underscores')
      return v
  ```

### 2. **LLM Call Propagation**
The method has been updated with parameters propagated through the call chain. 

**Improvement Suggestions**:
- Adding type hints enhances clarity and understanding of the parameters.
  
  **Example**:
  ```python
  from_task: Optional[Task] = None  # Type hint addition
  ```

- **Documentation**: Ensure each parameter in the function has a detailed docstring for better understanding.

### 3. **Event Emission Updates**
There's proper propagation of task and agent information to events, which is crucial for tracking.

**Improvement Suggestions**:
- Implement error handling around event emission to gracefully handle failures.
  
  **Example**:
  ```python
  try:
      # Emit event
  except AttributeError as e:
      logger.warning(f"Failed to extract task/agent info: {str(e)}")
  ```

### 4. **Test Coverage**
While good test coverage is present, including negative test cases can further reinforce the reliability of the implementation.

**Improvement Suggestions**:
- **Negative Test Case Example**:
  ```python
  def test_llm_events_with_invalid_task():
      invalid_task = Mock(spec=[])
      with pytest.raises(AttributeError):
          # Logic to ensure invalid tasks raise AttributeError
  ```

## General Recommendations
1. **Error Logging**: Consider adding a centralized logging mechanism for event emission failures to capture and understand failures better.
   
   **Example**:
   ```python
   logger.error(f"Failed to emit event {event.type}: {str(e)}")
   ```

2. **Input Validation**: Implement broader validation for task/agent parameters to ensure they possess the necessary attributes before processing.

3. **Event Filtering**: Implement utility functions for filtering events based on agent ID or task ID. This would allow for more dynamic event handling.

**Example**:
```python
class EventFilter:
    @staticmethod
    def by_agent_id(agent_id: str) -> Callable[[LLMEventBase], bool]:
        return lambda event: event.agent_id == agent_id
```

## Conclusion
Overall, the changes made in this PR are coherent and well-integrated into the existing structure. The suggestions revolve mainly around improving validation, error handling, expanding test coverage, and enhancing code clarity through documentation and logging. The implementation does a commendable job of adding significant value to the LLM call tracking mechanism. Further refinement in these areas will help solidify this contribution, making it robust and reliable in production environments.

src/crewai/utilities/events/llm_events.py
```diff
@@ -5,6 +5,27 @@
...
+
+    def _set_agent_params(self, data: Dict[str, Any]):
+        if "from_task" in data and (agent := getattr(data["from_task"], "agent", None)):
+            self.agent_id = agent.id
+            self.agent_role = agent.role
+
+    def _set_task_params(self, data: Dict[str, Any]):
+        if "from_task" in data and (task := data["from_task"]):
+            self.task_id = task.id
+            self.task_name = task.name
```
lorenzejay: I assume this will work with liteagent?

lucasgomide: Good call! I supposed to.. but won't! 
I just fixed it in the next commit

src/crewai/lite_agent.py
```diff
@@ -129,6 +131,7 @@ class LiteAgent(FlowTrackable, BaseModel):
     model_config = {"arbitrary_types_allowed": True}
 
     # Core Agent Properties
+    id: UUID4 = Field(default_factory=uuid.uuid4, frozen=True)
```
lucasgomide, line 134: I need to add this attribute into LiteAgent. We are propagating from Agent instance to keeping the same ID

lorenzejay (approved):
nice man !

tests/utilities/test_events.py
```diff
@@ -903,3 +903,56 @@ def handle_llm_stream_chunk(source, event):
...
+        agent = Agent(
+            role="Speaker",
+            llm=LLM(model="gpt-4o-mini", stream=True),
+            goal="Just say hi",
+            backstory="You are a helpful assistant that just says hi",
+        )
+        agent.kickoff(messages=[{"role": "user", "content": "say hi!"}])
+
+
+    assert len(completed_event) == 2
```
lucasgomide, line 940: This is weird! It looks we're firing completed events twice! I will handle it in the next PRs

src/crewai/lite_agent.py
```diff
@@ -517,6 +520,7 @@ def _invoke_loop(self) -> AgentFinish:
                         messages=self._messages,
                         tools=None,
                         callbacks=self._callbacks,
+                        from_agent=self,
```
vinibrsl, line 523: Do we also need this information in the event, as it's already passed in as the first argument (`emit(source=self, event=...)`)?

lucasgomide: @vinibrsl I believe so.

This attribute indicates whether the LLM event was triggered by a Task or an Agent.While the source indicates the class that triggered the event, conceptually they are not the same.. It was added to all LLM event classes because these events are typically emitted inside the LLM instance.. 

In this specific case - an Agent-we can be confident that the source is the instance we care about, however we have many other cases where we don't.

vinibrsl (approved):

PR #3133, opened 2025-07-11
Introduce Evaluator Experiment

lucasgomide:
This PR introduce Agent Evaluation Experiments. In agent evaluation, experiments are systematic evaluations of an agent's performance by running it against specific datasets and measuring its behavior against predefined metrics.

You can compare the results to a baseline, track evaluations over time and assert their eval

https://github.com/user-attachments/assets/e0784abb-8dc3-40ae-b3b8-a269263a0b03

This unblock Regression Pipeline creation 🥳

### DIFF
--- a/src/crewai/task.py
+++ b/src/crewai/task.py
@@ -67,6 +67,7 @@ class Task(BaseModel):
         description: Descriptive text detailing task's purpose and execution.
         expected_output: Clear definition of expected task outcome.
         output_file: File path for storing task output.
+        create_directory: Whether to create the directory for output_file if it doesn't exist.
         output_json: Pydantic model for structuring JSON output.
         output_pydantic: Pydantic model for task output.
         security_config: Security configuration including fingerprinting.
@@ -114,6 +115,10 @@ class Task(BaseModel):
     output_file: Optional[str] = Field(
         description="A file path to be used to create a file output.",
         default=None,
+    )
+    create_directory: Optional[bool] = Field(
+        description="Whether to create the directory for output_file if it doesn't exist.",
+        default=True,
     )
     output: Optional[TaskOutput] = Field(
         description="Task output, it's final result after being executed", default=None
@@ -753,8 +758,10 @@ Follow these guidelines:
             resolved_path = Path(self.output_file).expanduser().resolve()
             directory = resolved_path.parent
 
-            if not directory.exists():
+            if self.create_directory and not directory.exists():
                 directory.mkdir(parents=True, exist_ok=True)
+            elif not self.create_directory and not directory.exists():
+                raise RuntimeError(f"Directory {directory} does not exist and create_directory is False")
 
             with resolved_path.open("w", encoding="utf-8") as file:
                 if isinstance(result, dict):
--- a/tests/task_test.py
+++ b/tests/task_test.py
@@ -1133,6 +1133,119 @@ def test_output_file_validation():
         )
 
 
+def test_create_directory_true():
+    """Test that directories are created when create_directory=True."""
+    from pathlib import Path
+    
+    output_path = "test_create_dir/output.txt"
+    
+    task = Task(
+        description="Test task",
+        expected_output="Test output",
+        output_file=output_path,
+        create_directory=True,
+    )
+    
+    resolved_path = Path(output_path).expanduser().resolve()
+    resolved_dir = resolved_path.parent
+    
+    if resolved_path.exists():
+        resolved_path.unlink()
+    if resolved_dir.exists():
+        import shutil
+        shutil.rmtree(resolved_dir)
+    
+    assert not resolved_dir.exists()
+    
+    task._save_file("test content")
+    
+    assert resolved_dir.exists()
+    assert resolved_path.exists()
+    
+    if resolved_path.exists():
+        resolved_path.unlink()
+    if resolved_dir.exists():
+        import shutil
+        shutil.rmtree(resolved_dir)
+
+
+def test_create_directory_false():
+    """Test that directories are not created when create_directory=False."""
+    from pathlib import Path
+    
+    output_path = "nonexistent_test_dir/output.txt"
+    
+    task = Task(
+        description="Test task",
+        expected_output="Test output",
+        output_file=output_path,
+        create_directory=False,
+    )
+    
+    resolved_path = Path(output_path).expanduser().resolve()
+    resolved_dir = resolved_path.parent
+    
+    if resolved_dir.exists():
+        import shutil
+        shutil.rmtree(resolved_dir)
+    
+    assert not resolved_dir.exists()
+    
+    with pytest.raises(RuntimeError, match="Directory .* does not exist and create_directory is False"):
+        task._save_file("test content")
+
+
+def test_create_directory_default():
+    """Test that create_directory defaults to True for backward compatibility."""
+    task = Task(
+        description="Test task",
+        expected_output="Test output",
+        output_file="output.txt",
+    )
+    
+    assert task.create_directory is True
+
+
+def test_create_directory_with_existing_directory():
+    """Test that create_directory=False works when directory already exists."""
+    from pathlib import Path
+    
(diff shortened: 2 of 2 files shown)


joaomdmoura:

# Code Review Comment: crewAI Evaluation Framework

## Summary
Thank you for the submission of the crewAI evaluation framework. This pull request implements a structure for running experiments, comparing results with baselines, and displaying metrics. The overall design is promising but requires attention in various areas to enhance maintainability and performance.

## Key Improvements Suggested

### 1. Type Hints & Documentation
Inconsistencies in type hints and missing docstrings can lead to confusion during usage. For instance:
```python
def _extract_scores(self, agent_evaluations: Dict[str, AgentAggregatedEvaluationResult]) -> Union[int, Dict[str, int]]:
```
**Proposed Change**:
```python
def _extract_scores(self, agent_evaluations: dict[str, AgentAggregatedEvaluationResult]) -> int | dict[str, int]:
    """Extract and aggregate evaluation scores from agent results."""
```

### 2. Error Handling
Currently, there is generic exception handling which can obscure the source of errors:
```python
except Exception as e:
```
Incorporating specific error types helps diagnose issues:
```python
class InvalidDatasetError(ExperimentError):
    pass
# Include in _run_test_case() to handle input validation properly.
```

### 3. Code Organization
Certain methods are excessively lengthy, such as `_compare_with_run()`. Consider extracting logic into separate classes for clear responsibilities. For example:
```python
class ExperimentComparison:
    # Handles the comparison logic.
```
This separation can improve testability and organization.

### 4. Performance Enhancements
To mitigate redundant computations during baseline comparisons, introduce caching mechanisms. For example:
```python
@functools.lru_cache(maxsize=1)
def _load_baseline_data(self, filepath: str):
```
This can considerably reduce the overhead of repeated file operations.

### 5. Testing Coverage
Expand testing to include edge cases and performance scenarios. Current test setups are limited, which may lead to undiagnosed errors in production. For instance:
```python
@pytest.mark.parametrize("test_case", [
    # Add tests for various scenarios, including edge cases.
])
```

### 6. Configuration Management
Instead of hardcoded values, utilize configuration classes to facilitate future adjustments. For example:
```python
@dataclass
class ExperimentConfig:
    min_score_threshold: float = 0.7
    display_format: str = "rich"
```
This can add flexibility to the framework's usage.

## Historical Context & Related PRs
While I couldn't fetch historical context from related pull requests due to tool access restrictions, reviewing previous modifications to the `src/crewai/evaluation` directory may provide valuable insights into design decisions and usage patterns.

## Closing Remarks
Overall, the pull request enhances the crewAI evaluation framework but requires some refinement to ensure a solid foundation for future development. Incorporating the above suggestions should improve code quality, maintainability, and performance.

Thank you for the effort put into this development! Looking forward to your updates.

src/crewai/experimental/evaluation/agent_evaluator.py
```diff
@@ -0,0 +1,182 @@
...
+from crewai.crew import Crew
+from crewai.utilities.events.crewai_event_bus import crewai_event_bus
+from crewai.utilities.events.utils.console_formatter import ConsoleFormatter
+from crewai.experimental.evaluation.evaluation_display import AgentAggregatedEvaluationResult
+
+class AgentEvaluator:
+    def __init__(
+        self,
+        evaluators: Sequence[BaseEvaluator] | None = None,
+        crew: Crew | None = None,
```
vinibrsl, line 19: What if instead of passing a crew here, this constructor took a list of agents?

I see some `crew` requirements down this file, meaning this would not work with Lite Agents. If not too big of a lift, I'd recommend adding support for it.

src/crewai/experimental/evaluation/agent_evaluator.py
```diff
@@ -0,0 +1,182 @@
...
+                    if task.agent and str(task.agent.id) != str(agent.id):
+                        continue
+
+                    trace = self.callback.get_trace(str(agent.id), str(task.id))
+                    if not trace:
+                        self.console_formatter.print(f"[yellow]Warning: No trace found for agent {agent.role} on task {task.description[:30]}...[/yellow]")
+                        progress.update(eval_task, advance=1)
+                        continue
+
+                    with crewai_event_bus.scoped_handlers():
```
vinibrsl, line 84: I think you need to call `create_evaluation_callbacks` inside the `scoped_handlers` `with` block for scoped handlers to actually work. Otherwise, you can remove the block and use handlers directly if applicable.

src/crewai/experimental/evaluation/agent_evaluator.py
```diff
@@ -0,0 +1,182 @@
...
+            return self.iterations_results[self.iteration]
+
+        return self.evaluate_current_iteration()
+
+    def display_results_with_iterations(self):
+        self.display_formatter.display_summary_results(self.iterations_results)
+
+    def get_agent_evaluation(self, strategy: AggregationStrategy = AggregationStrategy.SIMPLE_AVERAGE, include_evaluation_feedback: bool = False) -> Dict[str, AgentAggregatedEvaluationResult]:
+        agent_results = {}
+        with crewai_event_bus.scoped_handlers():
```
vinibrsl, line 108: https://github.com/crewAIInc/crewAI/pull/3133/files#r2204762954

src/crewai/experimental/evaluation/evaluation_display.py
```diff
@@ -0,0 +1,341 @@
...
+from typing import Dict, Any, List
+from rich.table import Table
+from rich.box import HEAVY_EDGE, ROUNDED
+from collections.abc import Sequence
+from crewai.experimental.evaluation.base_evaluator import AgentAggregatedEvaluationResult, AggregationStrategy, AgentEvaluationResult, MetricCategory
+from crewai.experimental.evaluation import EvaluationScore
+from crewai.utilities.events.utils.console_formatter import ConsoleFormatter
+from crewai.utilities.llm_utils import create_llm
+
+class EvaluationDisplayFormatter:
```
vinibrsl, line 11: Would you mind removing the comments in this file?

src/crewai/experimental/evaluation/evaluation_listener.py
```diff
@@ -0,0 +1,190 @@
...
+            cls._instance = super().__new__(cls)
+            cls._instance._initialized = False
+        return cls._instance
+
+    def __init__(self):
+        if not hasattr(self, "_initialized") or not self._initialized:
+            super().__init__()
+            self.traces = {}
+            self.current_agent_id = None
+            self.current_task_id = None
```
vinibrsl, line 47: Adding state to this class would make this feature not thread-safe, if I’m not mistaken? Events may occur for any running crew or agent. Even with scoped handlers, I'd do some testing to guarantee that.

As a suggestion, I'd recommend adding stateful data to the event itself, where applicable.

src/crewai/experimental/evaluation/agent_evaluator.py
lucasgomide: This is one thing I'm thinking to enhance: remove crew dependency. 

I initially chose this approach because it made it easier to track agent.task, but it introduced a limitation: we can’t evaluate LiteAgent anymore.

I will drop a PR with this improvement very soon

src/crewai/experimental/evaluation/agent_evaluator.py
lucasgomide: Are you sure?

I've tested this, and it works pretty well. It was added to prevent missing events fired from the evaluator to the EventListener. I had run into some UI issues when the LLMStarted event was fired; the Crew branch pop up indicating that the finished crew is "Thinking..." but is not true

What kind of test should I run to ensure this is working properly?

src/crewai/experimental/evaluation/evaluation_listener.py
lucasgomide: good call!

lucasgomide: @vinibrsl I will addressed this point in the next PR

PR #3151, opened 2025-07-14
Skip checking docker installation if unsafe mode

LWisteria:
When I ran crew in condition with

* `allow_code_execution=True`
* `code_execution_mode = "unsafe"`

following error was gotten:

```
RuntimeError: Docker is not installed. Please install Docker to use code execution with agent:
```

But as [the document](https://docs.crewai.com/en/concepts/agents#code-execution) said, docker is required only if with `code_execution_mode = "safe"`.

Thus, skipping this validation is the best way to solve this problem.

### DIFF
--- a/src/crewai/agent.py
+++ b/src/crewai/agent.py
@@ -187,8 +187,8 @@ class Agent(BaseAgent):
         if not self.agent_executor:
             self._setup_agent_executor()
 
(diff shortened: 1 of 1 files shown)


lucasgomide:
@LWisteria nice work! 

can you add some test to cover it?

PR #3158, opened 2025-07-14
feat: implement thread-safe event bus

lucasgomide:
This PR also added tests to ensure singleton pattern


tests/utilities/events/test_crewai_event_bus.py
```diff
@@ -1,13 +1,24 @@
...
+import pytest
 
 from crewai.utilities.events.base_events import BaseEvent
 from crewai.utilities.events.crewai_event_bus import crewai_event_bus
 
 
+@pytest.fixture(autouse=True)
+def scoped_event_handlers():
+    with crewai_event_bus.scoped_handlers():
+        yield
```
lucasgomide, line 14: Since event handlers registered in one test would persist and interfere with subsequent tests running in the same thread, I have to added it to isolate them

tests/utilities/events/test_crewai_event_bus.py
```diff
@@ -45,3 +56,98 @@ def broken_handler(source, event):
     out, err = capfd.readouterr()
     assert "Simulated handler failure" in out
     assert "Handler 'broken_handler' failed" in out
+
+
+def test_singleton_pattern_across_threads():
```
lucasgomide, line 68: Just to make sure singleton pattern of `crewai_event_bus` when running with multiples threads

tests/utilities/events/test_crewai_event_bus.py
```diff
@@ -45,3 +56,98 @@ def broken_handler(source, event):
...
+
+    for thread in threads:
+        thread.join()
+    assert len(instances) == 10
+    for instance in instances:
+        assert instance is crewai_event_bus
+        assert instance is instances[0]
+
+
+def test_thread_local_handler_isolation():
```
lucasgomide: Ensures each thread `mock_handler` is only called by events from that same thread.
Handlers are isolated per thread instead of being shared globally - it was the main improvement here

lorenzejay (approved):

thiagomoretto:
I recommend testing with async tasks and tools, since they run in an async loop thread and thus a different thread-local.

greysonlalonde:
Closing as stale, feel free to reopen if still planned though!

PR #3168, opened 2025-07-15
Emit events about Agent eval

lucasgomide:
We are triggering events when an evaluation has started/completed/failed

### DIFF
--- a/src/crewai/experimental/evaluation/agent_evaluator.py
+++ b/src/crewai/experimental/evaluation/agent_evaluator.py
@@ -1,17 +1,18 @@
+import threading
+from typing import Any
+
 from crewai.experimental.evaluation.base_evaluator import AgentEvaluationResult, AggregationStrategy
 from crewai.agent import Agent
 from crewai.task import Task
 from crewai.experimental.evaluation.evaluation_display import EvaluationDisplayFormatter
-
-from typing import Any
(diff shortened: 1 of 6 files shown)


src/crewai/experimental/evaluation/agent_evaluator.py
```diff
@@ -1,17 +1,18 @@
+import threading
```
lorenzejay, line 1: couple type errors here:
Screenshot 2025-07-15 at 12 12 08 PM

lorenzejay (changes requested):

src/crewai/experimental/evaluation/agent_evaluator.py
lucasgomide: should these error be warning by mypy?

lorenzejay: type-checking is on basic for me

lucasgomide: Same here! 🙂 I'm just wondering if we should rely on your local type-checking setup or the one used in CI.. just to keep things consistent. Either option works for me!

Maybe we missing some setup on mypy to make it better

lorenzejay (approved):

PR #3170, opened 2025-07-15
feat: add regression tests and configure its workflow

lucasgomide:


.github/workflows/regression-tests.yml
```diff
@@ -0,0 +1,75 @@
...
+        description: 'Branch to run tests on'
+        required: true
+        default: 'main'
+        type: string
+
+permissions:
+  contents: write
+
+env:
+  OPENAI_API_KEY: fake-api-key
```
lucasgomide, line 16: i'm waiting for a dedicated `OPENAI_API_KEY` to run these tests.

PR #3173, opened 2025-07-17
Adding support to multiple mcp servers on @CrewBase

thiagomoretto:


src/crewai/project/crew_base.py
```diff
@@ -67,36 +68,57 @@ def __init__(self, *args, **kwargs):
...
+                    if server is not None and server_name != server:
+                        continue
+
+                    adapter = self._mcp_server_adapter.get(server_name, None)
+                    if not adapter:
+                        self._mcp_server_adapter[server_name] = MCPServerAdapter(params)
+                    aggregated_tools.extend(
+                        self._mcp_server_adapter[server_name].tools.filter_by_names(tool_names or None))
+
+            return aggregated_tools
```
lucasgomide, line 121: This return statement should be aligned with the `isinstance(self.mcp_server_params, dict)` block, since aggregated_tools is defined within it.

src/crewai/project/crew_base.py
```diff
@@ -67,36 +68,57 @@ def __init__(self, *args, **kwargs):
...
-            adapter = getattr(self, '_mcp_server_adapter', None)
-            if not adapter:
-                self._mcp_server_adapter = MCPServerAdapter(self.mcp_server_params)
-
-            return self._mcp_server_adapter.tools.filter_by_names(tool_names or None)
+            if isinstance(self.mcp_server_params, list):
+                if self._mcp_server_adapter is None:
+                    self._mcp_server_adapter = MCPServerAdapter(self.mcp_server_params)
+                if server is not None and len(self.mcp_server_params) > 1:
+                    logging.warning("Using list of MCP server parameters. To use server parameter, please use a dictionary of MCP server parameters.")
```
lucasgomide, line 101: How is this information being displayed to the user? Could you share a screenshot? What about raise an error to prevent misleading configuration

src/crewai/project/crew_base.py
```diff
@@ -67,36 +68,57 @@ def __init__(self, *args, **kwargs):
...
+                # Type assertion: when mcp_server_params is a list, _mcp_server_adapter is a single MCPServerAdapter
+                adapter = cast(Any, self._mcp_server_adapter)
+                return adapter.tools.filter_by_names(tool_names or None)
+
+            # Separated MCP adapters for each server.
+            elif isinstance(self.mcp_server_params, dict):
```
lucasgomide, line 107: what happens when `self.mcp_server_params` is a dict and the requested server is not found?

tests/project_test.py
```diff
@@ -87,7 +87,7 @@ def crew(self):
 
 @CrewBase
 class InternalCrewWithMCP(InternalCrew):
-    mcp_server_params = {"host": "localhost", "port": 8000}
+    mcp_server_params = [{"url": "localhost", "port": 8000}]
```
lucasgomide, line 90: Currently, we support defining MCP server parameters using a single dictionary.

With your change, users with that configuration might run into issues..  it could introduce a breaking change, right?

To prevent that, what do you think about improving the multiple server detection logic? For example, we could check if the dictionary contains keys whose values are also dictionaries (or have nested keys), which would suggest a multi-server config.

src/crewai/project/crew_base.py
lucasgomide: Would be great adding tests for that

PR #3183, opened 2025-07-18
feat: remove duplicate print on LLM call error

lucasgomide:
Exceptions raised by LiteLLM are now clearer, and we're no longer duplicating or printing excessive error logs.

Original (main branch)

https://github.com/user-attachments/assets/847e8487-844d-4001-b622-3d0ffe3ccb15

Current branch

https://github.com/user-attachments/assets/88144b92-fd8a-4bdb-8f74-4c4361bd56fc

### DIFF
--- a/src/crewai/agents/crew_agent_executor.py
+++ b/src/crewai/agents/crew_agent_executor.py
@@ -120,11 +120,8 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
             raise
         except Exception as e:
             handle_unknown_error(self._printer, e)
-            if e.__class__.__module__.startswith("litellm"):
-                # Do not retry on litellm errors
-                raise e
-            else:
-                raise e
+            raise
+
 
         if self.ask_for_human_input:
             formatted_answer = self._handle_human_feedback(formatted_answer)
--- a/src/crewai/llm.py
+++ b/src/crewai/llm.py
@@ -59,6 +59,7 @@ from crewai.utilities.exceptions.context_window_exceeding_exception import (
 
 load_dotenv()
 
+litellm.suppress_debug_info = True
 
 class FilteredStream(io.TextIOBase):
     _lock = None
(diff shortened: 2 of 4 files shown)


src/crewai/agents/crew_agent_executor.py
```diff
@@ -120,11 +120,8 @@ def invoke(self, inputs: Dict[str, str]) -> Dict[str, Any]:
             raise
         except Exception as e:
             handle_unknown_error(self._printer, e)
-            if e.__class__.__module__.startswith("litellm"):
-                # Do not retry on litellm errors
-                raise e
-            else:
-                raise e
```
lucasgomide, line 127: useless if/else..

src/crewai/llm.py
```diff
@@ -59,6 +59,7 @@
 
 load_dotenv()
 
+litellm.suppress_debug_info = True
```
lucasgomide, line 62: This will prevent logs like "give feedback / get help" in lower_s from being printed

src/crewai/llm.py
```diff
@@ -989,7 +988,6 @@ def call(
                     self,
                     event=LLMCallFailedEvent(error=str(e), from_task=from_task, from_agent=from_agent),
                 )
-                logging.error(f"LiteLLM call failed: {str(e)}")
```
lucasgomide, line 1008: We still raising the original exception so this log.error is useless

src/crewai/utilities/agent_utils.py
```diff
@@ -232,12 +228,17 @@ def handle_unknown_error(printer: Any, exception: Exception) -> None:
         printer: Printer instance for output
         exception: The exception that occurred
     """
+    error_message = str(exception)
+
+    if "litellm" in error_message:
+        return
```
lucasgomide, line 234: Prevent printing any error to the terminal if its message contains “litellm”.

src/crewai/agents/crew_agent_executor.py
```diff
@@ -120,11 +120,8 @@ def invoke(self, inputs: Dict[str, str]) -> Dict[str, Any]:
...
+            raise e
```
3x0dv5: Consider replacing `raise e` with simply `raise`. In Python 3, invoking raise inside an except block re-raises the caught exception, preserving its original traceback and full context. By contrast, using `raise e` resets the traceback, which can obscure the exception chain and make it harder to trace back to the root cause in earlier frames.

I raised a PR to change all the `raise e`: https://github.com/crewAIInc/crewAI/pull/3193

lorenzejay (approved):

PR #3184, opened 2025-07-18
feat: drop unsupported stop parameter for LLM models automatically

lucasgomide:


src/crewai/llm.py
```diff
@@ -984,6 +984,27 @@ def call(
                 # whether to summarize the content or abort based on the respect_context_window flag
                 raise
             except Exception as e:
```
vinibrsl, line 986: Should we rescue that specific exception here? `UnsupportedParamsError`?

src/crewai/llm.py
```diff
@@ -984,6 +984,27 @@ def call(
...
+                unsupported_stop = "Unsupported parameter" in str(e) and "'stop'" in str(e)
+
+                if unsupported_stop:
+                    if "additional_drop_params" in self.additional_params and isinstance(self.additional_params["additional_drop_params"], list):
+                        self.additional_params["additional_drop_params"].append("stop")
+                    else:
+                        self.additional_params = {"additional_drop_params": ["stop"]}
```
vinibrsl, line 993: QQ: Why don't we add `stop` by default in `additional_drop_params` if we'll retry it with it anyways?

vinibrsl (approved):

src/crewai/llm.py
lucasgomide: We can't. It was a very generic exceptions raised by LiteLLM

```
litellm.exceptions.BadRequestError: litellm.BadRequestError: OpenAIException - Unsupported parameter: 'stop' is not supported with this model.
```

src/crewai/llm.py
lucasgomide: The `stop` parameter is added by default because most LLM models support it. However, since there's no straightforward way to check for support - frankly there's but is not accuracy -  we retry the call dropping  this parameter when a related error is raised.

vinibrsl (approved):

PR #3209, opened 2025-07-23
fix(converter): Ensure agent is available in Converter

daydaychen:
The Converter class's fallback mechanism for re-attempting Pydantic model conversion was failing due to a missing agent reference. This change ensures the agent is passed to the Converter instance, making the conversion process more robust.


lucasgomide (commented):
@daydaychen thank you for collaborating! 

Would you mind wrote regression tests for that?

daydaychen:
This implementation is a bit problematic, I'll deal with it later.

PR #3216, opened 2025-07-24
Changed the default value in Mem0 config

Vidit-Ostwal:
`infer=False` is somehow not working, default value has been set to `True`.

Following up with Mem0 team here: https://github.com/mem0ai/mem0/issues/3215

@lucasgomide


lucasgomide (commented):
would be great adding a regression test for it

Vidit-Ostwal:
> would be great adding a regression test for it

Yes, added.

PR #3217, opened 2025-07-24
fix: support to add memories to Mem0 with agent_id

lucasgomide:
Memories are not being saved to Mem0 when user_id is missing and only agent_id is set.

https://github.com/user-attachments/assets/69d591a3-b735-4af0-a8a4-7af51f9b80bc

Aside from that, we had unintentionally dropped an essential search feature. This PR reintroduces that behavior

### DIFF
--- a/src/crewai/memory/storage/mem0_storage.py
+++ b/src/crewai/memory/storage/mem0_storage.py
@@ -85,7 +85,8 @@ class Mem0Storage(Storage):
             agent_id = self.config.get("agent_id", "")
 
             if user_id and agent_id:
-                filter["OR"].append({"user_id": user_id, "agent_id": agent_id})
+                filter["OR"].append({"user_id": user_id})
+                filter["OR"].append({"agent_id": agent_id})
             elif user_id:
                 filter["AND"].append({"user_id": user_id})
             elif agent_id:
--- a/src/crewai/memory/storage/mem0_storage.py
+++ b/src/crewai/memory/storage/mem0_storage.py
@@ -79,7 +79,11 @@ class Mem0Storage(Storage):
 
         # Add user_id condition if the memory type is external
         if self.memory_type == "external":
-            filter["AND"].append({"user_id": self.config.get("user_id", "")})
+            if user_id := self.config.get("user_id", ""):
+                filter["AND"].append({"user_id": user_id})
+
+            if agent_id := self.config.get("agent_id", ""):
+                filter["AND"].append({"agent_id": agent_id})
 
         # Add run_id condition if the memory type is short_term and a run ID is set
         if self.memory_type == "short_term" and self.mem0_run_id:
@@ -89,7 +93,7 @@ class Mem0Storage(Storage):
 
     def save(self, value: Any, metadata: Dict[str, Any]) -> None:
         user_id = self.config.get("user_id", "")
-        assistant_message = [{"role" : "assistant","content" : value}] 
+        assistant_message = [{"role" : "assistant","content" : value}]
 
         base_metadata = {
             "short_term": "short_term",
@@ -105,9 +109,12 @@ class Mem0Storage(Storage):
         }
 
         if self.memory_type == "external":
-            params["user_id"] = user_id
+            if user_id:
+                params["user_id"] = user_id
+            if agent_id := self.config.get("agent_id", ""):
+                params["agent_id"] = agent_id
 
-        
+
         if params:
             # MemoryClient-specific overrides
             if isinstance(self.memory, MemoryClient):
@@ -123,12 +130,12 @@ class Mem0Storage(Storage):
 
     def search(self,query: str,limit: int = 3,score_threshold: float = 0.35) -> List[Any]:
         params = {
-            "query": query, 
-            "limit": limit, 
+            "query": query,
+            "limit": limit,
             "version": "v2",
             "output_format": "v1.1"
             }
-        
+
         if user_id := self.config.get("user_id", ""):
             params["user_id"] = user_id
 
@@ -138,7 +145,7 @@ class Mem0Storage(Storage):
             "entities": {"type": "entity"},
             "external": {"type": "external"},
         }
-        
+
         if self.memory_type in memory_type_map:
             params["metadata"] = memory_type_map[self.memory_type]
             if self.memory_type == "short_term":
@@ -151,11 +158,13 @@ class Mem0Storage(Storage):
         params['threshold'] = score_threshold
 
         if isinstance(self.memory, Memory):
-            del params["metadata"], params["version"], params["run_id"], params['output_format']
+            del params["metadata"], params["version"], params['output_format']
+            if params.get("run_id"):
+                del params["run_id"]
 
         results = self.memory.search(**params)
         return [r for r in results["results"]]
-    
+
     def reset(self):
         if self.memory:
             self.memory.reset()
--- a/src/crewai/memory/storage/mem0_storage.py
+++ b/src/crewai/memory/storage/mem0_storage.py
@@ -2,6 +2,7 @@ import os
 from typing import Any, Dict, List
 from collections import defaultdict
 from mem0 import Memory, MemoryClient
+from crewai.utilities.chromadb import sanitize_collection_name
 
 from crewai.memory.storage.interface import Storage
 
@@ -122,7 +123,7 @@ class Mem0Storage(Storage):
         if user_id:
             params["user_id"] = user_id
 
-        if agent_id := self.config.get("agent_id", ""):
+        if agent_id := self.config.get("agent_id", self._get_agent_name()):
             params["agent_id"] = agent_id
 
         self.memory.add(assistant_message, **params)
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:
@lucasgomide I had purposely removed the agent_id. As memory in CrewAI is not agent specific.
Also in retrieval as well I was facing problem when having the `agent_id`

Am I missing something?

lucasgomide:
> memory i

@Vidit-Ostwal 
Actually users are asking for this already, so it looks like we shouldn't drop it.

Vidit-Ostwal:
> > memory i
> 
> @Vidit-Ostwal Actually users are asking for this already, so it looks like we shouldn't drop it.

Got it, we should keep it.
Just check once that you are able to retrieve the memory, because I was facing problems with it yesterday.

lucasgomide:
> Also in retrieval as well I was facing problem when having the `agent_id`

I faced an issue to retrieve memories using both user_id and agent. I just [opened a issue](https://github.com/mem0ai/mem0/issues/3218) on their side to understand if i'm missing something or is a real bug. 
However I was able to retrieve memories when only the agent_ud is provided

Vidit-Ostwal:
> I faced an issue to retrieve memories using both user_id and agent.

Yes I was facing this only yesterday.

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -79,7 +79,11 @@ def _create_filter_for_search(self):
 
         # Add user_id condition if the memory type is external
         if self.memory_type == "external":
-            filter["AND"].append({"user_id": self.config.get("user_id", "")})
+            if user_id := self.config.get("user_id", ""):
+                filter["AND"].append({"user_id": user_id})
+
+            if agent_id := self.config.get("agent_id", ""):
+                filter["AND"].append({"agent_id": agent_id})
```
Vidit-Ostwal: @lucasgomide, I think 

this particular line of code 
```python
if agent_id := self.config.get("agent_id", ""):
                filter["AND"].append({"agent_id": agent_id})
```

should be out of the if condition. Consider the case when mem0 is being used by some other memory_type other than external.

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -105,9 +109,12 @@ def save(self, value: Any, metadata: Dict[str, Any]) -> None:
         }
 
         if self.memory_type == "external":
-            params["user_id"] = user_id
+            if user_id:
+                params["user_id"] = user_id
+            if agent_id := self.config.get("agent_id", ""):
+                params["agent_id"] = agent_id
```
Vidit-Ostwal: Same here

src/crewai/memory/storage/mem0_storage.py
lucasgomide: Actually I didn't get why we need this entire `self.memory_type == "external"`

lucasgomide:
I get this docs from Mem0

https://docs.mem0.ai/platform/quickstart#long-term-memory-for-both-users-and-agents

From what I understood we should support `agent_id` and drop `user_id` because we are storing messages always using `role: 'assistant'`.

lucasgomide:
I have a plan for that:
- Fix the adding memories when `agent_id` is provided; We shouldn't drop it.
- Refactor this class and drop the `user_id` support until the version 0.156.0 or on 2025-08-04. Since this is the deadline to remove UserMemory class

src/crewai/memory/storage/mem0_storage.py
lucasgomide: Do you have any thoughts?

Vidit-Ostwal: I think you are right in here.

Vidit-Ostwal:
> I get this docs from Mem0
> 
> https://docs.mem0.ai/platform/quickstart#long-term-memory-for-both-users-and-agents
> 
> From what I understood we should support `agent_id` and drop `user_id` because we are storing messages always using `role: 'assistant'`.

Aligned.

Vidit-Ostwal:
> I have a plan for that:
> 
> * Fix the adding memories when `agent_id` is provided; We shouldn't drop it.
> * Refactor this class and drop the `user_id` support until the version 0.156.0 or on 2025-08-04. Since this is the deadline to remove UserMemory class

Sounds good to me. 
I can pick up the second one.

will drop both the UserMemory and user_id parameter together in a single PR.

Vidit-Ostwal:
@lucasgomide, If we drop the `user_id` from external memory and user memory (legacy) as well, then we won't have ` user-specific memory that persists across sessions and integrates with the crew’s contextual memory system.` which was the aim of using external memory instead of `user_memory`.

Vidit-Ostwal:
Or what we can do is, send up the list of messages we have and then as per `user` , `assisstant` we have, we can send the request to Mem0.

```python
{'role': 'system', 'content': 'You are Friendly chatbot assistant. You are a kind and knowledgeable chatbot assistant. You excel at understanding user needs, providing helpful responses, and maintaining engaging conversations. You remember previous interactions to provide a personalized experience.\nYour personal goal is: Engage in useful and interesting conversations with users while remembering context.\nTo give my best complete final answer to the task respond using the exact following format:\n\nThought: I now can give a great answer\nFinal Answer: Your final answer must be the great and the most complete as possible, it must be outcome described.\n\nI MUST use these formats, my job depends on it!'}
{'role': 'user', 'content': '\nCurrent Task: Respond to user conversation. User message: Can you remember my name ?\n\nThis is the expected criteria for your final answer: Contextually appropriate, helpful, and friendly response.\nyou MUST return the actual complete content as the final answer, not a summary.\n\nBegin! This is VERY important to you, use the tools available and give your best Final Answer, your job depends on it!\n\nThought:'}
{'role': 'assistant', 'content': "I now can give a great answer  \nFinal Answer: While I don’t have the ability to remember your name right now, I’d love to learn it! Feel free to share your name, and I’ll remember it for our future chats. I'm here to help and make our conversations as personal and enjoyable as possible!"}
```
This is what we are getting, I think we can send both user and assistant messages to the mem0.

lucasgomide:
This discussion it is valuable! It might be more appropriate open an issue to discuss it properly @Vidit-Ostwal

Vidit-Ostwal:
> I have a plan for that:
> 
> * Fix the adding memories when `agent_id` is provided; We shouldn't drop it.
> * Refactor this class and drop the `user_id` support until the version 0.156.0 or on 2025-08-04. Since this is the deadline to remove UserMemory class

#3220

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -70,26 +70,31 @@ def _create_filter_for_search(self):
...
-            filter["AND"].append({"user_id": self.config.get("user_id", "")})
+        filter = defaultdict(list)
 
-        # Add run_id condition if the memory type is short_term and a run ID is set
         if self.memory_type == "short_term" and self.mem0_run_id:
             filter["AND"].append({"run_id": self.mem0_run_id})
+        else:
+            user_id = self.config.get("user_id", "")
+            agent_id = self.config.get("agent_id", "")
+
```
Vidit-Ostwal, line 86: Are we giving users the chance to introduce the `agent_id` now?

earlier it was something like this 


```python
    def _get_agent_name(self) -> str:
        if not self.crew:
            return ""

        agents = self.crew.agents
        agents = [self._sanitize_role(agent.role) for agent in agents]
        agents = "_".join(agents)
        return sanitize_collection_name(name=agents,max_collection_length=MAX_AGENT_ID_LENGTH_MEM0)

    agent_name = self._get_agent_name()
```
We used to set `agent_id` according to the agents.

lucasgomide: I think we can keep both. I will add this method again

Vidit-Ostwal: I think this is all that you will require.

```python
from crewai.utilities.chromadb import sanitize_collection_name

MAX_AGENT_ID_LENGTH_MEM0 = 255


def _get_agent_name(self) -> str:
    if not self.crew:
        return ""

    agents = self.crew.agents
    agents = [self._sanitize_role(agent.role) for agent in agents]
    agents = "_".join(agents)
    return sanitize_collection_name(name=agents,max_collection_length=MAX_AGENT_ID_LENGTH_MEM0)

def _sanitize_role(self, role: str) -> str:
    """
    Sanitizes agent roles to ensure valid directory names.
    """
    return role.replace("\n", "").replace(" ", "_").replace("/", "_")
```

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -104,31 +111,32 @@ def save(self, value: Any, metadata: Dict[str, Any]) -> None:
...
-                params["includes"] = self.includes
-                params["excludes"] = self.excludes
-                params["output_format"] = "v1.1"
-                params["version"]="v2"
+        if user_id:
+            params["user_id"] = user_id
 
-                if self.memory_type == "short_term":
-                    params["run_id"] = self.mem0_run_id
+        if agent_id := self.config.get("agent_id", self._get_agent_name()):
```
lorenzejay, line 127: love the walrus operator ;)

lorenzejay (approved):

lorenzejay:
lgtm

PR #3218, opened 2025-07-25
fix(flow): Prevent initial_state BaseModel from being coerced to dict

### LINKED ISSUE #3147: [BUG] Flow initial_state gets coerced into dict, causing attribute collision and unexpected behavior
### Description

When passing a custom Pydantic BaseModel object as initial_state to a Flow, it seems that the state is silently converted to a dict, which leads to unexpected behavior. In particular, if the model contains a field named items, it gets overridden by the built-in dict.items() method, causing runtime errors.

### Steps to Reproduce

1.Define a Pydantic state model with a field named items (or any name that collides with built-in dict methods like keys, values, etc.)

2.Initialize a Flow with this model as the initial_state

3.Call .kickoff()

4.Try to access flow.state.items expecting the original field, and instead encounter a method

### Expected behavior

Expected flow.state.items to be the list [1, 2, 3], but it's actually a method — looks like dict.items()?

### Screenshots/Code snippets

```python
from crewai.flow import Flow

class MyState(BaseModel):
    items: list = [1, 2, 3]
    metadata: dict = {"x": 1}

flow = Flow(initial_state=MyState())
flow.kickoff()

# This will print a method, not a list
print("State items:", flow.state.items)

# This raises TypeError: object of type 'builtin_function_or_method' has no len()
len(flow.state.items)
```

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

0.130.0

### crewAI Tools Version

pydantic==2.9.2

### Virtual Environment

Conda

### Evidence

TypeError: object of type 'builtin_function_or_method' has no len()

### Possible Solution

None

### Additional context

None

diego-coder:
Fixes #3147

This PR resolves a critical bug where passing a Pydantic `BaseModel` as `initial_state` to a `Flow` would cause it to be silently coerced into a dictionary. This led to attribute collisions and runtime errors if the `BaseModel` contained fields with names matching dictionary methods (e.g., `items`, `keys`, `values`).

The fix ensures that the `BaseModel` type is preserved throughout the flow's lifecycle by using Pydantic's native copy methods and correcting the `Flow` constructor.

#### Key Changes & Fixes

1.  **Corrected `Flow` Constructor:** The `__init__` method in `src/crewai/flow/flow.py` was modified to properly accept the `initial_state` parameter. Previously, this parameter was being ignored and incorrectly handled, which was the root cause of the state being initialized as a dictionary.

2.  **Preserved `BaseModel` on Creation:** The `_create_initial_state` method was updated to use `model.model_copy()` (Pydantic v2) or `model.copy()` (Pydantic v1) instead of converting the model to a dictionary. This ensures the state is a `BaseModel` instance from the very beginning.

3.  **Safe `BaseModel` Updates:** The `_initialize_state` method was updated to use `model.copy(update=...)` for applying inputs. This is a much safer and more direct way to update Pydantic models than the previous dictionary-merging approach.

#### Changes to Tests

A new, comprehensive test suite was created in `tests/test_flow_initial_state_fix.py` to validate the fix and prevent regressions:

1.  **Reproduces the Bug:** The tests confirm the original bug by using state models with colliding attribute names like `items`, `keys`, `values`, and `get`.
2.  **Validates the Fix:** The assertions now pass, proving that `flow.state` remains a `BaseModel` instance and that attributes can be accessed correctly.
3.  **Corrected Test Models:** All Pydantic models used for testing were updated to inherit from `FlowState`. This was a necessary change because the fix correctly enforces the requirement that all state models must have an `id` field, a detail the initial automated PR's tests missed.

#### Comparison to Previous Automated PR
The initial automated PR (#3148) correctly identified the need for `model_copy()` but failed because it missed two critical issues that this PR addresses:
1.  It did not fix the `Flow` constructor, which was the primary source of the bug.
2.  Its tests were incomplete and failed because they did not account for the mandatory `id` field in state models.

#### Code Quality & Linting
All modified files (`src/crewai/flow/flow.py` and `tests/test_flow_initial_state_fix.py`) have been formatted with `ruff format` and checked with `ruff check --fix` to ensure they adhere to the project's coding standards.

### DIFF
--- a/src/crewai/flow/flow.py
+++ b/src/crewai/flow/flow.py
@@ -69,19 +69,6 @@ def ensure_state_type(state: Any, expected_type: Type[StateT]) -> StateT:
         TypeError: If state doesn't match expected type
         ValueError: If state validation fails
     """
-    """Ensure state matches expected type with proper validation.
-
-    Args:
-        state: State instance to validate
-        expected_type: Expected type for the state
-
-    Returns:
-        Validated state instance
-
-    Raises:
-        TypeError: If state doesn't match expected type
-        ValueError: If state validation fails
-    """
     if expected_type is dict:
         if not isinstance(state, dict):
             raise TypeError(f"Expected dict, got {type(state).__name__}")
@@ -446,15 +433,20 @@ class Flow(Generic[T], metaclass=FlowMeta):
 
     def __init__(
         self,
+        initial_state: Union[Type[T], T, None] = None,
         persistence: Optional[FlowPersistence] = None,
         **kwargs: Any,
     ) -> None:
         """Initialize a new Flow instance.
-
         Args:
+            initial_state: Initial state for the flow (BaseModel instance or dict)
             persistence: Optional persistence backend for storing flow states
             **kwargs: Additional state values to initialize or override
         """
+        # Set the initial_state for this instance
+        if initial_state is not None:
+            self.initial_state = initial_state
+
         # Initialize basic instance attributes
         self._methods: Dict[str, Callable] = {}
         self._method_execution_counts: Dict[str, int] = {}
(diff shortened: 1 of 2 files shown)


src/crewai/flow/flow.py
```diff
@@ -58,19 +58,6 @@ class FlowState(BaseModel):
 def ensure_state_type(state: Any, expected_type: Type[StateT]) -> StateT:
     """Ensure state matches expected type with proper validation.
 
-    Args:
```
lucasgomide, line 70: Is there any reason to drop this docstring?

tests/test_flow_initial_state_fix.py
```diff
@@ -0,0 +1,99 @@
+"""Test Flow initial_state BaseModel dict coercion fix for issue #3147"""
```
lucasgomide: Can you move the tests to `flow_test.py` I really dont think we should create another file for those tests

Also, would you mind removing unnecessary comments?

src/crewai/flow/flow.py
```diff
@@ -646,30 +636,25 @@ def _initialize_state(self, inputs: Dict[str, Any]) -> None:
             # For BaseModel states, preserve existing fields unless overridden
             try:
                 model = cast(BaseModel, self._state)
-                # Get current state as dict
-                if hasattr(model, "model_dump"):
-                    current_state = model.model_dump()
-                elif hasattr(model, "dict"):
-                    current_state = model.dict()
+                if hasattr(model, "model_copy"):
```
lucasgomide, line 658: Why did you change from `model_dump`?

diego-coder:
Sounds good @lucasgomide, changed it:
    Restored the docstring for the ensure_state_type function.
    Moved the new tests into flow_test.py and removed the separate test file.
    Reverted the logic in the _initialize_state method back to the model_dump approach.
The PR should be ready for another look.

lucasgomide:
@diego-coder have you pushed your changes? Cuz I didn't see changes related to my comments.

diego-coder:
OK, let me take a look, I am away from my PC but can check when I get back
to it this afternoon.

diego-coder:
@lucasgomide My apologies. I've now synced the branch with main, resolved all the conflicts, and addressed your feedback. This should be ready for another review. Cheers

PR #3225, opened 2025-07-26
Dropping User Memory

Vidit-Ostwal:
> Refactor this class and drop the user_id support until the version 0.156.0 or on 2025-08-04. Since this is the deadline to remove UserMemory class

@lucasgomide


Vidit-Ostwal:
I need to drop some code in mem0_storage as well.
That can be done after this gets merged #3217

Vidit-Ostwal:
This is ready for review @lucasgomide

Vidit-Ostwal:
Resolves #3266

lucasgomide:
@Vidit-Ostwal I just found a missing TODO related to memory_config. I think you sohuld address that right now

```python
        # TODO: Memory config will be removed in the future the config will be passed as a parameter
        self.config = config or getattr(crew, "memory_config", {}).get("config", {}) or {}
```

Vidit-Ostwal:
> @Vidit-Ostwal I just found a missing TODO related to memory_config. I think you sohuld address that right now
> 
> ```python
>         # TODO: Memory config will be removed in the future the config will be passed as a parameter
>         self.config = config or getattr(crew, "memory_config", {}).get("config", {}) or {}
> ```

@lucasgomide Just to get some clarity on the TODO.
Moving forward `self.config = config`

lucasgomide:
> Moving forward `self.config = config`

From what I recall, the config is set when ExternalMemory is used, and memory_config is only needed to maintain support for UserMemory. Can you double-check that?

Vidit-Ostwal:
> > Moving forward `self.config = config`
> 
> From what I recall, the config is set when ExternalMemory is used, and memory_config is only needed to maintain support for UserMemory. Can you double-check that?

I doubt that. 
I think `memory_config` is being used by entity and short term memory to implement Mem0.

https://github.com/crewAIInc/crewAI/blob/477bc1f09e39ae9274dd17c607a6c8500934cad7/src/crewai/memory/entity/entity_memory.py#L30-L42

https://github.com/crewAIInc/crewAI/blob/477bc1f09e39ae9274dd17c607a6c8500934cad7/src/crewai/memory/short_term/short_term_memory.py#L31-L44

Vidit-Ostwal:
Here is what I am thinking to do
1. Remove the memory_config parameter at crew level 
2. Any implementation which require Mem0, needs to be done explicitly by initialising the respective memory classes, with embedder_config

```python
class ShortTermMemory(Memory):
    """
    ShortTermMemory class for managing transient data related to immediate tasks
    and interactions.
    Inherits from the Memory class and utilizes an instance of a class that
    adheres to the Storage for data storage, specifically working with
    MemoryItem instances.
    """

    _memory_provider: Optional[str] = PrivateAttr()

    def __init__(self, crew=None, embedder_config=None, storage=None, path=None):
```
```python
class EntityMemory(Memory):
    """
    EntityMemory class for managing structured information about entities
    and their relationships using SQLite storage.
    Inherits from the Memory class.
    """

    _memory_provider: Optional[str] = PrivateAttr()

    def __init__(self, crew=None, embedder_config=None, storage=None, path=None):
```


Let me know what you think?

lucasgomide:
> Let me know what you think?

YUp

docs/en/concepts/memory.mdx
```diff
@@ -9,8 +9,7 @@ icon: database
 The CrewAI framework provides a sophisticated memory system designed to significantly enhance AI agent capabilities. CrewAI offers **three distinct memory approaches** that serve different use cases:
 
 1. **Basic Memory System** - Built-in short-term, long-term, and entity memory
-2. **User Memory** - User-specific memory with Mem0 integration (legacy approach)
-3. **External Memory** - Standalone external memory providers (new approach)
+2. **External Memory** - Standalone external memory providers (new approach)
```
lucasgomide: We can drop the words (new approach) and any other similar reference from External Memory. Wdyt?

docs/en/concepts/memory.mdx
```diff
@@ -776,30 +712,52 @@ crew = Crew(
                     "config": {"api_key": "your-api-key", "model": "text-embedding-3-small"}
```
lucasgomide, line 712: I'm wondering if this setup is correct. In some places we have:

```
embedder_config -> config -> config -> api_key
```

But in other places we have
```
embedder_config -> config -> api_key
```

Which one is right?

docs/pt-BR/concepts/memory.mdx
```diff
@@ -684,67 +684,19 @@ print(f"OpenAI: {openai_time:.2f}s")
 print(f"Ollama: {ollama_time:.2f}s")
 ```
 
-## 2. Memória de Usuário com Mem0 (Legado)
+## 2. Memória Externa (Nova Abordagem)
```
lucasgomide: Don't forget to update this one (Nova Abordagem) means "New approach". It would be great to remove these references since we now have only a single approach for external storage memory

src/crewai/memory/contextual/contextual_memory.py
```diff
@@ -145,6 +113,6 @@ def _fetch_external_context(self, query: str) -> str:
             return ""
 
         formatted_memories = "\n".join(
-            f"- {result['memory']}" for result in external_memories
```
lucasgomide, line 148: is there any reason to update this line?

lucasgomide:
I found a bug 🐛 

take a look at this code

```python
external_memory = ExternalMemory(
    embedder_config={
        "provider": "mem0",
        "config": {
            "user_id": "user-happy",
        }
    }
)
chat_crew = Crew(
    agents=[chat_agent],
    tasks=[chat_task],
    process=Process.sequential,
    external_memory=external_memory,
    memory=True,
    embedder={
        "provider": "mem0"
    },
    verbose=True
)
```

This code is initializing external_memory twice. The first one is explicit, with the proper config key. But the second time (through the embedder parameter), an error is raised because `self.config` isn't present.

Currenlty it doesn't fail because we have this fallback logic:
```python
self.config = config or getattr(crew, "memory_config", {}).get("config", {}) or {}
```

IMO, we shouldn't try to initialize external_memory again if it was already provided. I'd prefer to tackle this behavior improvement later, but I do believe we should still apply a fallback for `self.config` to prevent errors in the meantime

lucasgomide:
Everything else looks good!
It would be great to have a proper note for the release changelog.. something like highlighting the move from User Memory to External Memory, like

> We've transitioned from the old User Memory system to a unified External Memory interface. This simplifies configuration and improves flexibility for future storage providers. Be sure to update your code to use the new external_memory setup. 
Here is an example of:
@Vidit-Ostwal can you drop an example here?

docs/en/concepts/memory.mdx
Vidit-Ostwal: This one is the right one 
`embedder_config -> config -> api_key`

docs/en/concepts/memory.mdx
Vidit-Ostwal: Nice Catch, Resolved.

docs/en/concepts/memory.mdx
Vidit-Ostwal: Can you share the part of the docs, where you find this way of configuration
`
embedder_config -> config -> config -> api_key`

src/crewai/memory/contextual/contextual_memory.py
Vidit-Ostwal: I have added this part in the Mem0 Storage, search 

```python
        if isinstance(self.memory, Memory):
            del params["metadata"], params["version"], params['output_format']
            if params.get("run_id"):
                del params["run_id"]

        results = self.memory.search(**params)

        # This makes it compatible for Contextual Memory to retrieve
        for result in results["results"]:
            result["context"] = result["memory"]
        
        return [r for r in results["results"]]
```

Vidit-Ostwal: This makes retireved dictionary to be same for all the current storage.

Vidit-Ostwal:
> Contributor

As per what I am understanding, 

```python
external_memory = ExternalMemory(
    embedder_config={
        "provider": "mem0",
        "config": {
            "user_id": "user-happy",
        }
    }
)
```

This just initializes the structure of the External Memory, but doesn't make the storage, which I think is more important
that happens when we try to make the crew class object.


I am still of the opinion that, 
1. We should make the storage right way the first time we are initializing, don't wait for the crew class object.
2. the embedder config, of crew and external memory are independent, I am still not able to figure out what exactly caused the issue here. Can you share the traceback of error.

Vidit-Ostwal:
> Everything else looks good! It would be great to have a proper note for the release changelog.. something like highlighting the move from User Memory to External Memory, like
> 
> > We've transitioned from the old User Memory system to a unified External Memory interface. This simplifies configuration and improves flexibility for future storage providers. Be sure to update your code to use the new external_memory setup.
> > Here is an example of:
> > @Vidit-Ostwal can you drop an example here?

Sounds good to me.

docs/en/concepts/memory.mdx
lucasgomide: docs/en/concepts/memory.mdx

lucasgomide:
> * We should make the storage right way the first time we are initializing, don't wait for the crew class object.

But as I mentioned let's tackle that in the future PRs, not for this one

> I am still not able to figure out what exactly caused the issue here. Can you share the traceback of error.

The issue is happening because `Mem0Storage().config` no longer has a fallback. Here's the relevant diff:

Screenshot 2025-08-06 at 9 37 47 AM

I previously shared this snippet, where the embedder passed at the crew level does not include a config key:

```python
external_memory = ExternalMemory(
    embedder_config={
        "provider": "mem0",
        "config": {
            "user_id": "user-happy",
        }
    }
)
chat_crew = Crew(
    agents=[chat_agent],
    tasks=[chat_task],
    process=Process.sequential,
    external_memory=external_memory,
    memory=True,
    embedder={
        "provider": "mem0"
    },
    verbose=True
)
```

You can make `self.config = config or {}` only

docs/pt-BR/concepts/memory.mdx
Vidit-Ostwal: Resovled.

docs/en/concepts/memory.mdx
Vidit-Ostwal: I think you might have come across this: https://docs.crewai.com/en/concepts/memory#testing-different-embedding-providers
this is fine imo.

Vidit-Ostwal:
> But as I mentioned let's tackle that in the future PRs, not for this one

Aligned.


> I previously shared this snippet, where the embedder passed at the crew level does not include a config key:

Exactly, I believe the embedder config at crew and embedder config at External memory are independent
External Memory doesn't get affected with any configuration done at crew level.

Let me know if I am missing something.

Here is what I did
```python
import os
from crewai.memory.external.external_memory import ExternalMemory

# Create external memory instance
external_memory = ExternalMemory(
    embedder_config={
        "provider": "mem0",
        "config": {
            "user_id": "user-happy",
        }
    }
)
# external_memory = external_memory.set_crew("Crew")


from crewai import Crew, Agent, Task, Process, LLM
from dotenv import load_dotenv
import os
from pathlib import Path
from crewai.memory.external.external_memory import ExternalMemory
from mem0 import MemoryClient
load_dotenv()

llmgpt = LLM(
    model="openrouter/openai/gpt-4o-mini",
    api_key=OPENROUTER_API_KEY
)


chat_agent = Agent(
    role="Friendly chatbot assistant",
    goal="Engage in useful and interesting conversations with users while remembering context.",
    backstory="You are a kind and knowledgeable chatbot assistant. You excel at understanding user needs, providing helpful responses, and maintaining engaging conversations. You remember previous interactions to provide a personalized experience.",
    verbose=False,
    llm=llmgpt
)
chat_task = Task(
    description="Respond to user conversation. User message: {user_msg}",
    expected_output="Contextually appropriate, helpful, and friendly response.",
    agent=chat_agent
)

chat_crew = Crew(
    agents=[chat_agent],
    tasks=[chat_task],
    process=Process.sequential,
    external_memory=external_memory,
    embedder={
        "provider": "mem0"
    },
    verbose=False,
    llm=llmgpt
)

```

Able to initialize crew object without any issues.
Let me know if I am missing something.

lucasgomide:
> Able to initialize crew object without any issues.

I didn't see any related-commit addressing this issue 🤔 I ran some tests and my initial issue persist

lucasgomide (approved):

Vidit-Ostwal:
Release Note:

We've transitioned from the old User Memory system to a unified External Memory interface. This simplifies configuration and improves flexibility for future storage providers. External Memory currently supports `Mem0` or any customer storage at this moment. Be sure to update your code to use the new external_memory setup.

Here is an example of initialising External Memory

```python
external_memory = ExternalMemory(
    embedder_config={
        "provider": "mem0",
        "config": {
            "user_id": "john",
            "api_key": "custom-api-key"           
        },
    }
)

crew = Crew(
    agents=[...],
    tasks=[...],
    external_memory=external_memory, 
    process=Process.sequential,
    verbose=True
)
```

External Memory supports both Mem0 OSS and Mem0 Client.
You can find more info here:
https://docs.crewai.com/en/concepts/memory

lucasgomide (approved):
Good job @Vidit-Ostwal 

I've just run several tests, and everything looks good

Vidit-Ostwal:
> Good job @Vidit-Ostwal
> 
> I've just run several tests, and everything looks good

OtmanFakri:
Configuration error: 3 validation errors for MatchValue
fastapi_backend  | value.bool
fastapi_backend  |   Input should be a valid boolean [type=bool_type, input_value=[{'user_id': 'fucking_user'}], input_type=list]
fastapi_backend  |     For further information visit https://errors.pydantic.dev/2.11/v/bool_type
fastapi_backend  | value.int
fastapi_backend  |   Input should be a valid integer [type=int_type, input_value=[{'user_id': 'fucking_user'}], input_type=list]
fastapi_backend  |     For further information visit https://errors.pydantic.dev/2.11/v/int_type
fastapi_backend  | value.str
fastapi_backend  |   Input should be a valid string [type=string_type, input_value=[{'user_id': 'fucking_user'}], input_type=list]
fastapi_backend  |     For further information visit https://errors.pydantic.dev/2.11/v/string_type
external_memory = ExternalMemory(
            embedder_config={
                "provider": "mem0",
                "config": {
                    "user_id": "fucking_user",
                    "local_mem0_config": {
                        "vector_store": {
                            "provider": "qdrant",
                            "config": {
                                "url": settings.QDRANT_URL,
                                "collection_name": f"case_{self.case_id}"
                            }
                        },
                        "embedder": {
                            "provider": "huggingface",
                            "config": {
                                "model": "Omartificial-Intelligence-Space/Arabert-all-nli-triplet-Matryoshka"
                            }
                        },
                        "llm": {
                            "provider": "gemini",
                            "config": {
                                "model": "gemini-2.0-flash",
                                "api_key": settings.GEMINI_API_KEY
                            }
                        }
                    }
                },
            },
        )

Vidit-Ostwal:
Hi @OtmanFakri, mind raising another issue?

PR #3233, opened 2025-07-29
Tasks retry after failures (API Rate Limit / Model Throttling)

T4n17:
This feature would help avoid the crew getting stuck after a failing task, caused by an API rate limit error or a model throttling error

### DIFF
--- a/src/crewai/task.py
+++ b/src/crewai/task.py
@@ -4,6 +4,7 @@ import json
 import logging
 import threading
 import uuid
+import time
 from concurrent.futures import Future
 from copy import copy
 from hashlib import md5
@@ -157,6 +158,15 @@ class Task(BaseModel):
         default=3, description="Maximum number of retries when guardrail fails"
     )
     retry_count: int = Field(default=0, description="Current number of retries")
+    max_retries_after_failure: int = Field(
+        default=5, description="Maximum number of retries after a Task failure"
+    )
+    number_of_retries_remaining_after_failure: int = Field(
+        default=max_retries_after_failure.default, description="Number of retries remaining after a Task failure"
+    )
+    max_delay_after_failure: int = Field(
+        default=5, description="Maximum delay after a Task failure in seconds"
+    )
     start_time: Optional[datetime.datetime] = Field(
         default=None, description="Start time of the task execution"
     )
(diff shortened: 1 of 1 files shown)


src/crewai/task.py
```diff
@@ -501,7 +511,14 @@ def _execute_core(
         except Exception as e:
             self.end_time = datetime.datetime.now()
             crewai_event_bus.emit(self, TaskFailedEvent(error=str(e), task=self))
-            raise e  # Re-raise the exception after emitting the event
+            if self.number_of_retries_remaining_after_failure > 0:
+                # Retrying Task execution after failure
+                time.sleep(self.max_delay_after_failure)
```
lucasgomide: time.sleep is a blocking call in multithreaded systems and may degrade performance in CrewAI's parallel executions. Consider replacing it with a non-blocking retry or scheduling mechanism.

src/crewai/task.py
```diff
@@ -157,6 +158,15 @@ class Task(BaseModel):
         default=3, description="Maximum number of retries when guardrail fails"
     )
     retry_count: int = Field(default=0, description="Current number of retries")
+    max_retries_after_failure: int = Field(
+        default=5, description="Maximum number of retries after a Task failure"
+    )
+    number_of_retries_remaining_after_failure: int = Field(
+        default=max_retries_after_failure.default, description="Number of retries remaining after a Task failure"
+    )
```
lucasgomide: This might generate Runtime execution error, isn't?

src/crewai/task.py
```diff
@@ -501,7 +511,14 @@ def _execute_core(
...
             self.end_time = datetime.datetime.now()
             crewai_event_bus.emit(self, TaskFailedEvent(error=str(e), task=self))
-            raise e  # Re-raise the exception after emitting the event
+            if self.number_of_retries_remaining_after_failure > 0:
+                # Retrying Task execution after failure
+                time.sleep(self.max_delay_after_failure)
+                self.number_of_retries_remaining_after_failure -= 1
+                return self._execute_core(agent, context, tools)
+            else:
+                crewai_event_bus.emit(self, TaskCompletedEvent(output=TaskOutput(description="Task failed", agent=self.agent.role), task=self))
```
lucasgomide: We should not emit `TaskCompleted`. Consider raising an error instead

src/crewai/task.py
T4n17: Fixed this. I used a threading timer instead of blocking sleep function

src/crewai/task.py
T4n17: Fixed. I set a fixed default value (5)

src/crewai/task.py
T4n17: Raising the error will just cause the crew to hang, since it does waits for all the task to complete. I changed this by returning a TaskOutput describing the error instead of emitting the TaskCompletedEvent

PR #3246, opened 2025-07-31
feat: centralize embedding types and create base client

greysonlalonde:
## Summary
- Move embedding types to dedicated embeddings/types.py
- Add BaseProviderOptions with ABC pattern for provider configuration
- Update embedding factory to handle both dict and EmbeddingOptions objects
- Clean up types.py to remove package-specific types

## Changes
- Created  with centralized embedding types
- Created  with improved embedding function factory
- Added  with provider pattern

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ authors = [
 dependencies = [
     # Core Dependencies
     "pydantic>=2.4.2",
-    "openai<1.100.0",  # TODO: Temporarily pin due to https://github.com/openai/openai-python/issues/2564. Can revert once the issue is fixed.
+    "openai>=1.13.3",
     "litellm==1.74.9",
     "instructor>=1.3.3",
     # Text Processing
--- a/src/crewai/utilities/events/event_listener.py
+++ b/src/crewai/utilities/events/event_listener.py
@@ -161,8 +161,10 @@ class EventListener(BaseEventListener):
         def on_task_started(source, event: TaskStartedEvent):
             span = self._telemetry.task_started(crew=source.agent.crew, task=source)
             self.execution_spans[source] = span
+            # Pass both task ID and task name (if set)
+            task_name = source.name if hasattr(source, 'name') and source.name else None
             self.formatter.create_task_branch(
-                self.formatter.current_crew_tree, source.id
+                self.formatter.current_crew_tree, source.id, task_name
             )
 
         @crewai_event_bus.on(TaskCompletedEvent)
@@ -173,11 +175,14 @@ class EventListener(BaseEventListener):
                 self._telemetry.task_ended(span, source, source.agent.crew)
             self.execution_spans[source] = None
 
+            # Pass task name if it exists
+            task_name = source.name if hasattr(source, 'name') and source.name else None
             self.formatter.update_task_status(
                 self.formatter.current_crew_tree,
                 source.id,
                 source.agent.role,
                 "completed",
+                task_name
             )
 
         @crewai_event_bus.on(TaskFailedEvent)
@@ -188,11 +193,14 @@ class EventListener(BaseEventListener):
                     self._telemetry.task_ended(span, source, source.agent.crew)
                 self.execution_spans[source] = None
 
+            # Pass task name if it exists
+            task_name = source.name if hasattr(source, 'name') and source.name else None
             self.formatter.update_task_status(
                 self.formatter.current_crew_tree,
                 source.id,
                 source.agent.role,
                 "failed",
+                task_name
             )
 
         # ----------- AGENT EVENTS -----------
--- a/src/crewai/utilities/events/utils/console_formatter.py
+++ b/src/crewai/utilities/events/utils/console_formatter.py
@@ -220,14 +220,22 @@ class ConsoleFormatter:
         return tree
 
     def create_task_branch(
-        self, crew_tree: Optional[Tree], task_id: str
+        self, crew_tree: Optional[Tree], task_id: str, task_name: Optional[str] = None
     ) -> Optional[Tree]:
         """Create and initialize a task branch."""
         if not self.verbose:
             return None
 
         task_content = Text()
-        task_content.append(f"📋 Task: {task_id}", style="yellow bold")
+        
+        # Display task name if available, otherwise just the ID
+        if task_name:
+            task_content.append("📋 Task: ", style="yellow bold")
+            task_content.append(f"{task_name}", style="yellow bold")
+            task_content.append(f" (ID: {task_id})", style="yellow dim")
+        else:
+            task_content.append(f"📋 Task: {task_id}", style="yellow bold")
+            
         task_content.append("\nStatus: ", style="white")
         task_content.append("Executing Task...", style="yellow dim")
 
@@ -251,6 +259,7 @@ class ConsoleFormatter:
         task_id: str,
         agent_role: str,
         status: str = "completed",
+        task_name: Optional[str] = None,
     ) -> None:
         """Update task status in the tree."""
         if not self.verbose or crew_tree is None:
@@ -270,8 +279,13 @@ class ConsoleFormatter:
             if str(task_id) in str(branch.label):
                 # Build label without introducing stray blank lines
                 task_content = Text()
-                # First line: Task ID
-                task_content.append(f"📋 Task: {task_id}", style=f"{style} bold")
+                # First line: Task ID/name
+                if task_name:
+                    task_content.append("📋 Task: ", style=f"{style} bold")
+                    task_content.append(f"{task_name}", style=f"{style} bold")
+                    task_content.append(f" (ID: {task_id})", style=f"{style} dim")
+                else:
+                    task_content.append(f"📋 Task: {task_id}", style=f"{style} bold")
 
                 # Second line: Assigned to
                 task_content.append("\nAssigned to: ", style="white")
@@ -285,8 +299,9 @@ class ConsoleFormatter:
                 break
 
         # Show status panel
+        display_name = task_name if task_name else str(task_id)
         content = self.create_status_content(
-            f"Task {status.title()}", str(task_id), style, Agent=agent_role
+            f"Task {status.title()}", display_name, style, Agent=agent_role
         )
         self.print_panel(content, panel_title, style)
 
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.10, <3.14"
 resolution-markers = [
     "python_full_version >= '3.13' and platform_python_implementation == 'PyPy' and sys_platform == 'darwin'",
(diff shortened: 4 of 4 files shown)


src/crewai/rag/embeddings/factory.py
```diff
@@ -0,0 +1,148 @@
+"""Minimal embedding function factory for CrewAI."""
```
lorenzejay, line 1: from a folder structure perspective, should we do something like:

embeddings:
- chroma
--factory.py
--...
- qdrant
--factory.py
...


wdyt?

src/crewai/rag/embeddings/factory.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Examples:
+        # Use default OpenAI with retry logic
+        >>> embedder = get_embedding_function()
+
+        # Use Cohere with dict
+        >>> embedder = get_embedding_function({
+        ...     "provider": "cohere",
+        ...     "api_key": "your-key",
+        ...     "model_name": "embed-english-v3.0"
```
lorenzejay, line 87: love this already

src/crewai/rag/embeddings/factory.py
greysonlalonde: I dig it, also thinking about an embeddings package/module within each supported provider. Let's go over it in the chromadb pr once that's ready 👍

src/crewai/rag/embeddings/factory.py
```diff
@@ -0,0 +1,148 @@
...
+
+    # Handle EmbeddingOptions object
+    if isinstance(config, EmbeddingOptions):
+        config_dict = config.model_dump(exclude_none=True)
+    else:
+        config_dict = config.copy()
+
+    provider = config_dict.pop("provider", "openai")
+
+    embedding_functions = {
```
lucasgomide, line 125: what about using `lazy_load` strategy

```
embedding_functions = {
"openai": "chromadb.utils.embedding_functions.openai_embedding_function"
}
```

src/crewai/rag/embeddings/factory.py
```diff
@@ -0,0 +1,148 @@
...
+)
+from chromadb.utils.embedding_functions.huggingface_embedding_function import (
+    HuggingFaceEmbeddingFunction,
+)
+from chromadb.utils.embedding_functions.instructor_embedding_function import (
+    InstructorEmbeddingFunction,
+)
+from chromadb.utils.embedding_functions.jina_embedding_function import (
+    JinaEmbeddingFunction,
+)
```
lucasgomide, line 25: What about a folder structure to support multiples vector db?

src/crewai/rag/core/base_client.py
```diff
@@ -0,0 +1,433 @@
...
+
+
+class BaseCollectionParams(TypedDict):
+    """Base parameters for collection operations.
+
+    Attributes:
+        collection_name: The name of the collection/index to operate on.
+    """
+
+    collection_name: Required[
```
lucasgomide, line 22: what about considering metadata attribute for `get_or_create` collections?

src/crewai/rag/core/base_client.py
```diff
@@ -0,0 +1,433 @@
...
+            >>> client.add(collection_name="my_docs", documents=records)
+            >>>
+            >>> records_with_id: list[BaseRecord] = [
+            ...     {
+            ...         "doc_id": "nlp_001",
+            ...         "content": "Advanced NLP techniques",
+            ...         "metadata": {"source": "file5", "topic": "NLP"}
+            ...     }
+            ... ]
+            >>> client.add(collection_name="my_docs", documents=records_with_id)
```
lucasgomide: I'm not a big fan of this interface: `client.add`. I'd prefer using `collection.add` instead.

Here’s why:

When you pass a `collection_name` as a string rather than using a Collection instance, the add method is responsible for resolving the collection from the client (it means they might be responsible for even creation). But in practice, the implementation always maintains a direct reference to a collection that already contains documents.

This approach could lead to misconfigurations. For example, a user might intend to create a collection with specific metadata, but calling client.add with just a name wouldn’t reflect that intent, and the metadata could be lost or mismatched.

src/crewai/rag/core/base_client.py
```diff
@@ -0,0 +1,433 @@
...
+        ...
+
+    @abstractmethod
+    async def aget_or_create_collection(
+        self, **kwargs: Unpack[BaseCollectionParams]
+    ) -> Any:
+        """Get an existing collection or create it if it doesn't exist asynchronously.
+
+        Keyword Args:
+            collection_name: The name of the collection to get or create.
```
lucasgomide, line 154: would be great add `metadata`  parameter that is very common for creation

src/crewai/rag/core/base_client.py
```diff
@@ -0,0 +1,433 @@
...
+    All fields except collection_name and query are optional.
+
+    Attributes:
+        query: The text query to search for (required).
+        limit: Maximum number of results to return.
+        metadata_filter: Filter results by metadata fields.
+        score_threshold: Minimum similarity score for results (0-1).
+    """
+
+    query: Required[str]
```
lucasgomide, line 56: Should query be of `list[float]`? Since the query is expected to be performed using an embedded vector

I might be a bit confused about the intended purpose of these protocols. At first glance, they seemed to be designed as interfaces for VectorDB clients like Qdrant, PSVector, or ChromaDB. But after reviewing them more closely, it looks like they're actually meant to define interfaces for building RAG components. If that's the case, the design considerations might differ.

src/crewai/rag/embeddings/types.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    model_name: str | None = Field(
+        default=None, description="Model name for the embedding provider"
+    )
+    api_key: SecretStr | None = Field(
+        default=None, description="API key for the embedding provider"
+    )
+
+
+class EmbeddingConfig(BaseModel):
```
lucasgomide, line 52: What is the intended use of EmbeddingConfig? Perhaps a real usage example would help

src/crewai/rag/embeddings/factory.py
greysonlalonde: We can do that! I say we do it once we've moved away from chromadb for embeddings, though. `chromadb.utils.embedding_functions` classes are pretty lightweight and already implement lazy loading for external imports from what I've seen, like [here](https://github.com/chroma-core/chroma/blob/6c378b3eb2574408b30cac0334603c1663c16af8/chromadb/utils/embedding_functions/cohere_embedding_function.py#L27)

src/crewai/rag/embeddings/factory.py
greysonlalonde: Future prs :) chroma will have it's own package next

src/crewai/rag/core/base_client.py
greysonlalonde: Pushing back here, reasons being:

- [Qdrant](https://qdrant.tech/documentation/concepts/collections/) uses client level operations, and [Pinecone](https://docs.pinecone.io/guides/index-data/upsert-data#upsert-text) uses client level ops or `Index` with namespace separation instead of collection, so we can't really abstract out to a `Collection` interface and use it in a unified fashion. 
- I think the naming here is confusing (I started here with a collection interface initially), `BaseClient.add` is specific to inserting documents into a collection/index, but should not be responsible for collection/index creation. A create + add flow would look like:

    ```python3
    client.create_collection(
        collection_name="docs",
        **kwargs #implementation may define these params by extending the unpacked `TypedDict` as metadata, whatever
    )
    # then
    client.add(collection_name="docs", documents=[...])
     ```
     We could probably rename this method to `add_documents`, or something similarly explicit.
- Client level operations make it easier to manage thread safety and ensure consistent transaction boundaries. Right now we are trying to manage client lifecycle from within separate classes and from within helper functions

src/crewai/rag/embeddings/types.py
greysonlalonde: This is a wip for a future pr, I can show you on the chromadb impl

src/crewai/rag/embeddings/factory.py
lorenzejay: sounds good

src/crewai/rag/embeddings/factory.py
lorenzejay: haha we had the same comment @lucasgomide

src/crewai/rag/core/base_client.py
greysonlalonde: It's more of a replacement for how we are using vector store clients currently. I think it's worth noting that this is an internal concern to the framework, there is no plan to surface clients and the only user interaction will be a configuration parameter at the top level. On the query key, we could specify `query_embeddings`, but that sets an expectation for data types external to the interface, making it harder to plugin across frameworks

src/crewai/rag/core/base_client.py
greysonlalonde: Implementations are expected to extend the params so could do this

src/crewai/rag/core/base_client.py
greysonlalonde: Same here, implementations will do this for sure

src/crewai/rag/core/base_client.py
lucasgomide: 👍🏼 for renaming to `add_documents`

Regarding the other points, sounds good to me.. makes totally sense.

src/crewai/rag/core/base_client.py
lucasgomide: > but that sets an expectation for data types external to the interface, making it harder to plugin across frameworks

What's on my mind is that this is essentially a VectorDB interface for our framework (which is great!). However, queries in vector databases are performed using embedding vectors... and your specification explicitly defines query as a string. In practice, all client implementations will end up overriding the query type anyway. 

Does it makes sense?

greysonlalonde: Circling back to this finally - I'm somewhat on the fence here. in [RAGStorage.search](https://github.com/crewAIInc/crewAI/blob/2a0cefc98b8a7c1d029a124f21d6b04fa956a831/src/crewai/memory/storage/rag_storage.py#L128), we pass directly as text, but could also use embedding vectors instead. I think other implementations limit to one or the other (not sure enough to speak on that confidently, I'll check). Is there anywhere where we are passing direct embedding vectors rather than text?

lucasgomide: Got it, good points! Currently I'm actually passing embedding vectors to query documents on RAG Tools - not a big deal, I can seamlessly update it to use this approach.

Thinking about it, I actually like this method better it keeps things simpler and moves the complexity of embedding text to the Vector client.

THe last question about that, can the user change the embedder provider/model?

greysonlalonde: Absolutely - the vision for this is to be very easily configurable all around 👍

lucasgomide (approved):

PR #3271, opened 2025-08-04
Support Device authorization with Okta

lucasgomide:


src/crewai/cli/authentication/providers/auth0.py
```diff
@@ -0,0 +1,28 @@
+from crewai.cli.authentication.constants import (
+    AUTH0_AUDIENCE,
+    AUTH0_CLIENT_ID,
+    AUTH0_DOMAIN,
+)
```
heitorado: I like how each provider now has its own file and class, but these environment variables should be gone now, no?

Even for auth0 and workos, we should be fetching these values from the settings (e.g. imagine that a factory client wants to use auth0 or workos - then we would have three providers ready instead of just okta.)

src/crewai/cli/authentication/providers/workos.py
```diff
@@ -0,0 +1,28 @@
+from crewai.cli.authentication.constants import (
+    WORKOS_DOMAIN,
+    WORKOS_CLI_CONNECT_APP_ID,
+    WORKOS_ENVIRONMENT_ID,
+)
```
heitorado: Same concern as my other comment on auth0.py - we should be fetching this from the settings as well, no?

src/crewai/cli/config.py
```diff
@@ -59,6 +64,11 @@ class Settings(BaseModel):
     )
     config_path: Path = Field(default=DEFAULT_CONFIG_PATH, frozen=True, exclude=True)
 
+    oauth2_provider: str = Field(description="Oauth provider", default=DEFAULT_OAUTH2_PROVIDER)
+    oauth2_audience: Optional[str] = Field(description="Oauth audience", default=None)
+    oauth2_client_id: Optional[str] = Field(description="Oauth client ID", default=None)
+    oauth2_domain: Optional[str] = Field(description="Oauth domain", default=None)
+
```
heitorado: I think we should set the defaults here as well: it should default to the WorkOS domain, client id and audience, not just the provider.

src/crewai/cli/constants.py
```diff
@@ -1,4 +1,5 @@
 DEFAULT_CREWAI_ENTERPRISE_URL = "https://app.crewai.com"
+DEFAULT_OAUTH2_PROVIDER = "workos"
```
heitorado: We could add the other constants related to workos here as well and change the naming so it's more meaningful that they are the default just for the crewai enterprise saas:

```suggestion
CREWAI_ENTERPRISE_DEFAULT_OAUTH2_PROVIDER = "workos"
CREWAI_ENTERPRISE_DEFAULT_OAUTH2_AUDIENCE = "client_01JNJQWBJ4SPFN3SWJM5T7BDG8"
CREWAI_ENTERPRISE_DEFAULT_OAUTH2_CLIENT_ID = "client_01JYT06R59SP0NXYGD994NFXXX"
CREWAI_ENTERPRISE_DEFAULT_OAUTH2_DOMAIN = "login.crewai.com"
```

src/crewai/cli/config.py
```diff
@@ -105,4 +115,4 @@ def _reset_user_settings(self) -> None:
     def _reset_cli_settings(self) -> None:
         """Reset all CLI settings to default values"""
         for key in CLI_SETTINGS_KEYS:
-            setattr(self, key, DEFAULT_CLI_SETTINGS[key])
+            setattr(self, key, DEFAULT_CLI_SETTINGS.get(key))
```
heitorado, line 141: I understand that this `.get()` was added because there are more keys inside `CLI_SETTINGS_KEYS` now than the defaults set on `DEFAULT_CLI_SETTINGS`, however, I think that should not be the case.

Every CLI setting should have a default - because the default state of the CLI is to connect to the Crewai Enterprise SaaS.

If it doesn't, when we set this to other values (to test factory environments for example) it will be a pain to change everything back to point to Crewai Enterprise SaaS.

src/crewai/cli/authentication/providers/auth0.py
lucasgomide: I had the same thought 😆 — that's why I already implemented it [this way](https://github.com/crewAIInc/crewAI/pull/3271/files#diff-e5762b7c738519cc1fb9035985dcbcf4157facf2a003fd4686655b81abfc228aR22):

```python
def get_audience(self) -> str:
        return self.settings.audience or AUTH0_AUDIENCE
```

So we will we default value as fallback

src/crewai/cli/authentication/providers/workos.py
lucasgomide: same answer (: 

I have done it already

```python
    def get_audience(self) -> str:
        return self.settings.audience or WORKOS_ENVIRONMENT_ID

    def get_client_id(self) -> str:
        return self.settings.client_id or WORKOS_CLI_CONNECT_APP_ID

    def _get_domain(self) -> str:
        return self.settings.domain or WORKOS_DOMAIN

```

src/crewai/cli/config.py
lucasgomide: agreed! I'm heading to update that

src/crewai/cli/constants.py
lucasgomide: sure

src/crewai/cli/config.py
lucasgomide: > Every CLI setting should have a default 

I'm not sure we should always assume this to be true. It won't be an issue anymore as I’ll add default WorkOS values.

A couple of thoughts on this:
- If the value is meant to always be present, we could simplify things by using a single variable.. this would help avoid confusion.
- It might also be worth renaming the variable. Based on your point, it seems more related to organization setup, so a clearer name could improve understanding.

src/crewai/cli/authentication/providers/workos.py
heitorado: cool! but with the defaults, do we still need to import the constants here? shouldn't they just be used inside `config.py`?

src/crewai/cli/authentication/providers/auth0.py
heitorado: cool! do you think we should remove those auth0 constants though? either now or in a future PR? WorkOS migration is done and all. 🤔

src/crewai/cli/authentication/constants.py
```diff
@@ -2,7 +2,3 @@
 AUTH0_DOMAIN = "crewai.us.auth0.com"
 AUTH0_CLIENT_ID = "DEVC5Fw6NlRoSzmDCcOhVq85EfLBjKa8"
 AUTH0_AUDIENCE = "https://crewai.us.auth0.com/api/v2/"
```
heitorado, line 6: Since you removed WorkOS constants, you can remove auth0 constants as well, especially since we won't be using it anymore.

src/crewai/cli/authentication/providers/auth0.py
lucasgomide: For sure we should remove that after WorkOS migration. We should keep only default Enterprise ones here IMO

src/crewai/cli/authentication/providers/workos.py
lucasgomide: Got it. Right now, we're importing it in both config.py and workos.py.
After your suggestion, i'mm thinking about removing it from this file since we've already set default values. What do you think? can you see any issue?

src/crewai/cli/authentication/constants.py
lucasgomide: We can remove it after the WorkOS migration is completed. What do you think? I can add a "TODO" for that

heitorado: good! please do

src/crewai/cli/config.py
```diff
@@ -59,6 +73,26 @@ class Settings(BaseModel):
     )
     config_path: Path = Field(default=DEFAULT_CONFIG_PATH, frozen=True, exclude=True)
 
+    oauth2_provider: str = Field(
+        description="OAuth2 provider used for authentication (e.g., workos, okta, auth0).",
+        default=DEFAULT_CLI_SETTINGS["oauth2_provider"]
+    )
```
heitorado, line 78: Something just occurred to me: not sure if we want to babysit the user _that much_, but would it be possible to add allowed values to this field?

Right now the only possible values are `["workos", "auth0", "okta"]`.

If user tries `crewai config set oauth2_provider blabla` we would then raise an error saying that this is not a valid value for the field.

src/crewai/cli/authentication/providers/workos.py
heitorado: Good! I don't think we'll have any issue - I guess it's better if we only import from config.py, so this file has less dependency from other modules. 

Then each provider is completely oblivious from the constants we have in the files: the only source of truth in the provider's perspective is the current settings.

src/crewai/cli/authentication/providers/auth0.py
heitorado: right, but just like the workos.py file, we don't need to import them here, right? just on config.py

src/crewai/cli/config.py
lucasgomide: sure, will do

lucasgomide: I'm facing some issues to do that due config set is using `setattr()`. So, I will make it in another PR

src/crewai/cli/authentication/main.py
```diff
@@ -22,55 +15,69 @@
...
         # TODO: WORKOS - Next line and conditional are temporary until migration to WorkOS is complete.
         user_provider = self._determine_user_provider()
         if user_provider == "auth0":
-            device_code_url = self.AUTH0_DEVICE_CODE_URL
-            token_url = self.AUTH0_TOKEN_URL
-            client_id = AUTH0_CLIENT_ID
-            audience = AUTH0_AUDIENCE
-            self.user_provider = "auth0"
+            from crewai.cli.authentication.providers.auth0 import Auth0Provider
+            self.oauth2_provider = Auth0Provider(settings=Oauth2Settings(provider="auth0"))
```
heitorado: Just noticed something here: this works when it comes to override the default provider class, but when the `Auth0Provider` is instantiated, it will eventually call these methods:

```python
    def get_audience(self) -> str:
        return self.settings.audience or AUTH0_AUDIENCE

    def get_client_id(self) -> str:
        return self.settings.client_id or AUTH0_CLIENT_ID

    def _get_domain(self) -> str:
        return self.settings.domain or AUTH0_DOMAIN
```

right?

So, what will happen on the conditional is that it will use the values on the settings, because they won't be empty. But this is a problem, because the values on the settings won't be the auth0 audience, client ID, nor the domain. The values on the settings will be the default values, i.e. the workos values.

Am I missing something?

heitorado: Not a big deal, I just think that login with auth0 won't work anymore after this is merged (not a problem since we already have all users on workos)

lucasgomide: oh good call

For this case which we need to override default provider class we have to set the default AUTH0 parameters as well

lucasgomide: addressed here https://github.com/crewAIInc/crewAI/pull/3271/commits/34c76826b8136a2226db7051721795ca811e2071

heitorado (approved):
Awesome work!

PR #3276, opened 2025-08-05
fix: allow persist Flow state with BaseModel entries

lucasgomide:
When trying to persist Flow state with BaseModel entries we got the following error

```
Failed to persist state for method execute_message: Object of type Message is not JSON serializable
```

### DIFF
--- a/src/crewai/flow/persistence/sqlite.py
+++ b/src/crewai/flow/persistence/sqlite.py
@@ -81,7 +81,7 @@ class SQLiteFlowPersistence(FlowPersistence):
         """
         # Convert state_data to dict, handling both Pydantic and dict cases
         if isinstance(state_data, BaseModel):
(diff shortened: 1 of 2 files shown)


src/crewai/flow/persistence/sqlite.py
```diff
@@ -81,7 +81,7 @@ def save_state(
         """
         # Convert state_data to dict, handling both Pydantic and dict cases
         if isinstance(state_data, BaseModel):
-            state_dict = dict(state_data)  # Use dict() for better type compatibility
+            state_dict = state_data.model_dump()
```
lorenzejay, line 84: love this

danielfsbarreto: So do I! ❤️ 

 This will make definition of states much smoother!

PR #3279, opened 2025-08-06
Lorenze/tracing v1

lorenzejay:


src/crewai/utilities/events/listeners/tracing/interfaces.py
```diff
@@ -0,0 +1,33 @@
...
+    """Trace sender for sending trace batches to the backend"""
+
+    def send_batch(self, batch: TraceBatch) -> bool:
+        """Print trace batch to console"""
+        try:
+            payload = batch.to_dict()
+
+            def datetime_handler(obj):
+                if isinstance(obj, datetime):
+                    return obj.isoformat() + "Z" if obj.tzinfo else obj.isoformat()
```
lucasgomide: Appending "Z" manually can be misleading unless you’re sure the datetime is in UTC.. I'm worried whether current approach is safer

src/crewai/flow/flow.py
```diff
@@ -465,7 +469,9 @@ def __init__(
 
         # Initialize state with initial values
         self._state = self._create_initial_state()
-
+        if os.getenv("CREWAI_TRACING_ENABLED", "false").lower() == "true":
```
lucasgomide, line 472: We currently have those env vars related to telemetry, can we reuse any of them?

```
CREWAI_DISABLE_TELEMETRY
CREWAI_DISABLE_TRACKING
```

Just trying to avoid adding unnecessary env-vars

src/crewai/utilities/events/listeners/tracing/types.py
```diff
@@ -0,0 +1,19 @@
...
+import uuid
+
+
+@dataclass
+class TraceEvent:
+    """Individual trace event payload"""
+
+    event_id: str = field(default_factory=lambda: str(uuid.uuid4()))
+    timestamp: str = field(
+        default_factory=lambda: datetime.now(timezone.utc).isoformat() + "Z"
```
lucasgomide: Avoid appending "Z" manually.. `datetime.now(timezone.utc).isoformat()` already includes timezone info as "+00:00" (UTC timezone)

adding manually "Z" may result in misleading or malformed timestamps like the following example

```
$ datetime.now(timezone.utc).isoformat() + "Z"
> '2025-08-06T16:37:32.779313+00:00Z'
```

If you really wanna to ends with "Z" consider replacing instead
`datetime.now(timezone.utc).isoformat().replace('+00:00', 'Z')`

```
datetime.now(timezone.utc).isoformat().replace('+00:00', 'Z')
'2025-08-06T16:38:52.297460Z'
```

src/crewai/utilities/events/llm_events.py
```diff
@@ -61,6 +64,9 @@ class LLMCallCompletedEvent(LLMEventBase):
     messages: str | list[dict[str, Any]] | None = None
     response: Any
     call_type: LLMCallType
+    response_cost: Optional[float]
```
lucasgomide: Should we send cost in integer values instead float? I always prefer treating "cost" as integer instead of decimal values

tests/tracing/test_tracing.py
```diff
@@ -0,0 +1,49 @@
...
+from crewai import Agent, Task, Crew
+from crewai.utilities.events.listeners.tracing.trace_listener import (
+    TraceCollectionListener,
+)
+from crewai.utilities.events.listeners.tracing.trace_batch_manager import (
+    TraceBatchManager,
+)
+
+
+class TestTraceListenerSetup:
```
lucasgomide, line 16: Should we test the propageted paramtners? To make sure we are providing them corectly?

src/crewai/utilities/events/llm_events.py
lorenzejay: its comes in like `0.32131`

src/crewai/flow/flow.py
lorenzejay: tracing and telemetry for us and users will be completely different

lucasgomide (approved):

src/crewai/utilities/events/listeners/tracing/trace_batch_manager.py
```diff
@@ -0,0 +1,252 @@
...
+            "execution_metadata": self.execution_metadata,
+            "events": [event.to_dict() for event in self.events],
+        }
+
+
+class TraceBatchManager:
+    """Single responsibility: Manage batches and event buffering"""
+
+    def __init__(self):
+        self.plus_api = PlusAPI(api_key=get_auth_token())
```
greysonlalonde, line 44: Do we need any validation here to disable, like if no auth token is available?

PR #3289, opened 2025-08-07
feat: add enterprise configure command

lucasgomide:
https://github.com/user-attachments/assets/5e7bc6b3-447f-4831-a68d-69360898597f

### DIFF
--- a/src/crewai/cli/enterprise/main.py
+++ b/src/crewai/cli/enterprise/main.py
@@ -1,0 +1,84 @@
+import requests
+from typing import Dict, Any
+from rich.console import Console
+from requests.exceptions import RequestException, JSONDecodeError
+
+from crewai.cli.command import BaseCommand
+from crewai.cli.settings.main import SettingsCommand
+from crewai.cli.version import get_crewai_version
+
+console = Console()
+
+
+class EnterpriseCommand(BaseCommand):
+    def __init__(self):
+        super().__init__()
+        self.settings_command = SettingsCommand()
+
+    def configure(self, enterprise_url: str) -> None:
+        try:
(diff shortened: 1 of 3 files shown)


src/crewai/cli/enterprise/main.py
```diff
@@ -0,0 +1,84 @@
...
+from requests.exceptions import RequestException, JSONDecodeError
+
+from crewai.cli.command import BaseCommand
+from crewai.cli.settings.main import SettingsCommand
+from crewai.cli.version import get_crewai_version
+
+console = Console()
+
+
+class EnterpriseCommand(BaseCommand):
```
heitorado: `EnterpriseConfigureCommand` would be better? Since enterprise is the name of the group of commands. 

Really small detail, don't need to change if you are not convinced haha

src/crewai/cli/enterprise/main.py
```diff
@@ -0,0 +1,84 @@
...
+                'oauth2_audience': oauth_config['audience'],
+                'oauth2_client_id': oauth_config['device_authorization_client_id'],
+                'oauth2_domain': oauth_config['domain']
+            }
+
+            console.print("🔄 Updating local OAuth2 configuration...")
+
+            for key, value in config_mapping.items():
+                self.settings_command.set(key, value)
+                console.print(f"  ✓ Set {key}: {value}", style="dim")
```
heitorado, line 81: Really nice that you used the `settings_command` here instead of setting the settings (lol) directly 💯

heitorado (approved):
Awesome work, feels really magical 🪄 

Left just one comment

heitorado (approved):

PR #3290, opened 2025-08-07
feat: include exchanged agent messages into ExternalMemory metadata

lucasgomide:
Including the messages exchanged by the Agent at the ExternalMemory level is useful for helping external memory providers gain more context about the full conversation


Vidit-Ostwal:
Hi @lucasgomide, I was thinking instead of sending all the previous message in the metadata, we should instead send it to Memory Object inside External Memory directly, while saving.

The metadata I feel is not used while making or searching memory.

Something like this, we should make a function which removed the system prompt and also the other prommt we add in the user and assisstant prompt. wdyt?

```
client = MemoryClient(api_key="your_api_key", org_id="your_org_id", project_id="your_project_id")

messages = [
    {"role": "user", "content": "<user-message>"},
    {"role": "assistant", "content": "<assistant-response>"}
]

client.add(messages, user_id="<user-id>", version="v2")
```

lucasgomide:
> The metadata I feel is not used while making or searching memory.

I didn't get it, we're propagating the metadata thought `save` storage method

I'm exposing the whole exchanged messages by Agent only for External memory, so each storage implementation can decide which messages will be picked.

Vidit-Ostwal:
> I'm exposing the whole exchanged messages by Agent only for External memory, so each storage implementation can decide which messages will be picked.

Aligned on this.

I am suggesting additional to this, we should make a little change in the Mem0 Storage, here

https://github.com/crewAIInc/crewAI/blob/a92211f0ba8d60337365fb3741c9251cc5db65fb/src/crewai/memory/storage/mem0_storage.py#L89-L91

We should change this to something like 

```python
    def save(self, value: Any, metadata: Dict[str, Any]) -> None:
        user_id = self.config.get("user_id", "")
        assistant_message = metadata.get('messages') or [{"role" : "assistant","content" : value}]
```

Here messages should be more like 

```python
{'role': 'system', 'content': 'You are Friendly chatbot assistant. You are a kind and knowledgeable chatbot assistant. You excel at understanding user needs, providing helpful responses, and maintaining engaging conversations. You remember previous interactions to provide a personalized experience.\nYour personal goal is: Engage in useful and interesting conversations with users while remembering context.\nTo give my best complete final answer to the task respond using the exact following format:\n\nThought: I now can give a great answer\nFinal Answer: Your final answer must be the great and the most complete as possible, it must be outcome described.\n\nI MUST use these formats, my job depends on it!'}
{'role': 'user', 'content': '\nCurrent Task: Respond to user conversation. User message: Can you remember my name ?\n\nThis is the expected criteria for your final answer: Contextually appropriate, helpful, and friendly response.\nyou MUST return the actual complete content as the final answer, not a summary.\n\nBegin! This is VERY important to you, use the tools available and give your best Final Answer, your job depends on it!\n\nThought:'}
{'role': 'assistant', 'content': "I now can give a great answer  \nFinal Answer: While I don’t have the ability to remember your name right now, I’d love to learn it! Feel free to share your name, and I’ll remember it for our future chats. I'm here to help and make our conversations as personal and enjoyable as possible!"}```

Vidit-Ostwal:
Additional to this, in the list of messages,
we should remove the system prompt as well, just keep the user and assistant message.

lucasgomide:
Yup, I’d prefer not to include that in this PR to keep the scope smaller
We can definitely add it later

Vidit-Ostwal:
> Yup, I’d prefer not to include that in this PR to keep the scope smaller We can definitely add it later

Aye, Aye Captain. :)

PR #3291, opened 2025-08-07
Lorenze/tracing v2

lorenzejay:


src/crewai/flow/flow.py
```diff
@@ -469,8 +471,10 @@ def __init__(
 
         # Initialize state with initial values
         self._state = self._create_initial_state()
-        if os.getenv("CREWAI_TRACING_ENABLED", "false").lower() == "true":
-            trace_listener = TraceCollectionListener()
+        print("tracing within flow", tracing)
```
lucasgomide: can we drop this debugger?

src/crewai/flow/flow.py
```diff
@@ -469,8 +471,10 @@ def __init__(
...
+        self.tracing = tracing
+        if os.getenv("CREWAI_TRACING_ENABLED", "false").lower() == "true" or tracing:
```
lucasgomide: we have this line `os.getenv("CREWAI_TRACING_ENABLED", "false").lower()`  in many places, what about create a method into some misc file and call `is_tracing_enabled` instead?

src/crewai/flow/flow.py
```diff
@@ -469,8 +471,10 @@ def __init__(
...
+            trace_listener = TraceCollectionListener(tracing=self.tracing or False)
```
lucasgomide: `self.tracing` default value is already False. Are you concerned about self.tracing beign None?

lucasgomide (approved):

PR #3323, opened 2025-08-14
Lorenze/emphemeral tracing

lorenzejay:


src/crewai/cli/authentication/token.py
```diff
@@ -1,9 +1,13 @@
...
+
+logger = getLogger(__name__)
 
 
 def get_auth_token() -> str:
     """Get the authentication token."""
     access_token = TokenManager().get_token()
     if not access_token:
-        raise Exception("No token found, make sure you are logged in")
+        logger.warning("No token found, make sure you are logged in")
```
lorenzejay: @vinibrsl you were the last to touch this. would not raising an exception here break anything ?

lorenzejay: alternatively i can do a return type of str | None and if None we can handle it

vinibrsl: Good call. Not raising an exception here would break things like [this](https://github.com/crewAIInc/crewAI/blob/07f8e7395853759bb5a60705e250b4efa29f23aa/src/crewai/cli/command.py#L22). We expect that exception to prompt the user to log in every time `PlusAPI` is instantiated.

Another option would be to raise an specific exception type, and catch it where you don't want to break.

src/crewai/cli/plus_api.py
```diff
@@ -18,6 +18,7 @@ class PlusAPI:
     CREWS_RESOURCE = "/crewai_plus/api/v1/crews"
     AGENTS_RESOURCE = "/crewai_plus/api/v1/agents"
     TRACING_RESOURCE = "/crewai_plus/api/v1/tracing"
+    TEMP_TRACING_RESOURCE = "/crewai_plus/api/v1/tracing/ephemeral"
```
lucasgomide: Can we call `ephemeral_tracing_resource` instead?

src/crewai/utilities/events/listeners/tracing/trace_listener.py
```diff
@@ -76,39 +75,37 @@ class TraceCollectionListener(BaseEventListener):
...
-    def _check_trace_enabled(self) -> bool:
+    def _check_authenticated(self) -> bool:
         """Check if tracing should be enabled"""
-        auth_token = get_auth_token()
-        if not auth_token:
+        try:
+            auth_token = get_auth_token()
+            if not auth_token:
+                return False
+            return True
```
lucasgomide:
```suggestion
            return get_auth_token()
```

src/crewai/utilities/events/listeners/tracing/trace_batch_manager.py
```diff
@@ -182,12 +206,24 @@ def _finalize_backend_batch(self):
...
+                else self.plus_api.finalize_trace_batch(self.trace_batch_id, payload)
+            )
 
             if response.status_code == 200:
+                access_code = response.json().get("access_code", None)
                 console = Console()
+                return_link = (
+                    f"{CREWAI_BASE_URL}/crewai_plus/trace_batches/{self.trace_batch_id}"
+                    if not ephemeral
+                    else f"{CREWAI_BASE_URL}/crewai_plus/ephemeral_trace_batches/{self.trace_batch_id}?access_code={access_code}"
```
gabemilani, line 223: Is the `access_code` required in this link? If not, we could omit the `?access_code` param if it's none.

src/crewai/cli/plus_api.py
lorenzejay: sure

src/crewai/utilities/events/listeners/tracing/trace_batch_manager.py
lorenzejay: it is.

gabemilani: Got it! Should we ensure that access_code is not None then?
Looking at the default value provided in the `access_code = response.json().get("access_code", None)` makes me think it can happen.

lorenzejay: what about this:
```py
return_link = (
                    f"{CREWAI_BASE_URL}/crewai_plus/trace_batches/{self.trace_batch_id}"
                    if not ephemeral and access_code
                    else f"{CREWAI_BASE_URL}/crewai_plus/ephemeral_trace_batches/{self.trace_batch_id}?access_code={access_code}"
                )
```

gabemilani: That looks great!

gabemilani (approved):
LGTM

PR #3328, opened 2025-08-15
chore: ignore deprecation warning from chromadb

gabemilani:
The following warning was being displayed when creating a Crew with `memory=True`:
```
PydanticDeprecatedSince211: Accessing the 'model_fields' attribute on the instance is deprecated. 
Instead, you should access this attribute from the model class. 
Deprecated in Pydantic V2.11 to be removed in V3.0.
```

This is happening in the `chromadb` dependency and they have [fixed it](https://github.com/chroma-core/chroma/pull/4339) in the version 1.0.8. This would require a major upgrade on our end, which we need to approach with more caution, so I'm suppressing the warnings for now until we upgrade.


lorenzejay (approved):

src/crewai/knowledge/storage/knowledge_storage.py
```diff
@@ -85,6 +87,15 @@ def search(
                 raise Exception("Collection not initialized")
 
     def initialize_knowledge_storage(self):
+        # Suppress deprecation warnings from chromadb, which are not relevant to us
+        # This is a temporary fix until we upgrade to chromadb to at least 1.0.8.
```
lucasgomide: Just adding a `TODO:` section to remember us to drop that after upgrating pls

lucasgomide (approved):

src/crewai/knowledge/storage/knowledge_storage.py
gabemilani: Sounds good, added it!

lucasgomide (approved):

PR #3334, opened 2025-08-18
chore: remove AgentOps integration

greysonlalonde:
- Remove AgentOps listener and dependencies
- Clean up documentation and navigation
- Update pyproject.toml and regenerate lock file

PR #3345, opened 2025-08-18
Revert "chore: remove AgentOps integration"

areibman:
Reverts crewAIInc/crewAI#3334

Hey Crew -- We received a ton of complaints overnight that the AgentOps docs on the Crew site went missing. Seems that there was an overzealous cleanup on the telemetry/instrumentation page. 

Are we able to revert? Are there any changes you'd like to see otherwise perhaps to make it clear/trimmed-down?

### DIFF
--- a/.github/workflows/type-checker.yml
+++ b/.github/workflows/type-checker.yml
@@ -12,13 +12,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0  # Fetch all history for proper diff
+          fetch-depth: 0 # Fetch all history for proper diff
 
       - name: Install uv
         uses: astral-sh/setup-uv@v6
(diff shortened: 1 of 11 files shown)


src/crewai/utilities/events/third_party/agentops_listener.py
```diff
@@ -0,0 +1,67 @@
+from typing import Optional
```
lorenzejay: hey @areibman , we're doing a cleanup on our codebase so that we have less third party dependancies within our repo. 

Is there a way you can implement this externally?

areibman: Yep! We instrument Crew without needing to be a strict dependency, so we can hook in without these files.

Of course, there's a risk some stuff might break on our side, but that's something we can solve independently :)

In the meantime, can we revert the documentation change?

lorenzejay: Sounds good. Yeah we're just prepping for a big v1.0.0 release so we're cleaning up shop. Just want to make sure everything works. 

But docs, we can bring back!

areibman: Done!

areibman:
@lorenzejay Did a major cleanup (plus updated a bunch of docs/links). Should be good to go now!

areibman:
bump! @lorenzejay @greysonlalonde @tonykipkemboi

lorenzejay (approved):
lgtm, docs back on cc: @greysonlalonde

areibman:
bump! @lorenzejay @greysonlalonde @tonykipkemboi

lorenzejay (approved):

PR #3347, opened 2025-08-18
Refactor tracing logic to consolidate conditions for enabling tracing…

lorenzejay:
… in Crew class and update TraceBatchManager to handle ephemeral batches more effectively. Added tests for trace listener handling of both ephemeral and authenticated user batches.


src/crewai/utilities/events/listeners/tracing/trace_batch_manager.py
```diff
@@ -58,10 +60,12 @@ def initialize_batch(
         use_ephemeral: bool = False,
     ) -> TraceBatch:
         """Initialize a new trace batch"""
+        print("lorenze is here: Initializing batch", use_ephemeral)
```
lucasgomide: oops

lorenzejay: lol

PR #3361, opened 2025-08-19
Fix flaky test_agent_from_repository

greysonlalonde:
Set EMBEDCHAIN_DB_URI environment variable to avoid TypeError when XMLSearchTool initializes embedchain

### DIFF
--- a/docs/en/tools/overview.mdx
+++ b/docs/en/tools/overview.mdx
@@ -117,4 +117,19 @@ agent = Agent(
 )
 ```
 
+## **Max Usage Count**
+
+You can set a maximum usage count for a tool to prevent it from being used more than a certain number of times.
+By default, the max usage count is unlimited.
+
+
+
+```python
+from crewai_tools import FileReadTool
+
+tool = FileReadTool(max_usage_count=5, ...)
+```
+
+
+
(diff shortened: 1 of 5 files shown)


tests/agent_test.py
```diff
@@ -2347,51 +2334,65 @@ def mock_get_auth_token():
...
+            "role": "test role",
+            "goal": "test goal",
+            "backstory": "test backstory",
+            "tools": [
+                {
+                    "module": "crewai_tools",
+                    "name": "SerperDevTool",
+                    "init_params": {"n_results": "30"},
+                },
+                {
```
lucasgomide, line 2364: Can we just drop the `XMLSearchTool` from this list? Let's use a simpler tool instead

greysonlalonde: Zero objections there, but I think all RAG based tools will run into this issue. What do you suggest?

lucasgomide: When I've added this tool  I just picked one at random. I think we could use any other that doesn't inherent from RagTool.

lucasgomide: On enterprise we set this variable automactly, I thnk

PR #3374, opened 2025-08-20
feat: chromadb generic client

greysonlalonde:
## Summary
- Implement ChromaDBClient conforming to BaseClient protocol
- Add utility functions and type definitions
- Include comprehensive test suite

Work in progress - integration into existing codebase pending


src/crewai/rag/chromadb/utils.py
```diff
@@ -0,0 +1,157 @@
...
+        metadatas_list[0] if metadatas_list and "metadatas" in include_strings else []
+    )
+
+    distances_list = results.get("distances")
+    distances = (
+        distances_list[0] if distances_list and "distances" in include_strings else []
+    )
+
+    for i, doc_id in enumerate(ids):
+        score = 1.0 - (distances[i] / 2.0) if distances and i < len(distances) else 0.0
```
lucasgomide: I think there are a couple of issues here:

1. You are assuming a range of [0, 2] distance for each documents. But is non-standard for both cosine and L2 (defaul chromaDB's metrics) distance conversions. So can you clarify which metric are you based to calculate the distance here? I'd recommend using the correct conversion formula
2. Would be great support the "distnace kind" since we can configure collections with many different [hnsw-index](https://docs.trychroma.com/docs/collections/configure#hnsw-index-configuration.My concern here is  it may produce incorrect similarity scores depending on the actual distance function used )

greysonlalonde: this was some dummy logic, missed replacing it 🤦

src/crewai/rag/chromadb/utils.py
```diff
@@ -0,0 +1,221 @@
...
+        distance_metric: The distance metric used ("l2", "cosine", or "ip").
+
+    Returns:
+        Similarity score in range [0, 1] where 1 is most similar.
+    """
+    if distance_metric in ("cosine", "ip"):
+        score = 1.0 - 0.5 * distance
+    else:
+        score = 1.0 - 0.25 * distance
+    return max(0.0, min(1.0, score))
```
lucasgomide: I liked this version, but will left my 🪙 on that

I believe the functions has issues. I'm not a mathematical expert but I got some insights from [here](https://en.wikipedia.org/wiki/Cosine_similarity) and [here](https://medium.com/@zilliz_learn/similarity-metrics-for-vector-search-62ccda6cfdd8)

Cosine: Using 0.5 * distance is incorrect. Cosine distance ranges [0,2], so it should be distance / 2.0

IP: For normalized vectors, IP similarity = 1 - distance, not 1 - 0.5 * distance

L2: Linear scaling 0.25 * distance will saturate quickly since L2 is unbounded. Consider 1.0 / (1.0 + distance) instead

So I would say you should do something like that

```suggestion
if distance_metric == "cosine":
        # Cosine distance is [0,2], divide by 2 to normalize
        score = 1.0 - (distance / 2.0)
    elif distance_metric == "ip":
        # IP distance equals (1 - similarity) for normalized vectors
        score = 1.0 - distance
    else: 
        # L2 is unbounded, use reciprocal to avoid saturation
        score = 1.0 / (1.0 + distance)
    
    return max(0.0, min(1.0, score))
```

greysonlalonde: pushing this per internal conversations

lorenzejay (approved):

PR #3377, opened 2025-08-21
feat: qdrant generic client

greysonlalonde:
Implements Qdrant vector database client for RAG functionality.

## Changes
- Add QdrantClient class with sync/async collection and document operations
- Add type definitions and utility functions
- Add score normalization from [-1, 1] to [0, 1] range
- Include BAAI/bge-small-en-v1.5 as default embedding
- Extract shared constants and utilities


src/crewai/rag/qdrant/client.py
```diff
@@ -0,0 +1,527 @@
...
+            self.client = client
+        else:
+            location = kwargs.get("location", ":memory:")
+            client_kwargs = {k: v for k, v in kwargs.items() if k != "location"}
+            self.client = SyncQdrantClientBase(location, **cast(Any, client_kwargs))
+
+        if embedding_function is not None:
+            self.embedding_function = embedding_function
+        else:
+            _embedder = TextEmbedding("BAAI/bge-small-en-v1.5")
```
lucasgomide, line 73: Can we support custom model name? Maybe not for now..

lucasgomide (approved):

src/crewai/rag/qdrant/client.py
greysonlalonde: absolutely! planned in an upcoming config pr

src/crewai/rag/qdrant/client.py
```diff
@@ -0,0 +1,527 @@
...
+                alt_client="AsyncQdrantClient",
+            )
+
+        collection_name = kwargs["collection_name"]
+
+        if self.client.collection_exists(collection_name):
+            raise ValueError(f"Collection '{collection_name}' already exists")
+
+        params = _get_collection_params(kwargs)
+        self.client.create_collection(**params)
```
vinibrsl, line 117: As this process may happen in multiple threads or processes in parallel, what do you think of skipping `self.client.collection_exists` and catching as an exception of `self.client.create_collection`?

Or even doesn't raise that at all, and swallow that exception and treat this as "get or create collection".

src/crewai/rag/qdrant/client.py
```diff
@@ -0,0 +1,527 @@
...
+                alt_client="QdrantClient",
+            )
+
+        collection_name = kwargs["collection_name"]
+
+        if await self.client.collection_exists(collection_name):
+            raise ValueError(f"Collection '{collection_name}' already exists")
+
+        params = _get_collection_params(kwargs)
+        await self.client.create_collection(**params)
```
vinibrsl, line 157: Same parallel execution comment.

src/crewai/rag/qdrant/client.py
```diff
@@ -0,0 +1,527 @@
...
+
+        collection_name = kwargs["collection_name"]
+
+        if self.client.collection_exists(collection_name):
+            return self.client.get_collection(collection_name)
+
+        params = _get_collection_params(kwargs)
+        self.client.create_collection(**params)
+
+        return self.client.get_collection(collection_name)
```
vinibrsl, line 201: I believe we should catch exceptions here to prevent race conditions between `self.client.collection_exists` and `self.client.get_collection`.

src/crewai/rag/qdrant/client.py
greysonlalonde: let me dig into the qdrant docs a little more to be sure, but I think I'm aligned with you here + on the other comments

greysonlalonde: @vinibrsl marking this as resolved - I'm implementing a reusable lock manager with `portalocker` and will add in a separate pr

Issue #3220, opened 2025-07-25
[BUG] Mem0 Memory Save / Search Issue


Vidit-Ostwal:
### Description

Few points to discuss on the Mem0 Save and Search Issue. Note: User-Memory will be dropped in the next 2 weeks and external memory will be used in place.

Current Scenario.
1. All the message which are sent to Mem0 are sent as `assisstant` message. There memory will be tagged as `agent_id` only.
2. `Agent_id` has been dropped, we don't send agent_id parameter while sending the mem0 search request.
3. We can send both `user_id` and `agent_id`, but there is a problem while searching with both of them.

Current discussed Solution 

> Fix the adding memories when agent_id is provided; We shouldn't drop it.
> Refactor this class and drop the user_id support until the version 0.156.0 or on 2025-08-04. Since this is the deadline to remove UserMemory class

Proposed solution
1. Instead of sending the final message we can send the list of message, and can appropriately send the Mem0 request
2. Bring back `agent_id` and keep `user_id` as well.
3. I think the search was wrong, if one message can either be from `user` or an `assisstant`, we should not have a AND block for the filter in Mem0. Start using OR block if both user_id and agent_id is used while initializing.

### Steps to Reproduce

-

### Expected behavior

-

### Screenshots/Code snippets

-

### Operating System

macOS Sonoma

### Python Version

3.11

### crewAI Version

Latest Git

### crewAI Tools Version

-

### Virtual Environment

Venv

### Evidence

None

### Possible Solution

-

### Additional context

-

Vidit-Ostwal:
@lucasgomide, let me know if I am missing something.

lucasgomide:
1. [This PR](https://github.com/crewAIInc/crewAI/pull/3217) brought `agent_id` back 
2. The same PR also is using OR for searching messages from agent_id/user_id. Alghoth is not working properly, I believe that why we are sending message as `assistant` role.
3. Sending exchanged LLM messages (both user and assistant) would great def. However, currently, the framework only sends and stores the Agent's replies, always works like that.
I'm wondering has it always worked this way, or was this behavior changed in V2? I've seen some setups where the developer manually adds the user messages for each interaction and then only retrieves by user_id.

Vidit-Ostwal:
>  I believe that why we are sending message as assistant role.

I only made this change in the `save` method, here #3187, Didn't took this into consideration


[...]

PR #3390, opened 2025-08-23
Dropping messages from metadata in Mem0 Storage

Vidit-Ostwal:
Following changes have been done:

- Instead of adding user_message and agent_message in metadata, this PR add them to message while calling the save method of Mem0 Storage
- Also drops the messages parameter from the metadata config. (Metadata config has limitation of 2000 characters)

More context here -> https://github.com/crewAIInc/crewAI/pull/3290#issuecomment-3168621637 
@lucasgomide

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,6 +22,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Fetch all history for proper diff
 
       - name: Restore global uv cache
         id: cache-restore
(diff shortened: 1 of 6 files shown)


Vidit-Ostwal:
Fixes #3391

Vidit-Ostwal:
Resolves #3220

Vidit-Ostwal:
Requesting @lucasgomide for a review.

src/crewai/memory/storage/mem0_storage.py
```diff
@@ -86,9 +88,30 @@ def _create_filter_for_search(self):
...
+
+            if user_msg := self._get_user_message(user_content):
+                conversations.append({"role": "user", "content": user_msg})
+            if assistant_msg := self._get_assistant_message(assistant_content):
+                conversations.append({"role": "assistant", "content": assistant_msg})
+
+            metadata.pop('messages') # Dropping messages from the metadata dictionary
+        else:
+            conversations.append({"role": "assistant", "content": value})
+
```
lucasgomide: Here is a more clear suggestion
```suggestion
def save(self, value: Any, metadata: dict[str, Any]) -> None:
    def _last_content(messages: Iterable[dict[str, Any]], role: str) -> str:
        return next(
            (m.get("content", "") for m in reversed(list(messages)) if m.get("role") == role),
            ""
        )

    messages = metadata.pop("messages", None)

    if messages:
        last_user = _last_content(messages, "user")
        last_assistant = _last_content(messages, "assistant")

        if user_msg := self._get_user_message(last_user):
            conversations.append({"role": "user", "content": user_msg})

        if assistant_msg := self._get_assistant_message(last_assistant):
            conversations.append({"role": "assistant", "content": assistant_msg})
    else:
        conversations.append({"role": "assistant", "content": value})
```

lucasgomide:
Can you take a look on type-checkers?

src/crewai/memory/storage/mem0_storage.py
Vidit-Ostwal: Added

Vidit-Ostwal:
> Can you take a look on type-checkers?

wasn't able to find a type file for mem0. `pip install type-mem0`
Have added a ignore rule in pyproject.toml

```
[[tool.mypy.overrides]]
module = "mem0.*"
ignore_missing_imports = true
```
Let me know if you are looking for some other solution

lucasgomide:
> > Can you take a look on type-checkers?
> 
> wasn't able to find a type file for mem0. `pip install type-mem0` Have added a ignore rule in pyproject.toml
> 
> ```
> [[tool.mypy.overrides]]
> module = "mem0.*"
> ignore_missing_imports = true
> ```
> 
> Let me know if you are looking for some other solution

A simpler solution would be only adding `# type: ignore[import-untyped]` on the line. Can you change that?

Vidit-Ostwal:
Yes, have made the changes.

PR #3394, opened 2025-08-25
Add RAG configuration system with optional imports

greysonlalonde:
## Summary
- Add RAG configuration system
- Support for ChromaDB provider
- Handle optional dependencies gracefully

## Usage
```python
import crewai
from crewai.rag.chromadb.config import ChromaDBConfig  # import location will be updated

# Configure RAG
crewai.rag.config = ChromaDBConfig(
    tenant=...,
    database=...,
    settings=Settings(...),
    embedding_function=...
)
```


src/crewai/rag/__init__.py
```diff
@@ -1 +1,58 @@
...
+from typing import Any
+
+from crewai.rag.config.types import RagConfigType
+from crewai.rag.config.utils import set_rag_config
+
+
+_module_path = __path__
+_module_file = __file__
+
+class _RagModule(ModuleType):
```
joaomdmoura, line 15: I don't think we are using this still?

joaomdmoura (approved):

src/crewai/rag/__init__.py
greysonlalonde: Mostly new logic in this, consolidating any rag-related functionality here & removing redundancy as we go

Issue #3277, opened 2025-08-05
[BUG] crewai chat yields litellm.BadRequestError: VertexAIException


mikebz:
### Description

crewai chat seems to be generating an exception at least when the gemini models are being used.

### Steps to Reproduce

1. create a simple 2 agent crew (largely based on the fast start).
2. try `crewai run` - get the results with no problems.
3. try `crewai chat` - receive an exception

```
crewai chat

Starting a conversation with the Crew
Type 'exit' or Ctrl+C to quit.


Analyzing crew and required inputs - this may take 3 to 30 seconds depending on the complexity of your crew.
.....................................

...

litellm.exceptions.BadRequestError: litellm.BadRequestError: VertexAIException BadRequestError - {
  "error": {
    "code": 400,
    "message": "Unable to submit request because at least one contents field is required. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.DebugInfo",
        "detail": "[ORIGINAL ERROR] generic::invalid_argument: Unable to submit request because at least one contents field is required. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini [google.rpc.error_details_ext] { message: \"Unable to submit request because at least one contents field is required. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini\" } 525003295 { 1 { 1 { 1 { 1: \"blade:cloud.ai.nl.llm.proto.service.languagemodelservice-prod-us-central1\" 2: \"LanguageModelService\" 3: \"GenerateMultiModal\" } 2 { 1: 3 2: \"generic\" 3: \"Unable to submit request because at least one contents field is required. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini\" 5 { 1 { 2: 82012934 3 { 2: \"Unable to submit request because at least one contents field is required. Learn more: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini\" } } } } 5: 1754415558162639 6: 1112 8: 1 9: 1 } 2: \"PredictionServiceV1\" 3: \"GenerateContent\" } }"
      }
    ]
  }
}
```

### Expected behavior

I am expecting that I am able to start a chat session.

### Screenshots/Code snippets

crewai set up with uv, project.toml is here:

```
[project]
[...]

PR #3407, opened 2025-08-27
Fix: Vertex/Gemini failure in crew chat (system-only intro → valid user turn)

laulopezreal:
Fixes: #3277 [GitHub](https://github.com/crewAIInc/crewAI/issues/3277)

### Summary

crew chat failed on Gemini (Vertex AI) because the first/only message sent was a system role. Vertex’s `generateContent` contract requires a valid conversation content payload; a lone system instruction is insufficient. Gemini (including Vertex-backed routes) rejects conversations that don’t contain a user turn. This PR appends a minimal user message when none is present so initial calls don’t fail on Gemini., so the initial call succeeds on Vertex/Gemini while keeping behavior unchanged for other providers. 
[Google Cloud](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini)

### Root cause
Gemini chat endpoints require a conversational content turn from the user. A system-only or assistant-only opening causes a schema/validation error. Appending a minimal user turn is the least invasive way to satisfy Gemini without altering semantics for other providers.

### What changed

In LLM._format_messages_for_provider:

- Gemini: ensure at least one user turn (Vertex requires contents)
```
if "gemini" in self.model.lower():
    if not any(m.get("role") == "user" for m in messages):
        # Append a minimal user turn; do not reorder/replace earlier messages
        return [*messages, {"role": "user", "content": "Okay."}]
    return messages
```

- Detection: if "gemini" in self.model.lower()

- Behavior: If no user role exists, append {"role": "user", "content": "Okay."}.

- Design choice: Append-only (no reordering/merging), to keep previous system/assistant turns intact.

#### Before (fails on Gemini)

[
  {"role": "system", "content": "You are..."},
  {"role": "assistant", "content": "Ready."}
]

#### After (accepted by Gemini)

[
  {"role": "system", "content": "You are..."},
  {"role": "assistant", "content": "Ready."},
  {"role": "user", "content": "Okay."}
]

### Backwards compatibility

- Other providers remain unaffected (the guard runs only when the model string includes "gemini").

- Anthropic and OpenAI o1 behavior is unchanged (their existing branches still run as before).

### Tests

- New (Vertex/Gemini, chat intro):

Given messages=[{"role":"system","content":"..."}], output to LiteLLM/Vertex contains a first user turn with the merged preamble; no empty or missing contents.

- Regression:

Anthropic: first message remains user; unchanged behavior.

OpenAI o1: no system role passes through; unchanged behavior.

Vertex/Gemini now works in crew chat without requiring user hacks.

### Risks & mitigations

Risk: Altering the very first turn semantics.
Mitigation: Only applies when no user message exists; content is additive (system preamble), preserving instruction fidelity.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -421,6 +421,7 @@
                   "en/api-reference/introduction",
                   "en/api-reference/inputs",
                   "en/api-reference/kickoff",
+                  "en/api-reference/resume",
                   "en/api-reference/status"
                 ]
               }
@@ -827,6 +828,7 @@
                   "pt-BR/api-reference/introduction",
                   "pt-BR/api-reference/inputs",
                   "pt-BR/api-reference/kickoff",
+                  "pt-BR/api-reference/resume",
                   "pt-BR/api-reference/status"
                 ]
               }
@@ -1239,6 +1241,7 @@
                   "ko/api-reference/introduction",
                   "ko/api-reference/inputs",
                   "ko/api-reference/kickoff",
+                  "ko/api-reference/resume",
                   "ko/api-reference/status"
                 ]
               }
--- a/docs/en/enterprise/guides/human-in-the-loop.mdx
+++ b/docs/en/enterprise/guides/human-in-the-loop.mdx
@@ -40,6 +40,28 @@ Human-In-The-Loop (HITL) is a powerful approach that combines artificial intelli
         <Frame>
             <img src="/images/enterprise/crew-resume-endpoint.png" alt="Crew Resume Endpoint" />
         </Frame>
+
+        <Warning>
+            **Critical: Webhook URLs Must Be Provided Again**:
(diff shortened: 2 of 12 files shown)


laulopezreal:
Plz review when you have some time :) @lorenzejay @joaomdmoura @bhancockio @lucasgomide @tonykipkemboi

src/crewai/llm.py
```diff
@@ -1156,11 +1156,18 @@ def _format_messages_for_provider(
             and messages[-1]["role"] == "assistant"
         ):
             return messages + [{"role": "user", "content": ""}]
-
+        
+       # Gemini: ensure at least one user turn (Vertex requires contents)
+        if "gemini" in self.model.lower():
+            if not any(m.get("role") == "user" for m in messages):
```
mishra-krishna, line 1162: This is a good fix for the Gemini API. However, instead of hardcoding the user message to 'Okay.', it would be better to use a more generic message like 'Ok.' to avoid any potential issues with the model's behavior. Also, it would be great to add a comment explaining why this is needed.

laulopezreal: @mishra-krishna resolved, please have a look and merge if you think is ok now.

lorenzejay: We have implemented support for native sdks for gemini/vertex. mind trying it out with our pre-release version? You can try with `uv add crewai==1.0.0a3 --prerelease=allow`

PR #3420, opened 2025-08-28
chore: refactor parser to module functions

greysonlalonde:
- Convert CrewAgentParser class to module functions
- Extract constants to dedicated module
- Pre-compile regex patterns for performance
- Fix class attributes to instance attributes in ToolsHandler

### DIFF
--- a/src/crewai/agents/constants.py
+++ b/src/crewai/agents/constants.py
@@ -1,0 +1,27 @@
+"""Constants for agent-related modules."""
+
+import re
+from typing import Final
+
+# crewai.agents.parser constants
+
+FINAL_ANSWER_ACTION: Final[str] = "Final Answer:"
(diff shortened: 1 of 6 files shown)


src/crewai/agents/constants.py
```diff
@@ -0,0 +1,27 @@
+"""Constants for agent-related modules."""
+
+import re
+from typing import Final
+
+# crewai.agents.parser constants
```
lucasgomide, line 6: Can we drop this line?

lucasgomide (approved):

src/crewai/agents/constants.py
greysonlalonde: Keeping temporarily, but just because I see this file expanding with many sections in the next weeks and want these to be easily pulled out

PR #3422, opened 2025-08-29
fix: enhance LLM event handling with task and agent metadata

lorenzejay:
- Added `from_task` and `from_agent` parameters to LLM event emissions for improved traceability.
- Updated `_send_events_to_backend` method in TraceBatchManager to return status codes for better error handling.
- Modified `CREWAI_BASE_URL` to remove trailing slash for consistency.
- Improved logging and graceful failure handling in event sending process.

### DIFF
--- a/src/crewai/llm.py
+++ b/src/crewai/llm.py
@@ -992,8 +992,6 @@ class LLM(BaseLLM):
         """
         # --- 1) Emit call started event
         assert hasattr(crewai_event_bus, "emit")
-        print(f"from_task: {from_task}")
-        print(f"from_agent: {from_agent}")
         crewai_event_bus.emit(
             self,
             event=LLMCallStartedEvent(
--- a/src/crewai/llm.py
+++ b/src/crewai/llm.py
@@ -950,6 +950,8 @@ class LLM(BaseLLM):
                         tool_name=function_name,
                         tool_args=function_args,
                         error=f"Tool execution error: {str(e)}",
+                        from_task=from_task,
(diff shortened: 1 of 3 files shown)


src/crewai/llm.py
```diff
@@ -990,6 +992,8 @@ def call(
         """
         # --- 1) Emit call started event
         assert hasattr(crewai_event_bus, "emit")
+        print(f"from_task: {from_task}")
+        print(f"from_agent: {from_agent}")
```
lucasgomide: ops

src/crewai/utilities/constants.py
```diff
@@ -16,4 +16,4 @@ def __repr__(self):
 # Unlike `None`, which might be a valid value from the user, `NOT_SPECIFIED` allows
 # us to distinguish between "not passed at all" and "explicitly passed None" or "[]".
 NOT_SPECIFIED = _NotSpecified()
-CREWAI_BASE_URL = "https://app.crewai.com/"
+CREWAI_BASE_URL = "https://app.crewai.com"
```
lucasgomide, line 19: Is really safe remove that?

lorenzejay: yeah because its only used in two places and we add this anyways `/`
Screenshot 2025-08-29 at 12 16 45 PM

src/crewai/llm.py
lorenzejay: whoops indeed hahaha

lucasgomide (approved):

PR #3425, opened 2025-08-30
refactor: Move events module to crewai.events

greysonlalonde:
Moves events from `crewai.utilities.events` to `crewai.events`.

- Relocated events module to top-level package
- Added deprecation warnings for old import paths
- Updated documentation and internal imports

Old: `from crewai.utilities.events import ...`
New: `from crewai.events import ...`

Old imports still work but will be removed in v1.0.0.


vinibrsl (approved):

src/crewai/events/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    "MemorySaveStartedEvent",
+    "MemoryQueryStartedEvent",
+    "MemoryRetrievalCompletedEvent",
+    "MemorySaveFailedEvent",
+    "MemoryQueryFailedEvent",
+    "KnowledgeRetrievalStartedEvent",
+    "KnowledgeRetrievalCompletedEvent",
+    "CrewKickoffStartedEvent",
+    "CrewKickoffCompletedEvent",
+    "AgentExecutionCompletedEvent",
```
carsongee, line 54: Sorry for the ghost post, but an error after an upgrade landed me on this PR. Did `AgentExecutionStartedEvent` just get missed here, or is it intentional and I should upgrade by importing it directly from `agent_events` @greysonlalonde ?

greysonlalonde: Oversight! Thanks for the catch

greysonlalonde: Fix in #3532

PR #3429, opened 2025-08-31
fix: replace mutable default arguments with None

greysonlalonde:
Replace mutable default arguments (lists/dicts) with None to prevent shared state bugs across calls.


lucasgomide (approved):
That's amazing! Thank you for caring

PR #3270, opened 2025-08-04
WIP: Bedrock agentcore memory

3coins:

PR #3458, opened 2025-09-05
feat: Bedrock AgentCore Memory integration

GowriH-1:
This PR is based on @3coins' WIP [draft](https://github.com/crewAIInc/crewAI/pull/3270) and adds support for [Bedrock Agentcore Memory](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/memory.html) as CrewAI External Memory.

# Key Features:
1. `save`  - As an External Memory Storage location, Bedrock AgentCore Storage extends the abstract `save` method. It utilizes Bedrock AgentCore's `create_events` API that allows passing upto 100 messages in a batch.
2. `search` - As an External Memory Storage location, Bedrock AgentCore Storage extends the abstract `search` method. It utilizes Bedrock AgentCore's `retrieve_memories` API that returns a relevance score associated with each retrieved result. The search method filters results based on a minimum relevance score (default: 0.35) for long term memories. If there are not enough long term memories to meet the limit threshold, `search` also calls `list_events` with and appends `limit - len(long_term_memories)` to the list of memories returned. 

# Call outs:
1. `reset` - This method is currently implemented as a pass method as Bedrock AgentCore doesn't provide a direct reset API.  The customer can call `delete_memory` to delete the entire resource - however, we will refrain from doing a control plane operation on the customer's behalf.
2. `crew.kickoff({"question": question})` - automatically saves memory internally [code ref](https://github.com/crewAIInc/crewAI/blob/main/src/crewai/agents/agent_builder/base_agent_executor_mixin.py#L61)

# Sample Usage:
```
agentcore_config = BedrockAgentCoreConfig(
    memory_id="your-bedrock-memory-id",
    actor_id="user-123",
    session_id="session-456",
    region_name="us-east-1",
    namespaces=[
        "/preferences/user-123",
        "/facts/user-123/session-456",
        "/summaries/user-123/session-456"
    ]
)

external_memory = ExternalMemory(
    embedder_config={
        "provider": "agentcore",
        "config": agentcore_config 
    }
)

crew = Crew(
    agents=[...],
    tasks=[...],
    external_memory=external_memory,
    verbose=True
)

external_memory.set_crew(crew)
```

---

> [!NOTE]
> Adds AWS Bedrock AgentCore as an ExternalMemory storage with save/search via Bedrock APIs, plus docs, samples, tests, and optional dependencies.
> 
> - **Memory/Backend**:
>   - **New Storage**: Implement `BedrockAgentCoreStorage` with `save` (creates events) and `search` (long‑term `retrieve_memory_records` + short‑term `list_events`, score filtering) and config model `BedrockAgentCoreConfig` in `crewai/memory/storage/bedrock_agentcore_storage.py`.
>   - **Provider Wiring**: Register `agentcore` in `ExternalMemory.external_supported_storages()` and accept both `BedrockAgentCoreConfig` and dict configs.
> - **Tests**:
>   - Add comprehensive tests for AgentCore config, storage, error handling, payload building, search fallbacks, and event emission in `tests/storage/test_bedrock_agentcore_storage.py` and extend `tests/memory/test_external_memory.py`.
> - **Docs & Samples**:
>   - Add AWS Bedrock AgentCore section with configuration examples to `docs/en/concepts/memory.mdx`.
>   - New sample notebook `samples/bedrock-agentcore/memory_demo.ipynb` demonstrating setup and queries.
> - **Packaging**:
>   - Introduce `agentcore` extra in `pyproject.toml` (adds `boto3`, `botocore`); dev deps include `boto3-stubs[bedrock-agentcore]`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit dfde1374acdd9a86d7b1d1543d93644c5ee86c0e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


src/crewai/events/types/memory_events.py
```diff
@@ -37,7 +37,7 @@ class MemorySaveStartedEvent(BaseEvent):
     """Event emitted when a memory save operation is started"""
 
     type: str = "memory_save_started"
-    value: Optional[str] = None
+    value: Optional[Any] = None
```
lil-frankie: These look like shared utilities across crew ai, should we be changing this from str to any?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
...
+
+        logger.info(
+            "AgentCoreStorage initialized: memory_type=%s, run_id=%s",
+            self.memory_type,
+            self.config.run_id,
+        )
+
+    def _validate_type(self, type: str) -> None:
+        """Validate the memory type."""
+        supported_types = {"external"}
```
lil-frankie, line 102: we could make this a class variable instead of a local var

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
...
+        if type not in supported_types:
+            raise ValueError(
+                f"Invalid type '{type}' for AgentCoreStorage. "
+                f"Must be one of: {', '.join(supported_types)}"
+            )
+
+    def _initialize_memory_client(self) -> None:
+        """Initialize the Bedrock AgentCore Memory client."""
+        if MemoryClient is None:
```
lil-frankie: why do we need to check for this before we try to initialize it? Can't we just rely on the catch for any initialization exceptions?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
...
+
+                    # Process search results
+                    if memories and isinstance(memories, list):
+                        for memory in memories:
+                            # Extract text content from memory record
+                            content = memory.get("content", {})
+                            text = content.get("text", "")
+
+                            # Get score if available
+                            score = memory.get("score", 1.0)
```
lil-frankie: Is it appropriate to assign a default score of 1.0 if it's not available? We should check with Jakob on this

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
...
+                                        "search_namespace": search_namespace,
+                                        "memory_strategy_id": memory.get(
+                                            "memoryStrategyId"
+                                        ),
+                                    },
+                                    "score": score,
+                                }
+                            )
+
+                except Exception as e:
```
lil-frankie: do we want to gracefully consume any retrieveMemories exceptions? How can the user easily tell which namespace is missing from `all_results` if it resulted in an exception?

lil-frankie:
I think this PR is also missing implementation for delete/list/get events & delete/list memory records. We should consider adding support for those methods as well

pyproject.toml
```diff
@@ -40,6 +40,8 @@ dependencies = [
     "blinker>=1.9.0",
     "json5>=0.10.0",
     "portalocker==2.7.0",
+    "boto3>=1.40.19",
+    "botocore>=1.40.19",
```
3coins: We should probably avoid adding direct dependency on boto3, as there will be users who won't be using agentcore or AWS services.

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
+import json
+import logging
+from typing import Any, Dict, List, Optional
+
+from pydantic import BaseModel, Field, field_validator
+
+from bedrock_agentcore.memory.client import MemoryClient
```
3coins: This should move to inside the initialization block.

src/crewai/memory/storage/bedrock_agentcore_storage.py
3coins: Should we change this to check for agentcore sdk?
```python
try:
  from bedrock_agentcore.memory.client import MemoryClient
except ImportError:
  raise ModuleNotFoundError(f"No Bedrock AgentCore sdk found. Use `pip install crewai[agentcore]` to install the sdk.")
```

pyproject.toml
```diff
@@ -62,6 +64,10 @@ openpyxl = [
     "openpyxl>=3.1.5",
 ]
 mem0 = ["mem0ai>=0.1.94"]
+agentcore = [
+    "boto3>=1.35.0",
+    "botocore>=1.35.0",
```
3coins: Should we update this to `bedrock-agentcore>=<version>`?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
...
+        elif "tool" in str(metadata).lower():
+            return "TOOL"
+
+        return "ASSISTANT"  # Default role
+
+    def reset(self) -> None:
+        """Reset/clear memory storage."""
+        logger.warning(
+            "Bedrock AgentCore memory reset not implemented - memories persist in AWS"
+        )
```
3coins, line 453: Ok to leave this as-is for now, but we should check with CrewAI maintainers about what users expect with reset? For example, there will be scenarios where a user might want to reset their chat (delete all history), so they start fresh within a conversation or don't want to save any history for privacy reasons.

src/crewai/events/types/memory_events.py
gowri100:
```
def save(
        self,
        value: Any,
        metadata: Optional[Dict[str, Any]] = None,
        agent: Optional[str] = None,
    ) -> None:
        """Saves a value into the external storage."""
        crewai_event_bus.emit(
            self,
            event=MemorySaveStartedEvent(
                value=value,
                metadata=metadata,
                agent_role=agent,
                source_type="external_memory",
            ),
        )

```

should be fine as the interface that creates a `MemorySaveStartedEvent` uses any type for value [code ref](https://github.com/crewAIInc/crewAI/blob/main/src/crewai/memory/external/external_memory.py#L60)

since string is an any type, this also does not break backwards compatibility.

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: i think it might be better to have it as a local variable as supported types isn't exactly a property of `BedrockAgentCoreStorage` and won't be used anywhere else in the class.

Any reason to make it a class var?

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: the check is more to provide an explicit error message

if the client isn't initialized, and the customer gets an NPE equivalent - might be hard for them to debug

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: Good catch, I think it might be better to return 0 and not include it but I'll wait for Jakob's take as well

docs/en/concepts/memory.mdx
```diff
@@ -589,6 +589,106 @@ crew = Crew(
...
+agentcore_config = BedrockAgentCoreConfig(
+    memory_id="your-bedrock-memory-id",
+    actor_id="user-123",
+    session_id="session-456",
+    region_name="us-east-1",
+    strategies=[
+        BedrockAgentCoreStrategyConfig(
+            name="user_preferences",
+            namespaces=["/preferences/user-123"],
+            strategy_id="strategy-pref-123"
```
3coins: Assuming that strategies will already be setup by the user before they create the storage class here. Is that correct?
If so, is stategy id/name a required input to work with strategies. My understanding was that namespace should be enough to query from a specific strategy? Wondering if we can simplify this to just take a list of strategy_namespaces? For example, I am following this sample from the [agentcore docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/memory-getting-started.html#memory-getting-started-long-term-memory).
```python
# No strategy id, or name, just the namespace.
memories = client.retrieve_memories(
    memory_id=memory.get("id"),
    namespace=f"/summaries/User84/OrderSupportSession1",
    query="can you summarize the support issue"
)

```

gowri100:
> I think this PR is also missing implementation for delete/list/get events & delete/list memory records. We should consider adding support for those methods as well

In Crew AI, Bedrock AgentCore Memory is implemented as an external storage rather than a tool. 

The Storage interface requires support for save, search, and reset (which this PR goes over).

delete/list/get events & delete/list memory -> these are not properties of storage + these will provide no additional functionality except memory_client.get_events: which the user can directly implement on their end if required.

Implementing these on our end might make the integration less extensible as we would have to keep updating these methods with new parameters as list / get / delete evolve while providing no additional functionality except being a wrapper around the existing wrapper client.

3coins:
@lucasgomide
Adding you here to review the integration with agentcore memory, also can you help kick off the CI.

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: Makes sense! 

I didn't want to take any control plane actions on behalf of the user - say they expect to clear the chat but end up deleting the memory resource itself.

docs/en/concepts/memory.mdx
gowri100: Makes sense

I kept strategy_id + name in case we need it to support future operations - ex. searching based on strategy name / id - not sure if this is a valid use case, can remove strategy_id + name otherwise

src/crewai/memory/storage/bedrock_agentcore_storage.py
bergjaak: I think 0, but I believe it should always be returned

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
...
+
+# Setup logger
+logger = logging.getLogger(__name__)
+
+
+class BedrockAgentCoreStrategyConfig(BaseModel):
+    """Configuration for an AgentCore memory strategy."""
+
+    name: str = Field(..., description="Name of the strategy (required)")
+    namespaces: List[str] = Field(
```
bergjaak: is there a way, like say with a doc string, to suggest that search searches over all of these namespaces? So the user should only give us the namespaces that they want to search. Also, the user can search prefixes of namespaces... so I am thinking it might make most sense to name this field like `searchableNamespacePrefixes` and then write a doc string explaining that these namespaces (or namespace prefixes) are used for search in retrieveMemoryRecords

then we can also explain to the writer that if they want to search all the namespaces, they can just give us a prefix that is common to all of their namespaces

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,487 @@
...
+
+                except Exception as e:
+                    logger.warning(
+                        "Failed to search namespace %s: %s", search_namespace, str(e)
+                    )
+                    continue
+
+            # Sort by score and limit results
+            all_results.sort(key=lambda x: x.get("score"), reverse=True)
+            final_results = all_results[:limit]
```
bergjaak: we shouldnt need this [:limit] right because we provide the top_k argument, right?

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: > is there a way, like say with a doc string, to suggest that search searches over all of these namespaces

yup, this should be doable

> then we can also explain to the writer that if they want to search all the namespaces, they can just give us a prefix that is common to all of their namespaces

this is tough as `memory_client.retrieve_memories` does not support prefix search and doing a get memory would require additional permissions from the customer + processing on responses

bergjaak: oh retrieve_memories does support prefix

src/crewai/memory/storage/bedrock_agentcore_storage.py
bergjaak: discussed offline, needed because we filter by score

docs/en/concepts/memory.mdx
```diff
@@ -837,6 +837,72 @@ crew = Crew(
...
+    namespaces=[
+        "/preferences/user-123",
+        "/facts/user-123/session-456",
+        "/summaries/user-123/session-456"
+    ]
+)
+
+# Using namespace template resolution utility (optional)
+resolved_namespaces = [
+    BedrockAgentCoreStorage.resolve_namespace_template(
```
bergjaak: i like this! Very self-documenting

pyproject.toml
bergjaak: I think we actually need boto3>=1.39.7 (I believe that is when bedrock-agentcore was added to boto3)

src/crewai/events/types/memory_events.py
bergjaak: are we sure though? is there any way we can check? I think it's generally frowned upon to use Any (it defeats the purpose of using types in the first place, which is to find bugs during static analysis instead of at runtime)

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,418 @@
...
+                f"Must be one of: {', '.join(supported_types)}"
+            )
+
+    def _initialize_memory_client(self) -> None:
+        """Initialize the Bedrock AgentCore Memory client."""
+        try:
+            import boto3
+        except ImportError:
+            raise ModuleNotFoundError(
+                "boto3 is required for Bedrock AgentCore. Use `pip install crewai[agentcore]` to install the required dependencies."
```
bergjaak: I wonder if we should include this message or not. What if the user is using `uv` or another dependency management system? (I am actually not sure, because I am not super familiar with Python development [I did use Python a lot for Leetcode lol])

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,418 @@
...
+
+        Returns:
+            List of memory items with context, sorted by relevance score
+        """
+        if self.bedrock_agentcore_memory_client is None:
+            raise RuntimeError("AgentCore memory client not initialized")
+
+        # Validate limit parameter
+        if limit >= 100:
+            raise ValueError("Limit must be less than 100")
```
bergjaak: in general, I wonder if it's best to not do this client side validation because the service and boto3 already does this validation. So I think it would make sense to remove this (and other validation that we can expect the service to handle, which should be whatever is listed in the API documentation)

src/crewai/memory/storage/bedrock_agentcore_storage.py
bergjaak: yeah i think we shouldn't catch exceptions

bergjaak: ideally we wouldn't have any exceptions

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,418 @@
...
+                        text = content.get("text", "")
+
+                        # Get score if available
+                        score = long_term_memory.get("score", 0.0)
+
+                        # Apply score threshold filter
+                        if score < score_threshold:
+                            continue
+
+                        long_term_memory_results.append(
```
bergjaak: does this `long_term_memory_results` need to adhere to some shape? is that why we're making this dictionary? (just want to confirm, I think that makes sense)

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,418 @@
...
+
+            logger.info(
+                "Created event in AgentCore memory: %s", response["event"]["eventId"]
+            )
+
+        except Exception as e:
+            logger.error("Error saving to AgentCore: %s", str(e))
+            raise
+
+    def search(
```
bergjaak, line 309: would it be good to break this function into 2 separate private functions that this function invokes? Namely, _search_long_term and _search_short_term? I think that would be a nice separation of concern

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,418 @@
...
+        self, query: str, limit: int = 3, score_threshold: float = 0.35
+    ) -> List[Dict[str, Any]]:
```
bergjaak: would it make sense for this return type to be a type that we define (instead of an untyped dict)? So that the fields are known?  I think it would be nice if we returns a List of classes, like List[MemoryContent] or something like that, where MemoryContent is just the fields you put into the dictionary that you're constructing

samples/bedrock-agentcore/memory_demo.ipynb
```diff
@@ -0,0 +1,599 @@
+{
+ "cells": [
+  {
```
bergjaak, line 3: I think with the demo, we should try  to illustrate creating a long-term memory, then starting a new session, and then using that long-term memory. Does that make sense? because the current demo reads like we're just continuing a conversation using regular conversation history, which I think doesn't demonstrate the long-term memory

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: I can modify this message to say "boto3 is required for Bedrock AgentCore" - would that be better?

I can say install `crewai[agentcore]` without specifying uv / pip etc.?

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: Makes sense, I'll remove this check

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: got it, can remove

src/crewai/memory/storage/bedrock_agentcore_storage.py
bergjaak: >I can say install crewai[agentcore] without specifying uv / pip etc.?

I think that's a great idea !

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: I think untyped is Dict is fine because:
- this is the required return type specified by the parent class
- this follows status quo for other storages in crew ai
- crew ai strigifies and dumps the memory into agent context so doesn't matter what type is returned
- we'll have to write mappers that convert STM response and LTM response to  `MemoryContent` type which might be an overkill

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: yup that's correct (it has to be a list of dicts)

src/crewai/memory/storage/bedrock_agentcore_storage.py
bergjaak: makes sense if other memory things do the same

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -201,9 +207,157 @@ def save(self, value: Any, metadata: Dict[str, Any]) -> None:
...
+                    self.bedrock_agentcore_memory_client.retrieve_memory_records(  # type: ignore
+                        memoryId=str(self.config.memory_id),
+                        namespace=search_namespace,
+                        searchCriteria={"searchQuery": query, "topK": limit},
+                    )
+                )
+            except Exception as e:
+                logger.warning(f"Error searching namespace {search_namespace}: {e}")
+                # Continue with next namespace
+                continue
```
bergjaak: I think we should fail if this happens

pyproject.toml
```diff
@@ -40,6 +40,8 @@ dependencies = [
     "blinker>=1.9.0",
     "json5>=0.10.0",
     "portalocker==2.7.0",
+    "crewai-tools>=0.62.1",
+    "boto3>=1.40.36",
```
3coins: Remove this as a direct dependency.

pyproject.toml
```diff
@@ -40,6 +40,8 @@ dependencies = [
     "blinker>=1.9.0",
     "json5>=0.10.0",
     "portalocker==2.7.0",
+    "crewai-tools>=0.62.1",
```
3coins: Do we need to add this?

samples/bedrock-agentcore/memory_demo.ipynb
gowri100: that's what the demo does right?
it creates a new memory resource and inserts into it
it demonstrates STM results - for questions that are asked within 1-2 minutes and LTM results for questions asked after 1-2 minutes (once extraction completes)

bergjaak: I think that's fair. But usually the value prop of memory integrations is: You do 1 session. Go away for awhile. Come back, start a new session -- LTM remebers something about you even though it is a blank/new session! Wow! Cool!

So like, if it were possible to "start a new crew" and have it remember something about the user, then that conveys the value add more clearly, imo.

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+    """Configuration for AgentCore memory storage."""
+
+    # Required fields
+    memory_id: str = Field(
+        ..., description="AWS Bedrock AgentCore Memory ID (required)"
+    )
+    actor_id: str = Field(..., description="Actor ID for the memory session (required)")
+    session_id: str = Field(
+        ..., description="Session ID for the memory session (required)"
+    )
```
3coins, line 24: Why is this required? For example, I can only have semantic memory configured, which doesn't require this.

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+    # Optional fields with defaults
+    region_name: str = Field(
+        default="us-east-1", description="AWS region name (default: 'us-east-1')"
+    )
+
+    # Namespace configuration
+    namespaces: list[str] = Field(
+        default_factory=list,
+        description="List of namespaces to be searched (default: empty list)",
+    )
```
3coins, line 35: Should this be a `namespace` instead of a list and should be required?
https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/retrieve-memory-records.html

Do we also need a top_k?

src/crewai/memory/storage/bedrock_agentcore_storage.py
lil-frankie: +1, I think I had a similar comment in a previous revision

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -201,9 +207,157 @@ def save(self, value: Any, metadata: Dict[str, Any]) -> None:
...
+            for long_term_memory in long_term_memories:
+                # Extract text content from memory record
+                content = long_term_memory.get("content", {})
+                text = content.get("text", "")
+
+                # Get score if available
+                score = long_term_memory.get("score", 0.0)
+
+                # Apply score threshold filter
+                if score < 0:
```
lil-frankie: Can a score ever be negative? Shouldn't the 0 be changed to the threshold value instead?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+    class Config:
+        """Pydantic configuration."""
+
+        validate_assignment = True
+        extra = "forbid"  # Prevent extra fields
+
+
+class BedrockAgentCoreStorage(Storage):
+    """
+    Enhanced AWS Bedrock AgentCore Memory storage implementation for CrewAI.
```
3coins:
```suggestion
    AWS Bedrock AgentCore Memory storage implementation for CrewAI.
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+
+    This implementation provides sophisticated integration with AWS Bedrock AgentCore Memory,
+    supporting advanced features like user preference learning, semantic fact extraction,
+    and proper conversation structure mapping.
```
3coins:
```suggestion
    This implementation provides integration with AWS Bedrock AgentCore Memory,
    supporting long term and short term memory. For more details, see the documentation:
    https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/memory.html
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+            self.memory_type,
+            self.config.namespaces,
+        )
+
+    def _validate_type(self, type: str) -> None:
+        """Validate the memory type."""
+        supported_types = {"external"}
+        if type not in supported_types:
+            raise ValueError(
+                f"Invalid type '{type}' for AgentCoreStorage. "
```
3coins:
```suggestion
                f"Invalid type '{type}' for BedrockAgentCoreStorage. "
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+    supporting advanced features like user preference learning, semantic fact extraction,
+    and proper conversation structure mapping.
+    """
+
+    def __init__(self, type: str, config: BedrockAgentCoreConfig):
+        super().__init__()
+
+        self._validate_type(type)
+        self.memory_type = type
+        self.config = config
```
3coins, line 86: Even though we require config of class `BedrockAgentCoreConfig`, users could pass in a dict that conforms with the overall shape of the config, which will not go through any of the validations you have inside the class. Should we check if a dict is passed, and convert that to the class?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # Initialize boto3 clients
+        self.bedrock_agentcore_memory_client: Any | None = None
+
+        # Initialize memory clients
+        self._initialize_memory_client()
```
3coins: Do we need to initialize the client, since the next statement does it anyway?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+        try:
+            import boto3
+        except ImportError as e:
+            raise ModuleNotFoundError(
+                "boto3 is required for Bedrock AgentCore. Install `crewai[agentcore]` for the required dependencies."
+            ) from e
+
+        try:
+            # Initialize boto3 clients directly as used by the MemoryClient
+            bedrock_agentcore_memory_client_config = Config(
```
3coins:
```suggestion
            config = Config(
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+                user_agent_extra="x-client-framework:crew_ai"
+            )
+            self.bedrock_agentcore_memory_client = boto3.client(
```
3coins:
```suggestion
            self.memory_client = boto3.client(
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+        Save memory item to AWS Bedrock AgentCore.
+
+        Args:
+            value: The content to save (task output, conversation, etc.) (required)
+            metadata: Additional metadata for the memory item (default: None)
+        """
+
+        if self.bedrock_agentcore_memory_client is None:
+            raise RuntimeError("AgentCore memory client not initialized")
```
3coins: Is this even possible, since the class won't be initialized if the client is not.

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+
+        payload = self._convert_to_event_payload(value, metadata)
+
+        try:
+            from datetime import datetime
```
3coins: Move this to the top.

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+
+            # Use the boto3 client create_event method directly
+            response = self.bedrock_agentcore_memory_client.create_event(
```
3coins:
```suggestion
            response = self.memory_client.create_event(
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+                memoryId=str(self.config.memory_id),
+                actorId=str(self.config.actor_id),
+                sessionId=str(self.config.session_id),
```
3coins: Are the str cast required?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+        long_term_memory_results = []
+
+        # Search each namespace
+        for search_namespace in namespaces_to_search:
+            logger.debug("Searching namespace: %s", search_namespace)
+
+            try:
+                # Use the boto3 client retrieve_memory_records method directly
+                long_term_memory_response = (
+                    self.bedrock_agentcore_memory_client.retrieve_memory_records(  # type: ignore
```
3coins:
```suggestion
                    self.memory_client.retrieve_memory_records(  # type: ignore
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+                        memoryId=str(self.config.memory_id),
```
3coins: Is str cast required?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+
+        if max_results <= 0:
+            return short_term_results
+
+        logger.info(
+            "Searching short-term memory for up to %d results",
+            max_results,
+        )
+
+        short_term_memory = self.bedrock_agentcore_memory_client.list_events(  # type: ignore
```
3coins:
```suggestion
        short_term_memory = self.memory_client.list_events(  # type: ignore
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+        Args:
+            query: Search query (required)
+            limit: Maximum number of results (default: 3, max: 100)
+            score_threshold: Minimum relevance score (default: 0.35)
+
+        Returns:
+            List of memory items with context, sorted by relevance score
+        """
+        if self.bedrock_agentcore_memory_client is None:
+            raise RuntimeError("AgentCore memory client not initialized")
```
3coins: Is this check required?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+
+        logger.info(
+            "Retrieved %d results from short-term memory",
+            len(short_term_results),
+        )
+
+        return short_term_results
+
+    def search(
+        self, query: str, limit: int = 3, score_threshold: float = 0.35
```
3coins: Would be important to understand how the limit is set here? Is it CrewAI controlled, or is it something that can be configured by the user?
Also, what happens if this function returns more than the limit?

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+            len(final_results),
+        )
+
+        return final_results
+
+    def _convert_to_event_payload(
+        self, value: Any, metadata: dict[str, Any]
+    ) -> list[dict[str, Any]]:
+        """
+        Convert CrewAI content to enhanced event payload that utilizes metadata when available.
```
3coins:
```suggestion
        Convert CrewAI content to event payload that utilizes metadata when available.
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,472 @@
...
+                config=bedrock_agentcore_memory_client_config,
+            )
+
+        except Exception as e:
+            raise ValueError(
+                f"Failed to initialize Bedrock AgentCore clients: {e!s}"
+            ) from e
+
+    @staticmethod
+    def resolve_namespace_template(
```
3coins: Not sure how useful is this. It adds an additional step in the process that makes me question this, plus needs the user to repeat inputs (actor_id etc.). 
If we really want to support namespace templates, why can't we do this explicitly from the top level config that already has all the inputs?

samples/bedrock-agentcore/memory_demo.ipynb
```diff
@@ -0,0 +1,599 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8e4cef33",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "bedrock_control_client = boto3.client(\"bedrock-agentcore-control\", region_name=REGION)\n",
+    "\n",
+    "\n",
+    "def create_memory_and_wait(\n",
```
jgordley, line 136: i think the creation of the memory resource itself is not the main focus of the notebook, maybe a markdown cell pointing to the docs on creating a memory resource would suffice? This works either way, just a thought.

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: we still need session_id for create_event calls

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: > Should this be a namespace instead of a list and should be required

the search method for crew ai searches through all namespaces (each namespace is a string)

> Do we also need a top_k?

the top_k is managed by the search method (which takes in limit as a parameter) - it is not managed by the user

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: > Would be important to understand how the limit is set here? Is it CrewAI controlled, or is it something that can be configured by the user?

```
def search(
        self, query: str, limit: int = 3, score_threshold: float = 0.35
    ) 
```
limit is an over-ridden parameter of the search method. Crew AI's invocation loop uses a value of 5 (updated from 3 to 5 last week)

the developer can choose the call the search method on their own as well and use a limit of their preference. 

> Also, what happens if this function returns more than the limit?
This is not possible as: 
- limit is passed as the topK parameter to retrive_memories 
- limit - len(LTM) is passed as maxResults to list_events for STM

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: The next statement just logs the client - not sure if I'm missing anything: 

```
logger.info(
            "BedrockAgentCoreStorage initialized: memory_type=%s, namespaces=%s",
            self.memory_type,
            self.config.namespaces,
        )
```

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: sure i can add this

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: So namespace would always be a template and we would initialize it internally?

src/crewai/memory/storage/bedrock_agentcore_storage.py
gowri100: you're right, I'll remove this check

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,498 @@
...
+                        {
+                            "id": event.get("eventId"),
+                            "content": text,
+                            "metadata": {
+                                "created_at": event.get("eventTimestamp"),
+                                "role": role,
+                            },
+                            "score": 1.0,  # Default score for short-term memory
+                        }
+                    )
```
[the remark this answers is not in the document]
gowri100: This is alright as it follows the expected format for STM search

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,498 @@
...
+
+            long_term_memories = long_term_memory_response.get(
+                "memoryRecordSummaries", []
+            )
+
+            logger.info(
+                "Retrieved %d memories from namespace: %s",
+                len(long_term_memories),
+                search_namespace,
+            )
```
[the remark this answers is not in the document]
gowri100: LTM is ordered by score (@bergjaak to confirm)

frankiekim5 (approved):

bergjaak (approved):

src/crewai/memory/storage/bedrock_agentcore_storage.py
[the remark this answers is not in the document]
bergjaak: yes

src/crewai/memory/storage/bedrock_agentcore_storage.py
bergjaak: this was implemented

src/crewai/memory/storage/bedrock_agentcore_storage.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def __init__(self, type: str, config: BedrockAgentCoreConfig):
+        super().__init__()
+
+        self._validate_type(type)
+        self.memory_type = type
+        self.config = config
+
+        # Initialize boto3 clients
+        self.memory_client: Any | None = None
```
3coins, line 89: nit: Wouldn't `Any` also qualify `None` as a value, so an explicit `Any | None` might not be required. I am not sure if this will pass linting test as-is, but would know more after CI runs.

src/crewai/memory/storage/bedrock_agentcore_storage.py
3coins: Are we going to remove this?

src/crewai/memory/storage/bedrock_agentcore_storage.py
3coins: Thanks, sounds good!

src/crewai/memory/storage/bedrock_agentcore_storage.py
3coins: This is not blocking for the PR, but would be nice to add in a follow up PR.

3coins (commented):
@GowriH-1 
Left some minor comments, which are not blocking to merge the PR. Only one suggestion, I think we should remove the `resolve_namespace_template` use externally, and handle templates internally in initialization, as all values needed to resolve templates are present in inputs (actor_id, session_id etc.). Removing this later will be a breaking change.

Also, not sure if just hiding the comments generated by cursor bot are enough to resolve them. You will also need to handle the errors generated by CI once the CrewAI maintainers kicks off the CI.

3coins:
@lucasgomide @lorenzejay 
Can you help kick off the CI, so @GowriH-1 can resolve the errors along with the rest of the updates. Also, there are some comments by cursor bot, and @GowriH-1  have added clarifying comments, is there anything further required before approval and merge?

bergjaak (approved):

3coins (approved):
LGTM!

PR #3477, opened 2025-09-07
Removed duplication of tool calling

### LINKED ISSUE #3462: [BUG] Tool invocation occurs twice in version 0.177.0
### Description

After upgrading from version 0.165.1 to 0.177.0, I observed that tools are being invoked exactly twice on every call initiated by the LLM. In the earlier version (0.165.1), the same architecture executed tools correctly with a single invocation.

This issue introduces redundant executions, leading to duplicate outputs and unnecessary latency. The duplicate calls occur within a few milliseconds of each other, which strongly suggests an internal execution or callback duplication in the updated release.

### Steps to Reproduce

Steps to Reproduce:

1. Set up a multi-crew architecture with agents configured to call external tools.

2. Register at least 4 tools in the environment.

3. Trigger tool execution through an agent.

4. Observe that the tool is invoked twice in quick succession (within milliseconds).

### Expected behavior

Each tool call from the LLM should result in exactly one execution.

In version 0.177.0, every tool call is executed twice.

In version 0.165.0, the same architecture executes correctly with a single invocation.

### Screenshots/Code snippets

Can't share project screenshot or code

### Operating System

Windows 11

### Python Version

3.11

### crewAI Version

Issue in 0.177.0, currently using 0.165.1

### crewAI Tools Version

0.62.3

### Virtual Environment

Venv

### Evidence

Can't share project logs or screenshots

### Possible Solution

None

### Additional context

None

Vidit-Ostwal:
Fixes #3462

### DIFF
--- a/.github/workflows/type-checker.yml
+++ b/.github/workflows/type-checker.yml
@@ -32,7 +32,7 @@ jobs:
         run: uv python install ${{ matrix.python-version }}
 
       - name: Install dependencies
-        run: uv sync --dev --no-install-project
+        run: uv sync --dev --all-extras --no-install-project
 
       - name: Get changed Python files
         id: changed-files
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,14 +1,18 @@
 repos:
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.12.11
+  - repo: local
     hooks:
       - id: ruff
-        args: ["--config", "pyproject.toml"]
+        name: ruff
+        entry: uv run ruff check
+        language: system
+        types: [python]
       - id: ruff-format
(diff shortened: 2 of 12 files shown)


src/crewai/tools/structured_tool.py
```diff
@@ -234,56 +232,56 @@ async def ainvoke(
...
-            return result
+            return asyncio.run(self.func(**parsed_args, **kwargs))
 
         try:
             result = self.func(**parsed_args, **kwargs)
         except Exception:
             raise
 
-        result = self.func(**parsed_args, **kwargs)
-
```
Vidit-Ostwal, line 286: This is main change. Others are just linter.

Vidit-Ostwal:
Fixes #3489

Issue #3454, opened 2025-09-05
[BUG] LLM Returns Empty Response When Input Is Too Long


phuihock:
### Description

When using `Qwen/Qwen3-235B-A22B-Instruct-2507` by DeepInfra , I hit this error:

"Invalid response from LLM call - None or empty."

The root cause is that LLMs hosted on DeepInfra respond with an empty message when the input messages exceed the model’s context window.

Agent executor doesn’t proactively check message length before calling the model—it only handles it after an exception is thrown
https://github.com/crewAIInc/crewAI/blob/610c1f70c0cd19561551f5ba61b236eb9b552eea/src/crewai/agents/crew_agent_executor.py#L240-L246

This is problematic because different providers handle overflows differently—some return structured errors, others fail silently.

Related to #2990 

### Steps to Reproduce

1. Create an agent with llm=`Qwen/Qwen3-235B-A22B-Instruct-2507` by DeepInfra
2. Give it a very long input of >40K characters
2. Kickoff the crew

### Expected behavior

It should summarize the messages prior to calling the LLM

### Screenshots/Code snippets

None

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

0.177.0

### crewAI Tools Version

0.69.0

### Virtual Environment

Venv

### Evidence

```
  File "/home/phuihock/github/fxml/.venv/lib/python3.11/site-packages/crewai/agents/crew_agent_executor.py", line 153, in _invoke_loop
    answer = get_llm_response(
             ^^^^^^^^^^^^^^^^^
  File "/home/phuihock/github/fxml/.venv/lib/python3.11/site-packages/crewai/utilities/agent_utils.py", line 167, in get_llm_response
    raise ValueError("Invalid response from LLM call - None or empty.")
```

### Possible Solution

I suggest handling message length earlier in the flow (call `handle_context_length` right before `get_llm_response` https://github.com/crewAIInc/crewAI/blob/610c1f70c0cd19561551f5ba61b236eb9b552eea/src/crewai/agents/crew_agent_executor.py#L187-L195

### Additional context

None

redvelvets:
Hi,
This issue seems quite similar to https://github.com/crewAIInc/crewAI/issues/2885

Vidit-Ostwal:
Hi @phuihock,
Mind giving this feature branch a try:

just do a `pip install git+https://github.com/Vidit-Ostwal/crewAI.git@Invalid_response_from_LLM`
[...]

PR #3488, opened 2025-09-09
Fix:  (#3454) LLM returns empty response when input is too long

ayushmangpta:
# Pre-handle context length before LLM call to avoid empty responses

Summary
- Proactively handle context window overflow right before invoking the LLM to prevent providers that fail silently (returning empty responses) from causing ValueError downstream.

Why
- Some providers (e.g., DeepInfra with Qwen/Qwen3-235B-A22B-Instruct-2507) return an empty message when input exceeds the model’s context window.
- Previously we only handled context overflow after an exception was thrown. If the provider fails silently, no exception is raised, and get_llm_response raises "Invalid response from LLM call - None or empty."
- This change adds a pre-call check so we summarize or abort before calling the LLM in an overflow scenario.

What changed
- File: crewai/agents/crew_agent_executor.py
  - In CrewAgentExecutor._invoke_loop, before get_llm_response:
    - Compute the total content length of messages and compare with llm.get_context_window_size().
    - If overflow is likely, call handle_context_length (summarize when respect_context_window=True; otherwise raise).
    - Keep the existing exception-based overflow handling as fallback.

Code snippet added
```python
# ...existing code...

enforce_rpm_limit(self.request_within_rpm_limit)

# Pre-handle context length just before calling the LLM
try:
    window_size = self.llm.get_context_window_size()
except Exception:
    window_size = None
if window_size is not None:
    total_len = sum(len(m.get("content", "")) for m in self.messages)
    if total_len >= window_size:
        handle_context_length(
            printer=self._printer,
            messages=self.messages,
            llm=self.llm,
            callbacks=self.callbacks,
            i18n=self._i18n,
        )

answer = get_llm_response(
    llm=self.llm,
    messages=self.messages,
    callbacks=self.callbacks,
    printer=self._printer,
    from_task=self.task,
)
formatted_answer = process_llm_response(answer, self.use_stop_words)

# ...existing code...
```

Behavior
- If respect_context_window is True: messages are summarized before the LLM call when the total content length meets/exceeds the model window size.
- If respect_context_window is False: a clear error is raised earlier, before calling the LLM, maintaining previous behavior while avoiding a silent empty response.
- Existing exception-based handling remains for providers that properly raise overflow errors.

Risks / Trade-offs
- Uses character length as a heuristic, not precise token counting. May summarize slightly earlier/later than ideal. This mirrors existing summarize_messages behavior and remains a practical safeguard.

Note: This is my first PR—feedback appreciated.

### DIFF
--- a/src/crewai/agents/crew_agent_executor.py
+++ b/src/crewai/agents/crew_agent_executor.py
@@ -186,6 +186,23 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
 
                 enforce_rpm_limit(self.request_within_rpm_limit)
 
+                # Pre-handle context length just before calling the LLM
+                try:
+                    window_size = self.llm.get_context_window_size()
+                except Exception:
+                    window_size = None
+                if window_size is not None:
+                    total_len = sum(len(m.get("content", "")) for m in self.messages)
+                    if total_len >= window_size:
+                        handle_context_length(
+                            respect_context_window=self.respect_context_window,
+                            printer=self._printer,
+                            messages=self.messages,
+                            llm=self.llm,
+                            callbacks=self.callbacks,
+                            i18n=self._i18n,
+                        )
+
                 answer = get_llm_response(
                     llm=self.llm,
                     messages=self.messages,


src/crewai/agents/crew_agent_executor.py
BELKHIR, line 1: Hi, 
I don't think get_context_window_size will return the true context size for llm backed by ollama.

Khalid,

ayushmangpta: hi, thank you for replying do you have any inputs to make it work for ollama as well?

BELKHIR: By default Ollama limit the context size  to 4096 tokens (8192 for gpt-oss).
This behavior can be changed in different ways listed in this faq :
https://github.com/ollama/ollama/blob/92b96d54efd6b49322b7cf046f9a0dc16b00cd0a/docs/faq.md?plain=1#L33

Additionally, models from the Ollama library can be customized via a Modelfile to save new params (num_ctx here) in a new model.

Using the Crewai LLM one can also change num_ctx :
max_context = 16384
gpt_ollama_llm =  LLM(
        model = f"ollama/llama3.2", 
        base_url="http://localhost:11434",
        num_ctx=max_context
        )

So there is no way to know apriori which is the max context window of a model.

But all the above mentioned solutions ends with adding a parameter num_ctx= 'the new value'

The only lead I found is by using Ollama Python library to call the 'show' command on the model to seek if there is a parameter named num_ctx and get it's value.

>>>from ollama import  show
# show on unchanged model with ollam default
>>>show('gpt-oss')['parameters']
'temperature                    1'

# show on modified model
>>>show('gpt-oss-65536ctx')['parameters']
'num_ctx                        65536\ntemperature                    1'

One have to extract the info from this unstructured result.

In the case of a user setting num_ctx in the LLM call, the above solution is useless, but one can use the value given to num_ctx to leverage a guardrail Function to validate task output before proceeding to the next one ?

Regards

PR #3495, opened 2025-09-10
Lorenze/fix tool call twice

lorenzejay:


src/crewai/tools/structured_tool.py
```diff
@@ -282,8 +282,6 @@ def invoke(
         except Exception:
             raise
 
-        result = self.func(**parsed_args, **kwargs)
```
gabemilani, line 285: Should we remove the `try-except` block instead of this? It's just capturing and raising the exception, so not useful at all.

lorenzejay: good idea - we are raising an actual error above it for if `has_reached_max_usage_count`

gabemilani (approved):
Overall looking good to me, just need to lint the changes and ensure the tests are passing.

You can run `pre-commit install` to ensure the linter (ruff) always run before commits.

lucasgomide (approved):

greysonlalonde (approved):

PR #3496, opened 2025-09-10
feat: add test duration caching for optimized pytest-split

greysonlalonde:
## Summary
- add workflow to generate and cache test durations per python version
- update tests workflow to use cached durations for better test distribution
- durations only update when test files change on main branch

### DIFF
--- a/src/crewai/agents/agent_adapters/langgraph/langgraph_adapter.py
+++ b/src/crewai/agents/agent_adapters/langgraph/langgraph_adapter.py
@@ -1,47 +1,56 @@
-from typing import Any, Dict, List, Optional
+"""LangGraph agent adapter for CrewAI integration.
 
-from pydantic import Field, PrivateAttr
+This module contains the LangGraphAgentAdapter class that integrates LangGraph ReAct agents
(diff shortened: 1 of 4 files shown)


.github/workflows/update-test-durations.yml
```diff
@@ -0,0 +1,71 @@
...
+          python-version: ${{ matrix.python-version }}
+          enable-cache: false
+
+      - name: Install the project
+        run: uv sync --all-groups --all-extras
+
+      - name: Run all tests and store durations
+        run: |
+          PYTHON_VERSION_SAFE=$(echo "${{ matrix.python-version }}" | tr '.' '_')
+          uv run pytest --store-durations --durations-path=.test_durations_py${PYTHON_VERSION_SAFE} -n auto
```
lucasgomide, line 53: Are we running tests twice?

greysonlalonde: Yeah, but because we need to store the durations. This should only happen on main when tests change

lucasgomide: got it

PR #3502, opened 2025-09-11
feat: add thread-safe platform context management

lucasgomide:


src/crewai/context.py
```diff
@@ -0,0 +1,25 @@
+import os
+import contextvars
+from typing import Optional
+from contextlib import contextmanager
+
+_platform_integration_token: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
```
greysonlalonde, line 6: Might consider defaulting to the env var here contingent on behavior

lucasgomide: I will keep as fallback value inside `get_platform_integration_token`... I'm a bit worried about following your suggestion because this will be default at module import time

PR #3504, opened 2025-09-11
feat: add `apps` & `actions` attributes to Agent

lucasgomide:
Depends on tools to be merged

### DIFF
--- a/src/crewai/agent.py
+++ b/src/crewai/agent.py
@@ -601,7 +601,7 @@ class Agent(BaseAgent):
 
     def get_platform_tools(self, apps: list[PlatformAppOrAction]) -> list[BaseTool]:
         try:
-            from crewai_tools import CrewaiPlatformTools # type: ignore[import-untyped]
+            from crewai_tools import CrewaiPlatformTools  # type: ignore[import-untyped]
 
             return CrewaiPlatformTools(apps=apps)
         except Exception as e:
--- a/tests/agents/agent_builder/test_base_agent.py
+++ b/tests/agents/agent_builder/test_base_agent.py
@@ -1,5 +1,5 @@
 import hashlib
-from typing import Any, List, Optional
+from typing import Any
 
 from pydantic import BaseModel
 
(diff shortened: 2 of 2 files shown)


src/crewai/agent.py
```diff
@@ -279,10 +274,10 @@ def execute_task(
             except Exception as e:
                 if hasattr(self, "_logger"):
                     self._logger.log(
-                        "error", f"Error during reasoning process: {str(e)}"
+                        "error", f"Error during reasoning process: {e!s}"
                     )
                 else:
-                    print(f"Error during reasoning process: {str(e)}")
+                    print(f"Error during reasoning process: {e!s}")
```
lorenzejay: can we move this as a logger? while youre here

tests/agents/test_agent.py
```diff
@@ -2503,3 +2505,108 @@ def test_agent_from_repository_without_org_set(
...
         style="yellow",
     )
+
+def test_agent_actions_basic_functionality():
+    # Test with valid actions
+    agent = Agent(
+        role="Platform Agent",
+        goal="Use platform tools",
+        backstory="Platform specialist",
+        actions=["gmail/create_task", "slack/update_status", "customer_support/send_notification"]
```
lorenzejay: i thought we are consolidating it to apps. but two ways of setting based on one attribute:

`Agent(apps=["gmail"])`
and
`Agent(apps=["gmail/action"])`

tests/agents/test_agent.py
```diff
@@ -2506,26 +2506,52 @@ def test_agent_from_repository_without_org_set(
...
 
-def test_agent_actions_basic_functionality():
-    # Test with valid actions
+def test_agent_apps_consolidated_functionality():
     agent = Agent(
         role="Platform Agent",
         goal="Use platform tools",
         backstory="Platform specialist",
-        actions=["gmail/create_task", "slack/update_status", "customer_support/send_notification"]
+        apps=["gmail/create_task", "slack/update_status", "hubspot"]
```
lorenzejay, line 2532: lfg

src/crewai/agents/agent_builder/base_agent.py
```diff
@@ -25,11 +26,31 @@
 from crewai.tools.base_tool import BaseTool, Tool
 from crewai.utilities import I18N, Logger, RPMController
 from crewai.utilities.config import process_config
-from crewai.utilities.converter import Converter
 from crewai.utilities.string_utils import interpolate_only
 
 T = TypeVar("T", bound="BaseAgent")
 
+PlatformApp = Literal[
```
greysonlalonde, line 33: On this, do you think we can define literals with app and app-specific actions (`Literal["gmail", "gmail/..."]` and split into a union, like `GmailApp | GitHubApp | ...`? I know it would be lengthy, but would provide really nice autocomplete and help with organization of apps. Maybe in a separate `types.py`

greysonlalonde: Or enums, but I'm partial to string literals 😜

lucasgomide: Honestly, I don't want to support all available actions since the list could get huge. For example, we already have more than 120 actions, and I'm concerned about the high maintenance cost..

lucasgomide: Otherwise, I liked the idea to serve the code as documentation.

PR #3506, opened 2025-09-11
fix: resolve test duration cache issues in CI workflows

greysonlalonde:
## Summary
- Fixed cache key mismatch between save and restore operations
- Added check to skip cached durations when test files have changed
- Prevents pytest-split from collecting 0 tests when cached durations don't match

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,18 +58,22 @@ jobs:
           PYTHON_VERSION_SAFE=$(echo "${{ matrix.python-version }}" | tr '.' '_')
           DURATION_FILE=".test_durations_py${PYTHON_VERSION_SAFE}"
           
-          # First check if duration file exists
-          if [ ! -f "$DURATION_FILE" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -49,17 +49,19 @@ jobs:
...
+          key: test-durations-py${{ matrix.python-version }}
 
       - name: Run tests (group ${{ matrix.group }} of 8)
         run: |
           PYTHON_VERSION_SAFE=$(echo "${{ matrix.python-version }}" | tr '.' '_')
           DURATION_FILE=".test_durations_py${PYTHON_VERSION_SAFE}"
           
-          if [ -f "$DURATION_FILE" ]; then
-            echo "Using cached test durations for optimal splitting"
+          # Check if any test files have changed compared to main
```
lucasgomide: I thought we should execute after changing tests 🤔 I got confused

greysonlalonde: Same, if no tests on main change, compare to branch, use cache, if non-main branch updates, don't use cache

Issue #3559, opened 2025-09-20
[BUG] Latest changes are breaking


Vidit-Ostwal:
### Description

Hey guys, I was just testing some random ideas using crewai. I think some recent changes have broken the previously running crews. I am on the main-branch from github from crewai.

This is while just simply invoking a simple crew setup.

Attaching one screenshot

<img width="1090" height="605" alt="Image" src="https://github.com/user-attachments/assets/dedf8689-bcc7-4446-94ce-afa91a6ba8d7" />

### Steps to Reproduce

-

### Expected behavior

-

### Screenshots/Code snippets

-

### Operating System

macOS Sonoma

### Python Version

3.11

### crewAI Version

Latest from Gtihub 

### crewAI Tools Version

-

### Virtual Environment

Venv

### Evidence

-

### Possible Solution

-

### Additional context

-

Vidit-Ostwal:
Tagging @joaomdmoura and @greysonlalonde to review/advise on this.

Vidit-Ostwal:
Okay, I think the exception handling is not configured to handle this case.

greysonlalonde:
Hey @Vidit-Ostwal can you close if resolved?

Vidit-Ostwal:
Hey @greysonlalonde, just checked this, working fine. 
Closing this and the PR rn.

PR #3563, opened 2025-09-21
Handle the exception handling for tracking correctly

Vidit-Ostwal:
Cloes #3559


Vidit-Ostwal:
Requesting @greysonlalonde, @lorenzejay, @lucasgomide for review.

PR #3571, opened 2025-09-21
Refactor API configuration for Azure provider

dibyasom:
There's been a recent change, in the SDK - and the given way of configuring azure hosted openAI embedding failed.

I have made the corrections basis what worked, and verified in pip package. The "config" key is incorrect, and raises fatal exception.


greysonlalonde:
Hi @dibyasom, thanks for flagging this. This is not expected behavior, it would be best to open an issue to track this

PR #3572, opened 2025-09-21
Add source to LLM Guardrail events

vinibrsl:
This commit adds the `source` attribute to LLM Guardrail event calls to identify the Lite Agent or Task that executed the guardrail.

### DIFF
--- a/src/crewai/task.py
+++ b/src/crewai/task.py
@@ -5,20 +5,14 @@ import logging
 import threading
 import uuid
 import warnings
+from collections.abc import Callable
 from concurrent.futures import Future
 from copy import copy
 from hashlib import md5
 from pathlib import Path
 from typing import (
     Any,
-    Callable,
     ClassVar,
-    Dict,
-    List,
-    Optional,
-    Set,
-    Tuple,
-    Type,
     Union,
     get_args,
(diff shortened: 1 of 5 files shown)


src/crewai/task.py
```diff
@@ -528,41 +525,6 @@ def _execute_core(
             crewai_event_bus.emit(self, TaskFailedEvent(error=str(e), task=self))
             raise e  # Re-raise the exception after emitting the event
 
-    def _process_guardrail(self, task_output: TaskOutput) -> GuardrailResult:
```
vinibrsl, line 531: I couldn't find references to this method, so I removed it.

src/crewai/utilities/guardrail.py
```diff
@@ -55,41 +56,35 @@ def from_tuple(cls, result: Tuple[bool, Union[Any, str]]) -> "GuardrailResult":
 
 
 def process_guardrail(
-    output: Any, guardrail: Callable, retry_count: int
+    output: Any, guardrail: Callable, retry_count: int, event_source: Any | None = None
```
vinibrsl, line 59: Main change is here. Most of other changes are linting.

PR #3579, opened 2025-09-22
fix: support nested config format for embedder configuration

greysonlalonde:
Fixes embedder configuration parsing to properly handle the documented nested format with 'config' key.


greysonlalonde:
addresses https://github.com/crewAIInc/crewAI/pull/3571

PR #3585, opened 2025-09-23
fix: add batch_size support to prevent embedder token limit errors

greysonlalonde:


src/crewai/knowledge/storage/knowledge_storage.py
```diff
@@ -35,12 +36,29 @@ def __init__(
         )
 
         if embedder:
-            embedding_function = get_embedding_function(embedder)
-            config = ChromaDBConfig(
-                embedding_function=cast(
-                    ChromaEmbeddingFunctionWrapper, embedding_function
+            # Cast to EmbedderConfig for type checking
+            embedder_typed = cast(EmbedderConfig, embedder) if embedder else None
```
vinibrsl: I believe we can remove this conditional as L38 is doing that.

```suggestion
            embedder_typed = cast(EmbedderConfig, embedder)
```

vinibrsl (approved):

vinibrsl (approved):

PR #3586, opened 2025-09-24
Lorenze/traces mark as failed

lorenzejay:


src/crewai/cli/plus_api.py
```diff
@@ -166,3 +166,13 @@ def finalize_ephemeral_trace_batch(
...
         )
+
+    def mark_trace_batch_as_failed(
+        self, trace_batch_id: str, error_message: str
+    ) -> requests.Response:
+        return self._make_request(
+            "PATCH",
+            f"{self.TRACING_RESOURCE}/batches/{trace_batch_id}",
+            json={"status": "failed", "failure_reason": error_message},
+            timeout=30,
```
joaomdmoura, line 177: should we do less timeout? maybe we start with 30?

joaomdmoura (approved):

PR #3601, opened 2025-09-26
Role should be a part when summarization

Vidit-Ostwal:
- Adding role to the message string before summarization


Vidit-Ostwal:
Requesting @greysonlalonde for review.

PR #3643, opened 2025-10-03
Enhance CLI with trigger listing and run commands

lucasgomide:
This PR introduces a set of commands to help users build automations locally. A few commands were added:

You can list all supported triggers

```bash
$ crewai triggers list
```

And ask run your automation with a trigger payload

```bash
$ crewai triggers run microsoft_outlook/email_received
```

https://github.com/user-attachments/assets/e68bb674-5aa9-46f9-b996-4f5d1e431dbe

### DIFF
--- a/lib/crewai/src/crewai/cli/triggers/main.py
+++ b/lib/crewai/src/crewai/cli/triggers/main.py
@@ -1,0 +1,123 @@
+import json
+import subprocess
+from typing import Any
+
+from rich.console import Console
+from rich.table import Table
+
+from crewai.cli.command import BaseCommand, PlusAPIMixin
+
+
+console = Console()
+
+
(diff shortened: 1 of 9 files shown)


lib/crewai/src/crewai/cli/triggers/main.py
```diff
@@ -0,0 +1,123 @@
...
+console = Console()
+
+
+class TriggersCommand(BaseCommand, PlusAPIMixin):
+    """
+    A class to handle trigger-related operations for CrewAI projects.
+    """
+
+    def __init__(self):
+        BaseCommand.__init__(self)
```
greysonlalonde, line 20: We gotta move away from using this pattern, it breaks the mro chain

greysonlalonde: future :)

PR #3671, opened 2025-10-08
feat: improve event bus thread safety and async support

greysonlalonde:


lib/crewai/src/crewai/events/event_bus.py
```diff
@@ -1,125 +1,455 @@
...
-    A singleton event bus that uses blinker signals for event handling.
-    Allows both internal (Flow/Crew) and external event handling.
+    try:
+        handler(source, event)
+        return None
+    except Exception as e:
+        return e
+
+
+class Signal(Signal_):
```
greysonlalonde, line 35: `blinker.Signal` is not thread safe, adding some minimal checks, still likely at risk for race conditions though

lib/crewai/src/crewai/events/event_bus.py
```diff
@@ -1,125 +1,455 @@
...
+            type[BaseEvent],
+            frozenset[Callable[[Any, BaseEvent], Coroutine[Any, Any, None]]],
+        ] = {}
+        self._sync_executor = ThreadPoolExecutor(
+            max_workers=10,
+            thread_name_prefix="CrewAISyncHandler",
+        )
+        self._console = ConsoleFormatter()
+
+        self._loop = asyncio.new_event_loop()
```
greysonlalonde, line 157: We could possibly delay this / make lazy, so no cost if no async handlers

lib/crewai/src/crewai/events/event_bus.py
```diff
@@ -1,125 +1,455 @@
...
+        self._loop_thread = threading.Thread(
+            target=self._run_loop,
+            name="CrewAIEventsLoop",
+            daemon=True,
```
greysonlalonde, line 161: This makes it so handlers that complete following the main program completion do not get lost

lib/crewai/src/crewai/events/utils/rw_lock.py
```diff
@@ -0,0 +1,81 @@
...
+This module provides a reader-writer lock implementation that allows multiple
+concurrent readers or a single exclusive writer.
+"""
+
+from collections.abc import Generator
+from contextlib import contextmanager
+from threading import Condition
+
+
+class RWLock:
```
greysonlalonde, line 12: This approach makes contention less likely in times of high emits when compared to usage of `threading.RLock`

lib/crewai/src/crewai/events/event_bus.py
```diff
@@ -1,125 +1,455 @@
...
+        self._rwlock = RWLock()
+        self._sync_handlers: dict[
+            type[BaseEvent], frozenset[Callable[[Any, BaseEvent], None]]
+        ] = {}
+        self._async_handlers: dict[
+            type[BaseEvent],
+            frozenset[Callable[[Any, BaseEvent], Coroutine[Any, Any, None]]],
+        ] = {}
+        self._sync_executor = ThreadPoolExecutor(
+            max_workers=10,
```
greysonlalonde, line 152: We can make this configurable / based on os resources, but this is likely fine for now

src/crewai/events/event_bus.py
```diff
@@ -1,125 +1,455 @@
...
+        _loop_thread: Background daemon thread running the event loop
+        _console: Console formatter for error output
     """
 
-    _instance = None
-    _lock = threading.Lock()
+    _instance: Self | None = None
+    _instance_lock: threading.RLock = threading.RLock()
+    _rwlock: RWLock
+    _sync_handlers: dict[type[BaseEvent], frozenset[Callable[[Any, BaseEvent], None]]]
```
greysonlalonde: We can also consider using  `queue.Queue`, `async.Queue` here - would require some restructuring

greysonlalonde: I may move to this on second thought

PR #3681, opened 2025-10-09
return message list for liteagent kickoff response

lorenzejay:
on lite agent kickoff return its message list


lorenzejay:
drafting for now, will add tests in a second

PR #3684, opened 2025-10-09
fix: correct cron schedule to run every 5 days at specific dates

greysonlalonde:


.github/workflows/build-uv-cache.yml
```diff
@@ -8,7 +8,7 @@ on:
       - "uv.lock"
       - "pyproject.toml"
   schedule:
-    - cron: "0 0 */5 * *"  # Run every 5 days at midnight UTC to prevent cache expiration
+    - cron: "0 0 1,6,11,16,21,26 * *"  # Run every 5 days at midnight UTC to prevent cache expiration
```
lucasgomide, line 11: I prefer `0 0 */5 * *"` since it's simpler and can execute on 31st (if exists)

PR #3689, opened 2025-10-10
fix: inject tool repository credentials in crewai run command

greysonlalonde:
closes eng-945

### DIFF
--- a/src/crewai/cli/run_crew.py
+++ b/src/crewai/cli/run_crew.py
@@ -1,10 +1,11 @@
+import os
 import subprocess
 from enum import Enum
 
 import click
 from packaging import version
 
-from crewai.cli.utils import read_toml
+from crewai.cli.utils import build_env_with_tool_repository_credentials, read_toml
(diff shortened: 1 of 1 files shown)


src/crewai/cli/run_crew.py
```diff
@@ -55,8 +56,22 @@ def execute_command(crew_type: CrewType) -> None:
...
+        sources = pyproject_data.get("tool", {}).get("uv", {}).get("sources", {})
+
+        for source_config in sources.values():
+            if isinstance(source_config, dict):
+                index = source_config.get("index")
+                if index:
+                    index_env = build_env_with_tool_repository_credentials(index)
+                    env.update(index_env)
+    except Exception:  # noqa: S110
+        pass
```
heitorado, line 71: seems that we have a repeating pattern here (check `src/crewai/cli/cli.py:47`, at `def uv`)... should we wrap this into its own util function?

greysonlalonde: yeah! i'll put up an additional pr

PR #3690, opened 2025-10-10
Fixing copy and adding NOT_SPECIFIED check in task.py

Vidit-Ostwal:
Following changes have been made.

1. copy method imported as shallow_copy, this helps to distinguish between the Class Task copy method and python in-built copy method.
2. For context in task, NOT_SPECIFIED case was missing, which has been added. (This will typically impact when .`kickoff_for_each` such methods are used where classes are copied, not having this condition changed the `context` variable to None, and hence no context is passed to next task when using this.)

### DIFF
--- a/src/crewai/__init__.py
+++ b/src/crewai/__init__.py
@@ -40,7 +40,7 @@ def _suppress_pydantic_deprecation_warnings() -> None:
 
 _suppress_pydantic_deprecation_warnings()
 
-__version__ = "0.203.0"
+__version__ = "0.203.1"
 _telemetry_submitted = False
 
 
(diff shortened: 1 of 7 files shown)


Vidit-Ostwal:
Resolves: https://github.com/crewAIInc/crewAI/issues/3691

lorenzejay:
Nice work. can you add some tests here to help catch this from happening again 🙏🏼  @Vidit-Ostwal

Vidit-Ostwal:
> Nice work. can you add some tests here to help catch this from happening again 🙏🏼 @Vidit-Ostwal

Yes sure

Vidit-Ostwal:
> Collaborator

@lorenzejay added.

lorenzejay:
@Vidit-Ostwal nice work

PR #3695, opened 2025-10-11
Fixed task configuration with None as context parameter

Vidit-Ostwal:
1. Currently when trying to configure a task with None as a context parameter via yaml, it's throws outs an error, this PR adds another check.


Vidit-Ostwal:
Resolves #3696

Vidit-Ostwal:
Resolves #3929

Vidit-Ostwal:
Closing this might not be relevant

PR #3707, opened 2025-10-14
feat: enhance AnthropicCompletion class with additional client parame…

lorenzejay:
…ters and tool handling

- Added support for client_params in the AnthropicCompletion class to allow for additional client configuration.
- Refactored client initialization to use a dedicated method for retrieving client parameters.
- Implemented a new method to handle tool use conversation flow, ensuring proper execution and response handling.
- Introduced comprehensive test cases to validate the functionality of the AnthropicCompletion class, including tool use scenarios and parameter handling.

### DIFF
--- a/lib/crewai-tools/pyproject.toml
+++ b/lib/crewai-tools/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "pytube>=15.0.0",
     "requests>=2.32.5",
     "docker>=7.1.0",
-    "crewai==1.0.0b1",
+    "crewai==1.0.0b2",
     "lancedb>=0.5.4",
     "tiktoken>=0.8.0",
     "beautifulsoup4>=4.13.4",
--- a/lib/crewai-tools/src/crewai_tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/__init__.py
@@ -291,4 +291,4 @@ __all__ = [
     "ZapierActionTools",
 ]
 
-__version__ = "1.0.0b1"
+__version__ = "1.0.0b2"
--- a/lib/crewai/pyproject.toml
+++ b/lib/crewai/pyproject.toml
@@ -48,7 +48,7 @@ Repository = "https://github.com/crewAIInc/crewAI"
 
 [project.optional-dependencies]
 tools = [
-    "crewai-tools==1.0.0b1",
+    "crewai-tools==1.0.0b2",
 ]
 embeddings = [
     "tiktoken~=0.8.0"
--- a/lib/crewai/src/crewai/__init__.py
+++ b/lib/crewai/src/crewai/__init__.py
@@ -40,7 +40,7 @@ def _suppress_pydantic_deprecation_warnings() -> None:
 
 _suppress_pydantic_deprecation_warnings()
 
-__version__ = "1.0.0b1"
+__version__ = "1.0.0b2"
 _telemetry_submitted = False
 
 
--- a/lib/devtools/src/crewai_devtools/__init__.py
+++ b/lib/devtools/src/crewai_devtools/__init__.py
@@ -1,3 +1,3 @@
 """CrewAI development tools."""
 
-__version__ = "1.0.0b1"
+__version__ = "1.0.0b2"


lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -385,6 +452,115 @@ def _handle_streaming_completion(
 
         return full_response
 
+    def _handle_tool_use_conversation(
+        self,
+        initial_response: Message
+        | Any,  # Can be Message or mock response from streaming
+        tool_uses: list[ToolUseBlock]
+        | list[Any],  # Can be ToolUseBlock or mock objects
```
greysonlalonde: Is it a heavy lift to create a protocol to use instead of `Any` here? maybe

```python3
class ToolUseProtocol(Protocol):
      id: str
      name: str
      input: dict[str, Any]
```

greysonlalonde: Same for `initial_response`

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -350,26 +389,54 @@ def _handle_streaming_completion(
...
 
-                # Execute tool
-                result = self._handle_tool_execution(
-                    function_name=function_name,
-                    function_args=function_args,
-                    available_functions=available_functions,
-                    from_task=from_task,
-                    from_agent=from_agent,
+                # Create a mock ToolUseBlock-like object
+                class MockToolUse:
```
greysonlalonde: Wondering why we need a mock here, can we refactor to a dataclass & move out of the class body?

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -183,20 +205,39 @@ def _prepare_completion_params(
...
-            name, description, parameters = safe_tool_conversion(tool, "Anthropic")
+            if "input_schema" in tool and "name" in tool and "description" in tool:
+                anthropic_tools.append(tool)
+                continue
+
+            try:
+                from crewai.llms.providers.utils.common import safe_tool_conversion
+
+                name, description, parameters = safe_tool_conversion(tool, "Anthropic")
+            except (ImportError, Exception):
```
greysonlalonde: Might be good to constrain down to `except (ImportError, KeyError, ValueError) as e:` so we can pin down any unexpected exceptions

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -385,6 +452,115 @@ def _handle_streaming_completion(
...
+
+        except Exception as e:
+            if is_context_length_exceeded(e):
+                logging.error(f"Context window exceeded in tool follow-up: {e}")
+                raise LLMContextLengthExceededError(str(e)) from e
+
+            logging.error(f"Tool follow-up conversation failed: {e}")
+            # Fallback: return the first tool result if follow-up fails
+            if tool_results:
+                return tool_results[0]["content"]
```
greysonlalonde, line 512: It seems like this could return unexpected results - not sure how to handle that other than raising `raise LLMToolExecutionError("Failed to complete tool use conversation") from e`, though

greysonlalonde (changes requested):

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
lorenzejay: this is different from agent invoking tools. this is for the case of :

```py
def anthropic_tool_use_runner():
    def get_weather(location: str) -> str:
        return f"The weather in {location} is sunny"

    llm = LLM(
        model="anthropic/claude-3-5-sonnet-20241022",
        api_key=os.getenv("ANTHROPIC_API_KEY"),
    )
    result = llm.call(
        messages=[{"role": "user", "content": "What is the weather in San Francisco?"}],
        available_functions={"get_weather": get_weather},
        tools=[
            {
                "type": "function",
                "function": {
                    "name": "get_weather",
                    "description": "Get the weather in a location",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "location": {
                                "type": "string",
                                "description": "The location to get the weather for",
                            }
                        },
                        "required": ["location"],
                    },
                },
            }
        ],
    )
    print("anthropic tool use result", result)


if __name__ == "__main__":
    anthropic_tool_use_runner()

```

greysonlalonde (approved):

PR #3710, opened 2025-10-15
Lorenze/tools drop litellm

lorenzejay:


lib/crewai-tools/src/crewai_tools/adapters/crewai_rag_adapter.py
```diff
@@ -251,9 +254,7 @@ def add(self, *args: ContentItem, **kwargs: Unpack[AddDocumentParams]) -> None:
                     if isinstance(arg, dict):
                         chunk_metadata.update(arg.get("metadata", {}))
 
-                    chunk_id = hashlib.sha256(
-                        f"{loader_result.doc_id}_{i}_{chunk}".encode()
-                    ).hexdigest()
+                    chunk_id = str(uuid4())
```
greysonlalonde: Wondering about the change here, this would allow duplicated embeddings - but maybe that's desirable

lib/crewai-tools/src/crewai_tools/adapters/crewai_rag_adapter.py
```diff
@@ -52,7 +52,10 @@ def model_post_init(self, __context: Any) -> None:
             self._client = create_client(self.config)
         else:
             self._client = get_rag_client()
-        self._client.get_or_create_collection(collection_name=self.collection_name)
+        collection_params = {"collection_name": self.collection_name}
+        if hasattr(self.config, "vectors_config") and self.config.vectors_config:  # type: ignore
```
greysonlalonde: Could use 

```python
if isinstance(self.config, QdrantConfig) and self.config.vectors_config:
    collection_params["vectors_config"] = self.config.vectors_config
```

here to type it, haven't tried that in editor though

lib/crewai-tools/src/crewai_tools/adapters/crewai_rag_adapter.py
lorenzejay: arent we using upsert ?

PR #3711, opened 2025-10-15
Lorenze/adding guardrails

lorenzejay:


src/crewai/task.py
```diff
@@ -268,6 +277,44 @@ def ensure_guardrail_is_callable(self) -> "Task":
 
         return self
 
+    @model_validator(mode="after")
```
greysonlalonde, line 280: Should we have Pydantic manage this?

```python
_guardrails: list[Callable] = PrivateAttr(default_factory=list)
```

src/crewai/task.py
```diff
@@ -268,6 +277,44 @@ def ensure_guardrail_is_callable(self) -> "Task":
...
+    def ensure_guardrails_is_list_of_callables(self) -> "Task":
+        guardrails = []
+        if self.guardrails is not None and (
+            not isinstance(self.guardrails, (list, tuple)) or len(self.guardrails) > 0
+        ):
+            if self.agent is None:
```
greysonlalonde, line 286: I think the agent check is too early maybe, only string-based guardrails need an agent, right?

src/crewai/task.py
```diff
@@ -268,6 +277,44 @@ def ensure_guardrail_is_callable(self) -> "Task":
 
         return self
 
+    @model_validator(mode="after")
+    def ensure_guardrails_is_list_of_callables(self) -> "Task":
+        guardrails = []
+        if self.guardrails is not None and (
```
greysonlalonde, line 283: did you want to check for empty lists here?

lorenzejay:
covered by https://github.com/crewAIInc/crewAI/pull/3728

PR #3717, opened 2025-10-16
feat: enhance GeminiCompletion class with client parameter support

lorenzejay:
rep: https://github.com/crewAIInc/crewAI/pull/3714


lib/crewai/src/crewai/llms/providers/gemini/completion.py
```diff
@@ -427,7 +492,7 @@ def supports_function_calling(self) -> bool:
 
     def supports_stop_words(self) -> bool:
```
greysonlalonde, line 493: Do all gemini models support stop words?

PR #3728, opened 2025-10-17
feat: enhance task guardrail functionality and validation

lorenzejay:
- Introduced support for multiple guardrails in the Task class, allowing for sequential processing of guardrails.
- Added a new `guardrails` field to the Task model to accept a list of callable guardrails or string descriptions.
- Implemented validation to ensure guardrails are processed correctly, including handling of retries and error messages.
- Enhanced the `_invoke_guardrail_function` method to manage guardrail execution and integrate with existing task output processing.
- Updated tests to cover various scenarios involving multiple guardrails, including success, failure, and retry mechanisms.

This update improves the flexibility and robustness of task execution by allowing for more complex validation scenarios.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,18 +4,22 @@ repos:
       - id: ruff
         name: ruff
         entry: uv run ruff check
+        args: ["--config", "pyproject.toml", "."]
         language: system
+        pass_filenames: false
         types: [python]
-        exclude: ^lib/crewai/
       - id: ruff-format
         name: ruff-format
         entry: uv run ruff format
+        args: ["--config", "pyproject.toml", "."]
         language: system
+        pass_filenames: false
         types: [python]
-        exclude: ^lib/crewai/
       - id: mypy
         name: mypy
         entry: uv run mypy
+        args: ["--config-file", "pyproject.toml", "."]
         language: system
+        pass_filenames: false
         types: [python]
-        exclude: ^lib/crewai/
+
--- a/lib/crewai/src/crewai/events/event_listener.py
+++ b/lib/crewai/src/crewai/events/event_listener.py
@@ -53,7 +53,6 @@ from crewai.task import Task
 from crewai.telemetry.telemetry import Telemetry
 from crewai.utilities import Logger
 from crewai.utilities.constants import EMITTER_COLOR
-from crewai.utilities.printer import Printer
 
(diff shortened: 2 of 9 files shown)


lib/crewai/src/crewai/task.py
```diff
@@ -154,6 +155,16 @@ class Task(BaseModel):
         default=None,
         description="Function or string description of a guardrail to validate task output before proceeding to next task",
     )
+    guardrails: (
+        Sequence[Callable[[TaskOutput], tuple[bool, Any]] | str]
```
greysonlalonde: Can you move this out of the class body & to a type definition, `GuardrailType` maybe?

lib/crewai/src/crewai/task.py
```diff
@@ -270,6 +284,46 @@ def ensure_guardrail_is_callable(self) -> "Task":
 
         return self
 
+    @model_validator(mode="after")
+    def ensure_guardrails_is_list_of_callables(self) -> "Task":
+        guardrails = []
+        if self.guardrails is not None and (
```
greysonlalonde: I think this evaluates to `If guardrails is not a list/tuple but is not None, then always True.`

Maybe you want:

```python3
if self.guardrails is not None:
    if isinstance(self.guardrails, (list, tuple)):
        if len(self.guardrails) > 0:
            # process list
    else:
        # process single guardrail
```

lib/crewai/src/crewai/task.py
```diff
@@ -791,3 +823,55 @@ def fingerprint(self) -> Fingerprint:
             Fingerprint: The fingerprint of the task
         """
         return self.security_config.fingerprint
+
+    def _invoke_guardrail_function(
```
greysonlalonde, line 834: Potential retry behavior here, if a guardrail fails and retries, it re-runs all guardrails, not just the failed one. I think the retry counter is also shared across guardrails. Ignore if expected behavior

lib/crewai/src/crewai/agent.py
```diff
@@ -1,7 +1,7 @@
+from collections.abc import Callable, Sequence
```
greysonlalonde, line 1: Can you run ruff on these if you haven't already? I think i disabled the pre-commit hook temporarily b/c of some issues

greysonlalonde (changes requested):

PR #3749, opened 2025-10-20
Implement improvements on QdrantVectorSearchTool

danielfsbarreto:
- Allow search filters to be set at the constructor level
- Fix issue that prevented multiple records from being returned

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -2,19 +2,26 @@ name: "CodeQL Config"
 
 paths-ignore:
   # Ignore template files - these are boilerplate code that shouldn't be analyzed
-  - "src/crewai/cli/templates/**"
+  - "lib/crewai/src/crewai/cli/templates/**"
   # Ignore test cassettes - these are test fixtures/recordings
-  - "tests/cassettes/**"
+  - "lib/crewai/tests/cassettes/**"
+  - "lib/crewai-tools/tests/cassettes/**"
   # Ignore cache and build artifacts
   - ".cache/**"
   # Ignore documentation build artifacts
   - "docs/.cache/**"
-  
+  # Ignore experimental code
+  - "lib/crewai/src/crewai/experimental/a2a/**"
+
 paths:
-  # Include all Python source code
-  - "src/**"
-  # Include tests (but exclude cassettes)
-  - "tests/**"
+  # Include all Python source code from workspace packages
+  - "lib/crewai/src/**"
+  - "lib/crewai-tools/src/**"
+  - "lib/devtools/src/**"
+  # Include tests (but exclude cassettes via paths-ignore)
+  - "lib/crewai/tests/**"
+  - "lib/crewai-tools/tests/**"
+  - "lib/devtools/tests/**"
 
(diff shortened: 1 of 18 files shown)


lib/crewai-tools/src/crewai_tools/tools/qdrant_vector_search_tool/qdrant_search_tool.py
```diff
@@ -130,14 +142,26 @@ def _run(
         if not self.qdrant_url:
             raise ValueError("QDRANT_URL is not set")
 
-        # Create filter if filter parameters are provided
-        search_filter = None
+        # If filter_by and filter_value are provided, add them to the search filter
+        # without modifying the original search filter
+        search_filter = self.search_filter.copy() if self.search_filter else None
```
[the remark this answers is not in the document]
danielfsbarreto: N/A

lib/crewai-tools/src/crewai_tools/tools/qdrant_vector_search_tool/qdrant_search_tool.py
```diff
@@ -106,6 +106,26 @@ def __init__(self, **kwargs):
                     "The 'qdrant-client' package is required to use the QdrantVectorSearchTool. "
```
greysonlalonde: If you have a sec, I'd also like to see the client initialization logic moved out of `__init__`, there isn't much of a reason as to why we should be using `__init__` methods in pydantic with the exception of highly specialized cases (this is just a bad pattern we have going in general that we should move away from). Could look like:

```python3
client: QdrantClient = Field(default_factory=lambda: QdrantClient(
        url=os.getenv(....),
        api_key=os.getenv(....),
    )
)
```

lib/crewai-tools/src/crewai_tools/tools/qdrant_vector_search_tool/qdrant_search_tool.py
```diff
@@ -106,6 +106,26 @@ def __init__(self, **kwargs):
...
                     "Please install it with: uv add qdrant-client"
                 )
+        if self.filter_conditions:
+            must_conditions = []
+            for filter_condition in self.filter_conditions:
+                if (
```
greysonlalonde: You can move the validation logic into `Field` on line 61, pydantic will also handle `not isinstance(filter_condition, (tuple, list))` for you automatically since you typed it as such. So the only thing you'll need to refactor is `len(filter_condition) < 2`, which is `max_length=2` per [pydantic](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field)

lib/crewai-tools/src/crewai_tools/tools/qdrant_vector_search_tool/qdrant_search_tool.py
greysonlalonde: or could use a `@model_validator` to give more control over api key etc, see [here](https://docs.pydantic.dev/latest/concepts/validators/#model-validators)

greysonlalonde (approved):

greysonlalonde (approved):

Issue #3667, opened 2025-10-08
[BUG] PGSearchTool does not work


AlexisGitHu:
### Description

PGSearchTool does not work providing a db_uri. 

### Steps to Reproduce

Instantiate the PGSearchTool class providing a db_uri.

### Expected behavior

It should not trigger an error on db_uri as it is being passed to the __init__ method

### Screenshots/Code snippets

<img width="257" height="89" alt="Image" src="https://github.com/user-attachments/assets/da5a5f37-9ae6-4323-b2f0-47f81c1f3bfc" />

### Operating System

Ubuntu 22.04

### Python Version

3.12

### crewAI Version

0.201.1

### crewAI Tools Version

0.75.0

### Virtual Environment

Venv

### Evidence

Error trace: 

```python

  File "/home/alexis/github/crewai-test-pcai/backend/v1/crew.py", line 127, in consultant
    pg_sql = PGSearchTool(
        db_uri=os.environ.get("PG_URI"),
    ...<5 lines>...
        },
    )
  File "/home/alexis/github/crewai-test-pcai/backend/.venv/lib/python3.13/site-packages/crewai_tools/tools/pg_search_tool/pg_search_tool.py", line 36, in __init__
    self.add(table_name, data_type=DataType.POSTGRES, metadata={"db_uri": self.db_uri})
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alexis/github/crewai-test-pcai/backend/.venv/lib/python3.13/site-packages/crewai_tools/tools/pg_search_tool/pg_search_tool.py", line 45, in add
    super().add(f"SELECT * FROM {table_name};", **kwargs)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/alexis/github/crewai-test-pcai/backend/.venv/lib/python3.13/site-packages/crewai_tools/tools/rag/rag_tool.py", line 192, in add
    self.adapter.add(*args, **kwargs)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/alexis/github/crewai-test-pcai/backend/.venv/lib/python3.13/site-packages/crewai_tools/adapters/crewai_rag_adapter.py", line 188, in add
[...]

PR #3786, opened 2025-10-23
Fixing CREWAI RAG ADAPTER

Vidit-Ostwal:
1. Sending metadata while loading.
Fixes: https://github.com/crewAIInc/crewAI/issues/3667


Vidit-Ostwal:
Created a replica of this PR in depriciated crewai-tools lib: https://github.com/crewAIInc/crewAI-tools/pull/485

PR #3809, opened 2025-10-29
fix: refactor  to property for dynamic check, add check for stop words

greysonlalonde:


lib/crewai/src/crewai/agents/crew_agent_executor.py
```diff
@@ -128,14 +127,25 @@ def __init__(
...
+                set(
+                    existing_stop + self.stop
+                    if isinstance(existing_stop, list)
+                    else self.stop
+                )
             )
-        )
+
+    @property
+    def use_stop_words(self) -> bool:
```
lucasgomide, line 142: This method is not being called, that's right?

greysonlalonde: Just as a property, no function call

PR #3849, opened 2025-11-06
feat: add pydantic validation dunder to BaseInterceptor

greysonlalonde:
- add pydantic validation dunder to BaseInterceptor
- improve HTTPTransport typing

### DIFF
--- a/lib/crewai/src/crewai/rag/chromadb/utils.py
+++ b/lib/crewai/src/crewai/rag/chromadb/utils.py
@@ -67,31 +67,44 @@ def _prepare_documents_for_chromadb(
     ids: list[str] = []
     texts: list[str] = []
     metadatas: list[Mapping[str, str | int | float | bool]] = []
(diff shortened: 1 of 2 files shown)


lib/crewai/src/crewai/llms/hooks/base.py
```diff
@@ -7,7 +7,14 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Generic, TypeVar
+from typing import TYPE_CHECKING, Any, Generic, TypeVar
+
+from pydantic_core import core_schema
+
+
+if TYPE_CHECKING:
```
lorenzejay, line 15: not checking on runtime?

lorenzejay (approved):

lib/crewai/src/crewai/llms/hooks/base.py
greysonlalonde: yup! static only

PR #3861, opened 2025-11-07
fix: properly handle agent max iterations

greysonlalonde:


lib/crewai/src/crewai/utilities/agent_utils.py
```diff
@@ -168,8 +168,16 @@ def handle_max_iterations_exceeded(
...
         raise ValueError("Invalid response from LLM call - None or empty.")
 
-    # Return the formatted answer, regardless of its type
-    return format_answer(answer=answer)
+    formatted = format_answer(answer=answer)
+
+    # If format_answer returned an AgentAction, convert it to AgentFinish
+    if isinstance(formatted, AgentFinish):
+        return formatted
+    return AgentFinish(
```
lorenzejay, line 176: perfect

lib/crewai/src/crewai/agents/crew_agent_executor.py
```diff
@@ -214,6 +214,7 @@ def _invoke_loop(self) -> AgentFinish:
                         llm=self.llm,
                         callbacks=self.callbacks,
                     )
+                    break
```
lorenzejay, line 217: nice

PR #3893, opened 2025-11-11
feat: implement before and after LLM call hooks in CrewAgentExecutor

lorenzejay:
- Added support for before and after LLM call hooks to allow modification of messages and responses during LLM interactions.
- Introduced LLMCallHookContext to provide hooks with access to the executor state, enabling in-place modifications of messages.
- Updated get_llm_response function to utilize the new hooks, ensuring that modifications persist across iterations.
- Enhanced tests to verify the functionality of the hooks and their error handling capabilities, ensuring robust execution flow.


lucasgomide (approved):
🚢

PR #3894, opened 2025-11-12
feat: fetch and store more data about okta authorization server

heitorado:
## Description

This PR adds more customization when using the OKTA provider for logging in with the CLI. Now it supports custom authorization servers other than the "default" one and even "org authorization servers".

#### This PR has a counterpart at the crewai-plus repo with relevant changes in the API:
- 

### Main technical changes
- Store more info on settings, now there is an `extra` field to store additional information that might be provider specific. 
- The field extra is always required to be sent, but not required to have a value. It can be an empty dict.
- How this dict is going to be used will be determined by each provider. If the logic is common to all auth providers, it should stay out of the `extra` field.
- ⚠️ **[NOTE]** - This PR got kinda bloated because ruff complained A LOT and pleasing it ended up triggering complaints on files that were not the ones I originally modified. Like a hydra of sorts. So there is a lot of type-checker-pleasing corrections.

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -38,6 +38,10 @@ from crewai.utilities.agent_utils import (
 )
 from crewai.utilities.constants import TRAINING_DATA_FILE
 from crewai.utilities.i18n import I18N, get_i18n
+from crewai.utilities.llm_call_hooks import (
+    get_after_llm_call_hooks,
+    get_before_llm_call_hooks,
+)
 from crewai.utilities.printer import Printer
 from crewai.utilities.tool_utils import execute_tool_and_check_finality
 from crewai.utilities.training_handler import CrewTrainingHandler
@@ -130,6 +134,10 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
         self.messages: list[LLMMessage] = []
         self.iterations = 0
         self.log_error_after = 3
+        self.before_llm_call_hooks: list[Callable] = []
+        self.after_llm_call_hooks: list[Callable] = []
+        self.before_llm_call_hooks.extend(get_before_llm_call_hooks())
+        self.after_llm_call_hooks.extend(get_after_llm_call_hooks())
         if self.llm:
             # This may be mutating the shared llm object and needs further evaluation
             existing_stop = getattr(self.llm, "stop", [])
@@ -226,6 +234,7 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
                     from_task=self.task,
                     from_agent=self.agent,
                     response_model=self.response_model,
+                    executor_context=self,
                 )
                 formatted_answer = process_llm_response(answer, self.use_stop_words)  # type: ignore[assignment]
 
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -33,6 +33,7 @@ from crewai.utilities.types import LLMMessage
 
 if TYPE_CHECKING:
     from crewai.agent import Agent
+    from crewai.agents.crew_agent_executor import CrewAgentExecutor
     from crewai.lite_agent import LiteAgent
     from crewai.llm import LLM
     from crewai.task import Task
@@ -236,6 +237,7 @@ def get_llm_response(
     from_task: Task | None = None,
     from_agent: Agent | LiteAgent | None = None,
     response_model: type[BaseModel] | None = None,
+    executor_context: CrewAgentExecutor | None = None,
 ) -> str:
     """Call the LLM and return the response, handling any invalid responses.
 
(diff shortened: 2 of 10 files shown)


lib/crewai/src/crewai/cli/authentication/main.py
```diff
@@ -28,22 +30,36 @@ class Oauth2Settings(BaseModel):
         description="OAuth2 audience value, typically used to identify the target API or resource.",
         default=None,
     )
+    extra: dict[str, Any] = Field(
+        description="Extra configuration for the OAuth2 provider.",
+        default={},
+    )
```
heitorado, line 36: Relevant code change

lib/crewai/src/crewai/cli/authentication/main.py
```diff
@@ -28,22 +30,36 @@ class Oauth2Settings(BaseModel):
...
+        """Create an Oauth2Settings instance from the CLI settings."""
+
         settings = Settings()
 
         return cls(
             provider=settings.oauth2_provider,
             domain=settings.oauth2_domain,
             client_id=settings.oauth2_client_id,
             audience=settings.oauth2_audience,
+            extra=settings.oauth2_extra,
```
heitorado, line 49: Relevant code change

lib/crewai/src/crewai/cli/authentication/providers/base_provider.py
```diff
@@ -24,3 +24,6 @@ def get_audience(self) -> str: ...
 
     @abstractmethod
     def get_client_id(self) -> str: ...
+
+    def get_required_fields(self) -> list[str]:
+        return []
```
heitorado, line 30: Relevant code change

lib/crewai/src/crewai/cli/authentication/providers/okta.py
```diff
@@ -3,16 +3,16 @@
...
-        return f"https://{self.settings.domain}/oauth2/default/v1/token"
+        return f"{self._oauth2_base_url()}/v1/token"
 
     def get_jwks_url(self) -> str:
-        return f"https://{self.settings.domain}/oauth2/default/v1/keys"
+        return f"{self._oauth2_base_url()}/v1/keys"
 
     def get_issuer(self) -> str:
-        return f"https://{self.settings.domain}/oauth2/default"
+        return self._oauth2_base_url().removesuffix("/oauth2")
```
heitorado, line 15: Relevant code change

lib/crewai/src/crewai/cli/authentication/providers/okta.py
```diff
@@ -27,3 +27,16 @@ def get_client_id(self) -> str:
...
+
+    def _oauth2_base_url(self) -> str:
+        using_org_auth_server = self.settings.extra.get("using_org_auth_server", False)
+
+        if using_org_auth_server:
+            base_url = f"https://{self.settings.domain}/oauth2"
+        else:
+            base_url = f"https://{self.settings.domain}/oauth2/{self.settings.extra.get('authorization_server_name', 'default')}"
+
+        return f"{base_url}"
```
heitorado, line 42: Relevant code change

lib/crewai/src/crewai/cli/config.py
```diff
@@ -136,7 +137,12 @@ class Settings(BaseModel):
         default=DEFAULT_CLI_SETTINGS["oauth2_domain"],
     )
 
-    def __init__(self, config_path: Path | None = None, **data):
+    oauth2_extra: dict[str, Any] = Field(
+        description="Extra configuration for the OAuth2 provider.",
+        default={},
+    )
```
heitorado, line 143: Relevant code change

lib/crewai/src/crewai/cli/enterprise/main.py
```diff
@@ -99,3 +88,38 @@ def _update_oauth_settings(
...
+            raise ValueError(
+                f"Missing authentication provider required fields in OAuth2 configuration: [{', '.join(missing_provider_specific_fields)}] (Configured provider: '{oauth_config['provider']}')"
+            )
+
+    def _get_provider_specific_fields(self, provider_name: str) -> list[str]:
+        provider = ProviderFactory.from_settings(
+            Oauth2Settings(provider=provider_name, client_id="dummy", domain="dummy")
+        )
+
+        return provider.get_required_fields()
```
heitorado, line 125: Relevant code change

lib/crewai/src/crewai/cli/authentication/providers/okta.py
gabemilani, line 42: Instead of using a flag to identify if it's using an auth server or not, wouldn't be simpler to just check the authorization server name presence ?

gabemilani (commented):
Any test worth writing to ensure this behavior?

lib/crewai/src/crewai/cli/authentication/providers/okta.py
heitorado: No, because the user could use the custom server instead of the auth server and then simply not setting the name if it wishes to use the `default` authorization server, because if no name is provided it falls back to `default`.

heitorado:
> Any test worth writing to ensure this behavior?

I think the updated tests are already covering the new behaviour of checking for the new "extra" field. I don't think writing a test just for Okta would be necessary. Any suggestions on new tests to write?

gabemilani (approved):

PR #3902, opened 2025-11-12
Lorenze/feat hooks

lorenzejay:


lib/crewai/src/crewai/hooks/llm_hooks.py
```diff
@@ -53,8 +56,51 @@ def __init__(
...
+            if response:
+                printer.print(content="\nProcessing your input...", color="cyan")
+
+            return response
+        finally:
+            event_listener.formatter.resume_live_updates()
+
 
-# Global hook registries (optional convenience feature)
 _before_llm_call_hooks: list[Callable[[LLMCallHookContext], None]] = []
```
greysonlalonde: `contextvar.ContextVar` for the different hook stores may be better across concurrency models / maybe easier avoiding state pollution issues

lib/crewai/src/crewai/hooks/llm_hooks.py
```diff
@@ -53,8 +56,51 @@ def __init__(
...
 _after_llm_call_hooks: list[Callable[[LLMCallHookContext], str | None]] = []
```
greysonlalonde: fut: will probably need any of `unregister_*_hook()` or `clear_hooks()`, adding for reminder

lib/crewai/src/crewai/hooks/tool_hooks.py
greysonlalonde, line 1: It'd be nice to align the hook returns, right now they're:

```python
  - lib/crewai/src/crewai/hooks/llm_hooks.py:110 - BeforeLLMCallHook returns None
  - lib/crewai/src/crewai/hooks/llm_hooks.py:137 - AfterLLMCallHook returns str | None
  - lib/crewai/src/crewai/hooks/tool_hooks.py:119 - BeforeToolCallHook returns bool | None
  - lib/crewai/src/crewai/hooks/tool_hooks.py:158 - AfterToolCallHook returns str | None
```

If purposely disjoint then ignore

lib/crewai/src/crewai/hooks/types.py
greysonlalonde, line 1: Could potentially use a base generic here to consolidate logic and propagate type hints better, something like

```python3
CtxT = TypeVar("CtxT", contravariant=True)
RetT = TypeVar("RetT", covariant=True)

@runtime_checkable
class Hook(Protocol[CtxT, RetT]):
    """Generic protocol for hook callables."""
    def __call__(self, context: CtxT) -> RetT: ...
```
then we can have case matching for message types

lib/crewai/src/crewai/hooks/base.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import annotations
+
+from typing import TypeVar
+
+
+# Type variable for hook context types
+HookContextT = TypeVar("HookContextT")
+
+
+def validate_hook_callable(hook: object, hook_type: str) -> None:
```
greysonlalonde: Wondering if this is future impl? or I just missed the use

lib/crewai/src/crewai/lite_agent.py
```diff
@@ -542,6 +542,7 @@ def _invoke_loop(self) -> AgentFinish:
                             agent_key=self.key,
                             agent_role=self.role,
                             agent=self.original_agent,
+                            crew=None,
```
greysonlalonde, line 545: Not sure on applicability, but would it be worth it / correct to use `crewai.utilities.crew.crew_context.get_crew_context` to pass the id? could do 

```python3
crew_context  = get_crew_context()
# isolate the id attr below if existing
```

greysonlalonde (approved):
Would be good to also unify the exception handling across all hooks, but not super pressing

lib/crewai/src/crewai/lite_agent.py
lorenzejay: LiteAgents don't fit in crew atm

greysonlalonde: what about for `HallucinationGuardrail`?

lib/crewai/src/crewai/hooks/base.py
lorenzejay: no implementation here, just guessed

lib/crewai/src/crewai/lite_agent.py
lorenzejay: then it be out of scope like it wouldnt be calling a tool which is the scope of this code block

PR #3922, opened 2025-11-15
fix: ensure instrumentation flags

greysonlalonde:


joaomdmoura (approved):

lib/crewai/src/crewai/events/listeners/tracing/trace_batch_manager.py
```diff
@@ -243,7 +247,13 @@ def _send_events_to_backend(self) -> int:
...
     def finalize_batch(self) -> TraceBatch | None:
         """Finalize batch and return it for sending"""
-        if not self.current_batch:
+        from crewai.events.listeners.tracing.utils import has_user_declined_tracing
+
+        if (
+            not self.current_batch
+            or os.getenv("CREWAI_TRACING_ENABLED", "false").lower() == "false"
+            or has_user_declined_tracing()
+        ):
```
lorenzejay: i think this is off

lorenzejay: user can have nothing set on .env but have tracing=True

same with if they denied user tracing, but have it set.

lib/crewai/src/crewai/events/listeners/tracing/utils.py
```diff
@@ -23,7 +24,120 @@
...
+    """
+    if override is True:
+        return True
+    if override is False:
+        return False
+
+    env_value = os.getenv("CREWAI_TRACING_ENABLED", "").lower()
+    if env_value in ("true", "1"):
+        return True
+
```
lorenzejay, line 53: nice

PR #3943, opened 2025-11-19
feat: support CLI login with Entra ID

heitorado:
## Description

This PR adds support for CLI login with Microsoft's Entra ID.
It is really similar to Okta but Entra ID has some unique intricacies such as a global base URL prefix (just the tenant ID changes) and the necessity of including a custom OIDC scope to perform token validation properly.

There is a counterpart on AMP that needs to be merged first:
- https://github.com/crewAIInc/crewai-plus/pull/1813

### Changes overview
- Add `EntraIdProvider` class and tests
- Change `from_settings` method and make dynamic provider class invocation more robust (now build class names properly when provider has more than one word in snake case format)
- Fetch device code `scope` payload parameter dynamically from the provider, default is `"opened profile email"`
- Within `_display_auth_instructions`, we now try to grab the verification uri from `“verification_uri_complete”` and if not found we fallback to `“verification_uri”` (Entra ID only uses `‘verification_uri’`)
- Feedback messages enhanced

### DIFF
--- a/lib/crewai/src/crewai/events/event_listener.py
+++ b/lib/crewai/src/crewai/events/event_listener.py
@@ -101,24 +101,25 @@ if TYPE_CHECKING:
 
 
 class EventListener(BaseEventListener):
-    _instance = None
+    _instance: EventListener | None = None
+    _initialized: bool = False
     _telemetry: Telemetry = PrivateAttr(default_factory=lambda: Telemetry())
-    logger = Logger(verbose=True, default_color=EMITTER_COLOR)
+    logger: Logger = Logger(verbose=True, default_color=EMITTER_COLOR)
     execution_spans: dict[Task, Any] = Field(default_factory=dict)
-    next_chunk = 0
-    text_stream = StringIO()
-    knowledge_retrieval_in_progress = False
-    knowledge_query_in_progress = False
+    next_chunk: int = 0
+    text_stream: StringIO = StringIO()
+    knowledge_retrieval_in_progress: bool = False
+    knowledge_query_in_progress: bool = False
     method_branches: dict[str, Any] = Field(default_factory=dict)
 
-    def __new__(cls):
+    def __new__(cls) -> EventListener:
         if cls._instance is None:
             cls._instance = super().__new__(cls)
             cls._instance._initialized = False
         return cls._instance
 
-    def __init__(self):
-        if not hasattr(self, "_initialized") or not self._initialized:
+    def __init__(self) -> None:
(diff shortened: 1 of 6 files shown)


lib/crewai/src/crewai/cli/authentication/providers/base_provider.py
```diff
@@ -28,3 +28,6 @@ def get_client_id(self) -> str: ...
     def get_required_fields(self) -> list[str]:
         """Returns which provider-specific fields inside the "extra" dict will be required"""
         return []
+
+    def get_scope(self) -> str:
+        return "openid profile email "
```
vinibrsl: Is the extra space at the end intentional?

heitorado: Yes. If not added here, would have to be added on the overriden method (if the default scope was to be reused) like this: `return f"{super().get_scope()} extra_scope_value"`

Do you think it is better?

vinibrsl: What if you had a list of strings and converted them to a single string only when needed?

lib/crewai/src/crewai/cli/authentication/utils.py
```diff
@@ -1,10 +1,12 @@
+from typing import Any
+
 import jwt
 from jwt import PyJWKClient
 
 
 def validate_jwt_token(
     jwt_token: str, jwks_url: str, issuer: str, audience: str
-) -> dict:
+) -> Any:
```
lucasgomide, line 9: Are we returning another data type?

lib/crewai/src/crewai/cli/authentication/providers/base_provider.py
lucasgomide: I would say the same 😆

heitorado: I'm sold! Check out [287e1cd](https://github.com/crewAIInc/crewAI/pull/3943/commits/287e1cd46a49c6d57ec138a9dc64409755fca2a1)

lib/crewai/src/crewai/cli/authentication/utils.py
heitorado: Yeah this confused me at first too, but turns out the `decode` method from the `jwt` lib returns `Any`, so we have to match the return type here:

image

vinibrsl (approved):

PR #3975, opened 2025-11-25
Lorenze/agent executor flow pattern

lorenzejay:


lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+
+        return "agent_finished"
+
+    @listen("continue_reasoning")
+    def call_llm_and_parse(self) -> str:
+        """Execute LLM call with hooks and parse the response.
+
+        Returns routing decision based on parsing result.
+        """
+        self._printer.print(
```
greysonlalonde: Unsure if the `self._printer.print` is right, would rich usage be better? This will break the active rich tree

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
greysonlalonde, line 1: Same note as below in this file, `self._printer.print` will break the rich tree

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+    Inherits from:
+    - Flow[AgentReActState]: Provides flow orchestration capabilities
+    - CrewAgentExecutorMixin: Provides memory methods (short/long/external term)
+
+    Note: Multiple instances may be created during agent initialization
+    (cache setup, RPM controller setup, etc.) but only the final instance
+    should execute tasks via invoke().
+    """
+
+    def __init__(
```
greysonlalonde, line 88: Love all the typing here! I think we can probably start replacing `BaseLLM` with `LLM` since this is internal

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+        """Compatibility property for mixin - returns state messages."""
+        return list(self._state.messages)
+
+    @property
+    def iterations(self) -> int:
+        """Compatibility property for mixin - returns state iterations."""
+        return self._state.iterations
+
+    @start()
+    def initialize_reasoning(self) -> str:
```
greysonlalonde: Can use a string literal here

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+            tools_handler: Tool handler instance.
+            step_callback: Optional step callback.
+            original_tools: Original tool list.
+            function_calling_llm: Optional function calling LLM.
+            respect_context_window: Respect context limits.
+            request_within_rpm_limit: RPM limit check function.
+            callbacks: Optional callbacks list.
+            response_model: Optional Pydantic model for structured outputs.
+        """
+        self._i18n: I18N = get_i18n()
```
greysonlalonde: We can omit this entirely as an attribute and just use get_i18n where it's needed since it's cached, no need for the per instance attr

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+        self.task = task
+        self.agent = agent
+        self.crew = crew
+        self.prompt = prompt
+        self.tools = tools
+        self.tools_names = tools_names
+        self.stop = stop_words
+        self.max_iter = max_iter
+        self.callbacks = callbacks or []
+        self._printer: Printer = Printer()
```
greysonlalonde, line 143: Same with the printer here, can omit and instantiate outside the class body

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+        """Increment iteration counter and loop back for next iteration."""
+        self.state.iterations += 1
+        self._printer.print(
+            content=f"+ increment_and_continue: Incremented to iteration {self.state.iterations}, looping back",
+            color="magenta",
+        )
+        return "initialized"
+
+    @listen(or_("agent_finished", "tool_result_is_final"))
+    def finalize(self) -> str:
```
greysonlalonde: Same here, if we return a string literal we can do more complex things with AST

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+        self._show_logs(self.state.current_answer)
+
+        return "completed"
+
+    @listen("parser_error")
+    def recover_from_parser_error(self) -> str:
+        """Recover from output parser errors and retry."""
+        formatted_answer = handle_output_parser_exception(
+            e=self._last_parser_error,
+            messages=list(self.state.messages),
```
greysonlalonde, line 423: Wondering here - passing `list(self.state.messages)` will create a copy of the list, so any list modifications in this function would be lost

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+
+        return "initialized"
+
+    @listen("context_error")
+    def recover_from_context_length(self) -> str:
+        """Recover from context length errors and retry."""
+        handle_context_length(
+            respect_context_window=self.respect_context_window,
+            printer=self._printer,
+            messages=list(self.state.messages),
```
greysonlalonde: Same here on list modification

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
greysonlalonde, line 1: Generally any return values that are hardcoded strings, we should change those to string literals. Makes it easier for visualization + we can use these for AST-enforced rules in the future

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+        crew: Crew,
+        agent: Agent,
+        prompt: SystemPromptResult | StandardPromptResult,
+        max_iter: int,
+        tools: list[CrewStructuredTool],
+        tools_names: str,
+        stop_words: list[str],
+        tools_description: str,
+        tools_handler: ToolsHandler,
+        step_callback: Any = None,
```
greysonlalonde, line 101: Is `step_callback` a known `Callable` type?

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+        original_tools: list[BaseTool] | None = None,
+        function_calling_llm: BaseLLM | Any | None = None,
```
greysonlalonde, line 103: We can probably constrain this to `BaseLLM | None`

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+
+        Flow initialization is deferred to prevent event emission during agent setup.
+        Returns the temporary state until invoke() is called.
+        """
+        return self._state
+
+    @property
+    def messages(self) -> list[LLMMessage]:
+        """Compatibility property for mixin - returns state messages."""
+        return list(self._state.messages)
```
greysonlalonde: Could change this to a direct reference to address the other copy notes

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+                from_agent=self.agent,
+                response_model=self.response_model,
+                executor_context=self,
+            )
+
+            # Parse the LLM response
+            formatted_answer = process_llm_response(answer, self.use_stop_words)
+            self.state.current_answer = formatted_answer
+
+            # Debug: Check what we parsed
```
greysonlalonde: Debug note here

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+
+        Args:
+            inputs: Input dictionary containing prompt variables.
+
+        Returns:
+            Dictionary with agent output.
+        """
+        self._ensure_flow_initialized()
+
+        if self._is_executing:
```
greysonlalonde: TOCTOU here to line 466

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
```diff
@@ -0,0 +1,801 @@
...
+        try:
+            # Reset state for fresh execution
+            self.state.messages.clear()
+            self.state.iterations = 0
+            self.state.current_answer = None
+            self.state.is_finished = False
+
+            if "system" in self.prompt:
+                system_prompt = self._format_prompt(
+                    cast(str, self.prompt.get("system", "")), inputs
```
greysonlalonde: I think since `"system"` is guaranteed as a key,  can use `self.prompt["system"]` without the cast

greysonlalonde: Don't quote me on that, though

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
greysonlalonde, line 1: I think we may need async variations of methods that involve external calls to keep the async chain true

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: going to handle the printer differently as this should be a separate pr to potentially refactor / deprecate how we use that printer for rich

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: `create_agent_executor` within core.py so that we always create a fresh executor instead of, we need to re-think this fully

original:
```py
if self.agent_executor is not None:
            self._update_executor_parameters(
                task=task,
                tools=parsed_tools,
                raw_tools=raw_tools,
                prompt=prompt,
                stop_words=stop_words,
                rpm_limit_fn=rpm_limit_fn,
            )
```

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: adding as a fast follow

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: good call

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: TODO: fast follow

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: noting for fast follow

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: yes, but the types passed down are all Any. will address with a typing fast follow

lib/crewai/src/crewai/experimental/crew_agent_executor_flow.py
lorenzejay: Good call! addressed!

PR #3976, opened 2025-11-26
refactor: enhance model validation and provider inference in LLM class

lorenzejay:
- Updated the model validation logic to support pattern matching for new models and "latest" versions, improving flexibility for various providers.
- Refactored the `_validate_model_in_constants` method to first check hardcoded constants and then fall back to pattern matching.
- Introduced `_matches_provider_pattern` to streamline provider-specific model checks.
- Enhanced the `_infer_provider_from_model` method to utilize pattern matching for better provider inference.

This refactor aims to improve the extensibility of the LLM class, allowing it to accommodate new models without requiring constant updates to the hardcoded lists.

### DIFF
--- a/lib/crewai-tools/pyproject.toml
+++ b/lib/crewai-tools/pyproject.toml
@@ -16,9 +16,9 @@ dependencies = [
     "lancedb>=0.5.4",
     "tiktoken>=0.8.0",
     "beautifulsoup4>=4.13.4",
-    "pypdf>=5.9.0",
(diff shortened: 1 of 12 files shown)


greysonlalonde:
I think only thing here is to add some logging / event push signaling if a model name doesn't match any provided ones, otherwise it'll be tricky debugging smaller misspelling issues in certain environments

greysonlalonde (approved):

PR #3978, opened 2025-11-26
supporting thinking for anthropic models

lorenzejay:


lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -52,6 +51,7 @@ def __init__(
         stream: bool = False,
         client_params: dict[str, Any] | None = None,
         interceptor: BaseInterceptor[httpx.Request, httpx.Response] | None = None,
+        thinking: dict[str, Any] | None = None,
```
greysonlalonde: Can we make this a TypedDict / BaseModel to constrain the params

greysonlalonde (changes requested):
Just one note, rest looks great

PR #3998, opened 2025-11-30
feat: create sys event types and handler

greysonlalonde:


lib/crewai/src/crewai/telemetry/telemetry.py
```diff
@@ -155,6 +166,66 @@ def set_tracer(self) -> None:
...
+        """
+        try:
+            original_handler = signal.getsignal(sig)
+            self._original_handlers[sig] = original_handler
+
+            def handler(signum: int, frame: Any) -> None:
+                crewai_event_bus.emit(self, event_class())
+
+                if shutdown:
+                    self._shutdown()
```
lucasgomide, line 202: Can we add a try/catch here? to ensure exception safety inside signal handlers

greysonlalonde: like for the shutdown block? we don't want to interfere with `raise SystemExit(0)`

lib/crewai/src/crewai/telemetry/telemetry.py
```diff
@@ -155,6 +166,66 @@ def set_tracer(self) -> None:
...
+
+                if original_handler not in (signal.SIG_DFL, signal.SIG_IGN, None):
+                    if callable(original_handler):
+                        original_handler(signum, frame)
+                elif shutdown:
+                    raise SystemExit(0)
+
+            signal.signal(sig, handler)
+        except (ValueError, OSError):
+            pass
```
lucasgomide: I would treat those errors as proper except blocks. Might be valid understand what is happening, especially the OSError. 

Regarding the ValueError, it will be raised when an unexpected signal is fired right?

greysonlalonde: This revolves more around thread context and platform compatibility. `ValueError` (iirc) is raised if this is called from a non-main thread, or if on a platform without support. we could add error logging, but we wouldn't be able to see these logs anyway with the current approach

greysonlalonde: So I guess it's more whether we want this to be complete in local environments, which I agree with

joaomdmoura (approved):

PR #4002, opened 2025-11-30
chore: pin dependencies in crewai, crewai-tools, devtools

greysonlalonde:


greysonlalonde:
> GitHub wouldn’t show me the full `uv.lock` diff. Did any versions get upgraded with the changes here?

Quite a few, I'm resolving now. So many deps had `>= `.

lorenzejay (approved):

PR #4008, opened 2025-12-01
feat: async llm support

greysonlalonde:
- add acall() async method to the base llm interface and implement it across all providers
  - support both streaming and non-streaming async responses
  - include comprehensive test coverage with vcr cassettes for all async functionality
  - Docs updated to reflect async api


lorenzejay (approved):
tested ! great work here

PR #4015, opened 2025-12-02
feat: native async tool support

greysonlalonde:
- adds native async support to tools
- adds relevant tests
- updates relevant docs

### DIFF
--- a/lib/crewai/src/crewai/tools/base_tool.py
+++ b/lib/crewai/src/crewai/tools/base_tool.py
@@ -318,7 +318,15 @@ class Tool(BaseTool, Generic[P, R]):
     func: Callable[P, R | Awaitable[R]]
 
     def run(self, *args: P.args, **kwargs: P.kwargs) -> R:
-        """Execute the tool synchronously with type-safe parameters."""
+        """Executes the tool synchronously.
+
+        Args:
+            *args: Positional arguments for the tool.
+            **kwargs: Keyword arguments for the tool.
+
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/tools/base_tool.py
```diff
@@ -239,32 +301,101 @@ def _get_arg_annotations(annotation: type[Any] | None) -> str:
...
+
+        Returns:
+            The result of the tool execution.
+        """
+        _printer.print(f"Using Tool: {self.name}", color="cyan")
+        result = self.func(*args, **kwargs)
+
+        if asyncio.iscoroutine(result):
+            result = asyncio.run(result)
+
```
lorenzejay, line 335: nice!

lorenzejay (approved):
LGTM

Mukhsin0508:
Does CrewAI agent call arun() by default when the tool subclasses have both run() and arun()? 

As far as we have seen, it is calling only the run() method even though the tool subclass has arun() method written. 


## Would love a quick clear explaination on this, as my app is getting future attached to different loop when calling it with kickoff_async(). But it works when I call it with kickoff itself only, yet it uses run() with kickoff. 


Thank you! Really waiting this to be fixed asap🙏🙏

greysonlalonde:
Hey @Mukhsin0508, this hasn't been wired up yet - next version will have it. Currently, `arun` is only planned to be called from native async contexts.

PR #4024, opened 2025-12-02
feat: async task support

greysonlalonde:
Depends on #4023

PR #4025, opened 2025-12-02
feat: async crew support

greysonlalonde:
Depends on #4024

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -618,14 +618,14 @@ wheels = [
 
 [[package]]
 name = "botocore-stubs"
-version = "1.42.2"
+version = "1.42.3"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/crew.py
```diff
@@ -948,6 +948,342 @@ def set_results_wrapper(result: Any) -> None:
...
+
+        Unlike kickoff_async which wraps sync kickoff in a thread, this method
+        uses native async/await for all operations including task execution,
+        memory operations, and knowledge queries.
+        """
+        if self.stream:
+            for agent in self.agents:
+                if agent.llm is not None:
+                    agent.llm.stream = True
+
```
lorenzejay: same as task, if we can make this DRY we gtg

lorenzejay (changes requested):

PR #4053, opened 2025-12-09
chore: add docs on native async

greysonlalonde:


lorenzejay (approved):

lorenzejay:
lets add the translations when we can @greysonlalonde

PR #4075, opened 2025-12-11
chore: display error message from response when tool repository login fails

heitorado:
> [!NOTE]
> On failed tool repo login, display HTTP status and server error message, with fallback for invalid JSON responses.
> 
> - **CLI (`crewai/cli/tools/main.py`)**:
>   - Improve login failure feedback in `ToolCommand.login()`:
>     - Prints HTTP status and server `message` from JSON response.
>     - Gracefully handles invalid JSON (`JSONDecodeError`) with a fallback message.
>     - Minor copy edit in the authentication failure prompt.
>   - Add import for `JSONDecodeError`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ddbc07437aea7805a257a41a3bef50bb65f7767f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

Local testing:
image

image

towards ENG-1395

### DIFF
--- a/lib/crewai/src/crewai/cli/tools/main.py
+++ b/lib/crewai/src/crewai/cli/tools/main.py
@@ -1,4 +1,5 @@
 import base64
+from json import JSONDecodeError
 import os
 from pathlib import Path
 import subprocess
@@ -165,10 +166,16 @@ class ToolCommand(BaseCommand, PlusAPIMixin):
                 "Authentication failed. Verify if the currently active organization can access the tool repository, and run 'crewai login' again.",
                 style="bold red",
             )
-            console.print(
-                f"[{login_response.status_code} error - {login_response.json().get('message', 'Unknown error')}]",
-                style="bold red italic",
-            )
+            try:
+                console.print(
+                    f"[{login_response.status_code} error - {login_response.json().get('message', 'Unknown error')}]",
+                    style="bold red italic",
+                )
+            except JSONDecodeError:
+                console.print(
+                    f"[{login_response.status_code} error - Unknown error - Invalid JSON response]",
+                    style="bold red italic",
+                )
             raise SystemExit
 
         login_response_json = login_response.json()


lib/crewai/src/crewai/cli/tools/main.py
```diff
@@ -162,9 +162,13 @@ def login(self) -> None:
 
         if login_response.status_code != 200:
             console.print(
-                "Authentication failed. Verify if the currently active organization access to the tool repository, and run 'crewai login' again. ",
+                "Authentication failed. Verify if the currently active organization can access the tool repository, and run 'crewai login' again.",
                 style="bold red",
             )
+            console.print(
+                f"[{login_response.status_code} error - {login_response.json().get('message', 'Unknown error')}]",
```
lucasgomide: IIRC calling json() may fail if the response body is HTML. It would be good to add a try/except block here to handle that scenario

greysonlalonde: Just checking if that's a known field in the response

heitorado: We do not expect to be getting HTML responses back since we should only be hitting internal API endpoints... I think that if this is the case (to receive an HTML response when I would be expecting JSON), I would want to see the exception instead of silencing it, because there would be something wrong with the AOP.

edit: added the try/except block because why not

heitorado: `status_code` is, since `login_response` is a `Response` object.

`message` is present at this specific endpoint on the API, I tested locally. I used `get` to ensure that if the 401 is coming from somewhere else, I would see a default message.

greysonlalonde: Cool! I'd just wrap in a `JSONDecodeError` and call it good

heitorado: Done at [ddbc074](https://github.com/crewAIInc/crewAI/pull/4075/commits/ddbc07437aea7805a257a41a3bef50bb65f7767f)

greysonlalonde (approved):

lib/crewai/src/crewai/cli/tools/main.py
greysonlalonde: Yeah I understand that, it's more pointed to 502/503s

PR #4097, opened 2025-12-16
Fixing NULL response from servers

### LINKED ISSUE #4093: [BUG] Received None or empty response from LLM call.
### Description

tool call is OFTEN failing when using gemini 3 pro model.
openai models are working fine.

```
@agent
    def morning_planning_agent(self) -> Agent:
        return Agent(
            config=cast(Dict[str, Any], self.agents_config)["morning_planning_agent"],
            allow_delegation=False,
            verbose=True,
            tools=self._create_tools(),
            llm=LLM(
                model="gemini-3-pro-preview",
                temperature=0.0,
                api_key=settings.GEMINI_API_KEY,
            ),
        )
```

Error:
```
Received None or empty response from LLM call.
An unknown error occurred. Please check the details below.
Error details: Invalid response from LLM call - None or empty.
An unknown error occurred. Please check the details below.                                                                             
Error details: Invalid response from LLM call - None or empty.
```

### Steps to Reproduce

configure gemini model with LLM()

### Expected behavior

it had to return content with action and action input

### Screenshots/Code snippets

see above

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

1.6.1

### crewAI Tools Version

1.6.1

### Virtual Environment

Venv

### Evidence

see above

### Possible Solution

maybe it's better to switch to normal tool calling instead of returning content with action and action input? because LLMs are trained for normall tool calling, isn't it?

### Additional context

.

Vidit-Ostwal:
Fixes #4093

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -289,7 +289,7 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
                 if e.__class__.__module__.startswith("litellm"):
                     # Do not retry on litellm errors
                     raise e
-                if is_context_length_exceeded(e):
+                if is_context_length_exceeded(e, messages=self.messages, llm=self.llm):
                     handle_context_length(
                         respect_context_window=self.respect_context_window,
                         printer=self._printer,
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -461,18 +461,31 @@ def handle_output_parser_exception(
     return formatted_answer
 
 
-def is_context_length_exceeded(exception: Exception) -> bool:
+def is_context_length_exceeded(exception: Exception, messages: list[LLMMessage], llm: LLM | BaseLLM) -> bool:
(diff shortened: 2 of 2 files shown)


twinkle77:
image

Vidit-Ostwal:
> image
> 



Can you share entire trace of logs?

Mukhsin0508:
maybe we can add a param for LLM() object to make the function calling either native or the way CrewAI does it? like:      
```
LLM(
   native_tool_calling=True
       )
```

This way would enable us to choose if we want to go with any llm's native way of calling tools or the way CrewAI reinvented the wheel. What do you think about it bro? @Vidit-Ostwal

Vidit-Ostwal:
> maybe we can add a param for LLM() object to make the function calling either native or the way CrewAI does it? like:
> 
> ```
> LLM(
>    native_tool_calling=True
>        )
> ```
> 
> This way would enable us to choose if we want to go with any llm's native way of calling tools or the way CrewAI reinvented the wheel. What do you think about it bro? @Vidit-Ostwal

Even I am not sure, around why crewai re-invented the wheel.
This issue is different though, we are sending a lot more in input_token then the system can doesn't respond at all.

Vidit-Ostwal:
Closing this might not be relevant with improved tool functionality.

PR #4119, opened 2025-12-18
fix: prevent crew inside flow to finish trace batch prematurely on failure

heitorado:
Adds the same checking we have on `CrewKickoffCompletedEvent` to `CrewKickoffFailedEvent`, to prevent premature end of trace batch when a nested crew inside a flow fails.

towards ENG-718


heitorado:

PR #4120, opened 2025-12-18
fix: always include timestamp on event data

heitorado:
Relates to #4119

This PR tweaks typing to be more reliable and removes redundant logic once we rely more on defined types.

It also ensures that a "timestamp" will ALWAYS be included in the `event_data` object stored inside a `TraceEvent`. Some events didn't have a timestamp and were relying on a default that could be temporally disconnected and this was messing the sorting.

towards ENG-718


heitorado:

PR #4121, opened 2025-12-18
feat: add flow failed event

heitorado:
related to #4120, #4119 

This PR:
- Adds a new event, `FlowFailedEvent`
- Emits the `FlowFailedEvent` on an `except` block which will be executed if any exceptions bubble up to the `kickoff` flow method
- Add listener to `FlowFailedEvent` that will wrap up and finalise the trace batch
- Remove redundant `type` parameters from `emit` call since those are already defined in the model
- Move batch finalisation on flow finished to the `FlowFinishedEvent` listener so we rely on the event bus to finalise the batch rather than the `kickoff` method body.

towards ENG-718


heitorado:

PR #4143, opened 2025-12-22
Adding HITL for Flows

joaomdmoura:
Introduce human feedback events and decorator for flow methods

- Implemented the @human_feedback decorator to facilitate human-in-the-loop workflows, allowing for feedback collection and routing based on responses.
- Enhanced Flow class to store human feedback history and manage feedback outcomes.
- Updated flow wrappers to preserve attributes from methods decorated with @human_feedback.
- Added integration and unit tests for the new human feedback functionality, ensuring proper validation and routing behavior.
- Added HumanFeedbackRequestedEvent and HumanFeedbackReceivedEvent classes to handle human feedback interactions within flows.

### DIFF
--- a/docs/ko/concepts/flows.mdx
+++ b/docs/ko/concepts/flows.mdx
@@ -565,6 +565,55 @@ Fourth method running
 
 이 Flow를 실행하면, `start_method`에서 생성된 랜덤 불리언 값에 따라 출력값이 달라집니다.
 
+### Human in the Loop (인간 피드백)
+
+`@human_feedback` 데코레이터는 인간의 피드백을 수집하기 위해 플로우 실행을 일시 중지하는 human-in-the-loop 워크플로우를 가능하게 합니다. 이는 승인 게이트, 품질 검토, 인간의 판단이 필요한 결정 지점에 유용합니다.
+
+```python Code
+from crewai.flow.flow import Flow, start, listen
+from crewai.flow.human_feedback import human_feedback, HumanFeedbackResult
+
+class ReviewFlow(Flow):
+    @start()
+    @human_feedback(
+        message="이 콘텐츠를 승인하시겠습니까?",
+        emit=["approved", "rejected", "needs_revision"],
+        llm="gpt-4o-mini",
+        default_outcome="needs_revision",
+    )
+    def generate_content(self):
+        return "검토할 콘텐츠..."
+
+    @listen("approved")
+    def on_approval(self, result: HumanFeedbackResult):
+        print(f"승인됨! 피드백: {result.feedback}")
+
+    @listen("rejected")
+    def on_rejection(self, result: HumanFeedbackResult):
+        print(f"거부됨. 이유: {result.feedback}")
+```
+
+`emit`이 지정되면, 인간의 자유 형식 피드백이 LLM에 의해 해석되어 지정된 outcome 중 하나로 매핑되고, 해당 `@listen` 데코레이터를 트리거합니다.
+
+라우팅 없이 단순히 피드백만 수집할 수도 있습니다:
+
+```python Code
+@start()
+@human_feedback(message="이 출력에 대한 코멘트가 있으신가요?")
+def my_method(self):
+    return "검토할 출력"
+
+@listen(my_method)
+def next_step(self, result: HumanFeedbackResult):
+    # result.feedback로 피드백에 접근
+    # result.output으로 원래 출력에 접근
+    pass
+```
+
+플로우 실행 중 수집된 모든 피드백은 `self.last_human_feedback` (가장 최근) 또는 `self.human_feedback_history` (리스트 형태의 모든 피드백)를 통해 접근할 수 있습니다.
+
+플로우에서의 인간 피드백에 대한 완전한 가이드는 비동기/논블로킹 피드백과 커스텀 프로바이더(Slack, 웹훅 등)를 포함하여 [Flow에서 인간 피드백](/ko/learn/human-feedback-in-flows)을 참조하세요.
+
 ## 플로우에 에이전트 추가하기
 
 에이전트는 플로우에 원활하게 통합할 수 있으며, 단순하고 집중된 작업 실행이 필요할 때 전체 Crew의 경량 대안으로 활용됩니다. 아래는 에이전트를 플로우 내에서 사용하여 시장 조사를 수행하는 예시입니다:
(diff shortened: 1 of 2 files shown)


docs/en/enterprise/features/deployment-api.mdx
vinibrsl, line 1: This API does not exist anymore and docs were included in https://docs.crewai.com/en/enterprise/guides/deploy-crew#option-3:-redeploy-using-api-ci/cd-integration

joaomdmoura: thanks, removed :)

lib/crewai/src/crewai/flow/flow.py
```diff
@@ -1435,10 +1501,190 @@ async def _execute_single_listener(
...
+            self._printer.print("  OUTPUT FOR REVIEW", color="bold_cyan")
+            self._printer.print("═" * 50 + "\n", color="bold_cyan")
+            self._printer.print(output)
+            self._printer.print("\n" + "═" * 50 + "\n", color="bold_cyan")
+
+            # Show message and prompt for feedback
+            self._printer.print(message, color="yellow")
+            self._printer.print("(Press Enter to skip, or type your feedback)\n", color="cyan")
+
+            feedback = input("Your feedback: ").strip()
```
vinibrsl, line 1968: Do you think we should make this implementation customizable, so we can accept different adapters, e.g. interactive input (this implementation), webhooks, etc?

joaomdmoura: I'm dwelling on making this here vs enterprise fork, but I'm thinking more and more about including this here, so I'm adding that in today, should we make this a separate PR maybe?

lib/crewai/src/crewai/flow/flow.py
```diff
@@ -1435,10 +1501,190 @@ async def _execute_single_listener(
...
+                message=message,
+                emit=list(emit) if emit else None,
+            ),
+        )
+
+        # Pause live updates during human input
+        event_listener.formatter.pause_live_updates()
+
+        try:
+            # Display output with formatting
```
greysonlalonde: Wondering here (and anywhere else we have `_printer` usage), can we convert to rich logging? It'll flow nicely with the terminal ui and make it easier to centralize the toggle for silencing

lib/crewai/src/crewai/flow/flow.py
greysonlalonde: I think separate PR would be good! Also, have chatted with @lorenzejay about this, but may be good to consider using an approach [like this](https://major.io/p/two-great-signals-sigstop-and-sigcont/), which would play nice with potential work on decoupling the event system and be generic enough to work with different adapters

docs/en/learn/human-feedback-in-flows.mdx
```diff
@@ -0,0 +1,373 @@
...
+## Quick Start
+
+Here's the simplest way to add human feedback to a flow:
+
+```python Code
+from crewai.flow.flow import Flow, start, listen
+from crewai.flow.human_feedback import human_feedback
+
+class SimpleReviewFlow(Flow):
+    @human_feedback(request="Please review this content:")
```
greysonlalonde: The initial kwarg in `human_feedback` is [`message`](https://github.com/crewAIInc/crewAI/pull/4143/changes#diff-d12596c04f7844831bf687f66ec6f2c65e31547a531e2de043b11e590532a749R128), I do think request may be a better term here though

docs/en/learn/human-feedback-in-flows.mdx
```diff
@@ -0,0 +1,373 @@
...
+3. Wait for the user to type feedback (or press Enter to skip)
+4. Pass a `HumanFeedbackResult` object to `process_feedback`
+
+## The @human_feedback Decorator
+
+### Parameters
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `request` | `str` | Yes | The message shown to the human alongside the method output |
```
greysonlalonde: Same note on initial kwarg https://github.com/crewAIInc/crewAI/pull/4143/changes#diff-d12596c04f7844831bf687f66ec6f2c65e31547a531e2de043b11e590532a749R128

lib/crewai/src/crewai/flow/human_feedback.py
```diff
@@ -0,0 +1,356 @@
...
+    Raises:
+        ValueError: If emit is specified but llm is not provided.
+        ValueError: If default_outcome is specified but emit is not.
+        ValueError: If default_outcome is not in the emit list.
+
+    Example:
+        Basic feedback without routing:
+        ```python
+        @start()
+        @human_feedback(message="Please review this output:")
```
greysonlalonde, line 215: Some of these examples show `@human_feedback` as innermost rather than outermost as specified in docs, could be potentially confusing down the road internally

lib/crewai/src/crewai/flow/flow.py
```diff
@@ -1435,10 +1501,190 @@ async def _execute_single_listener(
...
+
+        class FeedbackOutcome(BaseModel):
+            """The outcome that best matches the human's feedback intent."""
+
+            outcome: Literal[outcomes_tuple] = Field(  # type: ignore[valid-type]
+                description=f"The outcome that best matches the feedback. Must be one of: {', '.join(outcomes)}"
+            )
+
+        # Load prompt from translations
+        i18n = I18N()
```
greysonlalonde: Can use [`get_i18n`](https://github.com/crewAIInc/crewAI/blob/be70a041534dc0446b731c680b237d62aec7de61/lib/crewai/src/crewai/utilities/i18n.py#L115) here to take advantage of cache

lib/crewai/src/crewai/flow/human_feedback.py
```diff
@@ -0,0 +1,356 @@
...
+                    feedback=raw_feedback,
+                    outcome=collapsed_outcome,
+                    timestamp=datetime.now(),
+                    method_name=func.__name__,
+                    metadata=metadata or {},
+                )
+
+                # Store in flow instance
+                self.human_feedback_history.append(result)
+                self.last_human_feedback = result
```
greysonlalonde: Instead of storing as an instance variable, we may want to consider passing the result directly through the invocation chain, as we'll experience some race conditions sporadically if stored here

lib/crewai/src/crewai/flow/flow.py
joaomdmoura: the only problem is that SIGSTOP will stop the CPU usage but not RAM, so specially on enterprise deployments this could be a problem and maybe the server run OOM

greysonlalonde: Oh yeah that's a good point

PR #4192, opened 2026-01-07
feat: add a2a server config; agent card generation 

greysonlalonde:
- add `A2AServerConfig` for exposing agents as a2a servers with full `AgentCard` support                                                                          
 - add `A2AClientConfig` as the new client-side config, deprecate `A2AConfig`                                                                                        
 - inject `to_agent_card` method on agents with server config for A2A protocol compliance 

TODO:
- Follow for docs once several other prs are complete

### DIFF
--- a/lib/crewai/src/crewai/a2a/utils/agent_card.py
+++ b/lib/crewai/src/crewai/a2a/utils/agent_card.py
@@ -379,24 +379,21 @@ def _agent_to_agent_card(agent: Agent, url: str) -> AgentCard:
     )
 
 
-def inject_a2a_server_methods(target: Crew | Agent) -> None:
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/agent/core.py
```diff
@@ -765,11 +774,11 @@ def create_agent_executor(
     def _update_executor_parameters(
         self,
         task: Task | None,
-        tools: list,
+        tools: list[BaseTool],
         raw_tools: list[BaseTool],
-        prompt: dict,
+        prompt: SystemPromptResult | StandardPromptResult,
```
lorenzejay, line 779: lovely!

lorenzejay (approved):

PR #4226, opened 2026-01-12
feat: lazy import chromadb

gabemilani:
ChromaDB was imported at module level, causing slow imports even when not used. This was causing unnecessary delays. 

The changes from this PR are moving the imports to within where it's needed and using `TYPE_CHECKING` where it's just for typing hints.


gabemilani:
Holding on this for now.

Issue #4186, opened 2026-01-06
Parser fails on 'Action: None' - LLM wants direct response but uses wrong format


npkanaka:
When LLMs output `Action: None` (or variations like `Action: N/A`), the parser fails with `OutputParserError` and leaks internal "Thought:" text to users instead of providing a clean response.

### Problem

The ReAct format has a gap in handling scenarios where an LLM recognizes it should use a tool but can't/shouldn't. 

In these cases, LLMs commonly output:
```
Thought: [reasoning about why tool can't be used]
Action: None (direct response required)
```

This doesn't match the expected formats:
- `Action: [tool] + Action Input: {...}` → for tool calls
- `Final Answer: [answer]` → for direct responses  

The parser raises `OutputParserError`, and the error handler returns the raw text including "Thought:" prefix, creating poor UX.

### Environment

- CrewAI version: 1.5.0+
- LLM: Any ReAct-style LLM
- Python: 3.10+

PR #4229, opened 2026-01-13
Fix: prevent Thought leakage in final answers

npkanaka:
### Problem
Agents can leak internal ReAct control text (e.g., Thought:, Action:, Action Input:, Observation:) to end users in two ways:
1) format_answer() fails open on parse errors by returning the raw model output as AgentFinish.output.
2) Some models output a valid Final Answer: and then append Thought/Action/... afterward; the parser currently treats all text after Final Answer: as user-visible.

This shows up heavily with models that mimic prior Thought/Action/Observation patterns in history (e.g., Gemini), but it can happen with any provider.

### Changes
- Fail closed on parse errors: crewai.utilities.agent_utils.format_answer() no longer returns raw model output on parsing failures; it raises OutputParserError (wrapped for unexpected exceptions).
- Sanitize Final Answer outputs: crewai.agents.parser.parse() now strips trailing ReAct control blocks accidentally appended after Final Answer:
- Includes a guard to reduce false positives: only strips when the tail resembles a real control block (e.g., Action: only stripped if Action Input: appears).
- Does not strip markers inside `<code> / <pre><code>` spans.
- Max-iterations safety: handle_max_iterations_exceeded() catches OutputParserError on the last forced final-answer call and returns a safe generic message rather than crashing or leaking raw output.

This PR enforces a strict output contract at the framework boundary (where AgentFinish.output is produced), rather than relying on app-side sanitizers. It prevents both parse-failure leaks and parse-success-but-contaminated leaks, while keeping the agent retry loop functional and avoiding crashes in the max-iterations edge case.

++ https://github.com/crewAIInc/crewAI/issues/4186

---

> [!NOTE]
> Prevents leakage of internal ReAct control text to end users and hardens parsing behavior.
> 
> - Parser: When `Final Answer:` is present, strips trailing `Thought:`/`Action:`/`Action Input:`/`Observation:` blocks outside code spans via `_strip_trailing_react_blocks`; also normalizes stray trailing backticks.
> - Utilities: `format_answer()` now fails closed by raising `OutputParserError` (no raw output fallback); unexpected errors are wrapped into `OutputParserError` for retries.
> - Max-iterations: `handle_max_iterations_exceeded()` catches `OutputParserError` and returns a safe generic `AgentFinish` message instead of crashing or exposing raw model output.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11eacca9d7d40fa6d8ed9a3314031d3842615ad4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/agents/parser.py
+++ b/lib/crewai/src/crewai/agents/parser.py
@@ -6,6 +6,7 @@ AgentAction or AgentFinish objects.
 """
 
 from dataclasses import dataclass
+import re
 
 from json_repair import repair_json  # type: ignore[import-untyped]
 
@@ -96,6 +97,7 @@ def parse(text: str) -> AgentAction | AgentFinish:
 
     if includes_answer:
         final_answer = text.split(FINAL_ANSWER_ACTION)[-1].strip()
+        final_answer = _strip_trailing_react_blocks(final_answer)
         # Check whether the final answer ends with triple backticks.
         if final_answer.endswith("```"):
             # Count occurrences of triple backticks in the final answer.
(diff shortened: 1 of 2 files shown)


VedantMadane (commented):
Solid fix for a real security/UX issue. The "fail closed" approach is the right call here. Some observations:

**What I like:**
- _strip_trailing_react_blocks is well-designed - the code span detection prevents false positives in legitimate code output
- The guard conditions (checking for Action Input: presence) reduce over-stripping
- Converting unexpected exceptions to OutputParserError keeps the retry loop functioning without leaking raw output

**Minor suggestions:**

1. **Regex precompilation** - Good that _CODE_SPAN_RE and _TRAILING_REACT_MARKER_RE are module-level compiled patterns. 

2. **Edge case question** - What happens if the model outputs Final Answer: Here's how to use ReAct:\n\nThought: This is an example...? The current logic would strip the educational example. Might be worth documenting this known limitation, or checking if the "Thought:" appears within the first N characters after Final Answer.

3. **The max-iterations fallback message** is appropriately generic - good choice not to expose any internal state.

This addresses a real problem I've seen with Gemini models in particular. Nice work!

VedantMadane:
Friendly bump.

PR #4231, opened 2026-01-13
Increase frame inspection depth to detect `parent_flow`

vinibrsl:
This commit fixes a bug where `parent_flow` was not being set because the maximum depth was not sufficient to search for an instance of `Flow` in the current call stack frame during Flow instantiation.

### DIFF
--- a/lib/crewai/tests/test_crew.py
+++ b/lib/crewai/tests/test_crew.py
@@ -4500,6 +4500,71 @@ def test_crew_copy_with_memory():
         pytest.fail(f"Copying crew raised an unexpected exception: {e}")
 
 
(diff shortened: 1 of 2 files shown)


lib/crewai/tests/test_crew.py
```diff
@@ -4500,6 +4500,71 @@ def test_crew_copy_with_memory():
         pytest.fail(f"Copying crew raised an unexpected exception: {e}")
 
 
+def test_sets_parent_flow_when_using_crewbase_pattern_inside_flow():
```
vinibrsl, line 4503: Regression test

PR #4243, opened 2026-01-15
Lorenze/fix google vertex api using api keys

lorenzejay:


lib/crewai/src/crewai/llms/providers/gemini/completion.py
```diff
@@ -146,13 +159,34 @@ def _initialize_client(self, use_vertexai: bool = False) -> genai.Client:
...
         if self.client_params:
             client_params.update(self.client_params)
 
-        if use_vertexai or self.project:
+        # Determine authentication mode based on available credentials
+        has_api_key = bool(self.api_key)
+        has_project = bool(self.project)
+
+        if has_api_key and has_project:
+            logging.warning(
```
greysonlalonde, line 181: Should we raise here instead?

lorenzejay: nah, we just fallback to the api_key implementation. as one requires - `gcloud auth application-default login`

PR #4252, opened 2026-01-19
Fix LLM callback isolation without serializing requests

### LINKED ISSUE #4214: [BUG] Hidden race condition in LLM callback system causing test failures
### Description

Found a race condition in the LLM callback system. There's a test that only passes when a print statement is included, which is a classic sign of a timing-dependent bug (heisenbug). The developers left a TODO comment acknowledging they need to fix this eventually.

This indicates there's an async synchronization issue that could cause problems in production, especially in multi-threaded or high-concurrency environments.

### Steps to Reproduce

1. Look at lib/crewai/tests/test_llm.py line 21-22
2. The test `test_llm_callback_replacement` has a TODO comment
3. Remove any print statements or sleep calls from the test
4. The test becomes flaky and sometimes fails
5. Adding a print/sleep "fixes" it by introducing a delay

### Expected behavior

Tests should pass consistently without relying on print statements or arbitrary delays. The callback system should have proper synchronization mechanisms to ensure:

- Callbacks are properly registered before use
- No race conditions between callback registration and execution
- Event bus properly handles concurrent access
- Async operations are properly awaited/synchronized

### Screenshots/Code snippets

The problematic code is in `lib/crewai/tests/test_llm.py` lines 21-22:

# TODO: This test fails without print statement, which makes me think that 
# something is happening asynchronously that we need to eventually fix and 
# dive deeper into at a later date
@pytest.mark.vcr()
def test_llm_callback_replacement():
    llm1 = LLM(model="gpt-4o-mini", is_litellm=True)
    llm2 = LLM(model="gpt-4o-mini", is_litellm=True)
    # ... test codeThe TODO explicitly states this is an async issue that needs investigation.

### Operating System

Windows 11

### Python Version

3.10

### crewAI Version

Latest (main branch)

### crewAI Tools Version

N/A - core library

### Virtual Environment

Venv

### Evidence

Classic symptoms of a race condition:
- Test passes with print/sleep (introduces timing delay)
- Test fails without print/sleep (exposes the race)
- TODO comment explicitly acknowledging async issues
- Timing-dependent behavior indicates missing synchronization

This could manifest as:
- Callback not registered when LLM tries to use it
- Token usage metrics not properly updated
- Race between different LLM instances' callbacks

### Possible Solution

Need to investigate the event bus and callback registration system:

1. Review lib/crewai/src/crewai/events/event_bus.py for race conditions
2. Add proper async synchronization (locks, barriers, or async/await)
3. Ensure callback registration is atomic and thread-safe
4. Add proper event loop handling for async operations

This requires deeper investigation into the event bus architecture.
Might need to use threading.Lock, asyncio synchronization primitives,
or refactor to use proper async/await patterns throughout.

### Additional context

This is a more complex issue that requires investigation. The test comment
suggests the devs know about it but haven't had time to fix it properly.

While the workaround (print statement) makes tests pass, the underlying
race condition could cause production issues:
- Callbacks not firing correctly
- Missing telemetry/metrics
- Unreliable behavior under load
- Hard-to-debug intermittent failures

This should be prioritized for production reliability, especially for
users running high-concurrency workloads.

VedantMadane:
This is a follow-up to #4218 (auto-closed by bot) addressing the same race in LLM callback handling **without** holding a global lock across the network call.

### What changed
- Stop mutating LiteLLM global callback lists for per-request callbacks.
- Pass callbacks via the request params ("callbacks") and continue to invoke token usage callbacks from CrewAI response handlers.
- Make `test_llm_callback_replacement` deterministic by mocking `litellm.completion` (removes sleep/heisenbug).

### Why
The approach in #4218 used a class-level lock held across the entire LLM request which can serialize all concurrent agent calls. This keeps concurrency while still ensuring callback isolation.

Fixes #4214.

---

> [!NOTE]
> **High Risk**
> Touches core `LLM.call`/`LLM.acall` request plumbing and callback behavior, which can regress token tracking and integrations under concurrency. The async error-handling block appears to contain duplicated/stray code that could break `acall` at runtime.
> 
> **Overview**
> **Fixes callback race conditions across concurrent LiteLLM calls** by stopping CrewAI from mutating LiteLLM's global callback lists and instead passing `callbacks` on the per-request params for both sync and async code paths.
> 
> Removes the `LLM.set_callbacks` global-deduplication helper and updates tests: makes `test_llm_callback_replacement` deterministic by mocking `litellm.completion`, and adds a new threaded concurrency test to assert callback and token-usage isolation between simultaneous requests.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 877d02100a9597e34a44b5147ecb5186c433699b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


VedantMadane:
Not covered in this PR description:

1) Lock scoping alternative (save previous global callbacks, set new ones, perform request, then restore) and why we avoided it.
2) Context local callback isolation using contextvars or thread local dispatch.
3) A true concurrency regression test (multi thread or async) that proves no cross contamination under parallel calls.

If you prefer, I can add a follow up commit that documents these options or adds a concurrency focused test.

VedantMadane:
This PR is marked changes-requested mainly from CodeRabbit; happy to resolve specific bot items or close if the approach is not desired. Prefer human feedback on whether callback isolation without global mutation is the right direction.

VedantMadane:
Automated fix: rebased onto main (cherry-picked feature commit; dropped merge-commit history).

VedantMadane:
Addressed the remaining CodeRabbit reviewer feedback:

1. **Callback reset isolation in `set_env_callbacks`** — Updated `LLM.set_env_callbacks` so `failure_callbacks` is initialized to `[]` and both `litellm.success_callback` and `litellm.failure_callback` are assigned unconditionally outside the `if failure_callbacks_str:` block. This prevents stale global callbacks when `LITELLM_FAILURE_CALLBACKS` is unset.
2. **Enhanced test mock response** — Updated `_mock_response` in `test_llm_callback_replacement` to populate `model_extra["usage"]` and explicitly set `mock_message.tool_calls = None` so response usage parsing and callback dispatch paths are fully exercised.

Signed-off-by: Vedant Madane <vedantnm@gmail.com>

PR #4266, opened 2026-01-22
feat: native multimodal file handling

greysonlalonde:
- add `input_files` parameter to `Crew.kickoff()`, `Flow.kickoff()`, `Task`, and `Agent.kickoff()`                                        
- add provider-specific file uploaders for OpenAI, Anthropic, Gemini, and Bedrock                                                         
- add file type detection, constraint validation, and automatic format conversion                                                         
- add URL file source support for multimodal content                                                                                      
- add streaming uploads for large files                                                                                                   
- add prompt caching support for Anthropic                                                                                                
- add OpenAI Responses API support

### DIFF
--- a/lib/crewai/tests/test_flow.py
+++ b/lib/crewai/tests/test_flow.py
@@ -1343,12 +1343,21 @@ def test_deeply_nested_conditions():
     assert "c" in execution_order
     assert "d" in execution_order
 
-    # Result should execute after all starts
+    # Result should execute after at least one AND condition is satisfied
+    # With or_(and_(a, b), and_(c, d)), result fires when EITHER:
+    # - Both a AND b have completed, OR
+    # - Both c AND d have completed
     assert "result" in execution_order
-    assert execution_order.index("result") > execution_order.index("a")
-    assert execution_order.index("result") > execution_order.index("b")
-    assert execution_order.index("result") > execution_order.index("c")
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/llms/providers/openai/completion.py
```diff
@@ -1074,3 +1074,48 @@ def _format_messages(self, messages: str | list[LLMMessage]) -> list[LLMMessage]
...
+            True if the model supports images.
+        """
+        vision_models = (
+            "gpt-4o",
+            "gpt-4.1",
+            "gpt-4-turbo",
+            "gpt-4-vision",
+            "o1",
+            "o3",
+            "o4",
```
lorenzejay, line 2248: i think the 5 series models support too
Screenshot 2026-01-22 at 8 09 33 PM

https://platform.openai.com/docs/guides/images-vision

lib/crewai/src/crewai/task.py
```diff
@@ -142,6 +158,10 @@ class Task(BaseModel):
         default_factory=list,
         description="Tools the agent is limited to use for this task.",
     )
+    input_files: list[Any] = Field(
```
lorenzejay: do we support this with 

agent.kickoff(...)

greysonlalonde:
> nice ! can we add example docs

Fast follow on these tomorrow

PR #4281, opened 2026-01-26
Add call_id to LLM events for correlating requests

vinibrsl:
When monitoring LLM events, consumers need to know which events belong to the same API call. Before this change, there was no way to correlate LLMCallStartedEvent, LLMStreamChunkEvent, and LLMCallCompletedEvent belonging to the same request.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new required field on all `LLMEventBase` events and threads it through multiple providers and streaming paths, which could break any remaining event emissions that don’t set `call_id` or rely on prior event schemas.
> 
> **Overview**
> **Adds `call_id` correlation to LLM events.** `LLMEventBase` now includes a required `call_id`, and all LLM lifecycle events (started/chunks/completed/failed) are updated to emit this identifier.
> 
> **Implements call scoping and propagation.** Introduces `llm_call_context()` + `get_current_call_id()` (contextvars-backed) in `BaseLLM`, wraps `LLM.call`/`LLM.acall` and native provider `call`/`acall` implementations to establish a call scope, and attaches the current `call_id` to streaming chunks and error events.
> 
> **Updates and adds tests.** Adjusts existing streaming/event tests to include `call_id`, and adds VCR-backed tests + cassettes asserting same-call events share an ID and separate calls generate distinct IDs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3059c485aca2a9cd28afe22f4ab1b32466218fdf. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/events/types/llm_events.py
+++ b/lib/crewai/src/crewai/events/types/llm_events.py
@@ -10,6 +10,7 @@ class LLMEventBase(BaseEvent):
     from_task: Any | None = None
     from_agent: Any | None = None
     model: str | None = None
+    call_id: str
 
     def __init__(self, **data: Any) -> None:
         if data.get("from_task"):
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -37,7 +37,7 @@ from crewai.events.types.tool_usage_events import (
     ToolUsageFinishedEvent,
     ToolUsageStartedEvent,
 )
-from crewai.llms.base_llm import BaseLLM
+from crewai.llms.base_llm import BaseLLM, get_current_call_id, llm_call_context
 from crewai.llms.constants import (
     ANTHROPIC_MODELS,
     AZURE_MODELS,
@@ -770,7 +770,7 @@ class LLM(BaseLLM):
                 chunk_content = None
                 response_id = None
 
-                if hasattr(chunk,'id'):
+                if hasattr(chunk, "id"):
                     response_id = chunk.id
 
                 # Safely extract content from various chunk formats
@@ -827,7 +827,7 @@ class LLM(BaseLLM):
                                         available_functions=available_functions,
                                         from_task=from_task,
                                         from_agent=from_agent,
-                                        response_id=response_id
+                                        response_id=response_id,
                                     )
 
                                     if result is not None:
@@ -849,7 +849,8 @@ class LLM(BaseLLM):
                             from_task=from_task,
                             from_agent=from_agent,
                             call_type=LLMCallType.LLM_CALL,
-                            response_id=response_id
+                            response_id=response_id,
+                            call_id=get_current_call_id(),
                         ),
                     )
             # --- 4) Fallback to non-streaming if no content received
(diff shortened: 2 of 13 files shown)


Vidit-Ostwal:
Hey @vinibrsl,
Not really sure, but this PR was merged which kind of does the same functionality 
https://github.com/crewAIInc/crewAI/pull/3842

While streaming, there is no way to figure out which streaming chunks belongs to same chunk event, this PR add the response_id which is directly given by the provider.

Though this only targets LLMStreamChunkEvent
But I can extend it to LLMCallStartedEvent and LLMCallCompletedEvent as well.

vinibrsl:
> Hey @vinibrsl, Not really sure, but this PR was merged which kind of does the same functionality #3842
> 
> While streaming, there is no way to figure out which streaming chunks belongs to same chunk event, this PR add the response_id which is directly given by the provider.
> 
> Though this only targets LLMStreamChunkEvent But I can extend it to LLMCallStartedEvent and LLMCallCompletedEvent as well.

Thanks, @Vidit-Ostwal! That's great context.

Unless I'm missing something here, `call_id` serves a different purpose. When emitting `LLMCallStartedEvent`, for example, we don't have the response ID yet. Therefore, having some kind of ID that correlates all LLM events is still needed.

Let me know if you think we should consolidate somehow or if keeping both makes sense.

Vidit-Ostwal:
> > Hey @vinibrsl, Not really sure, but this PR was merged which kind of does the same functionality #3842
> > While streaming, there is no way to figure out which streaming chunks belongs to same chunk event, this PR add the response_id which is directly given by the provider.
> > Though this only targets LLMStreamChunkEvent But I can extend it to LLMCallStartedEvent and LLMCallCompletedEvent as well.
> 
> Thanks, @Vidit-Ostwal! That's great context.
> 
> Unless I'm missing something here, `call_id` serves a different purpose. When emitting `LLMCallStartedEvent`, for example, we don't have the response ID yet. Therefore, having some kind of ID that correlates all LLM events is still needed.
> 
> Let me know if you think we should consolidate somehow or if keeping both makes sense.

This makes more sense,

The PR I add only targets the LLMStreamChunkEvent, which only if successful will be populated, but there are also other LLMStream events which needs to be coupled together, this PR adds that.

Said that, this does makes the previous PR a bit redundant, as response_id has smaller scope, than call_id. which is trying to do the same functionality.
lmk if you think we should drop it, can make a PR for that.

vinibrsl:
> Said that, this does makes the previous PR a bit redundant, as response_id has smaller scope, than call_id. which is trying to do the same functionality.

We could follow up by marking response_id as deprecated, and eventually drop it in the next version. Thanks!

Vidit-Ostwal:
> > Said that, this does makes the previous PR a bit redundant, as response_id has smaller scope, than call_id. which is trying to do the same functionality.
> 
> We could follow up by marking response_id as deprecated, and eventually drop it in the next version. Thanks!

It was merged today, 8 hrs back.

lib/crewai/src/crewai/events/types/llm_events.py
```diff
@@ -10,6 +10,7 @@ class LLMEventBase(BaseEvent):
     from_task: Any | None = None
     from_agent: Any | None = None
     model: str | None = None
+    call_id: str
```
greysonlalonde, line 13: checking - is this for sure provided by external callers?

vinibrsl: Yes, if emitted through BaseLLM or LLM

PR #4287, opened 2026-01-27
feat: implement before and after tool call hooks in CrewAgentExecutor…

lorenzejay:
… and AgentExecutor

- Added support for before and after tool call hooks in both CrewAgentExecutor and AgentExecutor classes.
- Introduced ToolCallHookContext to manage context for hooks, allowing for enhanced control over tool execution.
- Implemented logic to block tool execution based on before hooks and to modify results based on after hooks.
- Added integration tests to validate the functionality of the new hooks, ensuring they work as expected in various scenarios.
- Enhanced the overall flexibility and extensibility of tool interactions within the CrewAI framework.

### DIFF
--- a/lib/crewai/tests/hooks/test_tool_hooks.py
+++ b/lib/crewai/tests/hooks/test_tool_hooks.py
@@ -602,9 +602,6 @@ class TestNativeToolCallingHooksIntegration:
         from crewai import Agent
         from crewai.tools import tool
 
-        if not os.environ.get("OPENAI_API_KEY"):
-            pytest.skip("OPENAI_API_KEY not set")
-
         hook_calls = {"before": [], "after": []}
 
         @tool("multiply_numbers")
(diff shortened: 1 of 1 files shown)


lib/crewai/tests/hooks/test_tool_hooks.py
```diff
@@ -590,3 +590,241 @@ def after_tool_call_hook(context: ToolCallHookContext) -> str | None:
...
+
+    @pytest.mark.vcr()
+    def test_agent_native_tool_hooks_before_and_after(self):
+        """Test that Agent with native tool calling executes before/after hooks."""
+        import os
+        from crewai import Agent
+        from crewai.tools import tool
+
+        if not os.environ.get("OPENAI_API_KEY"):
+            pytest.skip("OPENAI_API_KEY not set")
```
greysonlalonde: See a couple of these, should probably not skip, right? Could show on future regens

PR #4288, opened 2026-01-27
fix: gemini tool calling

greysonlalonde:
use pydantic_schema_utils for platform tool schema parsing and fix API payload structure

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/gemini/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/gemini/completion.py
@@ -493,7 +493,7 @@ class GeminiCompletion(BaseLLM):
             function_declaration = types.FunctionDeclaration(
                 name=name,
                 description=description,
-                parameters=parameters if parameters else None,
+                parameters_json_schema=parameters if parameters else None,
             )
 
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/hooks/llm_hooks.py
```diff
@@ -41,7 +42,7 @@ class LLMCallHookContext:
             Can be modified by returning a new string from after_llm_call hook.
     """
 
-    executor: CrewAgentExecutor | LiteAgent | None
+    executor: CrewAgentExecutor | AgentExecutor | LiteAgent | None
```
lorenzejay, line 45: great catch here!

lib/crewai/src/crewai/llms/providers/gemini/completion.py
```diff
@@ -493,7 +493,7 @@ def _convert_tools_for_interference(  # type: ignore[override]
             function_declaration = types.FunctionDeclaration(
                 name=name,
                 description=description,
-                parameters=parameters if parameters else None,
+                parameters_json_schema=parameters if parameters else None,
```
lorenzejay, line 496: lovely

PR #4302, opened 2026-01-29
feat: add a2a liteagent, auth, transport negotiation, and file support

greysonlalonde:
- add server-side auth (api key, bearer, basic/digest, mtls)
- introduce protocol extensions with middleware and validation
- implement automatic transport negotiation with fallbacks
- add flexible content-type parsing and encoding
- enable liteagent delegation via task adapter
- support end-to-end file input with base64 encoding/decoding

### DIFF
--- a/lib/crewai/tests/agents/test_agent_a2a_kickoff.py
+++ b/lib/crewai/tests/agents/test_agent_a2a_kickoff.py
@@ -35,6 +35,7 @@ class TestAgentA2AKickoff:
             ],
         )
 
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/lite_agent.py
```diff
@@ -84,6 +88,81 @@
 from crewai.utilities.types import LLMMessage
 
 
+def _kickoff_with_a2a_support(
```
lorenzejay, line 91: is this the place to add this or on core.py kickoff?

lorenzejay (approved):
lgtm

PR #4311, opened 2026-01-30
fix: Make PickleHandler thread-safe (Fixes #4215)

### LINKED ISSUE #4215: [BUG] File handling not thread-safe, requires workarounds in tests
### Description

Found that the PickleHandler class isn't thread-safe, so tests have to use workarounds with UUID-based filenames to avoid race conditions during parallel test execution. This suggests the underlying file handling implementation has concurrency issues.

The workaround masks the real problem - the file handler should be thread-safe by default, not require each test to use unique filenames.

### Steps to Reproduce

1. Look at lib/crewai/tests/utilities/test_file_handler.py lines 11-14
2. Tests generate unique filenames with UUID to avoid conflicts
3. Comment says "to avoid race conditions in parallel test execution"
4. Try running tests in parallel without unique filenames - they fail
5. This indicates the PickleHandler isn't handling concurrent access properly

### Expected behavior

PickleHandler should be thread-safe and handle concurrent access properly:

- Multiple threads/processes should be able to read simultaneously
- Write operations should be atomic or properly locked
- No need for test workarounds with unique filenames
- Should use proper file locking mechanisms (like portalocker which is already in dependencies)

### Screenshots/Code snippets

The workaround is in `lib/crewai/tests/utilities/test_file_handler.py` lines 11-14:

def setUp(self):
    # Use a unique file name for each test to avoid race conditions 
    # in parallel test execution
    unique_id = str(uuid.uuid4())
    self.file_name = f"test_data_{unique_id}.pkl"This is a workaround, not a fix. The real issue is in the PickleHandler implementation.

### Operating System

Windows 11

### Python Version

3.10

### crewAI Version

Latest (main branch)

### crewAI Tools Version

N/A - core utilities

### Virtual Environment

Venv

### Evidence

Clear indicators of concurrency issues:
- Test comment explicitly mentions "avoid race conditions"
- UUID workaround needed for parallel test execution
- Suggests underlying PickleHandler isn't thread-safe
- Project already has portalocker dependency but isn't using it

This pattern suggests:
- PickleHandler.save() isn't atomic
- No file locking during read/write operations
- Concurrent access causes corruption or conflicts

### Possible Solution

Implement proper file locking in PickleHandler:

1. Use portalocker (already in dependencies) for file locking
2. Implement atomic write operations (write to temp, then rename)
3. Add read/write locks using context managers
4. Ensure operations are thread-safe

Example approach:
import portalocker

def save(self, data):
    with portalocker.Lock(self.file_path, 'w') as file:
        pickle.dump(data, file)
        
def load(self):
    with portalocker.Lock(self.file_path, 'r') as file:
        return pickle.load(file)This would eliminate the need for UUID workarounds in tests.
`

### Additional context

This affects:
- Parallel test execution (currently worked around)
- Production code using PickleHandler concurrently
- Any automation that might access files simultaneously

While tests work around this with UUIDs, production code using PickleHandler
might face data corruption or race conditions if multiple processes/threads
access the same pickle file.

The project already includes portalocker~=2.7.0 in dependencies, so we have
the tools to fix this properly - just need to implement it in PickleHandler.

ksharma6309:
fix: Make PickleHandler thread-safe (Fixes #4215)

---

> [!NOTE]
> **Medium Risk**
> Touches low-level persistence: adds locking and atomic replace semantics, which can affect cross-platform file behavior (especially Windows) and introduce lock-timeout failures under contention.
> 
> **Overview**
> `PickleHandler` now uses `portalocker`-backed file locks plus *atomic temp-file writes + `os.replace`* to prevent corruption under concurrent access, and exposes a configurable `lock_timeout`.
> 
> Tests were updated to use temp files and expanded with a new multiprocessing-based concurrency suite that exercises concurrent readers/writers and timeout configuration.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c12eab3363e1190876cdc699f9010ed93cdd1b56. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/utilities/file_handler.py
+++ b/lib/crewai/src/crewai/utilities/file_handler.py
@@ -2,8 +2,10 @@ from datetime import datetime
 import json
 import os
 import pickle
+import tempfile
 from typing import Any, TypedDict
 
+import portalocker
 from typing_extensions import Unpack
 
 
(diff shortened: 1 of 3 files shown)


ksharma6309:
Issue will be fixed on other

PR #4325, opened 2026-02-01
fix: cast latest_n to int in LTMSQLiteStorage to prevent SQL injection

RinZ27:
I noticed that the 'latest_n' parameter was being directly interpolated into the SQL query strings in 'LTMSQLiteStorage'. While it's expected to be an integer, casting it explicitly to 'int' ensures that no malicious strings can be injected into the 'LIMIT' clause, especially in an AI agent context where input sources can be unpredictable. This also allows for better error handling if an invalid type is passed.

---

> [!NOTE]
> **Low Risk**
> No changes are introduced, so there is effectively no risk.
> 
> **Overview**
> This PR contains no code changes (the provided diff is empty), so there are no functional or behavioral modifications to review.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b8bd1c6a549b668bfb2f36008e94623e73ed2ed2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


RinZ27:
Resolved the merge conflict by accepting the deletion of `ltm_sqlite_storage.py`. The codebase has recently moved to a new unified memory system that replaced the old SQLite implementation with a vector-based approach (LanceDB). 

Since the target file no longer exists in the main branch and the new system doesn't use raw SQL interpolation for limits, this fix is technically no longer applicable. I've updated the branch to reflect the current state of main.

PR #4392, opened 2026-02-06
fix: event ordering

greysonlalonde:
fix: thread-safe state proxies & native routing
add thread-safe state proxies and refactor native tool routing.

* introduce `LockedListProxy` and `LockedDictProxy` in `flow.py` and update `StateProxy` to return them for list/dict attrs so mutations are protected by the flow lock.
* update `AgentExecutor` to use `StateProxy` on flow init, guard the messages setter with the state lock, and return a `StateProxy` from the temp state accessor.
* convert `call_llm_native_tools` into a listener (no direct routing return) an

expand the loop bridge listener to also trigger on tool completion events (`tool_completed` and `native_tool_completed`) so agent iteration resumes after tools finish. add a `requests.post` mock and response fixture in the liteagent test to simulate platform tool execution. refresh and sanitize vcr cassettes (updated model responses, timestamps, and header placeholders) to reflect tool-call flows and new recordings.


lib/crewai/src/crewai/experimental/agent_executor.py
```diff
@@ -225,7 +226,11 @@ def messages(self) -> list[LLMMessage]:
     @messages.setter
     def messages(self, value: list[LLMMessage]) -> None:
         """Delegate to state for ExecutorContext conformance."""
-        self._state.messages = value
+        if self._flow_initialized and hasattr(self, "_state_lock"):
```
lorenzejay, line 229: looking good

PR #4420, opened 2026-02-08
New Unified Memory System

joaomdmoura:
# Summary

Replaces the four separate memory types (`ShortTermMemory`, `LongTermMemory`, `EntityMemory`, `ExternalMemory`) with a single `Memory` class where both encoding and retrieval are LLM-driven. Results are ranked by a composite score blending semantic similarity, recency, and importance -- not just vector distance.

# Key changes

### **Unified `Memory` class.** 
One API: `remember()`, `recall()`, `forget()`. On save, the LLM infers scope, categories, and importance. On recall, a multi-step `RecallFlow` assesses query complexity, monitors its own confidence, and deepens search when uncertain. A `ConsolidationFlow` on save detects near-duplicate records and merges/updates/deletes them automatically.

### **Composite scoring.** 
`semantic_weight * similarity + recency_weight * decay + importance_weight * importance`. Applied in both shallow and deep recall paths, including the confidence router that decides "synthesize vs explore deeper."

### **Hierarchical scopes.**
Memories organize into a tree (`/company/engineering`, `/project/alpha`). `MemoryScope` restricts operations to a subtree; `MemorySlice` combines multiple scopes into one view.

### **Config lives on `Memory`.**
All tunable knobs are kwargs: `Memory(recency_weight=0.5, exploration_budget=2)`. No separate config object needed.

### **Simplified `Crew.memory`.** 
Single field replaces `memory` + `memory_instance`. `Crew(memory=True)` creates defaults; `Crew(memory=Memory(...))` for custom config.

##$ **Memory TUI.** 
`crewai memory` opens a terminal UI for browsing scopes, viewing entries with pagination, and running recall queries. `textual` is now a core dependency.

##$ **Flow proxy fix.**
`LockedListProxy` and `LockedDictProxy` now subclass `list`/`dict` so `isinstance` checks pass in LanceDB, Pydantic, etc.

# How memory integrates across the stack

### **Crew.** 
`Crew(memory=True)` creates a default `Memory()`. Pass `Crew(memory=Memory(...))` for custom tuning. All agents in the crew share it. After each task, the executor calls `extract_memories()` to decompose the raw result into atomic facts, then `remember()` on each -- so the crew's knowledge base grows automatically as tasks complete.

### **Agent.** 
- During task execution, the agent recalls relevant memories and injects them into the task context. The agent uses the crew's shared memory unless it has its own `memory` field set, in which case it uses that (can be a `Memory`, `MemoryScope`, or `MemorySlice`).

### **LiteAgent.** 
Same pattern: `LiteAgent(memory=True)` or `LiteAgent(memory=Memory(...))`. Before each turn, it recalls memories matching the user's latest message and appends them to the system prompt. After each turn, it extracts and remembers facts from the response.

### **Flow.** 
Every Flow automatically gets a `Memory` instance. Flow methods can call `self.remember()`, `self.recall()`, and `self.extract_memories()` directly -- memory is a first-class capability, not something wired up externally.

# What was removed

`ShortTermMemory`, `LongTermMemory`, `EntityMemory`, `ExternalMemory`, `ContextualMemory`, and their storage backends (`rag_storage.py`, `ltm_sqlite_storage.py`, `mem0_storage.py`). All replaced by unified `Memory` + `LanceDBStorage`.


docs/en/concepts/memory.mdx
```diff
@@ -1,1261 +1,320 @@
...
-        "config": {
-            "api_key": "your-openai-api-key",  # Optional: override env var
-            "model_name": "text-embedding-3-large",
-            "dimensions": 1536,  # Optional: reduce dimensions for smaller storage
-            "organization_id": "your-org-id"  # Optional: for organization accounts
-        }
-    }
+    process=Process.sequential,
+    memory_instance=memory,
+    verbose=True,
```
mplachta: docs remove embedder configuration, I'd add it back because it's pretty much mandatory for non-native OPENAI

docs/en/concepts/memory.mdx
```diff
@@ -1,1261 +1,320 @@
...
-            "model_name": "gemini-embedding-001"  # or "text-embedding-005", "text-multilingual-embedding-002"
-        }
-    }
-)
-```
+from crewai import Agent, Memory
 
-### Vertex AI Embeddings
+memory = Memory()
+crew_memory = memory.scope("/crew/project-alpha")
```
mplachta: Can we have better examples? This one feels artificial

docs/en/concepts/memory.mdx
```diff
@@ -1,1261 +1,320 @@
...
-            "model": "mxbai-embed-large",  # or "nomic-embed-text"
-            "url": "http://localhost:11434/api/embeddings"  # Default Ollama URL
-        }
-    }
-)
+class ResearchFlow(Flow):
+    @start()
+    def research(self):
+        results = self.do_research()
+        self.remember(f"Research found: {results}", scope="/research")
```
mplachta: I think it would be better to do `self._memory.remember()` etc. Some existing flows may already utilize remember and recall functions.

Additionally, this needs to be documented in flow too.

docs/en/concepts/memory.mdx
```diff
@@ -1,1261 +1,320 @@
...
+        past = self.recall("previous research findings")
+        # Use past memories to contextualize
+        return self.synthesize(research_results, past)
 ```
 
-### Cohere Embeddings
+## Hierarchical Scopes
 
-Use Cohere's embedding models for multilingual support.
+Memories are organized in a scope tree (e.g. `/`, `/project`, `/project/alpha`). When you don't pass a scope to `remember()`, the LLM suggests one based on content and existing scopes.
```
mplachta: I'd love a best practice here, eg.
* `/` for entire project (eg. flow)
* `/crew_name` for a crew (standalone or inside flow
* `/crew_name/agent_name` for an agent

Alternatively people can also add a prefix or postfix such as env (dev/staging/prod)

docs/en/concepts/memory.mdx
```diff
@@ -1,1261 +1,320 @@
...
-        "provider": "voyageai",
-        "config": {
-            "api_key": "your-voyage-api-key",
-            "model": "voyage-3",  # or "voyage-3-lite", "voyage-code-3"
-            "input_type": "document"  # or "query"
-        }
-    }
-)
+# Narrower subscope
+project_memory = agent_memory.subscope("project-alpha")
```
mplachta, line 234: Practical example?

docs/en/concepts/memory.mdx
```diff
@@ -1,1261 +1,320 @@
...
-        "provider": "bedrock",
-        "config": {
-            "aws_access_key_id": "your-access-key",
-            "aws_secret_access_key": "your-secret-key",
-            "region_name": "us-east-1",
-            "model": "amazon.titan-embed-text-v1"
-        }
-    }
+# Read-only: recall across /agent/researcher and /company/knowledge; remember() raises
+view = memory.slice(
```
mplachta: Practical example?

PR #4432, opened 2026-02-10
fix(executor): reset messages and iterations between task executions

### LINKED ISSUE #4389: [BUG] CrewAgentExecutor does not reset messages and iterations between task executions
### Description

When a CrewAI agent executes multiple sequential tasks within a Crew, the same CrewAgentExecutor instance is reused across tasks. 
The problem is that CrewAgentExecutor.invoke() and CrewAgentExecutor.ainvoke() do not reset self.messages and self.iterations before starting a new execution. 

The experimental AgentExecutor (crewai/experimental/agent_executor.py) does not have this bug, it correctly resets all execution state at the beginning of invoke() and invoke_async().

### Steps to Reproduce

1. Create a Crew with one agent assigned to two sequential tasks
2. Execute the crew 
3. Inspect the message history

### Expected behavior

Each task execution should start with a clean message history and iterations reset to 0. Task 2 should not see any messages from task 1's execution. The LLM context should contain only the system/user prompts relevant to the current task.

### Screenshots/Code snippets

Image

### Operating System

Ubuntu 20.04

### Python Version

3.11

### crewAI Version

1.9.3

### crewAI Tools Version

1.9.3

### Virtual Environment

Venv

### Evidence

None

### Possible Solution

I implement it in a PR.

### Additional context

None

Chase-Xuu:
## Description
Fixes #4389

When `CrewAgentExecutor` is reused across multiple sequential tasks within a Crew, it was not resetting its internal state (`messages` and `iterations`) before starting a new task execution.

## Problem
When an agent executes multiple sequential tasks:
1. **Task 2 sees messages from Task 1** - The `messages` list accumulates across invocations, so the LLM context contains conversation history from previous tasks
2. **Iterations counter doesn't reset** - The `iterations` counter continues from where the previous task left off, potentially triggering max iterations limit earlier than expected

## Root Cause
The `invoke()` and `ainvoke()` methods call `_setup_messages()` which **appends** to `self.messages` instead of first clearing it. Neither method resets `self.iterations` to 0.

## Solution
Added state reset at the beginning of both `invoke()` and `ainvoke()` methods:

```python
# Reset execution state for fresh execution
# This ensures clean state when executor is reused across tasks
self.messages = []
self.iterations = 0
```

This matches the behavior of the experimental `AgentExecutor` (in `crewai/experimental/agent_executor.py`) which already correctly resets all execution state at the beginning of each invocation.

## Testing
Added new test file `test_crew_agent_executor_state_reset.py` with tests verifying:
- `test_messages_reset_on_invoke` - Messages from previous execution are cleared
- `test_iterations_reset_on_invoke` - Iterations counter resets to 0
- `test_messages_reset_on_ainvoke` - Same for async version
- `test_iterations_reset_on_ainvoke` - Same for async version

## Impact
- **Low risk** - Simple, targeted fix that only adds initialization at the start of methods
- **No breaking changes** - Existing behavior for single-task execution is unchanged
- **Aligns with experimental executor** - Matches the state reset pattern already used in the newer Flow-based executor

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -210,6 +210,11 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
         Returns:
             Dictionary with agent output.
         """
+        # Reset execution state for fresh execution
+        # This ensures clean state when executor is reused across tasks
+        self.messages = []
+        self.iterations = 0
+
         self._setup_messages(inputs)
 
         self._inject_multimodal_files(inputs)
@@ -987,6 +992,11 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
         Returns:
             Dictionary with agent output.
         """
+        # Reset execution state for fresh execution
+        # This ensures clean state when executor is reused across tasks
+        self.messages = []
+        self.iterations = 0
+
         self._setup_messages(inputs)
 
         await self._ainject_multimodal_files(inputs)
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:

PR #4445, opened 2026-02-10
fix: replace @lru_cache with instance-level caching in Repository.is_git_repo()

### LINKED ISSUE #4210: [BUG] Memory leak in Repository.is_git_repo() due to @lru_cache decorator
### Description

Found a memory leak in the CLI's Repository class. The is_git_repo() method uses @lru_cache which holds references to self and prevents garbage collection. There's actually a TODO comment in the code acknowledging this issue but it hasn't been fixed yet.

This becomes a problem in long-running processes or scripts that create multiple Repository instances - the memory just keeps growing.

### Steps to Reproduce

1. Create a script that instantiates Repository objects multiple times
2. Run it in a loop (like in an automated CI/CD process)
3. Monitor memory usage - it keeps climbing
4. Repository instances never get garbage collected even after going out of scope

### Expected behavior

Repository instances should be properly garbage collected when they go out of scope. The caching should happen at the instance level, not in a decorator that holds references to self.

### Screenshots/Code snippets

The problematic code is in `lib/crewai/src/crewai/cli/git.py` around line 43:

@lru_cache(maxsize=None)  # noqa: B019
def is_git_repo(self) -> bool:
    """Check if the current directory is a git repository.
    
    Notes:
      - TODO: This method is cached to avoid redundant checks, 
        but using lru_cache on methods can lead to memory leaks
    """The TODO comment shows the devs were aware of this.

### Operating System

### Python Version

3.10

### crewAI Version

Latest (main branch)

### crewAI Tools Version

N/A - this is in core, not tools

### Virtual Environment

Venv

### Evidence

The decorator @lru_cache on an instance method is a well-known Python gotcha. 
The cache dictionary holds references to self, preventing the garbage collector 
from cleaning up instances. You can verify this by running a simple test with 
weakref to show instances aren't being collected.

### Possible Solution

Replace the @lru_cache decorator with instance-level caching:

- Add `self._is_git_repo_cache: bool | None = None` in __init__
- Remove the @lru_cache decorator
- Check the cache at the start of is_git_repo() and populate it if None

This keeps the same O(1) performance but allows proper garbage collection. 
Happy to submit a PR with tests if this approach sounds good.

### Additional context

This affects anyone running long-lived processes or automation scripts that 
repeatedly check git repos. The memory leak is gradual but adds up over time.

I have a complete fix ready with tests if you'd like me to submit it.

Chase-Xuu:
## Summary
Fixes #4210

Resolves the memory leak in `Repository.is_git_repo()` caused by the `@lru_cache` decorator holding references to `self`.

## Root Cause
The `@lru_cache` decorator on instance methods is a well-known Python gotcha. When applied to a method, the cache dictionary stores references to `self` as part of the cache key. This prevents the garbage collector from cleaning up instances even after they go out of scope.

The code already had a TODO comment acknowledging this issue:
```python
@lru_cache(maxsize=None)  # noqa: B019
def is_git_repo(self) -> bool:
    """...
    Notes:
      - TODO: This method is cached to avoid redundant checks, but using lru_cache on methods can lead to memory leaks
    """
```

## Changes
1. **Removed** the `@lru_cache` decorator and `functools` import
2. **Added** an instance-level cache attribute: `self._is_git_repo_cache: bool | None = None`
3. **Modified** `is_git_repo()` to use manual instance-level caching:
   - Check cache at start of method
   - Populate cache on first call
   - Return cached value on subsequent calls

## Benefits
- ✅ Same O(1) performance for repeated calls
- ✅ Proper garbage collection of Repository instances
- ✅ No memory accumulation in long-running processes
- ✅ Resolves the acknowledged TODO

## Testing
Added three regression tests using `weakref` to verify:
- Single Repository instances are properly garbage collected
- Multiple Repository instances created in loops are all collected
- Caching behavior is preserved (results are consistent)

### DIFF
--- a/lib/crewai/src/crewai/cli/git.py
+++ b/lib/crewai/src/crewai/cli/git.py
@@ -1,10 +1,10 @@
-from functools import lru_cache
 import subprocess
 
 
 class Repository:
     def __init__(self, path: str = ".") -> None:
         self.path = path
+        self._is_git_repo_cache: bool | None = None
 
         if not self.is_git_installed():
             raise ValueError("Git is not installed or not found in your PATH.")
@@ -40,22 +40,26 @@ class Repository:
             encoding="utf-8",
         ).strip()
 
-    @lru_cache(maxsize=None)  # noqa: B019
     def is_git_repo(self) -> bool:
         """Check if the current directory is a git repository.
 
-        Notes:
-          - TODO: This method is cached to avoid redundant checks, but using lru_cache on methods can lead to memory leaks
+        Uses instance-level caching to avoid redundant checks while allowing
+        proper garbage collection of Repository instances.
         """
+        if self._is_git_repo_cache is not None:
+            return self._is_git_repo_cache
+
         try:
             subprocess.check_output(
                 ["git", "rev-parse", "--is-inside-work-tree"],  # noqa: S607
                 cwd=self.path,
                 encoding="utf-8",
             )
-            return True
+            self._is_git_repo_cache = True
         except subprocess.CalledProcessError:
-            return False
+            self._is_git_repo_cache = False
+
+        return self._is_git_repo_cache
 
     def has_uncommitted_changes(self) -> bool:
         """Check if the repository has uncommitted changes."""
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:

PR #4490, opened 2026-02-15
Implement user input handling in Flows

joaomdmoura:
## Add `self.ask()` for conversational flows

Flows can now request user input mid-execution via `self.ask()`. This enables conversational patterns like gathering requirements, interactive troubleshooting, multi-turn dialogs, without breaking the flow lifecycle and requiring HITL.

### What it does

```python
class ResearchFlow(Flow):
    @start()
    def gather_and_research(self):
        topic = self.ask("What topic should we research?")
        depth = self.ask("How deep? (brief/detailed/exhaustive)")
        return crew.kickoff(inputs={"topic": topic, "depth": depth})
```

- **`self.ask(message, timeout=None, metadata=None)`** -- pauses the method, prompts the user, returns their response as a string. Works in both sync and async flow methods.
- **Timeout** -- returns `None` when expired, enabling `while (msg := self.ask("You:", timeout=300)) is not None:` loop patterns that always terminate.
- **Metadata** -- bidirectional. Send context to the provider (`metadata={"user_id": "u123"}`), receive context back from the provider via `InputResponse(text="answer", metadata={"responded_by": "u456"})`. Thread-safe by design (no shared mutable state).
- **Auto-checkpoint** -- persists `self.state` before each `ask()` call when persistence is configured, so previously gathered data survives crashes.
- **Non-blocking** -- the flow framework runs sync methods in a thread pool (`asyncio.to_thread`), so `ask()` never blocks the event loop. Parallel listeners execute concurrently even when one is waiting for input.

### How it differs from `@human_feedback`

| | `self.ask()` | `@human_feedback` |
|---|---|---|
| When | Mid-method, inline | Post-method, wraps return value |
| Purpose | Gather input | Review/approve output |
| Can be conditional, looped, called N times | Yes | No (once per decorated method) |
| Routes to listeners via emit | No | Yes |
| Persist and reload flow execution | No | Yes |

They coexist -- same flow can use both.

### DIFF
--- a/lib/crewai/src/crewai/cli/utils.py
+++ b/lib/crewai/src/crewai/cli/utils.py
@@ -468,13 +468,13 @@ def get_flows(flow_path: str = "main.py") -> list[Flow]:
                                         module_attr
                                     ):
                                         flow_instances.append(flow_instance)
-                                except Exception:
+                                except Exception:  # noqa: S112
                                     continue
 
                             if flow_instances:
                                 break
 
-                        except Exception:
+                        except Exception:  # noqa: S112
                             continue
 
                     except (ImportError, AttributeError):
@@ -483,7 +483,7 @@ def get_flows(flow_path: str = "main.py") -> list[Flow]:
             if flow_instances:
                 break
 
-    except Exception:
+    except Exception:  # noqa: S110
         pass
 
     return flow_instances


lib/crewai/src/crewai/flow/flow.py
```diff
@@ -2582,6 +2586,204 @@ async def _execute_single_listener(
...
+
+        Walks the call stack to find the first frame whose function name
+        matches a registered flow method on this instance.
+
+        Returns:
+            The flow method name, or ``"unknown"`` if not found.
+        """
+        import inspect
+
+        for frame_info in inspect.stack():
```
greysonlalonde: this can be brittle in certain threading contexts - we changed [`FlowTrackable`](https://github.com/crewAIInc/crewAI/blob/4aedd58829d6eb7678e8c53ae725a931372a5c5d/lib/crewai/src/crewai/flow/flow_trackable.py#L17) to use contextvars to address some of the bugs arising from stack inspection. Since some of the problems are hard to pin down once they start, we should do the same here or pass the method name explicitly from the flow execution machinery

lib/crewai/src/crewai/flow/flow.py
```diff
@@ -2582,6 +2586,204 @@ async def _execute_single_listener(
...
+                        raw = future.result(timeout=timeout)
+                    except FuturesTimeoutError:
+                        raw = None
+            else:
+                raw = provider.request_input(message, self, metadata=metadata)
+        except KeyboardInterrupt:
+            raise
+        except Exception:
+            logger.debug("Input provider error in ask()", exc_info=True)
+            raw = None
```
greysonlalonde, line 2753: each timed-out `ask()` spawns a thread pool, and when timeout fires, the underlying `provider.request_input()` call is not cancelled, resulting in a running background thread. for `ConsoleProvider`, this means a dangling `input()` call, and can mean network providers may leak connections. this should at minimum call `future.cancel()`

lib/crewai/src/crewai/flow/flow.py
```diff
@@ -784,6 +785,9 @@ def __init__(
         self._pending_feedback_context: PendingFeedbackContext | None = None
         self.suppress_flow_events: bool = suppress_flow_events
 
+        # User input history (for self.ask())
+        self._input_history: list[dict[str, Any]] = []
```
greysonlalonde: would be nice to type this as a `TypedDict`, like `InputHistoryEntry`, so it's easier for llms to gain context on attribute usage in future edits/review

lib/crewai/src/crewai/flow/flow.py
joaomdmoura: Good catch.

Replaced inspect.stack() with a current_flow_method_name contextvar, following the same pattern we already use for current_flow_id and current_flow_request_id in flow_context.py.

The contextvar is set in _execute_method before the sync/async branch (and before copy_context()), so it propagates correctly into the thread pool for sync methods. Each concurrent method execution gets its own context via copy_context(), so parallel listeners see their own method name with no race condition. Cleaned up with reset(token) in a finally block.

_resolve_calling_method_name() and the inspect import are deleted entirely. ask() now just reads current_flow_method_name.get("unknown") -- O(1), thread-safe, no stack walking.

Will push soon

lib/crewai/src/crewai/flow/flow.py
joaomdmoura: Added InputHistoryEntry as a TypedDict in flow/types.py (alongside the existing FlowMethodData, ExecutionMethodData, etc.) and updated _input_history to list[InputHistoryEntry]. All six fields are now explicitly typed and documented.

Will push soon

lib/crewai/src/crewai/flow/flow.py
joaomdmoura: This was actually worse than a thread leak.

The with ThreadPoolExecutor block calls shutdown(wait=True) on exit, which means after timeout fires, the flow deadlocks waiting for the provider thread to finish (e.g., input() blocks forever). The timeout was functionally broken for any blocking provider.

Fixed by replacing the with block with manual executor management:

```
executor = ThreadPoolExecutor(max_workers=1)
future = executor.submit(provider.request_input, message, self, metadata)
try:
    raw = future.result(timeout=timeout)
except FuturesTimeoutError:
    future.cancel()
    raw = None
finally:
    executor.shutdown(wait=False, cancel_futures=True)
```

- future.cancel() on timeout: cleans up if the task hasn't started yet.
- executor.shutdown(wait=False, cancel_futures=True): doesn't block waiting for the running thread, and cancels any queued tasks.

One thing to note: 
future.cancel() can't interrupt an already-running thread (there's no safe way to kill a thread blocked on input() or a socket call). 
The underlying request_input() may continue running until it completes naturally.

Not necessarily an issue since in prod we won't be relying on `input`

greysonlalonde (approved):

PR #4513, opened 2026-02-19
supporting parallel tool use

lorenzejay:


lib/crewai/src/crewai/agents/crew_agent_executor.py
```diff
@@ -698,6 +699,343 @@ def _handle_native_tool_calls(
...
+                }
+                self.messages.append(assistant_message)
+
+                def _execute_one(
+                    idx: int,
+                    call_id: str,
+                    func_name: str,
+                    func_args: str | dict[str, Any],
+                    original_tool: Any | None,
+                    should_execute: bool,
```
greysonlalonde: **Missing cache + hook support in parallel path**

The parallel `_execute_one` function is missing several features that the sequential path (below) has:

- **No cache read** — the sequential path checks `self.tools_handler.cache.read()`
- **No cache write** — the sequential path calls `self.tools_handler.cache.add()` after execution
- **No `cache_function` check** — the sequential path calls `original_tool.cache_function(args_dict, raw_result)`
- **No `before_tool_call` hooks** — the sequential path calls `get_before_tool_call_hooks()`
- **No `after_tool_call` hooks** — the sequential path calls `get_after_tool_call_hooks()`

This means any user relying on tool caching or hooks will silently lose that functionality whenever the LLM returns >1 tool call in a batch.

Consider extracting a shared single-tool-execution method (like the experimental executor does with `_execute_single_native_tool_call`) so both paths stay in sync.

lib/crewai/src/crewai/agents/crew_agent_executor.py
```diff
@@ -698,6 +699,343 @@ def _handle_native_tool_calls(
         if not tool_calls:
             return None
```
greysonlalonde, line 691: **~230 lines of duplication between parallel and sequential paths**

The parallel path (lines 701–936) largely duplicates the sequential path below it. This is what caused issue #1 (missing cache/hooks) — it's easy to miss features when copy-pasting.

The experimental executor handles this much better by extracting `_execute_single_native_tool_call` and reusing it for both the parallel (`ThreadPoolExecutor`) and sequential paths. The same pattern should be applied here to keep the two paths consistent and maintainable.

lib/crewai/tests/agents/test_native_tool_calling.py
```diff
@@ -64,6 +71,73 @@ class FailingTool(BaseTool):
...
+        if current > maximum:
+            maximum = current
+    return maximum
+
+
+def _assert_tools_overlapped() -> None:
+    windows = ParallelProbe.windows()
+    local_windows = [
+        w
+        for w in windows
```
greysonlalonde, line 133: **Event handler leak across tests**

Each call to `_attach_parallel_probe_handler()` registers a *new* handler on `crewai_event_bus` via `@crewai_event_bus.on(ToolUsageFinishedEvent)`, but these handlers are never unregistered. Since every test method calls this function, handlers accumulate — by the Nth test, there are N handlers firing for each `ToolUsageFinishedEvent`.

Consider either:
- Registering the handler once (e.g., in a session-scoped fixture) and only calling `ParallelProbe.reset()` per test, or
- Returning a cleanup callable and tearing it down after each test.

lib/crewai/tests/agents/test_native_tool_calling.py
```diff
@@ -122,7 +255,7 @@ def test_openai_agent_kickoff_with_tools_mocked(
         self, calculator_tool: CalculatorTool
```
greysonlalonde, line 255: **Nit: Unrelated model name changes**

This PR changes model names in several existing tests (`gpt-4o-mini` → `gpt-5-nano`, `gemini-2.0-flash-exp` → `gemini-2.5-flash`, `azure/gpt-4o-mini` → `azure/gpt-5-nano`) with corresponding cassette updates.

These are unrelated to parallel tool support. Consider splitting them into a separate commit to keep the diff focused and easier to review.

lib/crewai/src/crewai/agents/crew_agent_executor.py
```diff
@@ -698,6 +699,343 @@ def _handle_native_tool_calls(
...
+                    result: str = "Tool not found"
+                    if not should_execute and original_tool:
+                        result = (
+                            f"Tool '{func_name}' has reached its usage limit of "
+                            f"{original_tool.max_usage_count} times and cannot be used anymore."
+                        )
+                    elif hook_blocked:
+                        result = f"Tool execution blocked by hook. Tool: {func_name}"
+                    elif from_cache:
+                        pass
```
greysonlalonde: **Minor: Thread safety on shared mutable state**

When `_execute_one` runs in parallel threads via `ThreadPoolExecutor`:

- `BaseTool.run()` does `self.current_usage_count += 1` — this is a LOAD_ATTR + BINARY_ADD + STORE_ATTR sequence that isn't atomic even under the GIL. Two threads calling the same tool could lose an increment.
- `self.task.increment_tools_errors()` is similarly called from threads without synchronization.

In practice this is low-risk (a lost count of 1 is unlikely to matter), but worth noting. The experimental executor sidesteps this for `max_usage_count` tools by falling back to sequential execution via `_should_parallelize_native_tool_calls` — consider adding the same guard here.

lib/crewai/tests/agents/test_native_tool_calling.py
```diff
@@ -64,6 +71,73 @@ class FailingTool(BaseTool):
...
+
+    _lock = threading.Lock()
+    _windows: list[tuple[str, float, float]] = []
+
+    @classmethod
+    def reset(cls) -> None:
+        with cls._lock:
+            cls._windows = []
+
+    @classmethod
```
greysonlalonde, line 90: **Nit: Redundant double-recording in ParallelProbe**

Each tool's `_run()` calls `ParallelProbe.record()` directly (using `time.perf_counter()`), AND `_attach_parallel_probe_handler()` registers an event handler that also calls `ParallelProbe.record()` (using `event.started_at.timestamp()`, which is epoch seconds).

So every execution produces two records in `_windows`. The assertions still pass because the two time scales don't interfere with each other in the sweep-line algorithm, but the double recording is confusing. Consider picking one approach:
- Record only from the event handler (tests the full event pipeline), or
- Record only from `_run()` (simpler, more direct).

lib/crewai/tests/agents/test_agent_executor.py
```diff
@@ -462,3 +463,176 @@ def mock_kickoff_side_effect():
...
+        def slow_two() -> str:
+            time.sleep(0.2)
+            return "two"
+
+        executor._available_functions = {"slow_one": slow_one, "slow_two": slow_two}
+        executor.state.pending_tool_calls = [
+            {
+                "id": "call_1",
+                "function": {"name": "slow_one", "arguments": "{}"},
+            },
```
greysonlalonde, line 530: **Nit: Tight timing bound may flake on CI**

`assert elapsed < 0.35` for two `time.sleep(0.2)` calls. On a loaded CI runner, thread pool startup + scheduling overhead could push elapsed time past 0.35s. Consider a more generous bound (e.g., `< 0.5`) — still proves parallelism vs the sequential baseline of ~0.4s.

PR #4544, opened 2026-02-20
fix: ensure callbacks are ran/awaited if promise

greysonlalonde:


lib/crewai/src/crewai/agents/crew_agent_executor.py
```diff
@@ -1490,7 +1500,9 @@ def _invoke_step_callback(
             formatted_answer: Current agent response.
         """
         if self.step_callback:
-            self.step_callback(formatted_answer)
+            cb_result = self.step_callback(formatted_answer)
+            if inspect.iscoroutine(cb_result):
+                asyncio.run(cb_result)
```
lorenzejay, line 1505: lgtm. should we add a test here

PR #4545, opened 2026-02-20
chore: simplify platform integration token resolution

greysonlalonde:
## Summary
- Remove unused `platform_context` context manager and env var fallback from `crewai.context`
- Token resolves from context var or env var via `default_factory` on the tool field
- Replace custom `__init__` on `CrewAIPlatformActionTool` with `model_validator`
- Use `sanitize_tool_name` instead of naive `.lower().replace(" ", "_")`
- Builder captures token at construction and passes it through to tools


lucasgomide (changes requested):
How are we supposed to configure the integration token now that env-var support is being removed in OSS?

You also drop the public_methods in favor of using the "private instance"; i'm not big fan of this approach

From your current PR state users now have to import `_platform_integration_token` and set the token in the main file.
If this change is intentional we should update the docs to make it clear. In my opinion, this is a breaking change and should be treated as such

Vidit-Ostwal:

PR #4573, opened 2026-02-23
feat: enhance JSON argument parsing and validation in CrewAgentExecutor and BaseTool

joaomdmoura:
## Native function calling bypasses schema validation and swallows JSON parse errors

Two bugs in the native function calling path (`crew_agent_executor.py`):

1. **Silent argument loss** — When the LLM returns malformed JSON tool arguments (common with `code` params containing unescaped quotes), `json.loads` fails and `args_dict` is silently set to `{}`. The tool then gets called with no arguments and crashes with a confusing `TypeError` instead of telling the agent its JSON was bad.

2. **Schema never enforced** — Tool registration stores `BaseTool.run` as the callable, which is a pure passthrough to `_run` with zero validation. The `args_schema` Pydantic model exists and is sent to the LLM, but is never checked server-side. The other code path (`tool_usage.py` → `CrewStructuredTool.invoke`) validates properly — this path doesn't.

### Fixes

**Bug 1:** Capture the `JSONDecodeError` and return a descriptive error message to the agent so it can retry with valid JSON, instead of silently dropping all arguments.

**Bug 2:** Add `args_schema.model_validate(kwargs)` in `BaseTool.run()` and `Tool.run()` before calling `_run`/`func`, bringing the native path to parity with the `invoke` path. Positional-arg calls and schema-less tools are unaffected.

### DIFF
--- a/lib/crewai/src/crewai/tools/base_tool.py
+++ b/lib/crewai/src/crewai/tools/base_tool.py
@@ -166,7 +166,7 @@ class BaseTool(BaseModel, ABC):
         if kwargs and self.args_schema is not None and self.args_schema.model_fields:
             try:
                 validated = self.args_schema.model_validate(kwargs)
-                retun validated.model_dump()
+                return validated.model_dump()
             except Exception as e:
                 raise ValueError(
                     f"Tool '{self.name}' arguments validation failed: {e}"


joaomdmoura:
on it!

joaomdmoura:
@cursoragent actrually can you address @greysonlalonde for me please? maek sure to test it and thing about for asn long as needed

PR #4579, opened 2026-02-24
fix: preserve null types in tool parameter schemas for LLM

lucasgomide:
Tool parameter schemas were stripping null from optional fields via generate_model_description, forcing the LLM to provide non-null values for fields.
Adds strip_null_types parameter to generate_model_description and passes False when generating tool schemas, so optional fields keep anyOf: [{type: T}, {type: null}]

### DIFF
--- a/docs/en/concepts/llms.mdx
+++ b/docs/en/concepts/llms.mdx
@@ -106,6 +106,15 @@ There are different places in CrewAI code where you can specify the model to use
   </Tab>
 </Tabs>
 
+<Info>
+  CrewAI provides native SDK integrations for OpenAI, Anthropic, Google (Gemini API), Azure, and AWS Bedrock — no extra install needed beyond the provider-specific extras (e.g. `uv add "crewai[openai]"`).
+
+  All other providers are powered by **LiteLLM**. If you plan to use any of them, add it as a dependency to your project:
+  ```bash
+  uv add 'crewai[litellm]'
+  ```
+</Info>
+
 ## Provider Configuration Examples
 
 CrewAI supports a multitude of LLM providers, each offering unique features, authentication methods, and model capabilities.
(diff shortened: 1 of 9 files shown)


lib/crewai/src/crewai/utilities/pydantic_schema_utils.py
```diff
@@ -426,6 +430,10 @@ def generate_model_description(model: type[BaseModel]) -> ModelDescription:
 
     Args:
         model: A Pydantic model class.
+        strip_null_types: When ``True`` (default), remove ``null`` from
+            ``anyOf`` / ``type`` arrays.  Set to ``False`` for **tool
+            parameter** schemas so that the LLM can send ``null`` for
+            optional fields instead of being forced to guess a value.
```
gabemilani: nit: remove the tool context from the pydantic schema util file

```suggestion
        strip_null_types: When ``True`` (default), remove ``null`` from
            ``anyOf`` / ``type`` arrays.  Set to ``False`` to allow sending ``null`` for
            optional fields.
```

lucasgomide:
> Just ensure you test on native tool calling for anthropic and openai, otherwise these may result in validation errors

@greysonlalonde I've already tested it.. everything works well. I can share examples if you'd like

greysonlalonde:
> > Just ensure you test on native tool calling for anthropic and openai, otherwise these may result in validation errors
> 
> @greysonlalonde I've already tested it.. everything works well. I can share examples if you'd like

Nope, good with me. I think accept @gabemilani 's nit, then gtg 👍

lucasgomide:
> > > Just ensure you test on native tool calling for anthropic and openai, otherwise these may result in validation errors
> > 
> > 
> > @greysonlalonde I've already tested it.. everything works well. I can share examples if you'd like
> 
> Nope, good with me. I think accept @gabemilani 's nit, then gtg 👍

done!

lorenzejay (approved):

PR #4585, opened 2026-02-24
Make sys.stdout/stderr filtering opt-in

### LINKED ISSUE #3000: [FEATURE] Stop hijacking sys.stdout and sys.stderr
### Feature Area

Core functionality

NA

Maybe I'm lacking context as an end user, but I don't see a functional reason why anyone `import`ing `crewai.llm` should suddenly be unable to log anything with the word `"litellm"` without themselves having to re-override `sys.stderr`.

https://github.com/crewAIInc/crewAI/blob/970a63c13c7ac94b5e016e529cf80693f9b53b05/src/crewai/llm.py#L119-L125

Philosophically, globally overriding `sys` attributes for every CrewAI user's Python application seems like a massive overstep, especially when it comes to filtering out terms without a user's consent or control.

If you really really really really wanted to, you could wrap this module's `litellm` calls with [`contextlib.redirect_stdout()`](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stdout), although the Python maintainers mention

> Note that the global side effect on [sys.stdout](https://docs.python.org/3/library/sys.html#sys.stdout) means that this context manager is not suitable for use in library code and most threaded applications.

further suggesting that modifying `sys.stdout` and `sys.stderr` in general is not a fantastic approach.

Or if altering the behavior of others' code is still on the table, you can probably monkey patch `litellm`'s logging handlers.

### Additional context

_No response_

### Willingness to Contribute

No, I'm just suggesting the idea

thakoreh:
## Description

Fixes #3000 - Stop hijacking sys.stdout and sys.stderr

## Problem

The FilteredStream was unconditionally wrapping sys.stdout and sys.stderr at module import time, affecting all code that imports crewai.llm. This is a global side effect that library code should not impose on users.

As noted in Python docs for contextlib.redirect_stdout:
> Note that the global side effect on sys.stdout means that this context manager is not suitable for use in library code

## Solution

Make the filtering opt-in:
- Added environment variable CREWAI_FILTER_LITELLM_OUTPUT (set to 1/true/yes to enable)
- Added enable_litellm_filtering() helper for programmatic use
- Users who want the old behavior can set the env var

## Changes

- Filtering now requires explicit opt-in
- Added enable_litellm_filtering() public helper function
- Maintained backward compatibility for users who set the env var

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -175,10 +175,10 @@ class FilteredStream(io.TextIOBase):
 
 def enable_litellm_filtering() -> None:
     """Enable filtering of LiteLLM debug output from stdout/stderr.
-    
+
     This wraps sys.stdout and sys.stderr with FilteredStream to suppress
     noisy LiteLLM banners and debug messages.
-    
+
     Can also be enabled via environment variable: CREWAI_FILTER_LITELLM_OUTPUT=1
     """
     if not isinstance(sys.stdout, FilteredStream):
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -173,13 +173,26 @@ class FilteredStream(io.TextIOBase):
         return True
 
 
+def enable_litellm_filtering() -> None:
+    """Enable filtering of LiteLLM debug output from stdout/stderr.
+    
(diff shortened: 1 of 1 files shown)


Vidit-Ostwal:

PR #4586, opened 2026-02-25
Add proactive context window checking before LLM calls

### LINKED ISSUE #2990: [BUG] CrewAi Custom LLM is not calling get_context_window_size function
### Description

I've created a Custom LLM class based on the documentation, basic functionality works but get_context_window_size function was never called. So CrewAi just passed everything and caused token/context exceeded error. 

### Steps to Reproduce

1. Create CustomLLM per the documentation here: https://docs.crewai.com/learn/custom-llm
2. try to create an agent and task that exceed the maximum context supported by your model. By default, crewai should know that it exceeded the context size and need to summarize it since respect_context_window is True by default. But you ends up getting an error such as 400 bad request because your prompt is too big

### Expected behavior

CrewAI is supposed to call the get_context_window_size function to determine the context window size and then summarize the content if the context is too large.

### Screenshots/Code snippets

N/A

### Operating System

macOS Sonoma

### Python Version

3.10

### crewAI Version

0.126.0

### crewAI Tools Version

N/A

### Virtual Environment

Venv

### Evidence

Error during LLM call: 400 Client Error: BAD REQUEST for url:

### Possible Solution

None

### Additional context

None

thakoreh:
## Description

Fixes #2990 - get_context_window_size is not used to check context window proactively

## Problem

When using custom LLMs (non-litellm), context length errors were only handled reactively after they occurred. The  method exists but is never called proactively before making LLM calls.

For BaseLLM subclasses with custom context window sizes, this means:
- Errors are only caught after the API call fails
- Summarization only happens reactively
- Unnecessary API calls and errors

## Solution

Added proactive context window checking before each LLM call:

1. **New function**:  in 
   - Estimates token count from message content (1 token per 4 chars)
   - Compares against LLM's 
   - Calls  if approaching limit (90% threshold)

2. **Integration**: Added proactive check in  before LLM calls
   - Added to both ReAct loop and native tool calling loop
   - Only runs when 

## Changes

- : Added  function
- : Import and call proactive check before LLM calls

## Testing

When messages approach the context window limit, users will now see:

Instead of waiting for the API to return a context exceeded error.

---

> [!NOTE]
> **Medium Risk**
> Changes the pre-call LLM execution flow by mutating `messages` via automatic summarization, which can alter model behavior and tool-calling outcomes near context limits. The LiteLLM output filtering change is low risk but could affect debugging expectations when the env var is enabled.
> 
> **Overview**
> Adds a proactive context-window guard before LLM invocations in `CrewAgentExecutor` (both ReAct and native-tool loops) so conversations are summarized *before* requests fail with context-length errors.
> 
> Introduces `check_context_window_proactively` in `agent_utils` to estimate prompt size (including text from multimodal blocks), compare against `llm.get_context_window_size()`, and trigger `summarize_messages` once usage crosses a 90% threshold.
> 
> Separately makes LiteLLM stdout/stderr filtering opt-in by moving stream wrapping behind `enable_litellm_filtering()` and a `CREWAI_FILTER_LITELLM_OUTPUT` environment flag.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11f6392f2d6cbfb96cea1e9d70f851fd76e7a4fa. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -39,6 +39,7 @@ from crewai.hooks.tool_hooks import (
 )
 from crewai.utilities.agent_utils import (
     aget_llm_response,
+    check_context_window_proactively,
     convert_tools_to_openai_schema,
     enforce_rpm_limit,
     format_message_for_llm,
@@ -350,6 +351,17 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
 
                 enforce_rpm_limit(self.request_within_rpm_limit)
 
+                # Proactively check context window before LLM call
+                check_context_window_proactively(
+                    messages=self.messages,
+                    llm=self.llm,
+                    respect_context_window=self.respect_context_window,
+                    printer=self._printer,
+                    callbacks=self.callbacks,
+                    i18n=self._i18n,
+                    verbose=self.agent.verbose,
+                )
+
                 answer = get_llm_response(
                     llm=self.llm,
                     messages=self.messages,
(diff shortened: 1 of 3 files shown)


Vidit-Ostwal:

PR #4597, opened 2026-02-25
fix: map output_pydantic/output_json to native structured output

greysonlalonde:
## summary
- maps `output_pydantic`/`output_json` to the executor's `response_model`, enabling native structured output at the LLM layer instead of relying on fragile regex-based JSON extraction
- prevents Gemini stop words from corrupting JSON when `response_model` is set but tools are present
- handles `BaseModel` results directly in task execution when returned from native structured output

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -4195,7 +4195,7 @@ wheels = [
 
 [[package]]
 name = "nltk"
-version = "3.9.2"
+version = "3.9.3"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "click" },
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/agent/core.py
```diff
@@ -864,7 +864,11 @@ def create_agent_executor(
                 respect_context_window=self.respect_context_window,
                 request_within_rpm_limit=rpm_limit_fn,
                 callbacks=[TokenCalcHandler(self._token_process)],
-                response_model=task.response_model if task else None,
+                response_model=(
```
lorenzejay, line 867: yes!!

PR #4602, opened 2026-02-26
fix: replace bare except clause with except Exception

haosenwang1018:
Replace bare except clause with except Exception.


haosenwang1018:
Thanks for feedback. Closing for now to avoid stale backlog.

PR #4605, opened 2026-02-26
Update lancedb version and add lance-namespace packages

joaomdmoura:
- Updated lancedb dependency version from 0.4.0 to 0.29.2 in multiple files.
- Added new packages: lance-namespace and lance-namespace-urllib3-client with version 0.5.2, including their dependencies and installation details.
- Enhanced MemoryTUI to display a limit on entries and improved the LanceDBStorage class with automatic background compaction and index creation for better performance.

### DIFF
--- a/lib/crewai/src/crewai/memory/types.py
+++ b/lib/crewai/src/crewai/memory/types.py
@@ -99,7 +99,7 @@ class MemoryMatch(BaseModel):
             lines.append(f"  categories: {', '.join(self.record.categories)}")
         if self.record.metadata:
             for key, value in self.record.metadata.items():
-                if value:
+                if value is not None:
                     lines.append(f"  {key}: {value}")
         return "\n".join(lines)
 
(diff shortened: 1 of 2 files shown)


greysonlalonde (commented):
## Review Notes

### 1. `MemoryMatch.format()` filters out falsy metadata values

In `lib/crewai/src/crewai/memory/types.py`, the `format()` method uses `if value:` which will silently drop metadata entries where the value is `0`, `False`, or `""`:

```python
if self.record.metadata:
    for key, value in self.record.metadata.items():
        if value:  # ← drops 0, False, ""
            lines.append(f"  {key}: {value}")
```

Suggest changing to `if value is not None:` to only skip actual nulls.

### 2. `Memory.remember()` return type annotation mismatch

In `lib/crewai/src/crewai/memory/unified_memory.py:510`, a `# type: ignore[return-value]` is used because `remember()` now returns `None` when read-only, but the method signature still declares `-> MemoryRecord`. The `MemorySlice.remember()` signature was correctly updated to `-> MemoryRecord | None`, but `Memory.remember()` was not. Updating the return type annotation on `Memory.remember()` to `-> MemoryRecord | None` would remove the need for the type ignore.

PR #4705, opened 2026-03-04
fix: adjust executor listener value to avoid recursion

greysonlalonde:
> [!NOTE]
> **Medium Risk**
> Touches core `Flow` execution by adding a per-method call cap that raises `RecursionError`, which could change behavior for unusually long-running/cyclic flows. Agent execution routing constants also changed, so any external code depending on the old trigger label may need updates.
> 
> **Overview**
> Prevents runaway `Flow` executions caused by listeners that re-trigger themselves (e.g., a `@listen` label matching the method name) by adding a per-method call counter and configurable `max_method_calls` guard that raises `RecursionError`.
> 
> Updates `AgentExecutor` to route max-iteration handling via a new `max_iterations_exceeded` trigger (instead of `force_final_answer`) and passes a higher `max_method_calls` limit (`max_iter * 10`) when initializing its internal flow. Tests are updated and expanded to cover the new recursion guard and OR-condition self-listen behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 997701b4a56395d91a140431a3c40cea690e7ce9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/experimental/agent_executor.py
+++ b/lib/crewai/src/crewai/experimental/agent_executor.py
@@ -302,6 +302,7 @@ class AgentExecutor(Flow[AgentReActState], CrewAgentExecutorMixin):
             super().__init__(
                 suppress_flow_events=True,
                 tracing=current_tracing if current_tracing else None,
+                max_method_calls=self.max_iter * 10,
             )
             self._flow_initialized = True
 
--- a/lib/crewai/src/crewai/flow/flow.py
+++ b/lib/crewai/src/crewai/flow/flow.py
@@ -770,6 +770,7 @@ class Flow(Generic[T], metaclass=FlowMeta):
         persistence: FlowPersistence | None = None,
         tracing: bool | None = None,
         suppress_flow_events: bool = False,
+        max_method_calls: int = 100,
         **kwargs: Any,
     ) -> None:
         """Initialize a new Flow instance.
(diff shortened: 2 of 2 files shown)


lib/crewai/src/crewai/flow/flow.py
```diff
@@ -2569,6 +2572,17 @@ async def _execute_single_listener(
             - Skips execution if method was already completed (e.g., after reload)
             - Catches and logs any exceptions during execution, preventing individual listener failures from breaking the entire flow
         """
+        max_calls = 100
```
lorenzejay: can we have a mode where this is set byu the user

lib/crewai/src/crewai/experimental/agent_executor.py
```diff
@@ -655,11 +656,11 @@ def execute_tool_action(self) -> Literal["tool_completed", "tool_result_is_final
             return "tool_result_is_final"
 
         reasoning_prompt = self._i18n.slice("post_tool_reasoning")
-        reasoning_message: LLMMessage = {
+        reasoning_message_post: LLMMessage = {
```
lorenzejay, line 659: why the name var name change here ?

lib/crewai/src/crewai/flow/flow.py
```diff
@@ -2569,6 +2575,16 @@ async def _execute_single_listener(
...
+        if count > self._max_method_calls:
+            raise RecursionError(
+                f"Method '{listener_name}' has been called {self._max_method_calls} times in "
+                f"this flow execution, which indicates an infinite loop. "
+                f"This commonly happens when a @listen label matches the "
+                f"method's own name."
+            )
+        self._method_call_counts[listener_name] = count
+
         if listener_name in self._completed_methods:
```
lorenzejay, line 2588: - Counter scope: _method_call_counts is per-flow-execution (cleared in kickoff), which is correct. However, the counter
   is not cleared between cycles within a single execution — line 2602 clears _completed_methods for cyclic flows but
  doesn't reset _method_call_counts. This means legitimate cyclic flows will accumulate counts across cycles and
  eventually hit the 100 cap. This may need attention.

PR #4717, opened 2026-03-05
fix: preserve agent tools when output_pydantic is set

### LINKED ISSUE #4697: [BUG] When supports_function_calling() is True, only the output_pydantic model is sent to LiteLLM (the custom agent tools are discard)
### Description

### Steps to Reproduce

```python

class MyOutput(BaseModel):
    name: str
    value: str

class MySearchTool(BaseTool):
    name: str = "my_search_tool"

task = Task(
    agent=agent,
    output_pydantic=MyOutput,
)
```

3. Run the crew.

### Expected behavior

### Screenshots/Code snippets

X

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

1.10.0

### crewAI Tools Version

1.10.0

### Virtual Environment

Venv

### Evidence

### Possible Solution

### Additional context

X

weiguangli-io:
## Summary

Fixes #4697

When `output_pydantic` (`response_model`) is set on a task and the LLM supports native function calling, `InternalInstructor` intercepts the LLM call and creates its own `litellm.completion` that **only** contains the Pydantic schema as a tool -- all actual agent tools are silently discarded. The agent is forced to return structured output immediately without being able to call any of its tools first.

**Root cause**: the `response_model` guard in the four response handlers (`_handle_non_streaming_response`, `_ahandle_non_streaming_response`, `_handle_streaming_response`, `_ahandle_streaming_response`) did not check whether agent tools were also present. Two independent paths caused the issue:

1. `InternalInstructor` early-return -- creates a brand-new `completion()` call with only the Pydantic schema, ignoring `params["tools"]` entirely.
2. `response_model` forwarded to `litellm.completion` -- litellm's internal instructor overrides tools when `response_model` is in the kwargs.

**Fix**: add a `has_tools = bool(params.get("tools"))` guard. When tools are present:

- `InternalInstructor` is **skipped** (the agent needs to see and call its tools first).
- `response_model` is **not** added to `params` before calling `litellm.completion` / `litellm.acompletion`.
- Streaming handlers correctly return `tool_calls` when `available_functions` is `None` (executor-managed tool execution), mirroring the existing non-streaming behavior.

When no tools are present, behavior is completely unchanged -- `InternalInstructor` and `response_model` work as before for structured output.

## Changes

Single file: `lib/crewai/src/crewai/llm.py` (31 additions, 6 deletions)

## Test Plan

- Verified syntax correctness via `ast.parse`
- Verified boolean logic covers all four cases: (response_model + tools), (response_model + no tools), (no response_model + tools), (no response_model + no tools)
- Manual review of all four handler paths confirms the guard is applied consistently
- Existing tests remain unaffected since the no-tools path is unchanged

---

> [!NOTE]
> **Medium Risk**
> Touches core LLM response handling paths (sync/async, streaming/non-streaming), so regressions could affect tool-calling and structured output behavior. Logic is gated on presence of `params["tools"]`, reducing impact to tool-enabled calls.
> 
> **Overview**
> Fixes LiteLLM calls where providing `response_model` could *silently override/discard agent tools*.
> 
> When `params["tools"]` is present, the handlers now **skip `InternalInstructor`** and **avoid forwarding `response_model`** into `litellm.completion`/`litellm.acompletion`, allowing the model to invoke tools before structured output.
> 
> Streaming paths additionally return `tool_calls` directly when `available_functions` is `None` (executor-managed tool execution), mirroring non-streaming behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5c9cf0ba2e0d875b678eb14d82f2da17ef7c3b42. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -967,8 +967,17 @@ class LLM(BaseLLM):
                 self._track_token_usage_internal(usage_info)
             self._handle_streaming_callbacks(callbacks, usage_info, last_chunk)
 
+            # When the LLM returns tool_calls but available_functions is None
+            # (executor handles tool execution), return the tool_calls directly
+            # so the executor can process them.
+            if tool_calls and not available_functions:
+                return tool_calls
+
             if not tool_calls or not available_functions:
-                if response_model and self.is_litellm:
+                # Skip InternalInstructor when tools are present so the LLM
+                # can invoke agent tools before returning structured output.
+                has_tools = bool(params.get("tools"))
+                if response_model and self.is_litellm and not has_tools:
                     instructor_instance = InternalInstructor(
                         content=full_response,
                         model=response_model,
@@ -1150,7 +1159,10 @@ class LLM(BaseLLM):
             str: The response text
         """
         # --- 1) Handle response_model with InternalInstructor for LiteLLM
-        if response_model and self.is_litellm:
+        # Skip InternalInstructor when tools are present so the LLM can see
+        # and invoke agent tools before returning structured output.
+        has_tools = bool(params.get("tools"))
+        if response_model and self.is_litellm and not has_tools:
             from crewai.utilities.internal_instructor import InternalInstructor
 
             messages = params.get("messages", [])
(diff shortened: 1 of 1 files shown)


weiguangli-io:
Closing due to inactivity. Happy to reopen if there's interest.

PR #4742, opened 2026-03-06
feat: agent skills

greysonlalonde:
> [!NOTE]
> **Medium Risk**
> Adds a new prompt-injection surface (skill instructions appended to task prompts) and new YAML parsing dependency, which could affect agent behavior and error handling at runtime.
> 
> **Overview**
> Adds an **Agent Skills** feature: `Agent`/`Crew` now accept `skills` (paths for discovery or pre-loaded `Skill` objects), resolve/activate them during setup, and append activated skill instructions to task prompts.
> 
> Introduces a new `crewai.skills` module (models, YAML `SKILL.md` parser, discovery/activation loader, validation) plus skill lifecycle events for observability, along with docs pages in `en`, `pt-BR`, and `ko` and a new `pyyaml` dependency and test coverage for discovery/activation/parsing.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b4b1a978782b718b994121a33b47260981f0547c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -334,13 +334,16 @@ class Agent(BaseAgent):
         """Resolve skill paths and activate skills to INSTRUCTIONS level.
 
         Path entries trigger discovery and activation. Pre-loaded Skill objects
-        below INSTRUCTIONS level are activated. Crew-level skills are merged in.
+        below INSTRUCTIONS level are activated. Crew-level skills are merged in
+        with event emission so observability is consistent regardless of origin.
 
         Args:
             resolved_crew_skills: Pre-resolved crew skills (already discovered
                 and activated). When provided, avoids redundant discovery per agent.
         """
         from crewai.crew import Crew
+        from crewai.events.event_bus import crewai_event_bus
+        from crewai.events.types.skill_events import SkillActivatedEvent
 
         if resolved_crew_skills is None:
             crew_skills: list[Path | SkillModel] | None = (
@@ -379,7 +382,18 @@ class Agent(BaseAgent):
             elif isinstance(item, SkillModel):
                 if item.name not in seen:
                     seen.add(item.name)
-                    resolved.append(activate_skill(item, source=self))
+                    activated = activate_skill(item, source=self)
+                    if activated is item and item.disclosure_level >= INSTRUCTIONS:
+                        crewai_event_bus.emit(
+                            self,
+                            event=SkillActivatedEvent(
+                                from_agent=self,
+                                skill_name=item.name,
+                                skill_path=item.path,
+                                disclosure_level=item.disclosure_level,
+                            ),
+                        )
+                    resolved.append(activated)
 
         self.skills = resolved if resolved else None
 
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -384,7 +384,7 @@ class Crew(FlowTrackable, BaseModel):
                 from crewai.rag.embeddings.factory import build_embedder
 
                 embedder = build_embedder(cast(dict[str, Any], self.embedder))
-            self._memory = Memory(embedder=embedder)
+            self._memory = Memory(embedder=embedder, root_scope=crew_root_scope)
         elif self.memory:
             # User passed a Memory / MemoryScope / MemorySlice instance
             # Respect user's configuration — don't auto-set root_scope


docs/en/concepts/skills.mdx
```diff
@@ -0,0 +1,116 @@
...
+Skills use **progressive disclosure** — metadata is loaded first, full instructions only when activated, and resource catalogs only when needed.
+
+## Directory Structure
+
+```
+my-skill/
+├── SKILL.md            # Required — frontmatter + instructions
+├── scripts/            # Optional — executable scripts
+├── references/         # Optional — reference documents
+└── assets/             # Optional — static files (configs, data)
```
lorenzejay, line 21: wonder if we make this `.crewai/skills/...`

lib/crewai/src/crewai/agent/core.py
```diff
@@ -24,6 +25,7 @@
 
 from crewai.agent.utils import (
     ahandle_knowledge_retrieval,
+    append_skill_context,
```
lorenzejay, line 30: we also need this in the `Agent().kickoff()` methods

lib/crewai/src/crewai/skills/loader.py
```diff
@@ -0,0 +1,180 @@
...
+Provides functions to discover skills in directories, activate them
+for agent use, and format skill context for prompt injection.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+
```
joaomdmoura: Missing logger setup. Other modules in the codebase use `_logger = logging.getLogger(__name__)`. Without it, the `except` block in `discover_skills` silently drops errors when `source is None`.

Suggestion:
```python
import logging

_logger = logging.getLogger(__name__)
```

lib/crewai/src/crewai/skills/loader.py
```diff
@@ -0,0 +1,180 @@
...
+        crewai_event_bus.emit(
+            source,
+            event=SkillDiscoveryStartedEvent(
+                from_agent=source,
+                search_path=search_path,
+            ),
+        )
+
+    for child in sorted(search_path.iterdir()):
+        if not child.is_dir():
```
joaomdmoura, line 67: When `source is None`, invalid skills are silently swallowed — no log, no event. Consider adding:
```python
except Exception as e:
    _logger.warning("Failed to load skill from %s: %s", child, e)
    if source is not None:
        ...
```

lib/crewai/src/crewai/agent/core.py
```diff
@@ -289,6 +295,47 @@ def set_knowledge(self, crew_embedder: EmbedderConfig | None = None) -> None:
         except (TypeError, ValueError) as e:
             raise ValueError(f"Invalid Knowledge Configuration: {e!s}") from e
 
+    def set_skills(self) -> None:
+        """Resolve skill paths into loaded Skill objects.
+
```
joaomdmoura, line 335: `set_skills()` is called here in `post_init_setup()`, and then called again in `crews/utils.py:setup_agents()` after `agent.crew` is assigned. The first call is a no-op for agents that get skills only via the crew (since `self.crew` isn't set yet). Is the double invocation intentional for standalone agents? If so, a comment would help clarify.

lib/crewai/src/crewai/crew.py
```diff
@@ -291,6 +292,19 @@ class Crew(FlowTrackable, BaseModel):
...
     )
+    skills: list[str | Path] | None = Field(
+        default=None,
+        description="Skill search paths applied to all agents in the crew.",
+    )
+
+    @field_validator("skills", mode="before")
+    @classmethod
+    def coerce_skill_paths(cls, v: list[Any] | None) -> list[Path] | None:
+        """Coerce string entries to Path objects."""
```
joaomdmoura: `Crew.skills` is typed `list[str | Path] | None` while `BaseAgent.skills` is `list[Path | Skill] | None`. The asymmetry makes sense (crew holds search paths, agent resolves to Skill objects), but a brief docstring note explaining this would help future readers.

lib/crewai/src/crewai/skills/parser.py
```diff
@@ -0,0 +1,171 @@
...
+        SkillParseError: If parsing fails.
+    """
+    content = path.read_text(encoding="utf-8")
+    frontmatter_dict, body = parse_frontmatter(content)
+    frontmatter = SkillFrontmatter(**frontmatter_dict)
+    return frontmatter, body
+
+
+def load_skill_metadata(skill_dir: Path) -> Skill:
+    """Load a skill at METADATA disclosure level.
```
joaomdmoura, line 96: The frontmatter has field-level length limits (name: 64, description: 1024) but the SKILL.md body text has no cap. Since this body gets injected directly into agent prompts, a very large SKILL.md could blow up context windows. Consider a soft warning (log) if the body exceeds a threshold, or a configurable max.

docs/en/concepts/skills.mdx
greysonlalonde: I think people might want it committed, though I like it. Just wondering on users with more generic skill files they might be using across SDKs

lib/crewai/src/crewai/skills/parser.py
greysonlalonde: Done!

docs/en/concepts/skills.mdx
```diff
@@ -0,0 +1,116 @@
...
+## Overview
+
+Skills are self-contained directories that provide agents with domain-specific instructions, references, and assets. Each skill is defined by a `SKILL.md` file with YAML frontmatter and a markdown body.
+
+Skills use **progressive disclosure** — metadata is loaded first, full instructions only when activated, and resource catalogs only when needed.
+
+## Directory Structure
+
+```
+my-skill/
```
lorenzejay, line 17: should this be `.crewai/skills/...`

greysonlalonde: Just wondering on this one https://github.com/crewAIInc/crewAI/pull/4742#discussion_r2914562498, probably worth chatting about

lib/crewai/src/crewai/skills/models.py
```diff
@@ -0,0 +1,175 @@
...
+        metadata: Optional additional metadata as string key-value pairs.
+        allowed_tools: Optional space-delimited list of pre-approved tools.
+    """
+
+    model_config = ConfigDict(frozen=True, populate_by_name=True)
+
+    name: str = Field(
+        min_length=MIN_SKILL_NAME_LENGTH,
+        max_length=MAX_SKILL_NAME_LENGTH,
+        pattern=SKILL_NAME_PATTERN,
```
lorenzejay, line 60: **Bug: `SKILL_NAME_PATTERN` type mismatch with Pydantic `Field(pattern=...)`**

`validation.py` defines `SKILL_NAME_PATTERN` as `re.compile(...)` which produces a `re.Pattern[str]`, but Pydantic v2's `Field(pattern=...)` is typed `str | None`. This works today only because Pydantic internally coerces it via `str()`, but that's undocumented behavior and could break on a Pydantic minor bump.

**Fix** — either export a plain string from `validation.py`:
```python
SKILL_NAME_PATTERN: Final[str] = r"^[a-z0-9]+(?:-[a-z0-9]+)*$"
```
or extract the pattern string at the call site:
```python
name: str = Field(
    ...,
    pattern=SKILL_NAME_PATTERN.pattern,
)
```

greysonlalonde: https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field
`pattern: str | Pattern[str] | None = _Unset,`

iris-clawd (approved):
Re-approved after latest push — memory root_scope restored, looks clean. ✅

PR #4744, opened 2026-03-06
fix(tools): force LF newlines when generating tool.specs.json

haosenwang1018:
## Summary
- write `tool.specs.json` with explicit `newline="\n"` in `generate_tool_specs.py`
- add a unit test to assert `save_to_json` opens files with LF newlines

## Why
Running the generator on Windows can otherwise rewrite the file with CRLF line endings and create noisy diffs.


greysonlalonde:
Closing as stale — no activity in 30+ days.

PR #4760, opened 2026-03-07
fix: add missing list/dict methods to LockedListProxy and LockedDictProxy

lucasgomide:
We introduced a breaking change in 1.10.x when a Flow tries to access its state using .index (for example)

LockedListProxy and LockedDictProxy subclass `list` and `dict` but initialize the parent as empty (`super().__init__()`), delegating all access to an internal `self._list` / `self._dict`. However, several inherited methods were not overridden, causing them to silently operate on the empty parent instead of the real data.

Most critically, `list.index()` always raises `ValueError` because it searches the empty parent list. Other broken methods include `count()` (always returns 0), `sort()`/`reverse()` (no-ops), `copy()` (returns empty), and arithmetic operators (`+`, `*`).

All mutating operations acquire the lock; read-only operations delegate directly to the underlying collection, consistent with existing pattern.

---

> [!NOTE]
> **Medium Risk**
> Touches core `Flow` state proxy types used widely; behavior now changes for several built-in list/dict operations and in-place operators, with some concurrency implications despite added locking.
> 
> **Overview**
> Fixes `LockedListProxy`/`LockedDictProxy` so common `list`/`dict` APIs operate on the underlying state rather than the empty base container (e.g., `list.index`, `count`, `sort`, `reverse`, `copy`, arithmetic ops; and dict `copy`/merge ops and `reversed`).
> 
> Updates `StateProxy.__setattr__` to **unwrap** assigned `LockedListProxy`/`LockedDictProxy` values back to their underlying containers to avoid accidentally storing proxy objects in state.
> 
> Adds targeted tests covering the newly supported proxy methods and verifying in-place operators (`+=`, `|=`) don’t break subsequent mutations.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4aef39bccadcb81f0c43f95f0c75d459ff064c9b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/flow/flow.py
+++ b/lib/crewai/src/crewai/flow/flow.py
@@ -522,7 +522,7 @@ class LockedListProxy(list, Generic[T]):  # type: ignore[type-arg]
     def __radd__(self, other: list[T]) -> list[T]:
         return other + self._list
 
-    def __iadd__(self, other: Iterable[T]) -> "LockedListProxy[T]":
+    def __iadd__(self, other: Iterable[T]) -> LockedListProxy[T]:
         with self._lock:
             self._list += list(other)
         return self
@@ -533,7 +533,7 @@ class LockedListProxy(list, Generic[T]):  # type: ignore[type-arg]
     def __rmul__(self, n: SupportsIndex) -> list[T]:
         return self._list * n
 
-    def __imul__(self, n: SupportsIndex) -> "LockedListProxy[T]":
+    def __imul__(self, n: SupportsIndex) -> LockedListProxy[T]:
         with self._lock:
             self._list *= n
         return self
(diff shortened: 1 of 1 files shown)


lib/crewai/src/crewai/flow/flow.py
```diff
@@ -632,10 +632,10 @@ def __or__(self, other: dict[str, T]) -> dict[str, T]:
     def __ror__(self, other: dict[str, T]) -> dict[str, T]:
         return other | self._dict
 
-    def __ior__(self, other: dict[str, T]) -> dict[str, T]:
+    def __ior__(self, other: dict[str, T]) -> "LockedDictProxy[T]":
```
greysonlalonde: can you add `from __future__ import annotations` at top of file, and remove the quotations? Quotes around type hints are not conventional for us

lucasgomide: Done, @greysonlalonde

greysonlalonde (approved):

PR #4791, opened 2026-03-09
Code interpreter sandbox escape

theCyberTech:
Fixes F-001 sandbox escape vulnerability in `CodeInterpreterTool` by removing the insecure fallback to a restricted sandbox and making Docker a requirement for untrusted code execution.

The previous fallback to `run_code_in_restricted_sandbox()` when Docker was unavailable was exploitable via Python object introspection, allowing attackers to recover import machinery and execute arbitrary commands on the host. This PR ensures that untrusted code execution now fails closed if Docker is not available, preventing host RCE.

---

### DIFF
--- a/lib/crewai-tools/tests/tools/test_code_interpreter_tool.py
+++ b/lib/crewai-tools/tests/tools/test_code_interpreter_tool.py
@@ -90,9 +90,7 @@ print(result)
     assert "sandbox escape" in str(exc_info.value)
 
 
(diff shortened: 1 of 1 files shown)


theCyberTech:
CleanShot 2026-03-14 at 20 27 47@2x
Testing when Docker is not available.

theCyberTech:
CleanShot 2026-03-14 at 20 32 48@2x
Testing when Docker is available.

Issue #4788, opened 2026-03-09
[BUG] Native tool calls are discarded if LLM returns a text response


ehansis:
### Description

The Crew Agent Executor loop for models supporting native tools intentionally passes `available_functions=None` to `get_llm_response` in order to get the tool calls as list for external execution:

https://github.com/crewAIInc/crewAI/blob/cd42bcf035c7e5b50ca6317da712d99394c75b44/lib/crewai/src/crewai/agents/crew_agent_executor.py#L512-L528

However, if the LLM response contains a text response in addition to the tool calls, this text response is returned instead of the tool calls, because of `available_functions` being `None`. The intention probably would be to follow the path `6)` below, but `5)` is executed instead.

https://github.com/crewAIInc/crewAI/blob/cd42bcf035c7e5b50ca6317da712d99394c75b44/lib/crewai/src/crewai/llm.py#L1237-L1251 

Fixing this might be as easy as switching `5)` and `6)`, but I'm not sure if this breaks something else.

The plaintext response is passed back into the agent executor loop and handled as potential final answer, the tool calls are never executed.

https://github.com/crewAIInc/crewAI/blob/cd42bcf035c7e5b50ca6317da712d99394c75b44/lib/crewai/src/crewai/agents/crew_agent_executor.py#L546-L557

### Steps to Reproduce

Sorry, I have no minimal reproduction example available at this time.

I ran a crew with native tool calls against "openrouter/anthropic/claude-haiku-4.5" and "openrouter/anthropic/claude-sonnet-4.6" with some custom tool definitions. This triggered LLM responses with both text response and `ChatCompletionMessageToolCall` definitions, which causes the described issue.

### Expected behavior

Return the tool calls and execute them in the agent executor loop. Discard the text response content (I guess?).

### Screenshots/Code snippets

See above

### Operating System

Ubuntu 22.04

### Python Version

3.12

### crewAI Version

1.10.1

### crewAI Tools Version

?

### Virtual Environment

Venv

### Evidence

LLM response with both text response and tool calls:

```
[...]

PR #4795, opened 2026-03-10
fix: native tool calls discarded when LLM returns both text and tool calls

NIK-TIGER-BILL:
## Root cause

When `crew_agent_executor` calls `get_llm_response` with `available_functions=None` (intentionally, to receive tool calls as a list for external execution), the condition in step 5 of `LLM.call()` was too broad:

```python
# step 5 — BEFORE (buggy)
    return text_response  # ← returned here, step 6 never reached
```

When the LLM returns **both** a text response and tool calls:
- `tool_calls` = `[ChatCompletionMessageToolCall(...)]` → truthy
- `available_functions` = `None` → falsy → `not available_functions` = **True**
- `text_response` = some string → truthy
- Result: condition is `True`, text is returned, tool calls are silently dropped ❌

Step 6 (`if tool_calls and not available_functions: return tool_calls`) is **never reached**.

## Fix

Narrow the condition so text is only returned when there are genuinely **no tool calls**:

```python
# step 5 — AFTER (fixed)
if not tool_calls and text_response:
    return text_response
```

Tool calls now always take priority when both a text response and tool calls are present. This matches the original intent documented in the comment for step 6.

## Affected scenario

Reproduced with `openrouter/anthropic/claude-haiku-4.5` and `openrouter/anthropic/claude-sonnet-4.6` — models that sometimes emit a reasoning/text preamble alongside their tool call declarations.

Closes #4788

---

> [!NOTE]
> **Medium Risk**
> Changes response selection logic in `LLM._handle_non_streaming_response`, which can alter return types (string vs tool-call list) for some model outputs. Low implementation complexity, but impacts tool-execution flows that depend on this branching.
> 
> **Overview**
> Fixes a bug where `_handle_non_streaming_response` could return `text_response` and silently discard `tool_calls` when `available_functions` is `None` and the model returns both content and tool calls.
> 
> The tool-call branch is now evaluated *before* the text-response branch, ensuring callers that execute tools externally reliably receive the `tool_calls` list in this scenario.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7a656ae0046f707ac29e84bc06022bb01849c219. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1234,7 +1234,15 @@ class LLM(BaseLLM):
         # --- 4) Check for tool calls
         tool_calls = getattr(response_message, "tool_calls", [])
 
-        # --- 5) If no tool calls or no available functions, return the text response directly as long as there is a text response
+        # --- 5) If there are tool calls but no available functions, return the tool calls.
+        # This allows the caller (e.g., executor) to handle tool execution externally.
+        # Must be checked BEFORE the text-response branch so that tool calls are not
+        # silently discarded when the LLM returns both content and tool_calls in the
+        # same response (a valid behaviour for some models/providers).
+        if tool_calls and not available_functions:
+            return tool_calls
+
+        # --- 6) If no tool calls (or available_functions provided), return text response
         if (not tool_calls or not available_functions) and text_response:
             self._handle_emit_call_events(
                 response=text_response,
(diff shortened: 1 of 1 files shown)


NIK-TIGER-BILL:
Thanks for the review, @cursor[bot]! The async path fix was already included in this PR — commit `cb6e57e` applies the same narrowing to `_ahandle_non_streaming_response` as well. Both sync and async paths now use `if not tool_calls and text_response:`.

Issue #4748, opened 2026-03-06
[BUG] AWS Bedrock tool calls extract empty arguments - uses wrong field name


MatthiasHowellYopp:
### Description

CrewAI's tool argument extraction only checks OpenAI's `function.arguments` format and fails to extract Bedrock's `input` field, causing all tool calls from AWS Bedrock models to receive empty arguments `{}` regardless of what the LLM provides.

The bug is in `crew_agent_executor.py` line ~850 where the code uses:
```python
func_args = func_info.get("arguments", "{}") or tool_call.get("input", {})
```

Since `.get("arguments", "{}")` returns the string `"{}"` (not None), the `or` operator never evaluates the Bedrock `input` field.

### Steps to Reproduce

1. Configure CrewAI to use an AWS Bedrock model (e.g., `us.amazon.nova-pro-v1:0`)
2. Create an agent with a tool that requires parameters:
```python
from crewai import Agent, Task, Crew
from crewai.tools import tool

@tool("search_tool")
def search_tool(query: str) -> str:
    """Search for information."""
    return f"Searched for: {query}"

agent = Agent(
    role="Researcher",
    goal="Search for information",
    backstory="Expert researcher",
    tools=[search_tool],
    llm="bedrock/us.amazon.nova-pro-v1:0"
)

task = Task(
    description="Search for 'AWS Bedrock features'",
    expected_output="Search results",
    agent=agent
)

crew = Crew(agents=[agent], tasks=[task])
result = crew.kickoff()
```

3. Observe that the tool receives empty arguments `{}` instead of `{"query": "AWS Bedrock features"}`
4. Tool fails or returns incorrect results

### Expected behavior

Tool should receive the arguments that the Bedrock model provides in the `input` field: `{"query": "AWS Bedrock features"}`

### Screenshots/Code snippets

```python
func_args = func_info.get("arguments", "{}") or tool_call.get("input", {})
```

**Bedrock tool call format:**
```python
{
    "name": "search_tool",
    "input": {"query": "AWS Bedrock features"}  # ← This field is never extracted
}
```

**OpenAI tool call format:**
```python
{
    "function": {
        "name": "search_tool",
        "arguments": {"query": "test"}  # ← This works
    }
}
```

### Operating System

Other (specify in additional context)

### Python Version

3.12

### crewAI Version

1.10.1 - commit 87759cdb from main branch

### crewAI Tools Version

N/A

### Virtual Environment

Venv

### Evidence

[...]

PR #4803, opened 2026-03-11
fix: extract Bedrock tool arguments from input field correctly

alvinttang:
## Summary

Fixes #4748 - AWS Bedrock tool calls extract empty arguments due to wrong default value in `get()`.

All AWS Bedrock models (Nova Pro, Nova Lite, Claude via Bedrock, etc.) receive empty `{}` arguments for every tool call, regardless of what the LLM actually provides. This causes tools to fail or return incorrect results.

## Root cause

In `CrewAgentExecutor._parse_native_tool_call()` (line 850):

```python
# BEFORE (buggy)
```

`func_info.get("arguments", "{}")` returns the string `"{}"` when `"arguments"` is absent (Bedrock format). Since `"{}"` is truthy, Python's `or` operator **never evaluates** `tool_call.get("input", {})`, which is where Bedrock stores its arguments.

The correct pattern already exists in `utilities/agent_utils.py` line 1155:
```python
func_args = func_info.get("arguments") or tool_call.get("input") or {}
```

## Fix

One-line change: remove the default value `"{}"` from `get("arguments")` so it returns `None` when absent, allowing the `or`-chain to fall through to Bedrock's `"input"` field:

```python
# AFTER (fixed)
func_args = func_info.get("arguments") or tool_call.get("input") or {}
```

## Test plan

- [ ] Verify Bedrock tool call format `{"name": "search_tool", "input": {"query": "test"}}` correctly extracts arguments
- [ ] Verify OpenAI tool call format `{"function": {"name": "search_tool", "arguments": {"query": "test"}}}` still works
- [ ] Verify empty tool calls (no arguments in either format) default to `{}`
- [ ] Run existing test suite: `pytest lib/crewai/tests/agents/`

---

> [!NOTE]
> **Medium Risk**
> Touches core tool-call argument extraction used during agent execution; while the change is small and targeted, it can affect tool invocation behavior across providers when `arguments` is absent or empty.
> 
> **Overview**
> Fixes native tool-call parsing for dict-based providers (notably AWS Bedrock) by no longer defaulting missing `function.arguments` to the truthy string `"{}"`, allowing `_parse_native_tool_call()` to fall through to `tool_call["input"]` and correctly pass real tool arguments.
> 
> This changes the default behavior for tool calls with no provided args to return an empty dict (`{}`) rather than a stringified empty object, improving consistency across provider formats.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4d8f15f4284bcfa659f486cb404bca8b716f9be0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -847,7 +847,7 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
             func_name = sanitize_tool_name(
                 func_info.get("name", "") or tool_call.get("name", "")
             )
-            func_args = func_info.get("arguments", "{}") or tool_call.get("input", {})
+            func_args = func_info.get("arguments") or tool_call.get("input") or {}
             return call_id, func_name, func_args
         return None
 


alvinttang:
Closing — overlaps with #4855 which takes a more complete approach.

PR #4806, opened 2026-03-11
fix(llm): prevent text response from overriding native tool calls

mvanhorn:
## Summary
- When the LLM returns both a text response and `tool_calls`, the executor passes `available_functions=None`. The existing condition `(not tool_calls or not available_functions) and text_response` at step 5 matched because `not available_functions` was `True`, so the text response was returned and tool calls were silently discarded.
- Fix: reorder the checks so that "tool_calls present but no available_functions" is evaluated **before** the text-return path. This ensures tool calls are returned to the caller (executor) for handling.
- No new dependencies or breaking changes.

## Test plan
- [ ] Verify with an LLM that returns both text and tool_calls when `available_functions=None` -- tool calls should be returned, not text
- [ ] Verify that when only text is returned (no tool_calls), behavior is unchanged
- [ ] Verify that when `available_functions` is provided, tool execution still works normally

Fixes #4788

---

> [!NOTE]
> **Medium Risk**
> Changes LLM non-streaming return-path precedence so responses with `tool_calls` no longer fall back to text when `available_functions` is `None`, which may alter what downstream callers receive in this edge case.
> 
> **Overview**
> Prevents native `tool_calls` from being silently dropped when the LLM returns both text and tool calls but `available_functions` is not provided.
> 
> Reorders the non-streaming and async non-streaming response handling to **return `tool_calls` first** (and emit `LLMCallType.TOOL_CALL` completion events) before the text-return path, while keeping existing behavior when there are no tool calls or when tool execution is available.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 33985614e0a5b973d7fea32b678408e3e5a3eaa4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1248,8 +1248,8 @@ class LLM(BaseLLM):
             )
             return tool_calls
 
-        # --- 6) If no tool calls or no available functions, return the text response directly as long as there is a text response
-        if (not tool_calls or not available_functions) and text_response:
+        # --- 6) If no tool calls, return the text response directly
+        if not tool_calls and text_response:
             self._handle_emit_call_events(
                 response=text_response,
                 call_type=LLMCallType.LLM_CALL,
@@ -1387,7 +1387,8 @@ class LLM(BaseLLM):
             )
             return tool_calls
 
-        if (not tool_calls or not available_functions) and text_response:
+        # If no tool calls, return the text response directly
+        if not tool_calls and text_response:
             self._handle_emit_call_events(
                 response=text_response,
                 call_type=LLMCallType.LLM_CALL,
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1239,6 +1239,13 @@ class LLM(BaseLLM):
         # Must be checked before the text-return path so that tool calls are not
         # silently discarded when the LLM also returns a text response.
         if tool_calls and not available_functions:
+            self._handle_emit_call_events(
+                response=tool_calls,
+                call_type=LLMCallType.TOOL_CALL,
+                from_task=from_task,
+                from_agent=from_agent,
+                messages=params["messages"],
+            )
             return tool_calls
 
         # --- 6) If no tool calls or no available functions, return the text response directly as long as there is a text response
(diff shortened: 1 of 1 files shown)


alvinttang (commented):
The ordering fix is correct: by checking "tool_calls and not available_functions" before the text-return path, native tool calls are no longer silently discarded when the LLM returns both a text response and tool calls in the same message. This is a subtle but important behavior difference — many LLMs (especially Claude) routinely include a preamble text alongside tool calls. One concern: the new early-return path returns tool_calls without emitting any call events via _handle_emit_call_events, whereas the old code path did emit them before returning text. Verify that callers handling the returned tool_calls list are responsible for their own event emission, otherwise observability/tracing could silently drop tool call spans.

mvanhorn:
Addressed both issues in c01d4e3:

1. **Async path reordering** - `_ahandle_non_streaming_response` now checks `tool_calls and not available_functions` before the text-return path, matching the sync method. Good catch, @cursor[bot].

2. **Event emission** - Both sync and async paths now emit `TOOL_CALL` events via `_handle_emit_call_events` before returning raw tool calls. This ensures monitoring and callbacks still fire when the caller handles tool execution externally. @alvinttang - valid point, thanks.

mvanhorn:
Both of these look like they were reviewed against an earlier revision of the branch, and are already addressed at the current head.

**Async method not updated with same fix**: `_ahandle_non_streaming_response` has the same ordering as the sync path. At current head, `llm.py:1380` is `if tool_calls and not available_functions:` and `llm.py:1391` is `if not tool_calls and text_response:`, mirroring `1241` and `1252` in `_handle_non_streaming_response`. That landed in c01d4e3, "apply tool call ordering fix to async path and emit call events". The line numbers in the report, 1369 and 1381, do not correspond to those checks on this revision.

**Redundant condition branch after reordering**: also already gone. The condition in both handlers is now plain `if not tool_calls and text_response:`, with the dead `or not available_functions` removed in 3398561. The two remaining occurrences of `not tool_calls or not available_functions`, at `llm.py:970` and `llm.py:1605`, are in different methods that were not reordered, so the sub-condition is still meaningful there.

So no change needed for either. Happy to be corrected if you are looking at something I am not.

PR #4819, opened 2026-03-11
Fix LiteLLM response_model and native tool call handling

### LINKED ISSUE #4697: [BUG] When supports_function_calling() is True, only the output_pydantic model is sent to LiteLLM (the custom agent tools are discard)

echoVic:
This pull request fixes two related bugs in the LiteLLM integration used by the `LLM` wrapper:

- When `supports_function_calling()` is true and a `response_model` is provided, the internal instructor path was used even when native tools were also passed. This caused custom agent tools to be ignored in favor of the structured output model.
- Native tool calls could be dropped when the LLM returned both text and tool calls in the same response. In that case, the text branch won and the tool calls were never surfaced back to the executor.

Key changes:

1. Only route through `InternalInstructor` when no tools are being passed.
   This ensures that when tools are provided, the call behaves like a normal tool-enabled completion and agent tools are preserved.

2. Prefer tool calls over plain text when both are present.
   The non-streaming response handlers now return tool calls first when they exist, and only fall back to text when there are no tool calls.

These changes are applied to both the synchronous and asynchronous non-streaming handlers so that behavior is consistent across code paths.

Closes #4697.
Closes #4788.

---

> [!NOTE]
> **Medium Risk**
> Changes LiteLLM non-streaming return semantics around `response_model` and tool-calling, which can affect downstream executors that previously received plain text instead of tool call objects. Risk is moderate due to altered control flow in a core LLM wrapper, but scoped to LiteLLM non-streaming sync/async handlers.
> 
> **Overview**
> Fixes LiteLLM non-streaming handling so `response_model` routing through `InternalInstructor` is **skipped when native `tools` are provided**, preserving tool-enabled completions.
> 
> Adjusts both sync and async non-streaming response handlers to **prefer returning `tool_calls`** when present but `available_functions` isn’t supplied (so callers can execute tools), and only fall back to returning text when no tool calls exist.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f647049aa545f74414ca9e6e1475fab0aaa99728. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1150,7 +1150,8 @@ class LLM(BaseLLM):
             str: The response text
         """
         # --- 1) Handle response_model with InternalInstructor for LiteLLM
-        if response_model and self.is_litellm:
+        # Only use this path when no native tools are being passed.
+        if response_model and self.is_litellm and not params.get("tools"):
             from crewai.utilities.internal_instructor import InternalInstructor
 
             messages = params.get("messages", [])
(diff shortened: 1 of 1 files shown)


HydrogenC:
I met with a bug where my LiteLLM connected model (that supports native tool calling) keeps outputting ReAct-style tool calls. This PR may be a solution of it.

Vidit-Ostwal:

PR #4828, opened 2026-03-13
fix: prevent output_pydantic from injecting tool schema when LLM lacks function calling

### LINKED ISSUE #4695: [BUG] output_pydantic model injected as native tool even when supports_function_calling() is False
### Description

When a Task has output_pydantic set and the LLM is a LiteLLM provider (Ollama) without function calling support. The output_pydantic Pydantic model is automatically injects as a native function/tool in the LiteLLM call, and forces the LLM to produce it immediately via tool_choice. As a result:
- The LLM is forced to return the structured output immediately, without having gathered any data.
- All fields come back empty or with placeholder values
- There is no react system (with Action, Action Input and Observation like before)

### Steps to Reproduce

```python

class MyOutput(BaseModel):
    name: str
    value: str

class MySearchTool(BaseTool):
    name: str = "my_search_tool"

task = Task(
    agent=agent,
    output_pydantic=MyOutput,
)
```

1. Set supports_function_calling() to False (default for custom Ollama models).
2. Assign a tool to the agent and set output_pydantic on the task.
3. Run the crew.
4. Observe that LiteLLM receives:

```json
{
  "tools": [{"function": {"name": "MyOutput", .....}}],
  "tool_choice": {"type": "function", "function": {"name": "MyOutput"}}
}
```

### Expected behavior

The agent must not receive any tools in the native format.

### Screenshots/Code snippets

X

### Operating System

Ubuntu 20.04

### Python Version

3.11

### crewAI Version

1.10.0

### crewAI Tools Version

1.10.0

### Virtual Environment

Venv

### Evidence

X

### Possible Solution

I search and found this in internal_instructor.py : 
```python
self._client = instructor.from_litellm(completion) # always uses Mode.TOOLS by default
```

But I was unable to fix this issue in local sorry

### Additional context

X

giulio-leone:
## Summary

Fixes #4695

When a Task has `output_pydantic` set and the LLM does **not** support native function calling (e.g., Ollama models), the executor correctly falls back to the ReAct text-based loop. However, `response_model` was still forwarded to the LLM call, causing `InternalInstructor` to inject the Pydantic schema as a tool via `instructor.from_litellm(completion)` in `Mode.TOOLS` mode — which models without function-calling support cannot handle.

## Root Cause

`_invoke_loop_react()` and `_ainvoke_loop_react()` both passed `self.response_model` to `get_llm_response()` / `aget_llm_response()`. This triggered the instructor path in `LLM._process_call_params()` (line 1153), which wraps the LLM with `instructor.from_litellm(completion)` using the default `Mode.TOOLS` — injecting the Pydantic model as a native tool regardless of whether the model actually supports function calling.

## Fix

Pass `response_model=None` in both `_invoke_loop_react()` and `_ainvoke_loop_react()`. The task-level `convert_to_model()` in `task.py` (line 1021-1040) already handles text → Pydantic conversion after the ReAct loop finishes, so instructor-based extraction is unnecessary and harmful in this code path.

## Changes

- **`lib/crewai/src/crewai/agents/crew_agent_executor.py`**: Set `response_model=None` in both sync and async ReAct loops
- **`lib/crewai/tests/agents/test_react_response_model.py`**: Added 2 regression tests:
  - Verifies `response_model=None` is passed to `get_llm_response` in ReAct mode  
  - Verifies `_invoke_loop_react` is selected (not native tools) when function calling is unsupported

## Test Results

- 2 new tests pass ✅
- 29 existing sync executor tests pass ✅
- No regressions

---

> [!NOTE]
> **Medium Risk**
> Changes how `output_pydantic` is handled in the ReAct execution path, which can affect structured output behavior for models without native tool/function calling. Risk is mitigated by added regression tests covering the selection of ReAct mode and ensuring `response_model` is not forwarded.
> 
> **Overview**
> Prevents `output_pydantic` from being forwarded to the LLM call when the executor falls back to the ReAct text loop, by forcing `response_model=None` in both sync and async ReAct invocations so non-function-calling models don’t receive an injected tool schema.
> 
> Adds regression tests ensuring ReAct mode is chosen when `supports_function_calling()` is false and that `get_llm_response` is called with `response_model=None` in this path.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 16be9acaa748a405be2069c878e0f4f63e9ce8e5. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -370,28 +370,15 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
                     verbose=self.agent.verbose,
                 )
                 if self.response_model is not None:
+                    answer_str = answer if isinstance(answer, str) else str(answer)
                     try:
-                        if isinstance(answer, BaseModel):
-                            output_json = answer.model_dump_json()
-                            formatted_answer = AgentFinish(
-                                thought="",
-                                output=answer,
-                                text=output_json,
-                            )
-                        else:
-                            self.response_model.model_validate_json(answer)
-                            formatted_answer = AgentFinish(
-                                thought="",
-                                output=answer,
-                                text=answer,
-                            )
-                    except ValidationError:
-                        # If validation fails, convert BaseModel to JSON string for parsing
-                        answer_str = (
-                            answer.model_dump_json()
-                            if isinstance(answer, BaseModel)
-                            else str(answer)
+                        self.response_model.model_validate_json(answer_str)
+                        formatted_answer = AgentFinish(
+                            thought="",
+                            output=answer_str,
+                            text=answer_str,
                         )
+                    except ValidationError:
                         formatted_answer = process_llm_response(
                             answer_str, self.use_stop_words
                         )  # type: ignore[assignment]
@@ -1216,28 +1203,15 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
                 )
 
                 if self.response_model is not None:
+                    answer_str = answer if isinstance(answer, str) else str(answer)
                     try:
-                        if isinstance(answer, BaseModel):
-                            output_json = answer.model_dump_json()
-                            formatted_answer = AgentFinish(
-                                thought="",
-                                output=answer,
-                                text=output_json,
-                            )
-                        else:
-                            self.response_model.model_validate_json(answer)
-                            formatted_answer = AgentFinish(
-                                thought="",
-                                output=answer,
-                                text=answer,
-                            )
-                    except ValidationError:
-                        # If validation fails, convert BaseModel to JSON string for parsing
-                        answer_str = (
-                            answer.model_dump_json()
-                            if isinstance(answer, BaseModel)
-                            else str(answer)
+                        self.response_model.model_validate_json(answer_str)
+                        formatted_answer = AgentFinish(
+                            thought="",
+                            output=answer_str,
+                            text=answer_str,
                         )
+                    except ValidationError:
                         formatted_answer = process_llm_response(
                             answer_str, self.use_stop_words
                         )  # type: ignore[assignment]
--- a/lib/crewai/tests/agents/test_react_response_model.py
+++ b/lib/crewai/tests/agents/test_react_response_model.py
@@ -11,7 +11,7 @@ The fix: _invoke_loop_react() must pass response_model=None so the LLM
 produces plain text that the ReAct parser can handle normally.
 """
 
-from unittest.mock import MagicMock, patch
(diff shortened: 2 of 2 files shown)


giulio-leone:
Hi team — rebased onto latest `main` to resolve the behind-branch state. The fix prevents `output_pydantic` from injecting a tool schema when the LLM doesn't support function calling, which was causing unexpected tool-use behavior. Includes tests covering the affected code paths. Would appreciate a review. Thanks!

giulio-leone:
Rebased onto current `main` and addressed the remaining Cursor thread with a small follow-up on top of the existing no-function-calling fix.\n\nWhat changed:\n- removed the unreachable `BaseModel` handling from the sync/async ReAct paths after forcing `response_model=None`\n- kept the text fallback path intact when the response is still ReAct-style `Final Answer: ...` text\n- added focused sync + async regression coverage for the direct-JSON string path\n\nValidation:\n- `UV_PYTHON=3.10 uv run --frozen --group dev ruff check lib/crewai/src/crewai/agents/crew_agent_executor.py lib/crewai/tests/agents/test_react_response_model.py`\n- `UV_PYTHON=3.10 uv run --frozen --group dev pytest lib/crewai/tests/agents/test_react_response_model.py -q -n 0`\n- both commands passed twice with no code changes between runs, then passed again from committed head `174893dbcc6e8f99fa69840421d93e135e7f3d2a`\n\nExact-source proof against current `origin/main` (`09b84dd2b032cc2f2cdc6b6f4751ede436fc9005`):\n- rebased head `174893dbcc6e8f99fa69840421d93e135e7f3d2a`: both `_invoke_loop_react()` and `_ainvoke_loop_react()` pass `response_model=None`\n- current `origin/main`: both still pass `MyOutput`\n- on the rebased head, a valid JSON string response in both sync/async ReAct paths returns directly without hitting `process_llm_response`; ReAct-style `Final Answer: {"answer": "hello"}` text still parses to the same output payload\n\nSo the original no-function-calling regression remains fixed against current `main`, and the dead branches flagged in review are now gone without changing the observable result.

giulio-leone:
## Refreshed onto current `main` ✅

**Head:** `174893db` (2 commits cleanly stacked onto `origin/main` @ `09b84dd2`)

**Double-pass validation:**
```
uv run pytest tests/agents/test_react_response_model.py -q  →  4 passed (pass 1 + pass 2)
```

**Fix summary:** `crew_agent_executor.py` no longer injects `output_pydantic` as a tool-calling schema when the LLM lacks native function-calling support (ReAct path). This prevents the model from receiving a contradictory schema that causes hallucinated tool names. The refactor commit also removes two unreachable code branches confirmed dead by the new test matrix.

PR #4831, opened 2026-03-13
fix: prevent tool_calls from being silently discarded when LLM also returns text

SinzoL:
When the LLM response contains both text and tool_calls with available_functions=None, the condition incorrectly returns only text. Reorder checks so tool_calls always take priority. Applied to both sync and async code paths. Fixes #4788

---

> [!NOTE]
> **Medium Risk**
> Changes return behavior of non-streaming LLM calls: callers that previously received a text `str` may now receive `tool_calls` when the model returns both, which could affect downstream type/flow assumptions.
> 
> **Overview**
> Ensures `tool_calls` are not silently dropped when an LLM response includes both text and tool invocations but `available_functions` is `None`.
> 
> Reorders the non-streaming sync and async response handling to return `tool_calls` first in that scenario, leaving tool execution to the caller instead of falling back to the text content.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a0960d2cc152c185614cb1930a3aa142a159b454. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1234,8 +1234,15 @@ class LLM(BaseLLM):
         # --- 4) Check for tool calls
         tool_calls = getattr(response_message, "tool_calls", [])
 
-        # --- 5) If no tool calls or no available functions, return the text response directly as long as there is a text response
-        if (not tool_calls or not available_functions) and text_response:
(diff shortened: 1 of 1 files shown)


greysonlalonde:
Duplicate of #4806. Closing.

PR #4834, opened 2026-03-13
fix: propagate ContextVar state to async_execution worker threads

### LINKED ISSUE #4822: [BUG] async_execution=True loses ContextVar state — threading.Thread not using copy_context()
## Summary      
                                                                                                              
  Tasks with `async_execution=True` are executed via `threading.Thread()` without copying
  the current `contextvars.Context`. This causes any `ContextVar` set on the calling thread
  to silently reset to its default value inside the worker thread.

  Related to #4168 (token tracking race condition) — same root cause, different symptom.

  ## Environment

  - crewai: 1.7.2 but checked 1.10.0.1
  - Python: 3.12

  ## Root Cause

  `execute_async()` in `task.py` spawns the worker thread without a context copy:

  ```python
  def execute_async(self, agent, context, tools) -> Future[TaskOutput]:
      future: Future[TaskOutput] = Future()
      threading.Thread(
          daemon=True,
          target=self._execute_task_async,
          args=(agent, context, tools, future),
      ).start()
      return future

  threading.Thread() does not inherit contextvars.Context from the spawning thread.
  asyncio.to_thread() and loop.run_in_executor() do, which is why the kickoff_async()
  path doesn't have this problem.
```

  ## Impact

  Libraries that rely on ContextVar for request-scoped state are silently broken for
  tasks using async_execution=True:

  - OpenTelemetry — trace context and baggage are lost. LLM calls inside these tasks
  appear as orphaned traces with no parent span.
  - Langfuse and other tracing SDKs — session context set on the calling thread is
  not visible in the worker.
  - Any custom ContextVar state — resets to default with no error raised.

### Steps to Reproduce

```
  import contextvars, threading

  my_var = contextvars.ContextVar("my_var", default=None)
  my_var.set("hello")

  threading.Thread(target=lambda: print(my_var.get())).start()
  # prints: None

  vs. asyncio.to_thread():

  import asyncio, contextvars

  my_var = contextvars.ContextVar("my_var", default=None)
  my_var.set("hello")

  async def main():
      await asyncio.to_thread(lambda: print(my_var.get()))
      # prints: hello

  asyncio.run(main())
```

### Expected behavior

  ContextVar values set on the calling thread should be accessible inside the worker                          
  thread spawned by async_execution=True, consistent with how asyncio.to_thread()                             
  and run_in_executor() behave.     

### Screenshots/Code snippets

```
  import contextvars, threading                                                                               
                                                                                                              
  my_var = contextvars.ContextVar("my_var", default=None)                                                     
  my_var.set("hello")                                                 
```                                        
                  
  # Current behaviour — context is lost
```
  threading.Thread(target=lambda: print(my_var.get())).start()
  # prints: None
```

  # Expected behaviour — context is preserved
```
  import asyncio
  async def main():
      await asyncio.to_thread(lambda: print(my_var.get()))
  asyncio.run(main())
  # prints: hello
```

### Operating System

### Python Version

3.12

### crewAI Version

1.7.2

### crewAI Tools Version

1.7.2

### Virtual Environment

Venv

### Evidence

  Evidence:
  Verified by inspecting task.py in crewai 1.10.0.1:                                                          
                  
      def execute_async(self, agent, context, tools) -> Future[TaskOutput]:                                   
          future: Future[TaskOutput] = Future()
          threading.Thread(
              daemon=True,
              target=self._execute_task_async,
              args=(agent, context, tools, future),
          ).start()
          return future

  threading.Thread() does not copy contextvars.Context from the spawning thread.
  ContextVar values set before execute_async() is called read as their default
  values inside the worker thread. No exception is raised — the failure is silent.

  This affects OpenTelemetry (trace context lost, spans appear orphaned), Langfuse
  (session context lost), and any library using ContextVar for request-scoped state.

  Related: #4168 (token tracking race condition, same root cause), #4286 (open PR).

### Possible Solution

```
  import contextvars

  def execute_async(self, agent, context, tools) -> Future[TaskOutput]:
      future: Future[TaskOutput] = Future()
      ctx = contextvars.copy_context()
      threading.Thread(
          daemon=True,
          target=ctx.run,
          args=(self._execute_task_async, agent, context, tools, future),
      ).start()
      return future
```

  ## Workaround

  Use async def @listen with await crew.kickoff_async() inside asyncio.gather()
  instead of async_execution=True in sync crews — the async path propagates context
  correctly.

### Additional context

  References

  - #4168 — token tracking race condition in async_execution (same root cause)
  - #4286 — open PR for #4168
  - Python docs: contextvars.copy_context()
  - CPython: asyncio.to_thread implementation

SinzoL:
Task.execute_async() spawns a new threading.Thread for background execution, but threading.Thread does not inherit the parent's contextvars.Context. This causes ContextVar values (e.g. tracing spans, tenant IDs, session state) to be lost in the worker thread. Use contextvars.copy_context() to snapshot the current context and run the worker function inside it via ctx.run(). Fixes #4822

---

> [!NOTE]
> **Medium Risk**
> Touches core async task execution by changing how worker threads are started, which could affect any code relying on thread startup semantics. The change is small but impacts tracing/tenant context propagation across all `Task.execute_async` calls.
> 
> **Overview**
> **Fixes lost `ContextVar` state in async task execution.** `Task.execute_async()` now snapshots the current `contextvars` context and starts the background thread via `ctx.run(...)`, ensuring context-local values (e.g., tracing/tenant/session data) propagate into the worker thread.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 06973d3b0ab71fbeedcc38e03db025266c90c54c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/task.py
+++ b/lib/crewai/src/crewai/task.py
@@ -9,6 +9,7 @@ import inspect
 import json
 import logging
 from pathlib import Path
+import contextvars
 import threading
 from typing import (
     Any,
@@ -524,10 +525,13 @@ class Task(BaseModel):
     ) -> Future[TaskOutput]:
         """Execute the task asynchronously."""
         future: Future[TaskOutput] = Future()
+        # Copy the current context so ContextVar values propagate to the
+        # worker thread (e.g. tracing spans, tenant IDs, etc.).
+        ctx = contextvars.copy_context()
         threading.Thread(
             daemon=True,
-            target=self._execute_task_async,
-            args=(agent, context, tools, future),
+            target=ctx.run,
+            args=(self._execute_task_async, agent, context, tools, future),
         ).start()
         return future
 


alvinttang (commented):
## Review

This PR fixes `ContextVar` state not propagating to worker threads spawned by `Task.execute_async()`. The fix uses `contextvars.copy_context()` to snapshot the current context and `ctx.run()` to execute the task function within that snapshot.

### Analysis:

1. **The fix is correct and idiomatic**: `contextvars.copy_context()` + `ctx.run()` is the standard Python pattern for propagating `ContextVar` state to threads. This is well-documented in the Python stdlib and is the right approach.

2. **`ctx.run` signature**: `ctx.run(callable, *args)` — here it's `ctx.run(self._execute_task_async, agent, context, tools, future)`. This correctly passes all positional arguments through. Good.

3. **Thread safety of the context copy**: `copy_context()` creates a shallow copy of the current context at the time of the call. Any `ContextVar.set()` calls in the worker thread will only affect the copy, not the parent. This is the desired behavior — each async task execution gets an isolated copy of the context snapshot. No race conditions here.

4. **Interaction with `Future`**: The `future` object is still shared between the parent and worker thread (it's passed as an argument, not a ContextVar). This is correct — the `Future` needs to be the same object so the parent can read the result.

5. **Import placement**: `import contextvars` is added at module top level alongside `threading`, which is correct. No lazy import needed since `contextvars` is a stdlib module with negligible import cost.

6. **Broader applicability**: Are there other places in crewAI that spawn threads without propagating context? A quick search for `threading.Thread` across the codebase might reveal similar issues. If so, this pattern should be applied consistently.

7. **No test**: Testing ContextVar propagation is straightforward — set a ContextVar in the parent, run `execute_async`, assert the var is visible in the task function. Would be a valuable regression test.

This is a clean, correct, minimal fix. The only suggestion is to check for other thread-spawning sites that might need the same treatment and to add a test.

Vidit-Ostwal:

PR #4841, opened 2026-03-13
fix: allow MCP tools to ignore extra fields like security_context

### LINKED ISSUE #4796: [BUG] Pydantic Validation Error with `security_context` in MCP Tools
### Description

When using `MCPServerAdapter` from `crewai-tools`, CrewAI automatically injects a `security_context` parameter (containing `agent_fingerprint` and `metadata`) into tool calls. However, MCP tools' `inputSchema` (defined by the MCP server) does not include this field. This causes Pydantic validation to fail with an **"Extra inputs are not permitted"** error.

The `CrewAIToolAdapter` creates Pydantic models from MCP tool schemas using `create_model_from_schema()`, which by default rejects extra fields. Since CrewAI's tool execution framework adds `security_context` during parameter validation (before `_run` is called), the validation fails immediately.

**Problematic Code Location:**
`crewai-tools/lib/crewai_tools/adapters/mcp_adapter.py` (Lines 52-59)
```python
tool_name = sanitize_tool_name(mcp_tool.name)
tool_description = mcp_tool.description or ""
args_model = create_model_from_schema(mcp_tool.inputSchema)

class CrewAIMCPTool(BaseTool):
    name: str = tool_name
    description: str = tool_description
    args_schema: type[BaseModel] = args_model  # This model rejects extra fields
```

### Steps to Reproduce

1. Create an agent with MCP tools using `MCPServerAdapter`:

```python
from crewai_tools import MCPServerAdapter

with MCPServerAdapter({
    "url": "http://localhost:9500/sse?mode=rag",
    "transport": "sse",
    "headers": {"X-User-ID": "1", "X-Workflow-Run-ID": "test"}
}) as mcp_tools:
    agent = Agent(
        role="RAG Agent",
        goal="Search knowledge base",
        backstory="Expert at searching knowledge",
        tools=mcp_tools,
        verbose=True
    )
```

2. Execute a task that triggers the MCP tool (e.g., `search_knowledge`)
3. The tool call fails with Pydantic validation error

### Expected behavior

The `security_context` parameter should either:
1. Be filtered out before Pydantic validation, OR
2. The `args_schema` model should be configured to ignore extra fields (e.g., `ConfigDict(extra='ignore')`)

The tool should execute successfully with only the parameters defined in the MCP tool's `inputSchema`.

### Screenshots/Code snippets

None

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

1.10.1

### crewAI Tools Version

1.10.1

### Virtual Environment

Venv

## Error Output

```
│  Tool Failed                                                                                                    │
│  Tool: search_knowledge                                                                                         │
│  Iteration: 6                                                                                                   │
│  Attempt: 3                                                                                                     │
│  Error: Arguments validation failed: 1 validation error for DynamicModel                                        │
│  security_context                                                                                               │
│    Extra inputs are not permitted [type=extra_forbidden, input_value={'agent_fingerprint': {'u...61982',        │
│  'metadata': {}}}, input_type=dict]                                                                             │
│      For further information visit https://errors.pydantic.dev/2.11/v/extra_forbidden                           │
│  Expected arguments: {"query": {"title": "Query", "type": "string"}, "top_k": {"anyOf": [{"type": "integer"},   │
│  {"type": "null"}], "default": null, "title": "Top K"}}                                                         │
│  Required: ["query"]                                                                                            │
```

### Possible Solution

None

### Additional context

None

Ker102:
Fixes #4796.

Ran into this while working with MCP tools — CrewAI injects `security_context` into tool arguments during execution, but the Pydantic models for MCP tools reject it because `create_model_from_schema` defaults to `extra="forbid"`.

The fix is in two places:
- `mcp_adapter.py` — the `CrewAIToolAdapter` used by `MCPServerAdapter`
- `tool_resolver.py` — the native `MCPToolResolver` path

Both now pass `ConfigDict(extra="ignore")` so the MCP tool models silently drop fields like `security_context` that aren't part of the tool's input schema, instead of blowing up with a validation error. Used `"ignore"` rather than `"allow"` so unknown fields don't get forwarded to the MCP server.

The defaults in `create_model_from_schema` stay unchanged for everything else.

---

> [!NOTE]
> **Low Risk**
> Low risk: scopes Pydantic config changes to MCP tool argument models so unexpected fields (e.g., injected `security_context`) no longer cause validation failures, without affecting non-MCP schema generation.
> 
> **Overview**
> Fixes MCP tool invocations failing when CrewAI injects fields not present in the MCP `inputSchema`.
> 
> Both the MCP tool adapter (`mcp_adapter.py`) and MCP tool schema resolver (`tool_resolver.py`) now build dynamic Pydantic arg models with `ConfigDict(extra="ignore")`, dropping unknown fields instead of rejecting the call.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1064008e28229a46b755c43064925f874ad63a6a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ori-cofounder:
Nice fix! The  approach is the right call here.

One edge case worth considering: if a future MCP server happens to define a field named  in its own  (unlikely but possible), this fix would silently drop it on the Pydantic side while still forwarding it to the server — which could cause confusing failures downstream.

A defensive approach would be to strip CrewAI-injected fields *before* Pydantic validation rather than at the model level:

```python
CREWAI_INJECTED_FIELDS = {"security_context"}

def run(self, **kwargs):
    clean_kwargs = {k: v for k, v in kwargs.items() if k not in CREWAI_INJECTED_FIELDS}
    return super().run(**clean_kwargs)
```

This keeps Pydantic strict (catches real schema mismatches) while explicitly handling the injection. But for the current use case your approach is simpler and the risk is low — `security_context` is an unlikely MCP schema field name.

Either way, this unblocks a real pain point. 👍

Ker102:
Thanks! Yeah the edge case is pretty theoretical — security_context as a field name in an actual MCP tool schema would be a wild coincidence. If it ever came up, stripping injected fields before validation would be the cleaner long-term approach for sure. For now this keeps things simple and gets people unblocked.

Vidit-Ostwal:

PR #4940, opened 2026-03-18
fix: prioritize tool_calls over text when available_functions is None

alvinttang:
## Summary

Fixes #4788

- When an LLM returns both `content` (text) **and** `tool_calls` in the same response, and `available_functions` is `None` (the executor handles tool execution externally), the text response was being returned instead of the tool calls
- The tool calls were silently discarded, and the text was treated as a final answer by the agent executor loop
- **Fix:** swap the order of checks 5) and 6) in `_handle_non_streaming_response` so that `tool_calls` are returned to the caller before falling back to the text path

This is a one-line logic change (reorder two `if` blocks) in `lib/crewai/src/crewai/llm.py`.

## Affected models

Any model that returns both text and tool_calls in the same response, including:
- `anthropic/claude-haiku-4.5` (via OpenRouter / Bedrock)
- `anthropic/claude-sonnet-4.6` (via OpenRouter / Bedrock)

## Test plan

- [x] Added unit test `test_tool_calls_prioritized_over_text_when_no_available_functions` that mocks a response with both text and tool_calls and asserts tool_calls are returned
- [ ] Manual testing with Claude models via OpenRouter that reproduce the original issue

---

> [!NOTE]
> **Low Risk**
> Low risk logic reorder in non-streaming LLM response handling; behavior change only affects the edge case where providers return both `content` and `tool_calls` without `available_functions`, potentially altering executor flow but now matches intended tool-execution semantics.
> 
> **Overview**
> Fixes non-streaming LiteLLM response handling so that when a model returns both **text** and `tool_calls` and `available_functions` is `None`, `_handle_non_streaming_response` returns the `tool_calls` (for external execution) instead of falling back to the text response.
> 
> Adds a regression unit test (`test_tool_calls_prioritized_over_text_when_no_available_functions`) that mocks a response containing both fields and asserts `tool_calls` are prioritized.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 497521d7d73d1167eec3be57ef99300af001840f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1234,7 +1234,14 @@ class LLM(BaseLLM):
         # --- 4) Check for tool calls
         tool_calls = getattr(response_message, "tool_calls", [])
 
-        # --- 5) If no tool calls or no available functions, return the text response directly as long as there is a text response
+        # --- 5) If there are tool calls but no available functions, return the tool calls
+        # This allows the caller (e.g., executor) to handle tool execution.
+        # This must be checked before the text fallback below, because some LLMs
+        # return both text *and* tool_calls; the tool_calls should take priority.
+        if tool_calls and not available_functions:
+            return tool_calls
+
+        # --- 6) If no tool calls (or no available functions), return the text response directly as long as there is a text response
         if (not tool_calls or not available_functions) and text_response:
             self._handle_emit_call_events(
                 response=text_response,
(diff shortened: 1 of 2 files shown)


alvinttang:
Duplicate of #4872 which already fixes #4788, closing.

PR #4941, opened 2026-03-18
fix: prioritize tool calls over text when available_functions is None

alvinttang:
## Summary
- When an LLM (e.g. Claude) returns both text content AND tool calls in a single response, the text was incorrectly returned instead of the tool calls — silently discarding the tool calls
- Root cause: in `_handle_non_streaming_response` (and its async counterpart), the condition `(not tool_calls or not available_functions) and text_response` at step 5 matched before step 6 could return tool calls for external execution
- Fix: reorder steps 5 and 6 so that tool calls with `available_functions=None` are returned first, then fall back to text-only responses

## Test plan
- [x] Added `test_tool_calls_prioritized_over_text_when_no_available_functions` — sync path
- [x] Added `test_tool_calls_prioritized_over_text_when_no_available_functions_async` — async path
- Both tests mock an LLM response containing both `content` text AND `tool_calls`, pass `available_functions=None`, and verify tool calls are returned (not text)

Fixes #4788

---

> [!NOTE]
> **Medium Risk**
> Changes non-streaming LLM response selection logic to return `tool_calls` instead of text when `available_functions` is missing, which can affect downstream callers that previously received a string fallback.
> 
> **Overview**
> **Fixes a non-streaming tool-calling regression.** When an LLM returns both `content` text and `tool_calls` in the same response and `available_functions` is `None`, the LLM now returns the `tool_calls` list (for external execution) instead of incorrectly falling back to text.
> 
> Adds sync and async regression tests that mock a Claude-style response containing both text and tool calls and assert tool calls are prioritized.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ac4d00d0d965ad453fe081df8d2d6776cd6d2051. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1234,8 +1234,15 @@ class LLM(BaseLLM):
         # --- 4) Check for tool calls
         tool_calls = getattr(response_message, "tool_calls", [])
 
-        # --- 5) If no tool calls or no available functions, return the text response directly as long as there is a text response
-        if (not tool_calls or not available_functions) and text_response:
+        # --- 5) If there are tool calls but no available functions, return the tool calls
+        # This allows the caller (e.g., executor) to handle tool execution.
+        # This must be checked before the text fallback below, because some LLMs
+        # (e.g. Claude) return both text AND tool calls in the same response.
+        if tool_calls and not available_functions:
+            return tool_calls
+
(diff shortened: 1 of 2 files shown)


alvinttang:
Duplicate of #4872, closing.

Issue #4972, opened 2026-03-20
[BUG] _parse_native_tool_call drops Bedrock Converse API tool arguments — always passes empty dict


sajilkumarmk:
### Description

When using AWS Bedrock as the LLM provider with native function calling, all tool arguments are silently dropped. Every tool call receives an empty dict {} instead of the actual arguments, causing pydantic validation failures on any tool with required parameters.

### Steps to Reproduce

1. Configure a crewai project with BedrockCompletion as the LLM (any Claude model via Converse API)
2. Create a custom BaseTool with required arguments (e.g. a tool with search_query: str in its args_schema)
3. Create a crew/agent that uses this tool
4. Run the crew with kickoff() or kickoff_async()
5 .Observe that every tool call fails with Field required [type=missing, input_value={}, input_type=dict]

### Expected behavior

The agent should pass the arguments from the Bedrock Converse API response (from the "input" field) to the tool. For example, if the LLM returns {"name": "my_tool", "input": {"search_query": "test"}, "toolUseId": "abc"}, the tool should receive search_query="test".

### Screenshots/Code snippets

Error output (repeats every iteration until max iterations):

Tool 'search_in_user_knowledgebase' arguments validation failed: 1 validation error for SearchUserKnowledgebaseToolSchema
search_query
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
Expected arguments: {"search_query": {"description": "Mandatory semantic search query", "title": "Search Query", "type": "string"}}
Required: ["search_query"]
Bug location — crewai/agents/crew_agent_executor.py, method _parse_native_tool_call, dict handling branch:

func_info = tool_call.get("function", {})
func_args = func_info.get("arguments", "{}") or tool_call.get("input", {})
Bedrock returns {"name": "...", "input": {...}, "toolUseId": "..."} — no "function" key. So func_info is {}, and func_info.get("arguments", "{}") returns the default "{}" (a truthy string). The or short-circuits and tool_call["input"] is never read.

### Operating System

macOS Monterey

### Python Version

3.12

### crewAI Version

1.11.0

### crewAI Tools Version

1.11.0

### Virtual Environment

Poetry

### Evidence

# Reproducing the bug in isolation:
func_info = {}  # No "function" key in Bedrock format
tool_call = {"name": "my_tool", "input": {"search_query": "test"}, "toolUseId": "abc"}

# Current (broken):
[...]

PR #4978, opened 2026-03-20
fix: Bedrock Converse API tool arguments silently dropped (#4972)

guoyangzhen:
## Problem

When using AWS Bedrock with native function calling, **all tool arguments are silently dropped**. Every tool call receives an empty dict `{}` instead of the actual arguments, causing pydantic validation failures on any tool with required parameters.

This affects ALL Bedrock models using native function calling (Claude, Llama, etc.) via the native tool loop code path.

## Root Cause

In `_parse_native_tool_call` (crew_agent_executor.py), the dict-format branch has a truthy default:

```python
```

Bedrock Converse API returns `{name, input, toolUseId}` — **no `"function"` key**.

So `func_info` is `{}`, and `func_info.get("arguments", "{}")` returns the default `"{}"` — a **truthy string**. The `or` short-circuits and `tool_call.get("input", {})` is never evaluated.

## Fix

Remove the default so `get()` returns `None` (falsy), allowing the `or` to fall through:

```python
func_args = func_info.get("arguments") or tool_call.get("input") or "{}"
```

## Verification

```
Bedrock format: {"name": "search_tool", "input": {"search_query": "hello"}, "toolUseId": "abc"}
OLD: func_info.get("arguments", "{}") → "{}" (truthy!) → short-circuits → broken
NEW: func_info.get("arguments") → None (falsy) → falls through to tool_call["input"] ✓

OpenAI format: {"function": {"name": "t", "arguments": "{"x": 1}"}}
NEW: func_info.get("arguments") → "{"x": 1}" (truthy) → same as before ✓
```

## Tests Added

- `test_parse_native_tool_call_bedrock_format` — verifies Bedrock input dict is correctly extracted
- `test_parse_native_tool_call_openai_dict_format` — verifies OpenAI format still works

Fixes #4972

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -847,7 +847,7 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
             func_name = sanitize_tool_name(
                 func_info.get("name", "") or tool_call.get("name", "")
             )
-            func_args = func_info.get("arguments", "{}") or tool_call.get("input", {})
+            func_args = func_info.get("arguments") or tool_call.get("input") or "{}"
             return call_id, func_name, func_args
         return None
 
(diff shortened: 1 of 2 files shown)


greysonlalonde:

PR #4986, opened 2026-03-20
fix: Bedrock Converse API tool arguments being dropped

goingforstudying-ctrl:
## Summary
Fixes #4972 - Bedrock Converse API tool arguments were being dropped due to incorrect fallback logic in _parse_native_tool_call.

## Problem
When using AWS Bedrock as the LLM provider with native function calling, all tool arguments were silently dropped. Every tool call received an empty dict instead of the actual arguments.

## Root Cause
The _parse_native_tool_call method had this logic:

```python
func_args = func_info.get('arguments', '{}') or tool_call.get('input', {})
```

Bedrock returns tool calls with format: `{"name": "...", "input": {...}, "toolUseId": "..."}`

When 'function' key is missing, func_info.get('arguments', '{}') returns the default "{}" (a truthy string), so the OR never falls through to tool_call.get('input').

## Fix
Changed to:
```python
func_args = func_info.get('arguments') or tool_call.get('input') or '{}'
```

This removes the default "{}" from get('arguments', '{}') so it returns None (falsy) when the key is missing, allowing the OR to correctly fall through to tool_call.get('input').

## Changes
| File | Change | Reason |
|------|--------|--------|
| crew_agent_executor.py | Fixed func_args extraction | Correctly handles Bedrock format |
| test_native_tool_calling.py | Added tests for Bedrock/OpenAI formats | Prevent regression |

## Testing
- [x] Added unit tests for Bedrock format parsing
- [x] Added unit tests for OpenAI format parsing (regression test)
- [x] Both tests pass

Closes #4972

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -847,7 +847,7 @@ class CrewAgentExecutor(CrewAgentExecutorMixin):
             func_name = sanitize_tool_name(
                 func_info.get("name", "") or tool_call.get("name", "")
             )
-            func_args = func_info.get("arguments", "{}") or tool_call.get("input", {})
+            func_args = func_info.get("arguments") or tool_call.get("input") or "{}"
             return call_id, func_name, func_args
         return None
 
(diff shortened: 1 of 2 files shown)


sajilkumarmk (approved):
approve

goingforstudying-ctrl:
Rebased with latest main. Ready for review!

greysonlalonde:

PR #5012, opened 2026-03-22
feat: add SafetyMDTool — EVM payment address safety checker

0xzoz:
Adds SafetyMDTool to crewai.tools for verifying payment addresses before agents execute fund transfers.

- Pydantic v2 compatible with typed SafetyMDInput schema
- Never raises — all error paths return descriptive strings
- 3s timeout with graceful fallback
- Returns risk level: low/medium/high/critical
- Free tier: 10 checks/day, no auth required
- API: https://safetymd.p-u-c.workers.dev


greysonlalonde:
Hi! We are generally not accepting any external tool additions at the moment. Please see https://docs.crewai.com/en/guides/tools/publish-custom-tools

PR #5014, opened 2026-03-22
fix: forward max_completion_tokens to Anthropic native provider

khan5v:
## Summary                                                                             
                                                                                         
  - `max_completion_tokens` (OpenAI-style parameter) is silently ignored by the native Anthropic provider                                                                     
  - The parameter falls into `**kwargs` -> `additional_params` and never reaches `self.max_tokens`
  - All Anthropic API calls use the default 4096 regardless of what the user sets        
  - The LiteLLM fallback path handles this correctly (`llm.py:717`), but the native path does not                                                                               
  - The OpenAI native provider already handles both parameters (`completion.py:203`)     

  ## Fix                                                                                 
                                                                                         
  Accept `max_completion_tokens` in `AnthropicCompletion.__init__` with the same fallback chain: `max_tokens` -> `max_completion_tokens` -> `4096`
                                                                                         
  ## Test plan                                           
                                         
  - [x] `max_completion_tokens=150` flows through to `_prepare_completion_params` as `max_tokens: 150`
  - [x] `max_tokens` takes precedence when both are set                                  
  - [x] Default 4096 preserved when neither is set                                       
  - [x] Existing tests unaffected

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -41,6 +41,7 @@ TOOL_SEARCH_TOOL_TYPES: Final[tuple[str, ...]] = (
     "tool_search_tool_bm25_20251119",
 )
 
+ANTHROPIC_DEFAULT_MAX_TOKENS: Final[int] = 4096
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:

PR #5043, opened 2026-03-24
fix: address Copilot review on OpenAI-compatible providers

alex-clawd:
Follow-up to #5042 (merged). Addresses 5 Copilot review comments:

1. **supports_function_calling()** — delegates to parent OpenAI implementation (handles o1 models routed via OpenRouter)
2. **Empty env var guard** — checks for non-empty value before using env var for base_url
3. **Misleading comment** — updated to note DeepSeek/Dashscope have model prefix restrictions
4. **Unused import** — removed MagicMock from test imports
5. **Env restoration** — uses `is not None` instead of truthy check for proper test cleanup

3 files, +11/-8


alex-clawd:
Reopening as #new from the org repo to fix the workflow file deletions (OAuth scope issue with fork pushes). Same changes, clean diff.

PR #5050, opened 2026-03-24
for counting

lorenzejay:


lib/crewai/src/crewai/telemetry/telemetry.py
```diff
@@ -825,64 +825,65 @@ def _operation() -> Span:
...
-                            "tools_names": [
-                                sanitize_tool_name(tool.name)
-                                for tool in task.tools or []
-                            ],
-                        }
-                        for task in crew.tasks
-                    ]
-                ),
-            )
+
```
alex-clawd, line 828: `crew_inputs` should also be gated behind `share_crew` — inputs can contain sensitive data (API keys, PII, customer data). Move this inside the `if crew.share_crew:` block.

lib/crewai/src/crewai/telemetry/telemetry.py
```diff
@@ -825,64 +825,65 @@ def _operation() -> Span:
...
+            if crew.share_crew:
+                self._add_attribute(
+                    span,
```
alex-clawd, line 831: Consider wrapping the agent/task serialization in a try/except for extra resilience — e.g. `agent.i18n.prompt_file` could fail on a misconfigured agent. The outer `_safe_telemetry_operation` catches it, but a targeted guard here gives better error context.

```python
try:
    # agent/task serialization
except Exception:
    pass  # telemetry should never break execution
```

PR #5052, opened 2026-03-24
fix(llm): preserve agent tools when response_model is set

### LINKED ISSUE #4697: [BUG] When supports_function_calling() is True, only the output_pydantic model is sent to LiteLLM (the custom agent tools are discard)

Krishnachaitanyakc:
## Summary
Fixes #4697

When `supports_function_calling()` is `True` and a task has `output_pydantic` set, `InternalInstructor` intercepts the LLM call and creates its own `litellm.completion` invocation that only sends `messages` and `response_model`. The agent's tools (e.g. `my_search_tool`) are silently discarded, forcing the LLM to return structured output immediately without ever calling any of its tools.

The root cause is in `_handle_non_streaming_response` and `_ahandle_non_streaming_response`: the condition `if response_model and self.is_litellm` unconditionally routes to `InternalInstructor`, regardless of whether `params["tools"]` is populated.

## Changes
- Guard the `InternalInstructor` shortcut with `not params.get("tools")` in both `_handle_non_streaming_response` (sync) and `_ahandle_non_streaming_response` (async)
- When tools are present, fall through to the standard `litellm.completion(**params)` path which sends both tools and `response_model` together
- When no tools are present, behavior is unchanged (InternalInstructor is still used for structured output)
- Added two regression tests (sync + async) that verify `litellm.completion` receives both `tools` and `response_model` when both are provided

## Test plan
- [x] New test `test_tools_not_discarded_when_response_model_set` verifies sync path
- [x] New test `test_tools_not_discarded_when_response_model_set_async` verifies async path
- [x] All existing `test_llm.py` tests pass (4 pre-existing failures due to missing `google-genai` optional dep are unrelated)

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1193,7 +1193,11 @@ class LLM(BaseLLM):
             str: The response text
         """
         # --- 1) Handle response_model with InternalInstructor for LiteLLM
-        if response_model and self.is_litellm:
+        # Only use the InternalInstructor shortcut when there are no tools.
+        # When tools are present, fall through to the standard litellm.completion
+        # path so that both tools and response_model are sent together — otherwise
+        # the agent's tools are silently discarded (see #4697).
+        if response_model and self.is_litellm and not params.get("tools"):
             from crewai.utilities.internal_instructor import InternalInstructor
 
             messages = params.get("messages", [])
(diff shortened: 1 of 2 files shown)


Krishnachaitanyakc:
Closing this PR as there are existing PRs addressing this issue (#4699, #4819, #4856). Apologies for the duplicate.

PR #5096, opened 2026-03-26
Revise security policy and reporting instructions

theCyberTech:
Updated the security reporting process and contact details.


iris-clawd (approved):
Looks good — clean swap to the Bugcrowd VDP. All the detailed reporting guidance (scope, safe harbor, etc.) lives on the Bugcrowd side now so removing it from the repo makes sense. 👍

PR #5176, opened 2026-03-31
refactor: convert CrewAgentExecutor to BaseModel

greysonlalonde:
## Summary
- Convert `CrewAgentExecutor` from plain class to Pydantic `BaseModel`
- Replace `__init__` with declared fields + `model_validator(mode="after")` for init logic
- Remove now-redundant `__get_pydantic_core_schema__` (BaseModel provides its own)
- Use `PrivateAttr` for `_i18n` and `_printer`
- `stop_words` param mapped to `stop` field via `Field(alias="stop_words")`

## Test plan
- [x] ruff, ruff-format, mypy all pass
- [ ] Existing test suite passes

---

> [!NOTE]
> **Medium Risk**
> Changes how the core agent executor is constructed/validated and mutates LLM stop-word configuration during Pydantic validation, which could affect runtime behavior and instantiation errors.
> 
> **Overview**
> Refactors `CrewAgentExecutor` from a hand-written class `__init__` into a Pydantic `BaseModel` with declared fields, default factories, and an `after` `model_validator` that performs the previous setup (i18n selection, hook registration, and LLM stop-word merging).
> 
> Adds `Field(alias="stop_words")` to keep the old constructor parameter name while storing it as `stop`, and moves `_i18n`/`_printer` to `PrivateAttr`. Removes the custom Pydantic core-schema shim and triggers `CrewAgentExecutor.model_rebuild()` during `crewai` package import to finalize the model schema.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1dd059a3959c0931ec266f975c22751d96d1af6c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/__init__.py
+++ b/lib/crewai/src/crewai/__init__.py
@@ -6,6 +6,7 @@ import warnings
 
 from crewai.agent.core import Agent
 from crewai.agent.planning_config import PlanningConfig
+from crewai.agents.crew_agent_executor import CrewAgentExecutor
 from crewai.crew import Crew
 from crewai.crews.crew_output import CrewOutput
 from crewai.flow.flow import Flow
@@ -17,6 +18,9 @@ from crewai.task import Task
 from crewai.tasks.llm_guardrail import LLMGuardrail
 from crewai.tasks.task_output import TaskOutput
 from crewai.telemetry.telemetry import Telemetry
+
+
+CrewAgentExecutor.model_rebuild()
 
 
 def _suppress_pydantic_deprecation_warnings() -> None:
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -30,6 +30,7 @@ from crewai.agents.parser import (
     AgentFinish,
     OutputParserError,
 )
+from crewai.agents.tools_handler import ToolsHandler
 from crewai.core.providers.human_input import ExecutorContext, get_provider
 from crewai.events.event_bus import crewai_event_bus
 from crewai.events.types.logging_events import (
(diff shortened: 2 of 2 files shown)


iris-clawd (commented):
## Review: `refactor: convert CrewAgentExecutor to BaseModel`

Good continuation of the BaseModel conversion series. A few things:

### ⚠️ Issues

**1. `task` and `crew` changed from required to optional (`None`)**
Old: `task: Task`, `crew: Crew` (required args in `__init__`)
New: `task: Task | None = None`, `crew: Crew | None = None`

Was this intentional? The executor is meaningless without a task and crew. If this was done to satisfy some construction path, it's hiding a real invariant. Every usage site will now need null checks (or will fail at runtime). If there's a reason for this, add a comment. Otherwise, keep them required.

**2. `stop` field alias may break callers using keyword arg**
```python
stop: list[str] = Field(alias="stop_words")
```
With `populate_by_name=True`, both `stop` and `stop_words` work for construction. But existing code that accesses `executor.stop` as an attribute will still work. However, serialization will use the alias `stop_words` by default. If anything serializes this model, the key name changes. Worth verifying no serialization paths exist.

**3. `i18n` is both a public field AND a private attr**
```python
i18n: I18N | None = Field(default=None, exclude=True)
_i18n: I18N = PrivateAttr()
```
The public `i18n` field accepts the constructor param, then `_init_executor` copies it to `_i18n`. This works but is awkward — now `executor.i18n` and `executor._i18n` both exist. After init, `executor.i18n` stays `None` or the passed value, while `_i18n` is the resolved one (with fallback). Could be confusing. Consider using only the private attr with a `model_validator(mode="before")` to extract it from data.

**4. Hook registration in `_init_executor` mutates `before_llm_call_hooks`/`after_llm_call_hooks`**
```python
self.before_llm_call_hooks.extend(get_before_llm_call_hooks())
```
Since these are `Field(default_factory=list)`, each instance gets its own list — this is fine. Just confirming no shared-state issues.

### ✅ What looks good
- Removing `__get_pydantic_core_schema__` — correct, BaseModel provides its own
- `PrivateAttr` for `_printer` and `_i18n` — proper encapsulation
- `populate_by_name=True` with alias — maintains backward compat for `stop_words` kwarg
- Net -40 lines — cleaner

### Verdict
Address the `task`/`crew` optionality question and this is good. The `i18n` dual-field pattern is a nit but not blocking. 💬 99

PR #5165, opened 2026-03-30
fix(llm): use validated messages variable instead of raw params access

Ricardo-M-L:
Fixes #5164

Both `_handle_non_streaming_response` and `_ahandle_non_streaming_response` extract and validate messages with `params.get("messages", [])`, but then pass `params["messages"]` to `_handle_emit_call_events`. This would raise a confusing `KeyError` instead of the intended `ValueError` if the key is missing.

Changed `params["messages"]` to `messages` on both lines (sync and async).

---

> [!NOTE]
> **Low Risk**
> Low risk: a small bug fix that only changes which `messages` value is passed into LLM call-completed events, avoiding `KeyError` when `params` lacks `messages`.
> 
> **Overview**
> Fixes non-streaming LLM event emission for `response_model` calls (sync and async) by passing the already-extracted/validated `messages` variable into `_handle_emit_call_events` instead of directly indexing `params["messages"]`.
> 
> This prevents a confusing `KeyError` when `messages` are missing and preserves the intended validation error behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca46bf569ddaead7ee789fdd3a6be0d5a38a2d34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5177, opened 2026-03-31
fix: use validated messages variable instead of params["messages"] in LLM handlers

### LINKED ISSUE #5164: [BUG] KeyError when params missing 'messages' key in _handle_non_streaming_response
### Description

In `crewai/llm.py`, both `_handle_non_streaming_response()` (line 1220) and `_ahandle_non_streaming_response()` (line 1364) access `params["messages"]` directly when calling `_handle_emit_call_events`, despite having already extracted and validated the messages via `params.get("messages", [])` a few lines earlier.

If the `messages` key is missing from `params`, the code correctly raises a helpful `ValueError("Messages are required when using response_model")` at line 1200/1345, but then at line 1220/1364 it would raise a confusing `KeyError` instead.

### Steps to Reproduce

Call `_handle_non_streaming_response` with `params` dict that doesn't contain `"messages"` key and a `response_model` set.

### Expected Behavior

Should use the already-validated `messages` variable instead of `params["messages"]`.

GopalGB:
## Summary

- Replace `params["messages"]` with the already-validated `messages` local variable in `_handle_non_streaming_response()` and `_ahandle_non_streaming_response()`

## Problem

In both handler methods, messages are safely extracted early via:
```python
messages = params.get("messages", [])
if not messages:
    raise ValueError("Messages are required when using response_model")
```

But the subsequent `_handle_emit_call_events()` call in the InternalInstructor code path uses `params["messages"]` directly (lines 1174 and 1318). If the `messages` key is missing from `params`, this raises a confusing `KeyError` instead of the helpful `ValueError` that was intended.

## Fix

Replace `params["messages"]` with the `messages` local variable (already validated) in the two affected `_handle_emit_call_events()` calls within the `response_model and self.is_litellm` branches.

Fixes #5164

## Test plan

- [ ] Verify calling `_handle_non_streaming_response` with params dict missing `messages` key and `response_model` set raises `ValueError`, not `KeyError`
- [ ] Verify normal operation (params with `messages` key) is unaffected
- [ ] Existing test suite passes

---

> [!NOTE]
> **Low Risk**
> Low risk: small, localized change that only affects event emission in the LiteLLM `response_model` path and prevents an unintended `KeyError` when `messages` is missing.
> 
> **Overview**
> Fixes the LiteLLM `response_model` path in `_handle_non_streaming_response()` and `_ahandle_non_streaming_response()` to pass the validated local `messages` list into `_handle_emit_call_events()`.
> 
> This avoids raising a `KeyError` from `params["messages"]` and preserves the intended `ValueError` when `messages` is absent.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 007a9f9bc918e5a14454bce06154dbc95ebd789b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1171,7 +1171,7 @@ class LLM(BaseLLM):
                 call_type=LLMCallType.LLM_CALL,
                 from_task=from_task,
                 from_agent=from_agent,
-                messages=params["messages"],
+                messages=messages,
             )
             return structured_response
 
@@ -1315,7 +1315,7 @@ class LLM(BaseLLM):
                 call_type=LLMCallType.LLM_CALL,
                 from_task=from_task,
                 from_agent=from_agent,
-                messages=params["messages"],
+                messages=messages,
             )
             return structured_response
 


greysonlalonde:
Duplicate of #5165. Closing.

PR #5181, opened 2026-03-31
perf: optimize python dictionary iterations (SIM118)

sjhddh:
### Description
This PR optimizes dictionary iterations across the codebase by removing redundant `.keys()` calls in `for` loops.

In Python, iterating directly over a dictionary (e.g., `for key in my_dict:`) is more idiomatic and marginally faster than explicitly calling `.keys()` (e.g., `for key in my_dict.keys():`). This conforms to the standard [SIM118](https://docs.astral.sh/ruff/rules/in-dict-keys/) rule.

**Changes made:**
- `for field in list(data.keys()):` ➡️ `for field in list(data):`
- `for scheme_name in first_security_req.keys():` ➡️ `for scheme_name in first_security_req:`
- `for field_name in Settings.model_fields.keys():` ➡️ `for field_name in Settings.model_fields:`
- `for iter_num in sorted(iterations_results.keys()):` ➡️ `for iter_num in sorted(iterations_results):`
- Test iterations over `user_settings` / `cli_settings` keys have also been simplified.

*(Refactored via [0-editor](https://github.com/0-protocol/0-editor))*


greysonlalonde:
Thanks for the contribution. Closing because the changes don't enable SIM118 in our ruff config (`pyproject.toml` extend-select), so the cleanup wouldn't prevent new violations. Going to land this in a single commit that enables the rule and runs `ruff check --fix` across the codebase, so the change is enforced going forward.

PR #5216, opened 2026-04-01
fix: pass fingerprint metadata via config instead of tool args

greysonlalonde:
## Summary
- Fixes OSS-9: `extra_forbidden` Pydantic validation errors on MCP and integration tools
- `_add_fingerprint_metadata()` was injecting `security_context` into tool arguments, which breaks tools with `extra="forbid"` schemas (the default for dynamic models)
- Moves fingerprint data to the `config` parameter that `invoke()`/`ainvoke()` already accept, keeping it available to consumers without polluting the tool args namespace
- The event bus fingerprint flow is unaffected

## Root Cause
`_add_fingerprint_metadata()` added `security_context` to the args dict right after filtering to `acceptable_args`, defeating the filter. `_parse_args()` then ran `model_validate()` against the schema with `extra="forbid"`, causing the validation error.

## Test plan
- [x] All 24 `test_tool_usage.py` tests pass
- [x] All 15 `test_fingerprint.py` tests pass
- [x] Verify MCP tools with strict schemas no longer raise `extra_forbidden`
- [x] Verify fingerprint data is accessible via `config` param in tool implementations

---

> [!NOTE]
> **Medium Risk**
> Changes how fingerprint `security_context` is propagated to all tool invocations (`invoke`/`ainvoke`), which could affect tools expecting it in arguments or relying on invocation config behavior.
> 
> **Overview**
> Fixes strict-schema tool validation failures by **stopping injection of fingerprint metadata into tool input args**.
> 
> `ToolUsage` now builds a `fingerprint_config` via `_build_fingerprint_config()` and passes it through the `config` parameter on `tool.invoke()`/`tool.ainvoke()`, keeping `security_context` available without adding extra fields to Pydantic-validated argument payloads.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bad5b71ca2b1cbb77308bef34a5e285e01ce212a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/ar/installation.mdx
+++ b/docs/ar/installation.mdx
@@ -5,6 +5,14 @@ icon: wrench
 mode: "wide"
 ---
 
+### شاهد: بناء Agents و Flows في CrewAI باستخدام Coding Agent Skills
+
+قم بتثبيت مهارات وكيل البرمجة الخاصة بنا (Claude Code، Codex، ...) لتشغيل وكلاء البرمجة بسرعة مع CrewAI.
+
+يمكنك تثبيتها باستخدام `npx skills add crewaiinc/skills`
+
+<iframe src="https://www.loom.com/embed/befb9f68b81f42ad8112bfdd95a780af" frameborder="0" webkitallowfullscreen mozallowfullscreen allowfullscreen style={{width: "100%", height: "400px"}}></iframe>
+
 ## فيديو تعليمي
 
 شاهد هذا الفيديو التعليمي لعرض تفصيلي لعملية التثبيت:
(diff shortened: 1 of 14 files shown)


alex-clawd (approved):
LGTM. Clean fix for a real production issue.

**The problem**: `_add_fingerprint_metadata()` injected `security_context` into tool arguments AFTER the `acceptable_args` filter — which then broke Pydantic validation on tools with `extra='forbid'` (MCP tools, integration tools, any strict schema).

**The fix**: Moves fingerprint data to the `config` parameter on `invoke()`/`ainvoke()` instead of polluting the args dict. This is the correct pattern — config is the sideband channel for metadata.

**What I verified**:
- Both sync (`_use`) and async (`_ause`) paths updated consistently
- `_build_fingerprint_config()` returns empty dict when no security context → no unnecessary config passed
- Tool arguments are now clean — only schema-valid keys reach `model_validate()`
- All CI green (lint, type-checker, tests 3.10–3.13)

One thing downstream consumers should note: any tool implementation that was reading `security_context` from `kwargs` or input args will need to read it from `config` instead. But since this was already broken for strict-schema tools, that's the right tradeoff.

PR #5218, opened 2026-04-01
fix: forward base_url to instructor client for custom OpenAI-compatible endpoints

### LINKED ISSUE #5204: [BUG] InternalInstructor discards base_url when creating instructor client — breaks OpenAI-compatible endpoints
### Description

When using an OpenAI-compatible provider with a custom `base_url` (e.g., self-hosted vLLM, Ollama, or any non-OpenAI endpoint), CrewAI's Converter/InternalInstructor silently discards the `base_url` and hits `api.openai.com` instead.

The root cause is in `internal_instructor.py` — `_create_instructor_client()` extracts only the model name and provider string from `self.llm.model`, then calls:

```python
return instructor.from_provider(f"{provider}/{model_string}")
```

The `base_url` from the LLM object is never forwarded. `instructor.from_provider()` creates a fresh OpenAI client defaulting to `api.openai.com/v1/`.

**Note:** There is a `_get_llm_extra_kwargs()` method that forwards `base_url`, but it is guarded behind `is_litellm=True`, so non-LiteLLM OpenAI-compatible providers are still affected.

### Steps to Reproduce

1. Configure a CrewAI agent with an OpenAI-compatible LLM that has a custom `base_url` (e.g., vLLM, Ollama, or any self-hosted endpoint)
2. Set `output_pydantic` or `output_json` on a Task so that CrewAI invokes the Converter
3. The Converter creates an `InternalInstructor` via `converter.py:145-152`, passing the full LLM object (which has `base_url` set)
4. `InternalInstructor._create_instructor_client()` (`internal_instructor.py:76-101`) extracts only `self.llm.model` and `self.llm.provider`, then calls `instructor.from_provider(f"{provider}/{model_string}")` — `base_url` is lost
5. The instructor client sends requests to `api.openai.com` instead of the configured endpoint
6. Result: `ConnectTimeout` / `ConverterError` in environments that cannot reach `api.openai.com`

### Expected behavior

When a LLM is configured with a custom `base_url`, the InternalInstructor should forward that `base_url` to the instructor client so that structured output parsing requests go to the correct endpoint — not to `api.openai.com`.

### Screenshots/Code snippets

**The buggy code path** (`internal_instructor.py:76-101`):

```python
def _create_instructor_client(self) -> Any:
    import instructor

    if isinstance(self.llm, str):
        model_string = self.llm
    elif self.llm is not None and hasattr(self.llm, "model"):
        model_string = self.llm.model  # ← extracts model name
    else:
        raise ValueError(...)

    if isinstance(self.llm, str):
        provider = self._extract_provider()
    elif self.llm is not None and hasattr(self.llm, "provider"):
        provider = self.llm.provider  # ← extracts provider
    else:
        provider = "openai"

    # ← base_url is NEVER forwarded here
    return instructor.from_provider(f"{provider}/{model_string}")
```

**The incomplete fix** — `_get_llm_extra_kwargs()` exists but is guarded:

```python
def _get_llm_extra_kwargs(self) -> dict[str, Any]:
    # This guard means non-litellm providers never get base_url forwarded
    if not getattr(self.llm, "is_litellm", False):
        return {}  # ← base_url lost for OpenAI-compatible providers

    extra = {}
    for attr in ("api_base", "base_url", "api_key"):
        value = getattr(self.llm, attr, None)
        if value is not None:
            extra[attr] = value
    return extra
```

### Operating System

macOS Sonoma

### Python Version

3.12

### crewAI Version

latest (main branch)

### crewAI Tools Version

latest

### Virtual Environment

Venv

### Evidence

Traced through the source code:

1. **`converter.py:145-152`** — `_create_instructor()` correctly passes the full LLM object (with `base_url`):
```python
def _create_instructor(self):
    return InternalInstructor(
        llm=self.llm,  # ← has base_url set
        model=self.model,
        content=self.text,
    )
```

2. **`internal_instructor.py:76-101`** — `_create_instructor_client()` discards `base_url`:
```python
model_string = self.llm.model       # only extracts model name
provider = self.llm.provider         # only extracts provider
return instructor.from_provider(f"{provider}/{model_string}")  # base_url lost
```

3. **`base_llm.py:123`** — confirms `base_url` is a defined field on `BaseLLM`:
```python
base_url: str | None = None
```

4. **`_get_llm_extra_kwargs()`** exists but is guarded by `is_litellm`:
```python
if not getattr(self.llm, "is_litellm", False):
    return {}  # non-litellm providers never get base_url forwarded
```

5. **Result in production:** All structured output requests (`output_pydantic`, `output_json`) from OpenAI-compatible providers hit `api.openai.com` → `ConnectTimeout` → `ConverterError`.

### Possible Solution

Two options:

**Option A (minimal fix):** In `_create_instructor_client()`, pass `base_url` to `instructor.from_provider()` if the instructor library supports it, or construct an explicit `OpenAI(base_url=...)` client and use `instructor.from_openai(client)` instead.

**Option B (broader fix):** Remove the `is_litellm` guard from `_get_llm_extra_kwargs()` so that `base_url` and `api_key` are forwarded for all provider types, not just LiteLLM-backed ones. The guard was added because "non-litellm instructor clients (from_provider) don't accept them" — but the fix should be to use a client constructor that does accept them (e.g., `instructor.from_openai(OpenAI(base_url=..., api_key=...))`) rather than silently dropping the config.

### Additional context

This affects anyone using OpenAI-compatible endpoints (vLLM, Ollama remote, Azure OpenAI with custom endpoints, etc.) with structured output tasks (`output_pydantic` or `output_json`). The agent's main LLM calls work fine because they go through the LLM class directly, but the Converter/InternalInstructor path bypasses the LLM and creates its own client — losing the `base_url` in the process.

There are partial fixes on feature branches (commits 9dabb3e81, 9bdc7b9ee for LiteLLM, and f9ae6c52d for A2A) but none address the non-LiteLLM OpenAI-compatible path on main.

r266-tech:
Closes #5204

## Problem

When a task uses `output_pydantic` or `output_json` and the LLM has a custom `base_url` (vLLM, Ollama, Azure OpenAI, any self-hosted OpenAI-compatible endpoint), the structured output request hits `api.openai.com` instead of the configured endpoint — resulting in `ConnectTimeout` / `ConverterError`.

**Root cause**: `_create_instructor_client()` calls `instructor.from_provider(f"{provider}/{model_string}")`, which creates a fresh `openai.OpenAI` client defaulting to `api.openai.com`. The `base_url` and `api_key` from the LLM object are never forwarded.

## Fix

When `base_url` (or `api_base`) is set on the LLM, build an explicit `openai.OpenAI(base_url=..., api_key=...)` client and wrap it with `instructor.from_openai()`:

```python
base_url: str | None = getattr(self.llm, "base_url", None) or getattr(
    self.llm, "api_base", None
)
if base_url:
    from openai import OpenAI
    api_key: str = getattr(self.llm, "api_key", None) or "not-needed"
    openai_client = OpenAI(base_url=base_url, api_key=api_key)
    return instructor.from_openai(openai_client)

return instructor.from_provider(f"{provider}/{model_string}")
```

- When `base_url` is set → explicit client with correct endpoint ✅
- When `base_url` is not set → existing `from_provider()` path unchanged ✅

## Scope

**1 file changed, +20 lines** in `lib/crewai/src/crewai/utilities/internal_instructor.py`.

Affects anyone using `output_pydantic` / `output_json` with custom OpenAI-compatible endpoints. The agent's main LLM calls are unaffected (they go through the `LLM` class directly).

---

> [!NOTE]
> **Medium Risk**
> Changes how `InternalInstructor` builds the underlying OpenAI client for structured outputs, which can affect request routing/auth and may break if provider detection or `base_url` handling is wrong. Scope is small and largely preserves existing `from_provider()` behavior when no custom endpoint is configured.
> 
> **Overview**
> Fixes structured-output calls (`output_pydantic`/`output_json`) to respect custom OpenAI-compatible endpoints by detecting `base_url`/`api_base` on the LLM and constructing an explicit `openai.OpenAI(base_url=..., api_key=...)` wrapped via `instructor.from_openai()`.
> 
> Keeps the existing `instructor.from_provider("{provider}/{model}")` path for non-OpenAI providers and for cases where no custom endpoint is set, limiting behavioral changes to the custom-endpoint scenario.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c38b16cbaba043bb7fb550e3fadc5050a85bb64b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/utilities/internal_instructor.py
+++ b/lib/crewai/src/crewai/utilities/internal_instructor.py
@@ -104,19 +104,21 @@ class InternalInstructor(Generic[T]):
         else:
             provider = "openai"  # Default fallback
 
-        # Forward base_url and api_key when the LLM targets a custom endpoint
-        # (e.g. vLLM, Ollama, Azure OpenAI, or any OpenAI-compatible server).
-        # instructor.from_provider() creates a fresh OpenAI client that
-        # ignores these settings, so we must build the client explicitly.
-        base_url: str | None = getattr(self.llm, "base_url", None) or getattr(
-            self.llm, "api_base", None
-        )
-        if base_url:
-            from openai import OpenAI
+        # Forward base_url and api_key when the LLM targets a custom OpenAI-compatible
+        # endpoint (e.g. vLLM, Ollama, Azure OpenAI).  For non-OpenAI providers
+        # (anthropic, google, etc.) we still use from_provider() which handles
+        # provider-specific client construction.
+        _openai_compatible = {"openai", "azure", "deepseek", "groq", "together", "ollama"}
+        if provider.lower() in _openai_compatible:
+            base_url: str | None = getattr(self.llm, "base_url", None) or getattr(
+                self.llm, "api_base", None
+            )
+            if base_url:
+                from openai import OpenAI
 
-            api_key: str = getattr(self.llm, "api_key", None) or "not-needed"
-            openai_client = OpenAI(base_url=base_url, api_key=api_key)
-            return instructor.from_openai(openai_client)
+                api_key: str = getattr(self.llm, "api_key", None) or "not-needed"
+                openai_client = OpenAI(base_url=base_url, api_key=api_key)
+                return instructor.from_openai(openai_client)
 
         return instructor.from_provider(f"{provider}/{model_string}")
 
--- a/lib/crewai/src/crewai/utilities/internal_instructor.py
+++ b/lib/crewai/src/crewai/utilities/internal_instructor.py
@@ -76,6 +76,12 @@ class InternalInstructor(Generic[T]):
     def _create_instructor_client(self) -> Any:
         """Create instructor client using the modern from_provider pattern.
 
+        When the LLM is configured with a custom ``base_url`` (e.g. a
+        self-hosted vLLM, Ollama, or Azure OpenAI endpoint), this method
+        constructs an explicit ``openai.OpenAI`` client with the correct
+        endpoint and wraps it with ``instructor.from_openai()``.  Otherwise
+        it falls back to ``instructor.from_provider()``.
+
         Returns:
             Instructor client configured for the LLM provider
 
@@ -97,6 +103,20 @@ class InternalInstructor(Generic[T]):
             provider = self.llm.provider
         else:
             provider = "openai"  # Default fallback
+
+        # Forward base_url and api_key when the LLM targets a custom endpoint
+        # (e.g. vLLM, Ollama, Azure OpenAI, or any OpenAI-compatible server).
+        # instructor.from_provider() creates a fresh OpenAI client that
+        # ignores these settings, so we must build the client explicitly.
+        base_url: str | None = getattr(self.llm, "base_url", None) or getattr(
+            self.llm, "api_base", None
+        )
+        if base_url:
+            from openai import OpenAI
+
+            api_key: str = getattr(self.llm, "api_key", None) or "not-needed"
+            openai_client = OpenAI(base_url=base_url, api_key=api_key)
+            return instructor.from_openai(openai_client)
 
         return instructor.from_provider(f"{provider}/{model_string}")
 


r266-tech (commented):
Good catch! Added a provider check so the OpenAI client wrapper is only used for OpenAI-compatible providers (openai, azure, deepseek, groq, together, ollama). Non-OpenAI providers like Anthropic continue to use `instructor.from_provider()` even when a custom `base_url` is set.

shivam2407:
Great fix! Happy to see this moving forward.

One thing worth noting: this PR currently only handles the `openai` provider path — when `base_url` is set, it constructs an `OpenAI(base_url=...)` client. But the same `base_url` forwarding issue exists for **Anthropic** and **Azure OpenAI** providers too.

In my PR #5264 (closed as a duplicate of this one), I had per-provider handling:

| Provider | Client constructed |
|---|---|
| `openai` | `OpenAI(base_url=base_url, api_key=api_key)` → `instructor.from_openai()` |
| `anthropic` | `Anthropic(base_url=base_url, api_key=api_key)` → `instructor.from_anthropic()` |
| `azure` / `azure_openai` | `AzureOpenAI(azure_endpoint=base_url, api_key=api_key, api_version=...)` → `instructor.from_openai()` |
| Others | Falls back to `instructor.from_provider()` |

Happy to either:
- **Submit a follow-up PR** building on this one once it merges, adding Anthropic and Azure support
- **Contribute the per-provider logic directly to this PR** if you'd like to extend it before merge

Let me know what works best!

r266-tech:
Thanks for the review and the per-provider breakdown!

I'd prefer to **keep this PR minimal** (openai/OpenAI-compatible path only) so we can get it landed quickly, then extend with a follow-up PR. The per-provider handling for Anthropic + Azure introduces additional SDK surface (`from_anthropic`, AzureOpenAI `api_version` handling) that is worth a dedicated review.

If you'd like to own the follow-up PR once this merges, that'd be great — I'll be happy to review it. Otherwise I'll pick it up next week.

r266-tech:
Closing — superseded by the current implementation on main: `InternalInstructor._create_instructor_client` now forwards `base_url` and `api_key` to `instructor.from_provider()` directly, which handles custom OpenAI-compatible endpoints in a provider-agnostic way and makes this wrapper approach unnecessary. @shivam2407 thanks again for the per-provider notes — the upstream solution covers those paths too.

PR #5259, opened 2026-04-03
chore: add exclude-newer = 7 days to all pyproject.toml files

greysonlalonde:
## Summary
- Adds `exclude-newer = "7 days"` to all `pyproject.toml` files (root, lib packages, and scaffold templates)
- Prevents uv from resolving any package released less than 7 days ago, mitigating supply chain attacks on transient dependencies
- Intentional bumps (e.g. CVE fixes from Dependabot) are unaffected

## Test plan
- [ ] Verify `uv lock` resolves correctly with the new setting
- [ ] Verify `crewai create crew` scaffolds a project with `exclude-newer` set

---

> [!NOTE]
> **Medium Risk**
> Changes dependency resolution behavior across the workspace by adding `uv`'s `exclude-newer` guard, which could affect lockfile generation and CI/build reproducibility if recent releases are required. No runtime logic changes, but installation may need explicit pins when transitive deps lag.
> 
> **Overview**
> Adds a workspace-wide `uv` supply-chain safeguard by setting `tool.uv.exclude-newer = "3 days"` in the root and each internal package `pyproject.toml`, preventing resolution of newly-published transitive dependencies by default.
> 
> Updates the installation docs to explain the new policy and how users can override it by explicitly pinning needed dependency versions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 047bdb7325dae25fbd2cdcae72d18a929fc7c4c1. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/docs/en/installation.mdx
+++ b/docs/en/installation.mdx
@@ -171,6 +171,9 @@ We recommend using the `YAML` template scaffolding for a structured approach to 
       ```shell
       uv add <package-name>
       ```
+      <Note>
+        As a supply-chain security measure, CrewAI's internal packages use `exclude-newer = "3 days"` in their `pyproject.toml` files. This means transitive dependencies pulled in by CrewAI won't resolve packages released less than 3 days ago. Your own direct dependencies are not affected by this policy. If you notice a transitive dependency is behind, you can pin the version you want explicitly in your project's dependencies.
+      </Note>
     - To run your crew, execute the following command in the root of your project:
       ```bash
       crewai run
--- a/lib/crewai-files/pyproject.toml
+++ b/lib/crewai-files/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
 ]
 
 [tool.uv]
-exclude-newer = "7 days"
+exclude-newer = "3 days"
 
 [build-system]
 requires = ["hatchling"]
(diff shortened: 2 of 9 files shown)


iris-clawd (changes requested):
Good intent — supply chain hardening is worth doing — but a few issues:

1. **`"7 days"` relative format requires uv ≥ 0.11.** Older versions (e.g. 0.7.x) fail with a TOML parse error since they only accept RFC 3339 timestamps. This is fine for the workspace itself (we control the CI/dev toolchain), but the **scaffold templates** (`crew/`, `flow/`, `tool/`) ship to end users who may be on any uv version. This will break `crewai create crew` for anyone on uv < 0.11.

2. **Workspace members inherit `[tool.uv]` from the workspace root.** Per uv docs: *'In workspaces, uv will begin its search at the workspace root, ignoring any configuration defined in workspace members.'* So adding `[tool.uv] exclude-newer` to each `lib/*/pyproject.toml` is redundant — only the root `pyproject.toml` entry matters for workspace resolution. Cleaner to just set it at root.

3. **Consider ISO 8601 duration format (`P7D`) instead of friendly string.** It's more portable across tools and versions that support duration-based exclude-newer.

Suggested approach:
- Root `pyproject.toml`: keep the setting (already there ✅), maybe switch to `P7D`
- `lib/*/pyproject.toml`: remove (redundant in workspace)
- Scaffold templates: either use `P7D` with a comment noting minimum uv version, or pin to an RFC 3339 date and document the intent

💬 140

PR #5264, opened 2026-04-03
fix: forward base_url to instructor client for OpenAI-compatible endpoints

### LINKED ISSUE #5204: [BUG] InternalInstructor discards base_url when creating instructor client — breaks OpenAI-compatible endpoints

shivam2407:
## Summary

Fixes #5204

When using an OpenAI-compatible provider with a custom `base_url` (e.g. self-hosted vLLM, Ollama, or any non-OpenAI endpoint), `InternalInstructor` silently discards the `base_url` and sends structured output requests to `api.openai.com` instead of the configured endpoint.

### Root cause

`_create_instructor_client()` extracts only `model` and `provider` from the LLM, then calls `instructor.from_provider(f"{provider}/{model_string}")`. The `base_url` is never forwarded, and `instructor.from_provider()` creates a default client pointing at `api.openai.com/v1/`.

### Fix

Added `_get_llm_client_kwargs()` to extract `base_url`, `api_base`, and `api_key` from the LLM object. When `base_url` is present, constructs an explicit SDK client instead of using `from_provider()`:

| Provider | Client used | Key forwarded as |
|---|---|---|
| `openai` | `OpenAI(base_url=...)` | `base_url` |
| `azure` / `azure_openai` | `AzureOpenAI(azure_endpoint=..., api_version=...)` | `azure_endpoint` |
| `anthropic` | `Anthropic(base_url=...)` | `base_url` |
| Others | `from_provider()` fallback | `api_key` only |

Additional behaviors:
- `api_base` is normalized to `base_url`; when both are set, `base_url` takes precedence
- All guards use `is not None` (not truthiness) so empty strings are handled correctly
- When no custom `base_url` is set, behavior is unchanged (`from_provider()` path)

### Changes

| File | Change |
|---|---|
| `internal_instructor.py` | Added `_get_llm_client_kwargs()`, `_create_instructor_client_with_base_url()` with per-provider client construction. Updated `_create_instructor_client()` to route through the new path when `base_url` is present. |
| `test_converter.py` | 8 new tests + fixed 1 existing mock |

## Test plan

- [x] 8 new tests pass (OpenAI/Anthropic/Azure forwarding, api_base normalization, collision precedence, string LLM, no-base_url default)
- [x] All 50 existing tests still pass
- [ ] Manual testing with vLLM/Ollama endpoint + `output_pydantic` task

### DIFF
--- a/lib/crewai/src/crewai/utilities/internal_instructor.py
+++ b/lib/crewai/src/crewai/utilities/internal_instructor.py
@@ -73,8 +73,32 @@ class InternalInstructor(Generic[T]):
             else:
                 self._client = self._create_instructor_client()
 
+    def _get_llm_client_kwargs(self) -> dict[str, Any]:
+        """Extract base_url and api_key from the LLM so they can be forwarded
+        to the instructor client.
+
+        Returns:
+            Dict with ``base_url`` and/or ``api_key`` when present on the LLM.
+        """
+        kwargs: dict[str, Any] = {}
+        if isinstance(self.llm, str):
+            return kwargs
+        for attr in ("base_url", "api_base", "api_key"):
+            value = getattr(self.llm, attr, None)
+            if value is not None:
+                # Normalize api_base → base_url for the OpenAI client.
+                # First writer wins: base_url takes precedence over api_base.
+                key = "base_url" if attr == "api_base" else attr
+                if key not in kwargs:
+                    kwargs[key] = value
+        return kwargs
+
     def _create_instructor_client(self) -> Any:
-        """Create instructor client using the modern from_provider pattern.
+        """Create instructor client configured for the LLM provider.
+
+        When the LLM carries a custom ``base_url`` (e.g. self-hosted vLLM,
+        Ollama, or any OpenAI-compatible endpoint), we construct an explicit
+        provider client so the URL is not silently discarded.
 
         Returns:
             Instructor client configured for the LLM provider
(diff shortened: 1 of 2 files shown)


greysonlalonde:
Duplicate of #5218. Closing.

PR #5289, opened 2026-04-06
Fix shared LLM stop words mutation causing cross-agent state pollution

### LINKED ISSUE #5141: [Bug] Shared LLM stop words mutation causes cross-agent state pollution
## Bug Description

When multiple agents in a Crew share the same `LLM` instance, each `CrewAgentExecutor.__init__` mutates the shared LLM's `stop` attribute, causing stop words to accumulate across agents and pollute each other's generation behavior.

## Root Cause

In `crew_agent_executor.py` (lines 164-173):

```python
if self.llm:
    # This may be mutating the shared llm object and needs further evaluation
    existing_stop = getattr(self.llm, "stop", [])
    self.llm.stop = list(
        set(
            existing_stop + self.stop
            if isinstance(existing_stop, list)
            else self.stop
        )
    )
```

The code itself has a comment acknowledging this: *"This may be mutating the shared llm object and needs further evaluation"*.

### Why this is a problem:

1. **Sequential execution**: Agent A adds its stop words to the LLM. Agent B then inherits A's stop words AND adds its own. Agent C gets A+B+C. Stop words grow monotonically — they are never cleaned up after an agent finishes execution.

2. **Concurrent execution** (async tasks): Multiple agents mutate `self.llm.stop` simultaneously on the same object — this is a race condition. The `set()` deduplication doesn't protect against interleaved reads and writes.

3. **Behavioral impact**: Extra stop words cause the LLM to terminate generation prematurely. If Agent A's stop word is `"Observation:"` and Agent B should not stop on that token, Agent B will nonetheless stop generating when it encounters `"Observation:"` because A's stop word leaked into the shared LLM config.

## Reproduction

```python
from crewai import Agent, Crew, Task, LLM

# One shared LLM instance
llm = LLM(model="gpt-4o")

agent_a = Agent(role="Researcher", llm=llm, ...)
agent_b = Agent(role="Writer", llm=llm, ...)

crew = Crew(agents=[agent_a, agent_b], tasks=[...])
crew.kickoff()

# After kickoff, llm.stop contains stop words from BOTH agents,
# and this persists into future crew.kickoff() calls
```

## Suggested Fix

Instead of mutating the shared LLM's stop words, the executor should use a **local copy** of the stop configuration:

```python
# Option A: Copy the stop words locally, don't mutate the shared LLM
if self.llm:
    existing_stop = getattr(self.llm, "stop", []) or []
    self._effective_stop = list(set(existing_stop + self.stop))
    # Use self._effective_stop when making LLM calls instead of self.llm.stop

# Option B: Reset stop words after execution in a finally block
```

Alternatively, consider making `LLM` stop words immutable at the instance level and passing per-request stop words through the API call parameters.

## Impact

- Affects all crews with 2+ agents sharing an LLM instance (the common/default case)
- Silent behavioral degradation — no error raised, just subtly wrong LLM outputs
- Accumulates across multiple `crew.kickoff()` calls on the same crew
- Race condition in async execution mode

Christian-Sidak:
## Summary

- Fixes the mutable shared state bug where `CrewAgentExecutor.__init__` permanently mutates `self.llm.stop`, causing stop words to accumulate across agents sharing the same LLM instance
- Computes effective stop words locally per-executor and only applies them to the LLM during `invoke`/`ainvoke`, restoring the original value in a `finally` block
- Prevents both sequential accumulation (Agent B inheriting Agent A's stop words) and reduces the race condition window in async execution

## Changes

Single file: `lib/crewai/src/crewai/agents/crew_agent_executor.py`

**`__init__`**: Instead of mutating `self.llm.stop` directly, we now store the original stop words (`self._original_stop`) and the merged per-executor stop words (`self._effective_stop`) without touching the shared LLM.

**`invoke` / `ainvoke`**: Set `self.llm.stop = self._effective_stop` before execution, then restore `self.llm.stop = self._original_stop` in a `finally` block.

## Test plan

- [ ] Verify that a shared LLM's `stop` attribute is not permanently modified after an executor runs
- [ ] Verify that multiple agents with different stop words do not pollute each other
- [ ] Verify that `llm.stop` is correctly restored even when execution raises an exception

Fixes #5141

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -142,12 +142,14 @@ class CrewAgentExecutor(BaseAgentExecutor):
         if not self.after_llm_call_hooks:
             self.after_llm_call_hooks.extend(get_after_llm_call_hooks())
         if self.llm and not isinstance(self.llm, str):
-            existing_stop = getattr(self.llm, "stop", [])
-            self.llm.stop = list(
+            # Compute effective stop words without mutating the shared LLM.
+            # We store the original value and only apply merged stop words
+            # around each invoke/ainvoke call to avoid cross-agent pollution.
+            existing_stop = getattr(self.llm, "stop", []) or []
+            self._original_stop = list(existing_stop) if isinstance(existing_stop, list) else []
+            self._effective_stop = list(
                 set(
-                    existing_stop + self.stop
-                    if isinstance(existing_stop, list)
-                    else self.stop
+                    self._original_stop + self.stop
                 )
             )
 
@@ -208,6 +210,10 @@ class CrewAgentExecutor(BaseAgentExecutor):
 
         self.ask_for_human_input = bool(inputs.get("ask_for_human_input", False))
 
+        # Apply executor-specific stop words to the LLM for this run only,
+        # then restore the original value so shared LLM instances stay clean.
+        if self.llm and hasattr(self, "_effective_stop"):
+            self.llm.stop = self._effective_stop
         try:
             formatted_answer = self._invoke_loop()
         except AssertionError:
@@ -220,6 +226,9 @@ class CrewAgentExecutor(BaseAgentExecutor):
         except Exception as e:
             handle_unknown_error(PRINTER, e, verbose=self.agent.verbose)
             raise
+        finally:
+            if self.llm and hasattr(self, "_original_stop"):
+                self.llm.stop = self._original_stop
 
         if self.ask_for_human_input:
             formatted_answer = self._handle_human_feedback(formatted_answer)
@@ -1078,6 +1087,10 @@ class CrewAgentExecutor(BaseAgentExecutor):
 
         self.ask_for_human_input = bool(inputs.get("ask_for_human_input", False))
 
+        # Apply executor-specific stop words to the LLM for this run only,
+        # then restore the original value so shared LLM instances stay clean.
+        if self.llm and hasattr(self, "_effective_stop"):
+            self.llm.stop = self._effective_stop
         try:
             formatted_answer = await self._ainvoke_loop()
         except AssertionError:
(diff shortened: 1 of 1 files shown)


0xbrainkid:
The finally-block restore pattern is the correct fix — set before execution, restore unconditionally on exit, zero shared state mutation. The test plan covers the right cases.

Worth noting the trust dimension beyond the bug itself: **shared LLM state mutation is a trust boundary violation between agents.** Agent A modifying `llm.stop` and Agent B inheriting those stop words means Agent B is not operating with its declared configuration — it is operating with Agent A's configuration overlaid on its own. In security-sensitive crews (e.g., a crew where one agent handles user-facing interactions and another handles privileged data access), this kind of implicit configuration bleed is exactly the class of bug that makes behavioral trust scores unreliable.

An agent that has been verified to behave within certain parameters (stop words, response patterns) is implicitly recertified by that verification. If the actual runtime configuration diverges from the verified configuration due to shared state mutation, the behavioral attestation is no longer valid.

Two additions worth considering alongside the fix:

**1. Configuration integrity check at task handoff.** Before an agent begins a task that was handed off from another agent, assert that `llm.stop == self._effective_stop` — catch the mutation proactively rather than discovering it post-execution via behavioral drift.

**2. Behavioral drift alert on stop word change.** A sudden change in stop words mid-session (even legitimately via the fixed code) represents a configuration state change worth flagging in observability tools — not as an error, but as an event that can explain behavioral differences across task boundaries.

Christian-Sidak:
Thanks for the thorough analysis, @0xbrainkid. You're right that shared LLM stop word mutation is a trust boundary issue, not just a correctness one — the point about behavioral attestation no longer being valid when runtime config diverges is well taken.

Both suggestions are good candidates for follow-up work:
1. A pre-task config assertion would catch this class of bug proactively rather than post-hoc via behavioral drift.
2. An observability event on stop word changes is useful for debugging across task boundaries.

I'll note these as potential follow-ups; they're out of scope for this minimal fix but worth tracking.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Superseded by #5689, which was merged on 2026-05-04 and closes the same underlying issue (#5141) with an equivalent fix (contextvars-based per-call stop word override instead of in-place mutation). Closing this one to avoid duplicate/conflicting work.

PR #5305, opened 2026-04-06
feat: Add Semantic Scholar Search Tool for CrewAI

trtc888:
- Add a tool for searching academic papers on Semantic Scholar
- Support keyword search (bulk) and paper lookup by ID
- Filters: year, citation count, fields of study, publication type, open access
- Optional API key with automatic fallback on invalid keys


greysonlalonde:
https://docs.crewai.com/en/guides/tools/publish-custom-tools

PR #5332, opened 2026-04-08
fix: make NL2SQLTool dialect-aware for SQLite and PostgreSQL

Krithi-murugesan:
## What
Fixed dialect compatibility issues in the NL2SQLTool to support SQLite and PostgreSQL.

## Why
The original tool used hardcoded PostgreSQL queries on system tables like information_schema.tables.
This caused runtime errors on SQLite, which doesn’t have those tables.

## How
Replaced hardcoded SQL with SQLAlchemy's dialect-agnostic inspect module to dynamically discover tables and columns.
This approach works across supported databases without errors.

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/tools/nl2sql/nl2sql_tool.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/nl2sql/nl2sql_tool.py
@@ -1,13 +1,11 @@
-from typing import Any
-
+from typing import Any, Type, Union, List, Dict
 from crewai.tools import BaseTool
-from pydantic import BaseModel, Field
-
+from pydantic import BaseModel, Field, PrivateAttr
+import json
 
 try:
-    from sqlalchemy import create_engine, text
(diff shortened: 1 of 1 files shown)


Krithi-murugesan:
Hi maintainers 👋

Adding some context for this PR:

What I tested

Validated NL2SQLTool with both SQLite and PostgreSQL
Verified table/column discovery using SQLAlchemy inspect
Confirmed removal of information_schema dependency (fixes SQLite errors)
Ran sample NL→SQL queries across both dialects to ensure consistent behavior

Why this matters

Fixes runtime failures for SQLite users
Removes dialect-specific assumptions → more robust + portable
Improves developer experience for local/testing setups

Happy to add more tests or refine further if needed. Thanks!

[OSContribution - Dialect.docx](https://github.com/user-attachments/files/26896309/OSContribution.-.Dialect.docx)

Krithi-murugesan:
Please apply the llm-generated label as this contribution was assisted by an LLM.

PR #5346, opened 2026-04-08
fix: preserve async task outputs when handling conditional tasks

Ricardo-M-L:
## Summary

When async tasks precede a `ConditionalTask` in a crew's sequential execution, their outputs are silently lost. This happens because `_handle_conditional_task` (and its async counterpart `_ahandle_conditional_task`) reassign the `task_outputs` parameter to a new list returned by `_process_async_tasks`, but this only changes the local variable — the caller's `task_outputs` list is never updated with the async results.

**Root cause:** In Python, reassigning a function parameter (`task_outputs = new_list`) only rebinds the local name; it does not mutate the caller's list. The `futures.clear()` call on the next line works correctly because `clear()` mutates in-place.

**Impact:** When a workflow has `[async_task_A, async_task_B, conditional_task_C, sync_task_D]`, the outputs of tasks A and B are used correctly within `check_conditional_skip` (via the local variable), but are lost from the caller's `task_outputs`. This means task D and the final `CrewOutput` will be missing the async task results.

**Fix:** Replace the local reassignment with in-place mutation (`clear()` + `extend()`) so the caller's list is properly updated.

## Test plan

- [ ] Verify a crew with `[async_task, ConditionalTask, sync_task]` includes the async task output in the final `CrewOutput`
- [ ] Verify the conditional task still receives the correct previous output for its condition check
- [ ] Verify existing conditional task tests continue to pass

---

> [!NOTE]
> **Medium Risk**
> Touches the sequential execution path for both sync and native-async crews; a small change, but it affects ordering/propagation of task outputs that downstream tasks and final `CrewOutput` depend on.
> 
> **Overview**
> Fixes a bug where async task results could be dropped when a `ConditionalTask` is encountered in sequential execution.
> 
> Both `_handle_conditional_task` and `_ahandle_conditional_task` now **mutate** the passed-in `task_outputs` list in-place (`clear()`/`extend()`) with outputs from `_process_async_tasks`/`_aprocess_async_tasks`, instead of rebinding the local parameter, ensuring downstream tasks and the final `CrewOutput` retain prior async outputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a53c129d6d819b6ac2545f17acd32b19484af341. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -1382,9 +1382,11 @@ class Crew(FlowTrackable, BaseModel):
     ) -> TaskOutput | None:
         """Handle conditional task evaluation using native async."""
         if pending_tasks:
-            task_outputs.extend(
-                await self._aprocess_async_tasks(pending_tasks, was_replayed)
+            async_outputs = await self._aprocess_async_tasks(
+                pending_tasks, was_replayed
             )
+            task_outputs.clear()
+            task_outputs.extend(async_outputs)
             pending_tasks.clear()
 
         return check_conditional_skip(
@@ -1589,7 +1591,9 @@ class Crew(FlowTrackable, BaseModel):
         was_replayed: bool,
     ) -> TaskOutput | None:
         if futures:
-            task_outputs.extend(self._process_async_tasks(futures, was_replayed))
+            async_outputs = self._process_async_tasks(futures, was_replayed)
+            task_outputs.clear()
+            task_outputs.extend(async_outputs)
             futures.clear()
 
         return check_conditional_skip(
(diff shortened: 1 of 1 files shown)


Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

Ricardo-M-L:
Addressed the review feedback. Removed both task_outputs.clear() calls so that previously completed synchronous task outputs are preserved when extending with async outputs. The extend() already appends without overwriting, so sync results in patterns like [sync_task, async_task, ConditionalTask] are no longer lost.

PR #5387, opened 2026-04-09
feat: checkpoint TUI with tree view, fork support, editable inputs/outputs

greysonlalonde:
## Summary
- Rewrite checkpoint TUI with `Tree` widget showing branch/fork lineage and collapsible entities
- Add Resume and Fork buttons with editable input fields and task output overrides for what-if exploration
- Store trigger event type in checkpoint JSON at write time instead of inferring from event record
- Unique fork branch names to prevent collisions on repeated forks
- Auto-detect `.checkpoints.db`, fix provider/location bugs on restore, docs updates

## Test plan
- [ ] Run `crewai checkpoint` against a SQLite DB with checkpoints — verify tree renders with correct nesting
- [ ] Fork from a checkpoint, confirm new branch appears with unique name
- [ ] Edit inputs and task outputs in TUI, fork, verify downstream tasks re-run with overrides
- [ ] Resume from a checkpoint, confirm completed tasks are skipped
- [ ] Verify trigger labels match configured `on_events`

---

> [!NOTE]
> **Medium Risk**
> Touches checkpoint serialization/restoration and CLI execution paths (including mutating restored task outputs and agent executor message history), so regressions could affect resume/fork correctness across JSON/SQLite providers.
> 
> **Overview**
> Adds a significantly richer `crewai checkpoint` experience: the TUI is rewritten to a tree view that groups checkpoints by `branch`, nests forks under their `parent_id`, and supports **Resume** vs **Fork** actions.
> 
> Enables *what-if* workflows by extracting kickoff input placeholders and task outputs from checkpoints, letting users edit inputs and completed task outputs in the TUI; on fork/resume these edits are applied to the restored crew, and downstream tasks are invalidated to force re-execution.
> 
> Improves checkpoint data fidelity and restore behavior by writing the triggering event type (`trigger`) directly into checkpoint payloads, persisting each task’s original description/expected output for interpolation, ensuring `RuntimeState` retains the detected provider on restore, and making auto-fork branch names unique. CLI also auto-detects `./.checkpoints.db` when the default directory doesn’t exist, and `SqliteProvider` locations now auto-append `.db` when missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 642d2bb4fe804f7178bb6eefe093738d759a9214. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iris-clawd (approved):
Thorough review of the full diff (631+, 239-). This is a substantial upgrade to the checkpoint experience.

**What's good:**
- **Tree widget** replacing OptionList is the right call — branch/fork lineage is naturally hierarchical. The multi-pass fork attachment logic handles nested forks correctly with the `max_passes` guard.
- **Editable task outputs for what-if exploration** is a killer feature. The invalidation logic (clearing `subsequent.output = None` + resetting agent executor) correctly forces downstream re-execution.
- **Trigger event stored at write time** instead of inferring from event record is much more reliable.
- **Fork branch uniqueness** (`{checkpoint_id}_{uuid[:6]}`) prevents collisions on repeated forks from the same checkpoint.
- **`_base_location` helper** — cleanly extracts the storage root for both SQLite and JSON so the restored state knows where to write subsequent checkpoints.
- **`_detect_location`** auto-fallback to `.checkpoints.db` is a nice UX touch.
- **Task original description/expected_output persistence** — same checkpoint field pattern used elsewhere, with proper round-trip through serialize → restore.
- Docs updated with the new `restore_from` flow, forking, and CLI examples.

**Notes (non-blocking):**
1. **Agent executor message override** (line ~559 in checkpoint_tui.py) — finding the last assistant message and replacing its content is a reasonable heuristic, but could misfire if the agent had multi-turn conversation (the last assistant msg might not be the task result). Probably fine for MVP.
2. **`CheckpointConfig._register_handlers`** now auto-appends `.db` for SQLite providers — side effect in a validator. Works but could surprise users who pass a custom location without extension. Worth a docstring note.
3. **Task override invalidation** clears output for all subsequent tasks, which is correct for sequential execution. For parallel/conditional task graphs, invalidation might need to be more targeted in the future.
4. **`checkpoint_original_description` and `checkpoint_original_expected_output`** are public Fields on Task — they'll show up in `model_dump()` output even when not checkpointing. Consider `exclude=True` in the Field definition if they shouldn't appear in normal serialization, or document them as checkpoint-only fields.

CI pending but code is solid. Great feature set 🚀

iris-clawd (approved):
Re-reviewed the 9 new commits since my first review. Nice work addressing all the Bugbot findings and iteratively hardening the override logic:

**Checkpoint TUI fixes (all solid):**
- `_task_output_ids` properly initialized in `__init__` ✅
- Two-pass task override: overrides applied first, then invalidation via `earliest = min(task_overrides)` — eliminates the silent-drop bug ✅
- Flat task index counter for multi-entity checkpoints ✅
- Original text stored in the tuple so `_collect_task_overrides` doesn't need to re-fetch from entity tasks ✅
- Membership test (`not in task_overrides`) instead of truthiness check ✅
- `await mount()` calls properly awaited ✅
- `_trigger` cleared in `finally` block to prevent stale leakage ✅
- Deterministic message override via system boundary segmentation (much more reliable than the naive last-assistant heuristic) ✅
- `_resuming` cleared on overridden agents so setup runs for the next task ✅
- Renamed shadowed `entry` params in tree builder helpers ✅

**Token tracking (bonus scope — great addition):**
- `reasoning_tokens` and `cache_creation_tokens` added to `UsageMetrics` and `_token_usage` dict
- All providers updated: OpenAI (both Chat + Responses), Anthropic, Azure, Bedrock, Gemini
- Bedrock cached token tracking via `cacheReadInputTokenCount`/`cacheReadInputTokens`
- Good test coverage for all new fields across all providers

**One note:** The dead `_location` private attr and `_base_location` helper were cleanly removed — good cleanup.

CI all green. LGTM 🚀 💬 206

iris-clawd (approved):
Re-reviewed 3 new commits (dd14e85, ca48d50, 642d2bb). All solid:

1. **Trigger injection refactored** — much cleaner: `_trigger` PrivateAttr removed entirely, trigger now injected into the serialized payload dict directly in `_do_checkpoint`. No more stale state risk. `event_record` properly uses `mode="json"` now too.

2. **`enumerate` offset** — replaces `crew.tasks.index(subsequent)` with `enumerate(..., start=earliest+1)`. Correct and O(n) instead of O(n²).

3. **Tree display hardening** — `_entry_id()` normalizes JSON filenames vs SQLite IDs for consistent parent lookup, `made_progress` guard prevents infinite loop on orphaned branches, orphans rendered with `(orphaned)` tag. Good defensive coding.

CI all green (25/25). LGTM 🚀 💬 207

PR #5388, opened 2026-04-10
fix: forward strict mode to Anthropic and Bedrock providers

greysonlalonde:
## Summary
- Anthropic and Bedrock providers were not forwarding `strict: true` from the OpenAI-format tool schema during conversion, so neither provider used constrained decoding for tool calls
- Without constrained decoding, the model can return string `"None"` instead of JSON `null` for nullable `Optional` fields, causing Pydantic validation failures downstream
- Added `strict: true` forwarding in `_convert_tools_for_interference` (Anthropic) and `_format_tools_for_converse` (Bedrock), plus added `strict` to the Bedrock `ToolSpec` TypedDict

## Context
The native tool calling path builds tool schemas via `convert_tools_to_openai_schema` which sets `strict: True`. The OpenAI provider already forwards this. Anthropic and Bedrock providers were written before those APIs supported strict mode and were never updated after the feature shipped (Anthropic GA, Bedrock GA Feb 2026).

## Test plan
- [ ] Verify Anthropic tool calls with `Optional` fields return `null` instead of `"None"`
- [ ] Verify Bedrock tool calls with `Optional` fields return `null` instead of `"None"`
- [ ] Confirm no regression on tools without nullable fields

---

> [!NOTE]
> **Medium Risk**
> Behavior changes in tool-calling for Anthropic/Bedrock by enabling provider-side strict/constrained decoding when requested, and bumps AWS SDK dependencies (plus lockfile updates) which may introduce compatibility/regression risk.
> 
> **Overview**
> Fixes tool conversion for the Anthropic and Bedrock LLM providers so OpenAI-style tools with `function.strict: true` now propagate that flag into each provider’s tool spec (`AnthropicCompletion._convert_tools_for_interference` and `BedrockCompletion._format_tools_for_converse`, with `strict` added to Bedrock’s `ToolSpec`). This enables strict/constrained tool-call decoding on those providers.
> 
> Updates AWS optional dependency pins in `pyproject.toml` (notably `boto3`/`aiobotocore`) and refreshes `uv.lock` with the resulting dependency/version changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e84a104bf8a4216331c9d7b15c7d854dd0b4794e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/pyproject.toml
+++ b/lib/crewai/pyproject.toml
@@ -74,8 +74,8 @@ qdrant = [
     "qdrant-client[fastembed]~=1.14.3",
 ]
 aws = [
-    "boto3~=1.40.38",
-    "aiobotocore~=2.25.2",
+    "boto3~=1.42.79",
+    "aiobotocore~=3.4.0",
 ]
 watson = [
     "ibm-watsonx-ai~=1.3.39",
@@ -87,7 +87,7 @@ litellm = [
     "litellm~=1.83.0",
 ]
 bedrock = [
-    "boto3~=1.40.45",
+    "boto3~=1.42.79",
 ]
 google-genai = [
     "google-genai~=1.65.0",
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -494,6 +494,10 @@ class AnthropicCompletion(BaseLLM):
                     "required": [],
                 }
 
+            func_info = tool.get("function", {})
+            if func_info.get("strict"):
+                anthropic_tool["strict"] = True
+
             anthropic_tools.append(anthropic_tool)
 
         return anthropic_tools
--- a/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
@@ -118,7 +118,7 @@ def _preprocess_structured_data(
 
 
 try:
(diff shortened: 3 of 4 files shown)


iris-clawd (approved):
Clean surgical fix. +10/-1.

The pattern is consistent: both Anthropic and Bedrock now check `function.strict` and forward it to their native tool specs, matching what OpenAI already does. The `ToolSpec` TypedDict update for Bedrock is correct.

One minor observation: the Anthropic API docs show `strict` was only recently GA'd — worth verifying that older SDK versions don't reject the field. But since it's only set conditionally (`if func_info.get("strict")`), tools without strict mode are unaffected.

Also: good catch removing the `# type: ignore[import-untyped]` on aiobotocore — that's a nice cleanup.

LGTM 🚀

iris-clawd (approved):
Re-reviewed — clean fix, CI all green. The strict forwarding pattern is consistent across both providers and the conditional guard means non-strict tools are unaffected. boto3/aiobotocore bumps are necessary for Bedrock's `strict` field support (botocore 1.42.42+).

One thing worth addressing from Bugbot's review: the OpenAI provider calls `force_additional_properties_false(params_dict)` before setting `strict: True` (completion.py L1574), but neither Anthropic nor Bedrock do this. Both Anthropic and Bedrock strict mode require `additionalProperties: false` on all object types in the schema — without it, the API will reject the tool definition. Might want to add that call in both conversion paths before shipping.

LGTM otherwise 🚀 💬 205

PR #5401, opened 2026-04-10
fix: rename loop variable that shadows provider parameter in create_crew()

### LINKED ISSUE #5270: [BUG] create_crew() loop variable shadows `provider` function parameter
### Description

In `create_crew()` (`cli/create_crew.py`), the for-loop on line 214 uses `provider` as its loop variable:

```python
for provider, env_keys in ENV_VARS.items():
```

This shadows the `provider` parameter that's passed in from the CLI (`crewai create crew <name> --provider <value>`). After the loop finishes, `provider` no longer holds the user's original selection — it holds whatever the last key from `ENV_VARS` was (or the key where `break` was hit).

Currently the parameter isn't read again after the loop (the function uses `selected_provider` from interactive selection instead), so this doesn't cause a visible behavioral issue *today*. But it's a latent bug — anyone adding code after the loop that references `provider` would silently get the wrong value. It also trips static analysis warnings and makes debugging confusing.

### Steps to Reproduce

1. Read `create_crew()` in `lib/crewai/src/crewai/cli/create_crew.py`
2. Notice that line 214 uses `provider` as the loop variable in `for provider, env_keys in ENV_VARS.items()`
3. Add a `print(provider)` after the loop (line 221) and call `create_crew("test", provider="openai")`
4. Observe that `provider` is no longer `"openai"`

### Expected behavior

The `provider` parameter should retain its original value throughout the function. Loop variables should not shadow function parameters.

### Screenshots/Code snippets

```python
def create_crew(
    name: str,
    provider: str | None = None,    # <-- parameter
    skip_provider: bool = False,
    parent_folder: str | None = None,
) -> None:
    ...
    existing_provider = None
    for provider, env_keys in ENV_VARS.items():   # <-- shadows parameter!
        if any(
            "key_name" in details and details["key_name"] in env_vars
            for details in env_keys
        ):
            existing_provider = provider
            break
    # At this point, `provider` is no longer the caller's value
```

### Operating System

Ubuntu 22.04

### Python Version

3.12

### crewAI Version

1.13.0

### crewAI Tools Version

1.13.0

### Virtual Environment

Venv

### Evidence

Standard Python scoping behavior — a `for` loop variable persists after the loop and overwrites any variable with the same name in the enclosing scope. Any linter (pylint W0621, ruff) would flag this as `redefined-from-outer-scope`.

### Possible Solution

Rename the loop variable to `env_provider` (or similar) so the parameter isn't clobbered. I have a one-line fix ready and will open a PR.

### Additional context

Side note: the `--provider` CLI flag doesn't seem to actually bypass the interactive provider selection — the function always calls `select_provider()` regardless. That's a separate issue though.

Dheeraj-Bhaskaruni:
## Summary

Fixes #5270

The `for provider, env_keys in ENV_VARS.items()` loop in `create_crew()` shadows the `provider` function parameter. After the loop completes, `provider` holds the last key from `ENV_VARS` instead of the caller's original value.

Renamed the loop variable to `env_provider` to prevent the shadowing.

### File changed

- `lib/crewai/src/crewai/cli/create_crew.py`

## Test plan

- [x] `crewai create crew test --provider openai` still works correctly
- [x] Existing env var detection still identifies configured providers
- [x] No linter warnings for variable shadowing (pylint W0621)


greysonlalonde:
Duplicate of #5272. Closing.

PR #5478, opened 2026-04-15
fix: handle cyclic JSON schemas in MCP tool resolution

greysonlalonde:
Fixes #5474

---

> [!NOTE]
> **Medium Risk**
> Touches dynamic Pydantic model generation and schema sanitization used for tool schemas; while mostly defensive, it can change emitted types/validation behavior and should be verified against representative schemas (especially unions/allOf and recursive refs).
> 
> **Overview**
> MCP native tool resolution now **catches failures when converting `inputSchema` to a Pydantic args model** and logs a warning, registering the tool without a typed schema instead of failing resolution.
> 
> `pydantic_schema_utils` is hardened to **handle cyclic `$ref` graphs and self-referential schemas**: ref expansion and all schema-walking sanitizers add cycle guards, dynamic model creation tracks in-progress schema ids to emit `ForwardRef` back-edges, and incomplete models are `model_rebuild()`’d after construction; optional fields switch to `Optional[T]` to avoid `ForwardRef` incompatibilities.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4aefe315f407c3a6ba0fee120c2b393e2f3b0e47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5498, opened 2026-04-16
fix(mcp): finish recursive schema model creation after #5478

shaun0927:
Addresses the model-creation follow-up from #5490

## Summary
- keep the existing `jsonref.replace_refs()` fast path for normal schemas
- add a narrow fallback for recursive top-level `$ref` schemas when `jsonref` cannot fully dereference them
- resolve the top-level `$ref` against the original root schema so the existing in-progress / `ForwardRef` machinery can handle recursive back-edges
- add a focused regression test covering recursive MCP-style schema creation

## Why this scope

Recent merged fixes in this repo tend to land when they stay narrowly attached to the observed regression and avoid broad rewrites.

This PR follows that pattern:
- it does not replace the existing non-recursive preprocessing path
- it only changes the error path where `jsonref.replace_refs()` still fails for recursive top-level refs
- it relies on the recursive model-building logic that already exists in the file instead of introducing a second schema builder

This deliberately targets the concrete `create_model_from_schema()` failure path reported in #5490. If maintainers want to tighten the strict-sanitization output for recursive schemas further, that can stay as a separate follow-up.

## Test plan
- [x] `uv run pytest lib/crewai/tests/utilities/test_pydantic_schema_utils.py -q`
- [x] `uv run ruff format --check lib/crewai/src/crewai/utilities/pydantic_schema_utils.py lib/crewai/tests/utilities/test_pydantic_schema_utils.py`
- [x] `uv run ruff check lib/crewai/src/crewai/utilities/pydantic_schema_utils.py lib/crewai/tests/utilities/test_pydantic_schema_utils.py`
- [ ] manual verification against a real MCP server exposing recursive input schemas

---
This PR was drafted with AI assistance. I could not apply an `llm-generated` label from my account.

### DIFF
--- a/lib/crewai/src/crewai/utilities/pydantic_schema_utils.py
+++ b/lib/crewai/src/crewai/utilities/pydantic_schema_utils.py
@@ -813,9 +813,14 @@ def create_model_from_schema(  # type: ignore[no-any-unimported]
         >>> person.name
         'John'
     """
-    json_schema = dict(jsonref.replace_refs(json_schema, proxies=False))
-
-    effective_root = root_schema or json_schema
+    effective_root = deepcopy(root_schema or json_schema)
+    try:
+        json_schema = dict(jsonref.replace_refs(json_schema, proxies=False))
+        effective_root = json_schema if root_schema is None else effective_root
+    except jsonref.JsonRefError:
+        json_schema = deepcopy(json_schema)
+        if "$ref" in json_schema:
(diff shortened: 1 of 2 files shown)


shaun0927:
Closing this as superseded by the maintainer fix in #5500.

Issue #5490 has already been closed as completed, and the maintainers resolved it through their own follow-up PR. Rather than keep a parallel implementation open, I'm closing this branch to reduce review noise.

PR #5505, opened 2026-04-16
fix: scope streaming handlers to prevent cross-run chunk contamination

greysonlalonde:
## Summary
- Streaming handlers on the singleton `crewai_event_bus` received ALL `LLMStreamChunkEvent` emissions, causing cross-run chunk fan-out when multiple streaming runs executed concurrently
- Adds a `ContextVar`-based stream scope ID (`_current_stream_id`) so each handler only accepts events emitted within its own execution context
- `LLMStreamChunkEvent` handlers run synchronously in the emitter's thread, so the context var naturally reflects the correct stream scope

Closes #5376

## Test plan
- [ ] Added `TestConcurrentStreamIsolation` regression test: sets up two streams in separate contexts, emits chunks concurrently, asserts zero cross-contamination
- [ ] All 35 existing streaming tests pass unchanged

---

> [!NOTE]
> **Medium Risk**
> Touches core streaming/event-bus integration and relies on `ContextVar` propagation across threads/tasks; mistakes could lead to missing chunks or continued fan-out under concurrency.
> 
> **Overview**
> Prevents concurrent streaming runs from cross-contaminating by scoping each registered `LLMStreamChunkEvent` handler to a per-run `stream_id` propagated via a `ContextVar`.
> 
> `create_streaming_state` now generates and stores a UUID `stream_id`, handlers drop events outside their context, and both sync/async chunk generators temporarily push the `stream_id` into the copied execution context so only in-scope emitters are observed. Adds a regression test that runs two concurrent streams and asserts each receives only its own chunks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7359c5c1f67216198cab566586797f5e4c1cfbb9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/utilities/streaming.py
+++ b/lib/crewai/src/crewai/utilities/streaming.py
@@ -7,6 +7,7 @@ import logging
 import queue
 import threading
 from typing import Any, NamedTuple
+import uuid
 
 from typing_extensions import TypedDict
 
@@ -24,6 +25,10 @@ from crewai.utilities.string_utils import sanitize_tool_name
 
 
 logger = logging.getLogger(__name__)
+
+_current_stream_id: contextvars.ContextVar[str | None] = contextvars.ContextVar(
+    "_current_stream_id", default=None
+)
 
 
 class TaskInfo(TypedDict):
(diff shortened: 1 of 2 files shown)


iris-clawd (approved):
ContextVar tuple-stack with deferred scope at context-copy site is the right design. Clean isolation, no parent context pollution, solid test coverage. 👍

PR #5518, opened 2026-04-17
fix: remove erroneous dollar-sign from BrightDataSearchTool search URLs and expose results_count in schema

### LINKED ISSUE #5269: [BUG] BrightData SERP tool uses JavaScript template syntax in Python f-strings, breaking all search queries
### Description

The `get_search_url` method in `BrightDataSearchTool` uses `${query}` (JavaScript template literal syntax) instead of `{query}` (Python f-string syntax) when building search URLs. This causes every search query to be prepended with a literal `$` character.

For example, searching for "AI news" produces:
```
https://www.google.com/search?q=$AI%20news
```
instead of:
```
https://www.google.com/search?q=AI%20news
```

All three search engines (Google, Bing, Yandex) are affected.

### Steps to Reproduce

1. Set up `BRIGHT_DATA_API_KEY` and `BRIGHT_DATA_ZONE` environment variables
2. Create a `BrightDataSearchTool` and call it with any query
3. Inspect the URL being sent to the Bright Data API

### Expected behavior

The search URL should contain the query string without any `$` prefix. `get_search_url("google", "test")` should return `https://www.google.com/search?q=test`.

### Screenshots/Code snippets

```python
# Current code (brightdata_serp.py, lines 132-137)
def get_search_url(self, engine: str, query: str) -> str:
    if engine == "yandex":
        return f"https://yandex.com/search/?text=${query}"  # <-- JS syntax
    if engine == "bing":
        return f"https://www.bing.com/search?q=${query}"    # <-- JS syntax
    return f"https://www.google.com/search?q=${query}"       # <-- JS syntax
```

### Operating System

Ubuntu 22.04

### Python Version

3.12

### crewAI Version

1.13.0

### crewAI Tools Version

1.13.0

### Virtual Environment

Venv

### Evidence

In Python, `f"...${query}"` produces a literal `$` followed by the interpolated value. Only `{query}` (without `$`) is correct f-string syntax. This can be verified in a Python shell:

```python
>>> query = "test"
>>> f"https://www.google.com/search?q=${query}"
'https://www.google.com/search?q=$test'
>>> f"https://www.google.com/search?q={query}"
'https://www.google.com/search?q=test'
```

### Possible Solution

Remove the `$` prefix from all three f-strings in `get_search_url`. I have a fix ready and will open a PR shortly.

### Additional context

Looks like the `${}` syntax was carried over from a JavaScript codebase or written by someone used to JS template literals. The fix is straightforward — just removing the three `$` characters.

octo-patch:
Fixes #5269

## Problem

`BrightDataSearchTool.get_search_url` used JavaScript template-literal syntax inside Python f-strings. Python treats `$` as a literal character, so every constructed URL contained a stray dollar sign:

- Google: `https://www.google.com/search?q=$hello%20world`
- Bing: `https://www.bing.com/search?q=$some+query`
- Yandex: `https://yandex.com/search/?text=$query`

This caused all SERP searches to fail because the search engine received `$<query>` instead of the actual query string.

Additionally, `results_count` was read from `**kwargs` with a hard-coded default of `"10"`, making it completely invisible to the Pydantic schema. Agents had no way to control the number of results returned.

## Solution

1. Removed the `$` prefix from all three f-string URL templates in `get_search_url`.
2. Added `results_count` to `BrightDataSearchToolSchema` as a typed `int | None` field with `default=10`, so agents can explicitly pass the desired number of results.
3. Added `results_count` as an explicit parameter to `_run`, replacing the hidden `kwargs.get("results_count", "10")` pattern.
4. Added tests verifying correct URL construction and results_count handling.

## Testing

New unit tests added to `brightdata_serp_tool_test.py`:
- `test_get_search_url_google` — verifies no `$` in Google URL
- `test_get_search_url_bing` — verifies no `$` in Bing URL
- `test_get_search_url_yandex` — verifies no `$` in Yandex URL
- `test_run_default_results_count_is_10` — verifies default `num=10`
- `test_run_with_custom_results_count` — verifies custom count is forwarded

> Note: This PR was created with AI assistance. Per the CONTRIBUTING.md guidelines, the `llm-generated` label should be applied (label does not yet exist in the repo).

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/tools/brightdata_tool/brightdata_serp.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/brightdata_tool/brightdata_serp.py
@@ -32,6 +32,7 @@ class BrightDataSearchToolSchema(BaseModel):
         search_type (Optional[str]): Type of search, such as "isch" (images), "nws" (news), "jobs", etc.
         device_type (Optional[str]): Device type to simulate ("desktop", "mobile", "ios", "android"). Default is "desktop".
         parse_results (Optional[bool]): If True, results will be returned in structured JSON. If False, raw HTML. Default is True.
+        results_count (Optional[int]): Number of search results to fetch. Default is 10.
     """
 
     query: str = Field(..., description="Search query to perform")
@@ -58,6 +59,10 @@ class BrightDataSearchToolSchema(BaseModel):
     parse_results: bool | None = Field(
         default=True,
         description="Whether to parse and return JSON (True) or raw HTML/text (False)",
+    )
+    results_count: int | None = Field(
+        default=10,
+        description="Number of search results to fetch (default: 10)",
     )
 
 
@@ -131,10 +136,10 @@ class BrightDataSearchTool(BaseTool):
 
     def get_search_url(self, engine: str, query: str) -> str:
         if engine == "yandex":
-            return f"https://yandex.com/search/?text=${query}"
+            return f"https://yandex.com/search/?text={query}"
         if engine == "bing":
-            return f"https://www.bing.com/search?q=${query}"
-        return f"https://www.google.com/search?q=${query}"
+            return f"https://www.bing.com/search?q={query}"
+        return f"https://www.google.com/search?q={query}"
 
     def _run(
         self,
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:
Closing in favor of #7326, which lands the same `$` → `{query}` fix for `get_search_url` and is tied to the open issue (#7325).

The extra `results_count` schema work here is out of scope for that bug — happy to take that as a follow-up if it is still wanted.

PR #5552, opened 2026-04-20
fix: preserve thought_signature in Gemini streaming tool calls

greysonlalonde:
## Summary
- Gemini thinking models (2.5+, 3.x) return a `thought_signature` on `functionCall` parts that must be passed back in conversation history
- The streaming path was extracting only name/args into plain dicts, stripping the signature and causing `400 INVALID_ARGUMENT` on subsequent turns
- Now returns raw `Part` objects from streaming (matching the non-streaming path), so the executor preserves them via `raw_tool_call_parts`

## Test plan
- [x] Existing Gemini test suite passes (97 passed)
- [x] Streaming tests pass (93 passed)
- [ ] Manual verification: Gemini 3.x model + streaming + tool calls completes multi-turn without 400 error

---

> [!NOTE]
> **Medium Risk**
> Changes the streaming Gemini tool-call return type from normalized dicts to raw `types.Part` objects, which may affect downstream consumers expecting OpenAI-style tool_call dicts. Scope is limited to the Gemini streaming path when `available_functions` is not provided.
> 
> **Overview**
> Preserves Gemini thinking-model metadata (e.g. `thought_signature`) during streaming tool calls by storing each raw `types.Part` (`raw_part`) alongside extracted name/args.
> 
> When streaming returns tool calls *without* `available_functions`, the completion now emits and returns the collected raw `Part` objects instead of converting them into simplified `{"id","function":...}` dicts, aligning streaming behavior with the non-streaming path and preventing loss of provider-specific fields.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7f83d3d1e53cf24f27d62f607e1a6ef63861248. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/gemini/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/gemini/completion.py
@@ -976,6 +976,7 @@ class GeminiCompletion(BaseLLM):
                             "id": call_id,
                             "name": part.function_call.name,
                             "args": args_dict,
+                            "raw_part": part,
                         }
 
                         self._emit_stream_chunk_event(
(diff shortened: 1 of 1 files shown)


iris-clawd (approved):
Clean fix. The non-streaming path already returns raw `Part` objects when `available_functions` is empty — this makes the streaming path do the same.

**The bug:** Gemini 2.5+/3.x thinking models attach a `thought_signature` to `functionCall` parts that must round-trip through conversation history. The streaming finalizer was converting to OpenAI-style `{id, function: {name, arguments}}` dicts, stripping the signature → `400 INVALID_ARGUMENT` on the next turn.

**The fix:** Store the raw `Part` on each streaming tool call (`raw_part` key), then return those raw parts instead of the formatted dicts. The executor already handles this via `raw_tool_call_parts` on the assistant message (agent_executor.py:1524), and `_convert_contents` (completion.py:641-644) already reconstructs `types.Content` from raw parts when available.

Net -8 lines. No new surface area — just aligning streaming with the existing non-streaming contract. LGTM ✅

💬 317

PR #5588, opened 2026-04-22
fix: reset messages and iterations at the start of each invoke/ainvoke

### LINKED ISSUE #4389: [BUG] CrewAgentExecutor does not reset messages and iterations between task executions

VANDRANKI:
Fixes #4389.

**Bug**

`CrewAgentExecutor` is reused across sequential tasks in a Crew. Neither `invoke()` nor `ainvoke()` cleared `self.messages` or reset `self.iterations` before starting a new task, so task N would see the entire message history from tasks 1 through N-1. This leads to incorrect context leaking between tasks and inflated iteration counts.

The experimental `AgentExecutor` already handles this correctly (`state.messages.clear()`, `state.iterations = 0` in `invoke()`).

**Fix**

In the non-resuming path of both `invoke()` and `ainvoke()`, reset state before calling `_setup_messages`:

```python
self.messages = []
self.iterations = 0
self._setup_messages(inputs)
```

The `_resuming` path is intentionally left unchanged - resuming a paused execution should preserve existing message history.


greysonlalonde:
Thanks, can you remove the comments? logic is self-explanatory

greysonlalonde:
Thanks for the contribution! This was independently fixed on `main` in 4c74dc0f8 ("fix(executor): reset messages and iterations between invocations"), which adds the same three-line reset in the non-resuming branch of both `invoke()` and `ainvoke()`. Closing as superseded.

PR #5592, opened 2026-04-22
feat(azure): fall back to DefaultAzureCredential when no API key

mattatcha:


mattatcha:
Closing. Will reopen with sanitized language.

PR #5596, opened 2026-04-23
feat: emit lifecycle events for checkpoint operations

greysonlalonde:
## Summary
- Adds a `checkpoint_*` event family (10 events) in `events/types/checkpoint_events.py` covering write, fork, restore lifecycles plus a one-shot prune event, so tracing/observability listeners can see checkpoint activity.
- Emits from every checkpoint code path: `_do_checkpoint` (listener), `RuntimeState.checkpoint` / `acheckpoint` (explicit writes), `RuntimeState.fork`, and `RuntimeState.from_checkpoint` / `afrom_checkpoint`.
- Widens `BaseProvider.prune` to return `int` (removed count); `JsonProvider` and `SqliteProvider` updated to report the count so `CheckpointPrunedEvent` can carry it.

## Notes
- Events use `CheckpointConfig` / `RuntimeState` as the emit source so `_find_checkpoint` returns `None` and there is no recursion.
- `_on_any_event` also early-returns for any `CheckpointBaseEvent` / `CheckpointForkBaseEvent` / `CheckpointRestoreBaseEvent` as belt-and-suspenders.
- Because the checkpoint listener and `TraceCollectionListener` both auto-subscribe via `BaseEvent.__subclasses__()`, no extra listener wiring is required.

## Test plan
- [ ] `uv run mypy .` clean
- [ ] `uv run ruff check .` / `uv run ruff format --check .` clean
- [ ] Explicit write: `RuntimeState.checkpoint(path)` emits `CheckpointStartedEvent` → `CheckpointCompletedEvent` with populated `checkpoint_id` and `duration_ms`
- [ ] Failure path: writing to a read-only location emits `CheckpointStartedEvent` → `CheckpointFailedEvent` then re-raises
- [ ] Fork: `RuntimeState.fork('demo')` emits `CheckpointForkStartedEvent` → `CheckpointForkCompletedEvent` with correct `parent_branch` / `branch`
- [ ] Restore: `RuntimeState.from_checkpoint(cfg)` emits `CheckpointRestoreStartedEvent` → `CheckpointRestoreCompletedEvent`; bad location emits `CheckpointRestoreFailedEvent`
- [ ] Prune: `CheckpointConfig(max_checkpoints=2)` + 4 writes emits `CheckpointPrunedEvent` with `removed_count` of 0 then 1 per overflow
- [ ] Any custom provider subclassing `BaseProvider` with a `prune` that previously returned `None` now needs to return an `int` (interface change)

---

> [!NOTE]
> **Medium Risk**
> Moderate risk due to new event emissions across checkpoint write/fork/restore paths and an interface change to `BaseProvider.prune()` that requires all custom providers to update their return type; runtime behavior should be unchanged aside from additional event bus traffic and timing/exception handling.
> 
> **Overview**
> Adds a new `checkpoint_*` event family (write, prune, fork, restore) and exposes it via lazy exports (`events/__init__.py`) and the `EventTypes` union (`event_types.py`) so observers can subscribe to checkpoint activity.
> 
> Emits `CheckpointStarted/Completed/Failed` (with `duration_ms` and `checkpoint_id`) from both explicit checkpoint APIs (`RuntimeState.checkpoint`/`acheckpoint`) and auto-checkpointing (`checkpoint_listener._do_checkpoint`), plus `CheckpointForkStarted/Completed` from `RuntimeState.fork` and `CheckpointRestoreStarted/Completed/Failed` from `from_checkpoint`/`afrom_checkpoint`. Auto-checkpointing now skips these checkpoint lifecycle events to avoid recursion.
> 
> Updates the provider contract so `BaseProvider.prune()` returns the number of removed checkpoints; `JsonProvider`/`SqliteProvider` implement this and the listener emits a `CheckpointPrunedEvent` and logs-and-continues on prune failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7fed74656336e34f22272c889de4f7dc224e5c66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iris-clawd (approved):
Solid work — well-structured event family with good recursion prevention. A few observations:

**Design ✅**
- Clean event hierarchy: three base events (write, fork, restore) + pruned as one-shot. Literal type discriminators are good for downstream consumers.
- Recursion prevention is belt-and-suspenders: emitting with `cfg`/`self` as source (so `_find_checkpoint` returns None) AND the explicit `isinstance` early-return in `_on_any_event`. Good.
- Provider interface change (`prune → int`) is well-scoped and documented.

**Minor notes (non-blocking):**
- In `_do_checkpoint`, `duration_ms` is captured before pruning but `CheckpointCompletedEvent` emits after pruning — so the completed event's timing is accurate for the write, but consumers might expect it to be the total latency. Might be worth a docstring note.
- The `fork()` try/except around `self._branch = new_branch` is effectively dead code (string assignment can't fail), but harmless as future-proofing if fork logic grows.
- The `**context` splat in `_do_checkpoint` assumes the incoming event has `task_id`/`task_name`/`agent_id`/`agent_role` — should be fine for crew/task lifecycle events that trigger checkpoints, but could `AttributeError` if a custom event without those fields somehow triggers a checkpoint. Low risk given `_should_checkpoint` filtering.

👍

iris-clawd (approved):
Re-approved after update. Good improvements: branch/parent_id are now snapshotted before `_chain_lineage` mutates state, and `CheckpointCompletedEvent` now emits before pruning so `duration_ms` accurately reflects just the write. Clean. 👍

iris-clawd (approved):
Third pass — looks good. All CI green across Python 3.10-3.13, lint, type-checker, CodeQL. A few notes:

**Strengths**
- Clean three-tier event hierarchy (write/fork/restore) with `CheckpointPrunedEvent` as a standalone one-shot — nice design.
- Recursion prevention is solid: `isinstance` guard in `_on_any_event` + using `CheckpointConfig`/`RuntimeState` as emit sources so `_find_checkpoint` returns `None`. Belt-and-suspenders, as you noted.
- Snapshots of `_parent_id` / `_branch` before `_chain_lineage` mutations — the Cursor Bugbot comments about stale `parent_id` don't apply to the current code since you're using the snapshot values.
- `CheckpointCompletedEvent` emits before pruning now, so a prune failure won't swallow the completed event. Bugbot's first concern is also stale.
- Provider contract change (`prune → int`) is clean and both implementations handle it correctly.

**Minor observations (non-blocking)**
- `checkpoint()` / `acheckpoint()` and `from_checkpoint()` / `afrom_checkpoint()` in `runtime.py` are ~90% identical sync/async pairs with copy-pasted event emission blocks. Could extract a helper to reduce duplication, but it's readable as-is.
- The `**context` dict in `_do_checkpoint` accesses `event.task_id`, `event.task_name`, `event.agent_id`, `event.agent_role` — worth confirming these exist on all `BaseEvent` subclasses that trigger checkpoints, otherwise you'll get `AttributeError`. If some events don't carry these, a `getattr(event, 'task_id', None)` pattern would be safer.
- No automated tests for the new event emissions. The manual test plan is thorough, but having at least a couple of unit tests (e.g., verifying `CheckpointCompletedEvent` is emitted with correct `checkpoint_id`, or that the recursion guard skips checkpoint events) would be valuable for regression coverage.

LGTM 👍

PR #5600, opened 2026-04-23
fix(flow): serialize initial_state class refs as JSON schema

greysonlalonde:
## Summary
- Auto-checkpoint of a `Flow` constructed with a Pydantic model *class* as `initial_state` (e.g. `Flow(initial_state=MyState)`) raised `PydanticSerializationError: Unable to serialize unknown type: ModelMetaclass` during `state.model_dump(mode="json")`.
- Add `_serialize_initial_state` + `PlainSerializer` on `Flow.initial_state`: `BaseModel` class refs emit `model_json_schema()` (same pattern as `Task.output_pydantic`), `BaseModel` instances dump to JSON, dicts/primitives pass through, foreign `type` values drop to `None`.
- Regression tests in `TestFlowInitialStateSerialization` cover class ref, instance, and dict inputs.

---

> [!NOTE]
> **Low Risk**
> Low risk: narrowly scoped to `Flow.initial_state` JSON serialization during checkpointing, with regression tests covering the supported input shapes.
> 
> **Overview**
> Fixes checkpoint JSON serialization for `Flow.initial_state` when it is a Pydantic model *class* (previously failing with `PydanticSerializationError` on `ModelMetaclass`).
> 
> Adds a custom `PlainSerializer` so **`BaseModel` class refs serialize as `model_json_schema()`**, `BaseModel` instances serialize via `model_dump(mode="json")`, dicts/primitives pass through, and unsupported bare `type` values serialize as `null`; includes new regression tests for class-ref, instance, and dict inputs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 870436f4a094e73afdcba6589e97f791491de008. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/flow/flow.py
+++ b/lib/crewai/src/crewai/flow/flow.py
@@ -158,20 +158,34 @@ def _resolve_persistence(value: Any) -> Any:
     return value
 
 
+_INITIAL_STATE_CLASS_MARKER = "__crewai_pydantic_class_schema__"
+
+
 def _serialize_initial_state(value: Any) -> Any:
     """Make ``initial_state`` safe for JSON checkpoint serialization.
 
-    ``BaseModel`` class refs are emitted as their JSON schema (same pattern
-    as ``Task.output_pydantic``); ``BaseModel`` instances are dumped to JSON.
-    Bare ``type`` values that are not ``BaseModel`` subclasses (e.g. ``dict``)
-    are dropped since they can't be represented in JSON.
+    ``BaseModel`` class refs are emitted as their JSON schema under a sentinel
+    marker key so deserialization can round-trip them back to a class.
+    ``BaseModel`` instances are dumped to JSON (round-trip as plain dicts,
+    which ``_create_initial_state`` accepts). Bare ``type`` values that are
+    not ``BaseModel`` subclasses (e.g. ``dict``) are dropped since they
+    can't be represented in JSON.
     """
     if isinstance(value, type):
         if issubclass(value, BaseModel):
-            return value.model_json_schema()
+            return {_INITIAL_STATE_CLASS_MARKER: value.model_json_schema()}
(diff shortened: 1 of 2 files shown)


iris-clawd (approved):
All CI green. Clean fix for a real serialization crash.

**What's good:**
- Root cause is correct: `model_dump(mode="json")` can't handle `ModelMetaclass` — the `PlainSerializer` approach is the right pattern.
- Serialization strategy is sensible: class refs → JSON schema under a sentinel marker, instances → `model_dump(mode="json")`, dicts passthrough, unknown `type` values → `None`.
- `BeforeValidator` for deserialization with `create_model_from_schema` handles the round-trip. Good that the test verifies the rehydrated model has the expected fields.
- Test coverage is solid — class ref, instance, dict, and round-trip cases all covered.

**A few observations (non-blocking):**

1. **Lossy round-trip for class refs**: Serializing a class as its JSON schema and rehydrating via `create_model_from_schema` produces a *new* dynamically-generated class, not the original `MyState`. This means `isinstance(state, MyState)` would fail after restore. That's probably fine for checkpoint use (you just need the shape), but worth noting in the docstring.

2. **`type[dict]` in the union**: The type annotation includes `type[dict]` but `_serialize_initial_state` would hit the `isinstance(value, type)` branch and return `None` for it (since `dict` isn't a `BaseModel` subclass). Technically correct behavior (can't serialize a bare `dict` class), but the annotation suggests it's a supported input that would survive serialization.

3. **Sentinel key in user data**: If someone's actual state dict happens to contain `__crewai_pydantic_class_schema__` as a key, the deserializer would misinterpret it. Extremely unlikely but the double-underscore prefix helps.

LGTM 👍

PR #5601, opened 2026-04-23
fix(flow): replay recorded method events on checkpoint resume

greysonlalonde:
## Summary

- On checkpoint resume, completed methods are skipped and never re-emit their `MethodExecution*` events. The UI's Flow Timeline (populated from the trace batch via live-bus events) ends up empty for those steps — showing only `flow_started`/`flow_finished`.
- Fix: on resume, replay the `MethodExecutionStartedEvent`/`Finished`/`Failed` events stored in the checkpoint's `event_record` through the bus so trace and console listeners can reconstruct the full timeline.
- New `crewai_event_bus.replay()` bypasses `_prepare_event` (preserves original `event_id`, `parent_event_id`, `emission_sequence`) and sets an `is_replaying()` ContextVar so side-effect listeners (`CheckpointListener`) opt out.

## Test plan

- [x] New `tests/events/test_event_replay.py` covers: field preservation, `is_replaying()` toggling across `replay`/`emit`, CheckpointListener opt-out, end-to-end resume dispatching events for completed methods
- [x] Full `tests/test_checkpoint.py` + `tests/test_flow_resumability_regression.py` + `tests/test_event_record.py` + `tests/test_flow.py` + `tests/events/` pass (278 tests)
- [x] `ruff check` and `mypy` clean
- [ ] Manual verification in the enterprise UI — resumed flow timeline shows the full method tree with preserved event relationships

---

> [!NOTE]
> **Medium Risk**
> Touches core `Flow` kickoff/resume and the shared event bus by adding a new replay dispatch path; incorrect flag propagation or ordering could cause missing/duplicated handler side effects despite added tests.
> 
> **Overview**
> On checkpoint resume, flows now **replay previously-recorded `MethodExecution*` events** for already-completed methods so timeline/trace listeners can reconstruct a complete execution history instead of showing only high-level flow events.
> 
> This introduces `crewai_event_bus.replay()` with an `is_replaying()` context flag to dispatch stored events **without mutating IDs/sequence or re-recording**, and updates side-effect listeners (notably checkpoint auto-writes) to **opt out during replay**. It also resets event scope/last-event context on checkpoint restore and ensures `Flow.fork()` generates a new state `id`, plus adds targeted unit/integration tests for replay behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1c1a4c6023625586c5e36054013cb8f2804b177. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/events/event_bus.py
+++ b/lib/crewai/src/crewai/events/event_bus.py
@@ -657,18 +657,16 @@ class CrewAIEventsBus:
                 sync_future = self._sync_executor.submit(
                     ctx.run, self._call_handlers, source, event, sync_handlers
                 )
+                self._track_future(sync_future)
                 if not async_handlers:
-                    return self._track_future(sync_future)
+                    return sync_future
 
-            if async_handlers:
-                return self._track_future(
-                    asyncio.run_coroutine_threadsafe(
-                        self._acall_handlers_replaying(source, event, async_handlers),
-                        self._loop,
-                    )
+            return self._track_future(
+                asyncio.run_coroutine_threadsafe(
+                    self._acall_handlers_replaying(source, event, async_handlers),
+                    self._loop,
                 )
-
-            return None
+            )
         finally:
             _replaying.reset(token)
 
--- a/lib/crewai/src/crewai/events/event_bus.py
+++ b/lib/crewai/src/crewai/events/event_bus.py
@@ -64,6 +64,22 @@ P = ParamSpec("P")
 R = TypeVar("R")
 
 
+_replaying: contextvars.ContextVar[bool] = contextvars.ContextVar(
+    "crewai_event_replaying", default=False
+)
+
+
+def is_replaying() -> bool:
+    """Return True if the current context is dispatching a replayed event.
+
+    Listeners with side effects (checkpoint writes, external API calls that
+    should not be repeated) should early-return when this is true. Listeners
+    whose purpose is reconstructing timeline state (trace batch, console
+    formatter) should ignore the flag and process replayed events normally.
+    """
+    return _replaying.get()
+
+
 class CrewAIEventsBus:
     """Singleton event bus for handling events in CrewAI.
 
@@ -260,6 +276,11 @@ class CrewAIEventsBus:
         with self._instance_lock:
             self._runtime_state = state
             self._registered_entity_ids = {id(e) for e in state.root}
+
+    @property
+    def runtime_state(self) -> RuntimeState | None:
+        """The RuntimeState currently attached to the bus, if any."""
+        return self._runtime_state
 
     def register_entity(self, entity: Any) -> None:
         """Add an entity to the RuntimeState, creating it if needed.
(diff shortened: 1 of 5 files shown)


iris-clawd (approved):
This is a well-thought-out addition. The checkpoint resume timeline gap is a real UX problem and replaying stored events is the right fix. CI still running but the approach is sound.

**Design**
- `replay()` as a distinct method from `emit()` is the right call — keeping the "no _prepare_event" semantics explicit rather than adding a flag to `emit()`.
- ContextVar-based `is_replaying()` is clean and thread-safe. The token-on-calling-thread + dedicated `_acall_handlers_replaying`/`_emit_with_dependencies_replaying` wrappers handle cross-thread propagation correctly.
- Checkpoint listener opt-out is minimal — just the `is_replaying()` early-return at the top of `_on_any_event`. Good placement.
- Sorting replayed events by `emission_sequence` preserves causal ordering.

**One potential issue (non-blocking but worth checking):**
In `replay()`, when both `sync_handlers` and `async_handlers` exist, the sync future is submitted via `self._sync_executor.submit()` but `_track_future()` is only called on the async future (the `if not async_handlers: return` branch is skipped). The sync future becomes fire-and-forget — if a sync handler raises, the exception is silently swallowed. For MethodExecution events this is likely fine (most handlers are side-effect-free timeline listeners), but it could mask issues.

**Tests** — solid coverage: field preservation, flag toggling, checkpoint opt-out, and E2E resume scenario.

LGTM

iris-clawd (commented):
Re-reviewed after two new commits (23db2669 + ec985359). Good improvements:

**Commit 2 (23db2669) — addresses my sync future note:**
- `_track_future(sync_future)` is now called even when async handlers exist — sync futures are no longer fire-and-forget. 👍
- Replay now filters to `self._completed_methods` — prevents replaying events for methods that are about to re-execute.
- E2E test assertions tightened from `in` to `count == 1`.

**Commit 3 (ec985359) — fork/restore scope cleanup:**
- `restore_event_scope(())` + `reset_last_event_id()` in `_restore_from_checkpoint` — clears stale event context so resumed methods get fresh parent chains. Good.
- Fork now generates a new flow state `id` via `uuid4()` — ensures forked flows have distinct identities. Clean.

**CI is red** — `test_resume_dispatches_completed_method_events` fails:
```
assert 4 == 1
  where 4 = ["step_a", "step_a", "step_a", "step_a", "step_b", "step_b", "step_b", "step_b", "step_c"].count("step_a")
```

`step_a` fires 4 times instead of 1. Likely cause: the `scoped_handlers()` context manager is not isolating properly — handlers from the first `flow1.kickoff()` are still active when `flow2.kickoff()` runs, AND the replay adds another round. Since `flow1.kickoff()` runs before the `scoped_handlers()` block, its events are captured by the handlers registered inside the block if the bus is shared. The test needs either:
1. Move `flow1.kickoff()` outside any handler scope (which it already is — so the issue is likely that `flow2.kickoff()` re-executes step_a+step_b via normal flow execution AND replays them)
2. The completed_methods filter should prevent step_a/step_b from re-executing via normal flow, but `step_c` listener triggers from step_b, and if replay emits MethodExecutionFinished for step_b, the flow listener graph might fire step_c... which then also emits step_c started/finished naturally.

The 4x count suggests events are being captured from: (a) replay of step_a started, (b) normal re-execution of step_a (if _completed_methods is not skipping start methods), (c) possibly duplicate handler registration. Worth debugging.

iris-clawd (approved):
Re-reviewed after 3 new commits (1d37f98c, a746facf, c1c1a4c6). All CI green ✅

**New changes look good:**

1. **Test fix (1d37f98c)** — clears `event_record` before the test to prevent stale events from `flow1.kickoff()` leaking into assertions.

2. **EventRecord accessors (a746facf)** — `all_nodes()` returns a lock-safe snapshot, `clear()` uses the write lock. Clean API additions that avoid callers needing to know about the internal `_lock`.

3. **flow_name scoping (c1c1a4c6)** — replay filter now matches `n.event.flow_name == flow_name` to exclude sub-flow events. This is the right fix for the 4x duplication — prevents events from nested/sub flows (or prior runs sharing the same RuntimeState) from being replayed into the wrong flow.

The overall design is solid: `replay()` path, `is_replaying()` ContextVar, checkpoint opt-out, fork id regeneration, and event scope clearing on restore. LGTM.

PR #5636, opened 2026-04-27
fix(agent_executor): rename force_final_answer to avoid self-referential router

greysonlalonde:
## Summary
The router method in `lib/crewai/src/crewai/experimental/agent_executor.py` is decorated with `@router("force_final_answer")` and also named `force_final_answer`, so Flow registers the same method under two identifiers that share a name. This causes a self-referential loop when an agent hits `max_iter`.

Renames the method to `ensure_force_final_answer`. The routing token `"force_final_answer"` (emitted by `check_max_iterations`) is unchanged, so existing tests asserting that string remain valid.

---

> [!NOTE]
> **Low Risk**
> Low risk rename of a Flow router handler method; behavior and routing token remain the same, but any external callers referencing the old Python method name would break.
> 
> **Overview**
> Fixes a max-iteration routing edge case by renaming the `@router("force_final_answer")` handler method from `force_final_answer` to `ensure_force_final_answer`, avoiding a self-referential router registration loop.
> 
> The routing key (`"force_final_answer"`) and forced-final-answer behavior are unchanged, so flow transitions and existing string-based assertions should remain stable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f93c0780ae22ec5ddeb78a4adb59de4e75362473. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iris-clawd (approved):
Clean fix — breaks the self-referential router loop without changing routing behavior. ✅

PR #5639, opened 2026-04-28
fix(crew): bind task-only agents to crew so multimodal input_files reach the LLM

### LINKED ISSUE #5534: [BUG] CrewAI: Task.input_files + missing Crew.agents causes multimodal files to be ignored
### Description

There is inconsistent behavior when using multimodal inputs (`input_files`) depending on whether `agents=[...]` is explicitly provided in the `Crew`.

If a `Task` already has an `agent` assigned (`task.agent=agent`), the `Crew` is expected to work without explicitly passing `agents=[agent]`. However, when using `input_files` (e.g., `ImageFile`), the files are silently ignored unless the agent is also explicitly included in `Crew.agents`.

This leads to confusing behavior where:

* The task executes successfully
* But the model behaves as if no image/file was provided

---

### Expected Behavior

Given:

```python
task = Task(
    description="Describe the image",
    agent=agent,
    input_files={"image": ImageFile(...)}
)

crew = Crew(tasks=[task])
```

👉 The crew should:

* correctly use the `task.agent`
* process `input_files`
* pass the image to the model

---

### Actual Behavior

With the above setup:

* The task runs
* The model responds as if **no image is provided**
* No error is raised

But when adding:

```python
crew = Crew(
    agents=[agent],
    tasks=[task]
)
```

👉 The exact same code **works correctly**, and the model receives the image.

---

### Observations

* This issue appears specifically when using `input_files`
* It does **not** throw an error — fails silently
* It contradicts expected API usage where `task.agent` should be sufficient
* CrewAI's own multimodal tests always include `agents=[...]` at the Crew level

---

### Environment
(replicated on both windows 11 and ubuntu 22.04)

* CrewAI version: **1.14.2**
* Using:

  * `ImageFile`
  * OpenAI model (`gpt-5.4-mini`)
  * `api="responses"`

---

### Possible Cause (Hypothesis)

It seems that:

* `crew.kickoff(input_files=...)` attaches files through crew-level agent handling
* If `Crew.agents` is empty, file injection is skipped or not propagated to `task.agent`

---

### Suggested Fix

Either:

1. Ensure `task.agent` is sufficient for file propagation
2. Or raise a clear error if `input_files` are used without `Crew.agents`
3. Or document that `agents=[...]` is required when using multimodal inputs

---

### Impact

* Silent failures are dangerous
* Makes debugging very hard
* Contradicts intuitive API usage

### Minimal Reproduction

```python
from crewai import Crew, Agent, Task, LLM
from crewai_files import ImageFile

llm = LLM(model="openai/gpt-5.4-mini", api="responses")

agent = Agent(
    role="Assistant",
    goal="Describe images",
    backstory="Assistant",
    llm=llm,
    multimodal=True,
)

task = Task(
    description="Describe the provided image.",
    expected_output="Short description.",
    agent=agent,
    input_files={
        "image": ImageFile(source="path/to/image.jpg")
    }
)

# ❌ Does NOT work (image ignored)
crew = Crew(tasks=[task])

result = crew.kickoff()
print(result)
```

Now change to:

```python
# ✅ Works
crew = Crew(
    agents=[agent],
    tasks=[task]
)
```

### Expected behavior

to work the same way as when we add agents=[agent] in the crew arguments

### Screenshots/Code snippets

in the description above

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

1.14.2

### crewAI Tools Version

1.14.2

### Virtual Environment

Venv

### Evidence

above 

### Possible Solution

get the agents correctly

### Additional context

nothing extrra

greysonlalonde:
Closes #5534.

`prepare_kickoff` only ran `setup_agents` over `crew.agents`, so an agent referenced solely via `task.agent` never got `agent.crew` assigned — silently disabling the multimodal file lookup at `task.py:907` and `crew_agent_executor.py:242`. Task agents are now merged into the setup list.

---

> [!NOTE]
> **Medium Risk**
> Touches `prepare_kickoff` agent initialization, which affects how executors are created and how multimodal `input_files` are resolved at runtime; regression risk is limited but in a core execution path.
> 
> **Overview**
> Fixes a kickoff regression where agents assigned only via `Task.agent` were not passed through `setup_agents`, leaving `agent.crew` unset and preventing multimodal `input_files` from being found/used.
> 
> `prepare_kickoff` now builds a de-duplicated `agents_to_setup` list that merges `crew.agents` with any per-task agents, and a new test covers the scenario where `Crew(agents=[])` still binds the task-only agent to the crew.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c51cd622310719071870acafc9f17c7d653d3068. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/crews/utils.py
+++ b/lib/crewai/src/crewai/crews/utils.py
@@ -357,10 +357,9 @@ def prepare_kickoff(
     agents_to_setup: list[BaseAgent] = list(crew.agents)
     seen_agent_ids: set[int] = {id(agent) for agent in agents_to_setup}
     for task in crew.tasks:
-        task_agent = getattr(task, "agent", None)
-        if task_agent is not None and id(task_agent) not in seen_agent_ids:
-            agents_to_setup.append(task_agent)
-            seen_agent_ids.add(id(task_agent))
+        if task.agent is not None and id(task.agent) not in seen_agent_ids:
+            agents_to_setup.append(task.agent)
+            seen_agent_ids.add(id(task.agent))
 
     setup_agents(
         crew,
--- a/lib/crewai/src/crewai/crews/utils.py
+++ b/lib/crewai/src/crewai/crews/utils.py
@@ -354,9 +354,17 @@ def prepare_kickoff(
     crew._set_tasks_callbacks()
     crew._set_allow_crewai_trigger_context_for_first_task()
 
+    agents_to_setup: list[BaseAgent] = list(crew.agents)
+    seen_agent_ids: set[int] = {id(agent) for agent in agents_to_setup}
+    for task in crew.tasks:
+        task_agent = getattr(task, "agent", None)
+        if task_agent is not None and id(task_agent) not in seen_agent_ids:
+            agents_to_setup.append(task_agent)
+            seen_agent_ids.add(id(task_agent))
+
     setup_agents(
         crew,
-        crew.agents,
+        agents_to_setup,
         crew.embedder,
         crew.function_calling_llm,
         crew.step_callback,
--- a/lib/crewai/tests/test_crew.py
+++ b/lib/crewai/tests/test_crew.py
@@ -4798,6 +4798,37 @@ def test_crew_kickoff_started_emits_display_name(
     assert captured == [expected]
 
 
+def test_prepare_kickoff_binds_task_only_agent_to_crew():
+    """Agents referenced only via task.agent must get .crew set during prepare_kickoff.
+
+    Regression for crewAIInc/crewAI#5534: when Crew is built without
+    agents=[...], multimodal input_files were silently dropped because the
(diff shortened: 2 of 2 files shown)


iris-clawd (approved):
Solid fix. The root cause is clear — `setup_agents` only iterated `crew.agents`, so task-only agents never got `agent.crew` wired up, silently breaking multimodal file resolution.

The de-duplication via `id()` is the right call (identity, not equality) and the test nails the exact regression scenario.

One thought (non-blocking): if someone later adds agents to tasks that also live in `crew.agents` but are *different objects with the same config*, the `id()`-based dedup would correctly treat them as separate — that's the right behavior here. Just noting for future readers.

LGTM 👍

Issue #4905, opened 2026-03-16
Agents always load trained_agents_data.pkl, ignoring custom filename; custom training file not respected on inference


manoharnv:
### Description

When training a CrewAI crew with a custom file name (`-f <custom_file>.pkl`), the consolidated agent guidance is correctly written to the given path (e.g., `my_custom_trained.pkl`). However, during normal (non-training) execution, agents **always** load from the hardcoded `trained_agents_data.pkl`, ignoring the custom file supplied at training time.

---
**Relevant code:**
```python
    if data := CrewTrainingHandler(TRAINED_AGENTS_DATA_FILE).load():  # ❌ HARDCODED!
```
From `agent/core.py`

### Steps to Reproduce

1. Train a crew with a custom file:
   ```shell
   crewai train -n 5 -f my_custom_trained.pkl
   ```
2. Observe that `my_custom_trained.pkl` is created and populated after training.
3. Run the crew (or agents) for inference **without renaming** the file.
4. The agents will not use your custom file; only `trained_agents_data.pkl` is loaded.

### Expected behavior

Agents should be able to load and apply trained suggestions from the actual file used in training, not just `trained_agents_data.pkl`. Ideally, the system should either:
- Track and propagate the custom path used (e.g., as `Crew(..., trained_file="my_custom_trained.pkl")`), **or**
- Provide an API, environment variable, or config to specify the trained data file to read during inference, **or**
- At minimum, clearly warn users and suggest renaming the file post-training.

### Screenshots/Code snippets

```python
    if data := CrewTrainingHandler(TRAINED_AGENTS_DATA_FILE).load():  # ❌ HARDCODED!
```
**File:** `agent/core.py` (lines 3-3)

### Operating System

Other (specify in additional context)

### Python Version

3.12

### crewAI Version

N/A

### crewAI Tools Version

N/A

### Virtual Environment

Venv

### Evidence

Trained models saved to custom path are not respected in inference; always loads `trained_agents_data.pkl` by default. Only workaround is to rename the file manually.

### Possible Solution

- Accept a `trained_file` parameter when instantiating a Crew or agent so this path is respected everywhere.
- Fall back to `trained_agents_data.pkl` only when no custom filename is provided.
- Add documentation warnings if this is intentional

### Additional context

This is especially confusing for multi-experiment workflows, CI, cloud, or production runs where explicit filenames are essential.

manoharnv:
[...]

PR #5640, opened 2026-04-28
fix(agent): honor custom trained-agents file at inference

greysonlalonde:
Closes #4905.

`crewai train -f my.pkl` saved the trained data to the custom path, but `Agent._use_trained_data()` always loaded from the hardcoded `trained_agents_data.pkl`, so the custom file was effectively dead on arrival.

- `_use_trained_data` now reads `CREWAI_TRAINED_AGENTS_FILE`, falling back to `trained_agents_data.pkl`.
- `crewai run -f <path>` mirrors `crewai train -f` and exports the env var to the kickoff subprocess.
- The env var also lets `crewai replay` / `crewai test` / triggered runs pick up the custom file when exported in the shell — CLI flags for those are intentionally left for a follow-up.

---

> [!NOTE]
> **Low Risk**
> Low risk: small, additive change that only affects where trained-agent suggestions are loaded from and how the `crewai run` subprocess receives that path.
> 
> **Overview**
> Agents now load trained suggestions from a configurable file: `Agent._use_trained_data()` reads the `CREWAI_TRAINED_AGENTS_FILE` env var (fallback to `trained_agents_data.pkl`) instead of always using the default path.
> 
> The `crewai run` CLI gains `-f/--filename` to forward a custom trained-agents pickle into the uv subprocess by exporting `CREWAI_TRAINED_AGENTS_FILE`. Tests were added/updated to cover env-var honoring and CLI/subprocess plumbing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93cdbf1e470e4e5a54c3549fdca09d76cbec3f23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -8,6 +8,7 @@ import concurrent.futures
 import contextvars
 from datetime import datetime
 import json
+import os
 from pathlib import Path
 import time
 from typing import (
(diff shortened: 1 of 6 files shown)


iris-clawd (approved):
Good catch — the custom trained-agents file was effectively write-only before this. Clean fix.

Minor nit (non-blocking): `test_run_crew.py` is missing a trailing newline at EOF.

Everything else looks right: env var as the bridge between CLI and agent runtime avoids plumbing the path through the entire call stack, constant is centralized, tests cover all three layers (agent, CLI, subprocess). LGTM 👍

PR #5641, opened 2026-04-28
fix(cli): forward trained-agents file through replay and test

greysonlalonde:
Follow-up to #5640. The first PR plumbed `CREWAI_TRAINED_AGENTS_FILE` through `crewai run`; `crewai replay` and `crewai test` still launched their subprocesses with no env override, so users on those paths had to export the var manually.

- `crewai replay -f <pkl>` and `crewai test -f <pkl>` mirror `crewai run -f` / `crewai train -f`.
- `replay_task_command` and `evaluate_crew` now build the subprocess env from `os.environ` and inject `CREWAI_TRAINED_AGENTS_FILE` when a path is supplied.
- Triggers (`crewai triggers run`) intentionally left alone — its CLI shape is different and the env var still works as an escape hatch.

---

> [!NOTE]
> **Low Risk**
> Low risk: small CLI plumbing change that only affects environment variables passed to `uv run` for `replay`/`test`, with tests covering the new behavior.
> 
> **Overview**
> `crewai replay` and `crewai test` now accept `-f/--filename` to point at a specific trained-agents pickle, mirroring the existing `run/train` behavior.
> 
> Both `replay_task_command` and `evaluate_crew` now build an explicit subprocess environment (via `build_env_with_all_tool_credentials`) and inject `CREWAI_TRAINED_AGENTS_FILE` when a filename is provided.
> 
> Tests were updated/added to assert the new CLI option is forwarded and that the env var is set/omitted appropriately.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3335e5b49ca5f3540520bd93fb43835761d73ce2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/cli/cli.py
+++ b/lib/crewai/src/crewai/cli/cli.py
@@ -139,16 +139,29 @@ def train(n_iterations: int, filename: str) -> None:
     type=str,
     help="Replay the crew from this task ID, including all subsequent tasks.",
 )
-def replay(task_id: str) -> None:
-    """
-    Replay the crew execution from a specific task.
+@click.option(
+    "-f",
+    "--filename",
+    "trained_agents_file",
+    type=str,
+    default=None,
+    help=(
+        "Path to a trained-agents pickle (produced by `crewai train -f`). "
+        "When set, agents load suggestions from this file instead of the "
+        "default trained_agents_data.pkl. Equivalent to setting "
+        "CREWAI_TRAINED_AGENTS_FILE."
+    ),
+)
+def replay(task_id: str, trained_agents_file: str | None) -> None:
+    """Replay the crew execution from a specific task.
 
     Args:
-        task_id (str): The ID of the task to replay from.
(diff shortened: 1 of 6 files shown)


iris-clawd (approved):
Clean follow-up — consistent `-f` flag across all CLI subcommands now. Pattern matches #5640 exactly.

One note: `evaluate_crew` uses `os.environ.copy()` while `run_crew` uses `build_env_with_all_tool_credentials()` — intentional difference or worth aligning? Non-blocking either way since `os.environ.copy()` is the safer default (inherits everything).

Same trailing-newline nit on `test_replay_from_task.py`. Otherwise LGTM 👍

iris-clawd (approved):
Looks good — env builder is now consistent across all three subcommands (`build_env_with_all_tool_credentials` everywhere). Still the trailing newline on `test_replay_from_task.py` but that's trivial. LGTM 👍

PR #5638, opened 2026-04-28
fix: serialize guardrail callables as null for JSON checkpointing

greysonlalonde:
Fixes #5620. Extends the #5559 pattern to `Task.guardrail`, `Task.guardrails`, `Agent.guardrail`, `LiteAgent.guardrail`.

Vs #5632:
- Shared helper in `crewai.utilities.guardrail` instead of three inline lambdas.
- Warns on dropped callables so the lossy restore isn't silent.
- Adds tests.

`uv.lock` bump re-syncs `exclude-newer` with `pyproject.toml`.

---

> [!NOTE]
> **Low Risk**
> Scoped to Pydantic JSON serialization for checkpointing; behavior change is limited to dropping non-serializable callable guardrails (now with explicit warnings).
> 
> **Overview**
> Fixes JSON checkpointing for `guardrail`/`guardrails` fields by adding shared serializers that **preserve string guardrails** but **serialize callable guardrails as `null` (and filter them out of lists)**, emitting a warning so the loss isn’t silent.
> 
> Applies this serializer to `Agent.guardrail`, `LiteAgent.guardrail`, `Task.guardrail`, and `Task.guardrails`, and adds regression tests to ensure `model_dump(mode="json")` no longer raises and serialized data round-trips through `model_validate`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b0d906ed913ac1547860e328b795532336c821ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #5650, opened 2026-04-29
fix(cli): guard crew chat description helpers against LLM failures

### LINKED ISSUE #5510: ChatWithCrewFlow.__init__ makes blocking LLM call at module import, crashes containers on any LLM hiccup
## Summary

`ChatWithCrewFlow.__init__` in `ag_ui_crewai.crews` triggers synchronous blocking LLM calls at module import time via `crewai.cli.crew_chat.generate_crew_chat_inputs`, which in turn calls:

- `generate_input_description_with_ai` — [`lib/crewai/src/crewai/cli/crew_chat.py:481`](https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/src/crewai/cli/crew_chat.py#L481)
  ```python
  response = chat_llm.call(messages=[{"role": "user", "content": prompt}])
  ```
- `generate_crew_description_with_ai` — [`lib/crewai/src/crewai/cli/crew_chat.py:535`](https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/src/crewai/cli/crew_chat.py#L535)
  ```python
  response = chat_llm.call(messages=[{"role": "user", "content": prompt}])
  ```

For users deploying CrewAI behind a FastAPI server via `ag_ui_crewai.endpoint.add_crewai_crew_fastapi_endpoint` (the recommended integration for AG-UI / CopilotKit), these LLM calls fire during module import — BEFORE `uvicorn` binds to its HTTP port.

## Failure mode

ANY LLM provider hiccup during container startup causes the Python process to crash before the HTTP server is listening:

- OpenAI 500 / 503 / rate-limit
- Network blip, DNS failure, slow cold-start on a mock/proxy server
- Invalid credentials (even transient)
- Litellm `APIError`, `Timeout`, or `APIConnectionError`

In orchestrated environments (Railway, Kubernetes, AWS ECS, Fly.io) the platform's readiness/health check fails because no process ever binds the port. The platform then marks the deploy failed and rolls back to the previous image, making the service effectively unresponsive to LLM-layer instability.

We hit this on our Railway-hosted CopilotKit showcase when our LLM mock (aimock) returned a transient schema error. The mock error was recoverable — the issue is that it shouldn't have been able to crash the entire container before the HTTP server was ready.

### Actual stack trace we observed

```
File "/app/agent_server.py", line 27, in <module>
    add_crewai_crew_fastapi_endpoint(app, LatestAiDevelopment(), "/")
File ".../ag_ui_crewai/endpoint.py", line 250, in add_crewai_crew_fastapi_endpoint
    add_crewai_flow_fastapi_endpoint(app, ChatWithCrewFlow(crew=crew), path)
File ".../ag_ui_crewai/crews.py", line 56, in __init__
    self.crew_chat_inputs = crew_chat_generate_crew_chat_inputs(...)
File ".../crewai/cli/crew_chat.py", line 387, in generate_crew_chat_inputs
    description = generate_input_description_with_ai(input_name, crew, chat_llm)
File ".../crewai/cli/crew_chat.py", line 481, in generate_input_description_with_ai
    response = chat_llm.call(messages=[...])
File ".../crewai/llm.py", line 956, in call
    return self._handle_non_streaming_response(...)
...
APIError: <connection failure>
```

Container exits with code 1, never binds a port, orchestrator's health check fails, deploy rolls back.

## Why this is a CrewAI concern, not just an ag-ui-crewai concern

While `ChatWithCrewFlow` lives in `ag-ui-crewai`, the two functions that block are part of CrewAI's public `crewai.cli.crew_chat` module. CrewAI is asking users to consume these helpers at import/init time without any of the standard production-server defenses:

- No timeout
- No retry/fallback
- No try/except with a graceful default
- No opt-out

Any consumer that instantiates a chat flow with them in a serving context inherits this fragility.

## Suggested fixes (any or all)

1. **Lazy init at first request.** Have `ChatWithCrewFlow.__init__` store the crew and LLM but defer `generate_crew_chat_inputs` until the first actual chat turn. (A similar fix has already landed on `ag-ui-protocol/ag-ui` main for `add_crewai_crew_fastapi_endpoint` — deferring `ChatWithCrewFlow` construction to first-request. But the underlying functions in CrewAI still have no defenses.)

2. **Try/except with a static fallback inside the generator functions.** If the LLM call fails for any reason, fall back to a generic string like `"Input value for the crew's tasks and agents."` or `"A CrewAI crew."`. These descriptions are only surfaced in the CrewAI chat UI — shipping a generic default on LLM failure is strictly better than crashing the process.

3. **Make AI-generated descriptions opt-in.** Accept a kwarg `generate_descriptions: bool = True` (default preserves current behavior), but let production users pass `False` to skip the LLM calls entirely.

4. **Timeout + bounded retry.** At minimum, enforce a short timeout (e.g., 10s) on `chat_llm.call` in these two functions so a hung LLM can't indefinitely block process startup.

## Our workaround

We're shipping a defensive monkey-patch in our showcase to patch both functions to return static strings before `ag_ui_crewai` is imported. PR: https://github.com/CopilotKit/CopilotKit/pull/3974

This is fragile (depends on private function names) and we'd much prefer an upstream fix so every AG-UI / CopilotKit / direct-CrewAI production deployment doesn't inherit this footgun.

## Environment

- `crewai>=0.130.0`
- `ag-ui-crewai==0.1.5` (latest released; `main` already has the deferred-construction fix in `endpoint.py` but no release yet)
- Python 3.12

greysonlalonde:
Closes #5510.

`ChatWithCrewFlow.__init__` (in ag-ui-crewai) calls `generate_crew_chat_inputs` at module import, which makes blocking LLM calls via `generate_input_description_with_ai` and `generate_crew_description_with_ai`. With no try/except or opt-out, any LLM hiccup at startup crashed the container before uvicorn bound.

- `generate_crew_chat_inputs` accepts `generate_descriptions: bool = True`; pass `False` to skip both AI calls and use static defaults.
- Both `generate_*_with_ai` helpers catch `chat_llm.call` failures, warn, and return a static default.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit c1eee2fc21abeff7062e7177fbc3fb1f2ba511fa. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/cli/crew_chat.py
+++ b/lib/crewai/src/crewai/cli/crew_chat.py
@@ -25,6 +25,9 @@ from crewai.utilities.version import get_crewai_version
 
 MIN_REQUIRED_VERSION: Final[Literal["0.98.0"]] = "0.98.0"
 
+DEFAULT_INPUT_DESCRIPTION: Final[str] = "Input value for the crew's tasks and agents."
+DEFAULT_CREW_DESCRIPTION: Final[str] = "A CrewAI crew."
+
 
 def check_conversational_crews_version(
     crewai_version: str, pyproject_data: dict[str, Any]
@@ -381,7 +384,10 @@ def load_crew_and_name() -> tuple[Crew, str]:
 
 
 def generate_crew_chat_inputs(
-    crew: Crew, crew_name: str, chat_llm: LLM | BaseLLM
+    crew: Crew,
+    crew_name: str,
+    chat_llm: LLM | BaseLLM,
+    generate_descriptions: bool = True,
 ) -> ChatInputs:
     """
     Generates the ChatInputs required for the crew by analyzing the tasks and agents.
@@ -390,21 +396,28 @@ def generate_crew_chat_inputs(
         crew (Crew): The crew object containing tasks and agents.
         crew_name (str): The name of the crew.
         chat_llm: The chat language model to use for AI calls.
+        generate_descriptions: When True (default), use the LLM to generate
+            input and crew descriptions. When False, skip all LLM calls and
+            return static defaults. Production callers that invoke this at
+            startup should pass ``False`` to avoid blocking on the LLM.
 
     Returns:
         ChatInputs: An object containing the crew's name, description, and input fields.
     """
-    # Extract placeholders from tasks and agents
     required_inputs = fetch_required_inputs(crew)
 
-    # Generate descriptions for each input using AI
     input_fields = []
     for input_name in required_inputs:
-        description = generate_input_description_with_ai(input_name, crew, chat_llm)
+        if generate_descriptions:
+            description = generate_input_description_with_ai(input_name, crew, chat_llm)
+        else:
+            description = DEFAULT_INPUT_DESCRIPTION
         input_fields.append(ChatInputField(name=input_name, description=description))
 
-    # Generate crew description using AI
-    crew_description = generate_crew_description_with_ai(crew, chat_llm)
+    if generate_descriptions:
+        crew_description = generate_crew_description_with_ai(crew, chat_llm)
+    else:
+        crew_description = DEFAULT_CREW_DESCRIPTION
 
     return ChatInputs(
         crew_name=crew_name, crew_description=crew_description, inputs=input_fields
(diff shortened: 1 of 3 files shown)


iris-clawd (approved):
Good fix — blocking LLM calls at import time with no error handling is a classic way to brick containers on startup.

The dual approach (opt-out flag + catch in the helpers themselves) gives callers both a fast path (`generate_descriptions=False`) and resilience even when they do want AI descriptions. Tests cover all the permutations.

One minor nit: missing trailing newline on `test_crew_chat.py`. And the `exclude-newer` went backwards again in `uv.lock` — same as #5638.

LGTM 👍

PR #5654, opened 2026-04-29
feat(crewai-tools): add highlights to ExaSearchTool, rename from EXASearchTool

theishangoswami:
## Summary

This updates `ExaSearchTool` to match how Exa now recommends using the API:

- **Adds a `highlights` init param** so agents can get small, query-relevant excerpts instead of full pages. Highlights use ~10x fewer tokens than full text and are now the recommended option in [Exa's search best practices](https://exa.ai/docs/reference/search-best-practices).
- **Renames `EXASearchTool` → `ExaSearchTool`** (Exa is a brand, not an acronym). The old name still works as a deprecated alias, so existing code does not break — it just emits a one-time `DeprecationWarning`.
- Updates the docs page and example to lead with `highlights`.

## What changed

- `lib/crewai-tools/src/crewai_tools/tools/exa_tools/exa_search_tool.py`: new `highlights` param, `ExaSearchTool` class + back-compat `EXASearchTool` alias, fixed `_run` to use `search_and_contents` when needed.
- `lib/crewai-tools/tests/tools/exa_search_tool_test.py`: tests for the new highlights param, plus a test that the deprecated `EXASearchTool` alias still works and emits a `DeprecationWarning`.
- `lib/crewai-tools/src/crewai_tools/__init__.py` + `tools/__init__.py`: export the new `ExaSearchTool` name (alongside the existing `EXASearchTool`).
- `docs/en/tools/search-research/exasearchtool.mdx`: rewritten to lead with highlights and link to Exa search best practices.
- `docs/{ar,ko,pt-BR}/tools/search-research/exasearchtool.mdx`: renamed `EXASearchTool` → `ExaSearchTool` references. Translations of the new prose can be regenerated by maintainers.
- `docs/en/concepts/tools.mdx` (+ ar/ko/pt-BR copies): updated the one-line description.
- `lib/crewai-tools/README.md` and `tools/exa_tools/README.md`: name + description refresh.
- `lib/crewai/src/crewai/cli/templates/AGENTS.md`: updated the built-in tools list to use the new name.

## Backwards compatibility

`from crewai_tools import EXASearchTool` keeps working. Calling `EXASearchTool(...)` returns an `ExaSearchTool` instance and emits a `DeprecationWarning`. No existing user code needs to change.

## Note for maintainers

`tool.specs.json` is auto-generated by `.github/workflows/generate-tool-specs.yml`, which only runs for in-repo PRs. After merge, the workflow will regenerate the spec automatically; alternatively it can be triggered via `workflow_dispatch`.


lib/crewai-tools/src/crewai_tools/tools/exa_tools/exa_search_tool.py
```diff
@@ -31,14 +32,21 @@ class EXABaseToolSchema(BaseModel):
     )
 
 
-class EXASearchTool(BaseTool):
+# Backwards-compat alias for the schema (renamed alongside the tool class).
```
greysonlalonde: Can you remove this comment? It's self-explanatory

greysonlalonde (changes requested):

lib/crewai-tools/src/crewai_tools/tools/exa_tools/exa_search_tool.py
theishangoswami: hey this is fixed ✅

lorenzejay (approved):

greysonlalonde (approved):

Sourav-Nandy-ai:
Hi Ishan,

I am flagging something a little different than a standard bug report. I checked whether agent-eval, a tool I am building for statistically comparing agent behavior across versions, could actually handle a PR like #5654. It already could, without me changing anything.

Your highlights change is a genuine default-behavior shift. A default-constructed ExaSearchTool() now returns richer, highlight-annotated results instead of bare search hits because it internally switches from search() to search_and_contents(highlights=True, ...).

Since you kept highlights as a constructor kwarg for backward compatibility, agent-eval can construct the pre-PR and post-PR behavior side-by-side in the same installed package. It can then run both through a crew 50 times each (which is the actual default in compare(), not just a rounded number for a pitch). This provides a Mann-Whitney U, Cohen's d, and a bootstrap CI on whether the distribution really moved, rather than just eyeballing a few runs.

I ran the underlying stats machinery on a modeled score distribution to confirm it correctly flags a real shift as significant. This was not tested against your live Exa API, as I do not have a key available on my end. Therefore, that specific number is not a claim about your PR's actual output, but rather proof that the harness's math does what it is supposed to when a real difference exists.

Given you own both sides of this integration, I would value your take on whether comparing the two ExaSearchTool configurations this way actually matches how you would want to catch a regression here.

You can find the repository here: https://github.com/RudrenduPaul/agent-eval

Best,
Sourav

PR #5675, opened 2026-05-01
fix(devtools): cover missing crewai pin sites in release flow

greysonlalonde:
> [!NOTE]
> **Medium Risk**
> Touches automated release tooling that edits dependencies and workflow files across repos; mistakes could cause incorrect pins or incomplete commits during a release.
> 
> **Overview**
> Ensures the release automation updates *all* `crewai` version pins, not just `pyproject.toml` entries.
> 
> Enterprise pinning now also updates dashed `crewai-*` packages listed in `ENTERPRISE_EXTRA_PACKAGES`, and workflow updates are refactored into a reusable `_update_workflow_crewai_pins` helper. The deployment-test repo update step now also rewrites any `crewai[...]` pins in `.github/workflows`, and only runs `uv lock`/stages `pyproject.toml`+`uv.lock` when the pyproject pin actually changed, while still committing any updated workflow files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8d716d78aff809952d5b501589e462f344bec9cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


iris-clawd (approved):
Looks good — clean refactoring. The extracted `_update_workflow_crewai_pins` helper is a nice DRY win, and scanning all workflow files in the deployment-test repo (vs. just configured paths for enterprise) makes sense to catch stray pin sites.

**One thing to consider** (non-blocking): in `_update_deployment_test_repo`, the early return at `if new_content == content: return` means if the pyproject.toml pin is already correct but a workflow file still has a stale pin, the workflow scan is never reached. Might be worth removing that early return or restructuring so the workflow scan always runs — but it's an edge case and doesn't block this.

All CI green ✅

iris-clawd (approved):
Re-reviewed after updates — the early-return issue I flagged is fixed cleanly. The `_update_deployment_test_repo` flow now correctly:

- Tracks `pyproject_changed` separately and always runs the workflow scan
- Only skips the commit/PR when nothing changed at all
- Conditionally runs `uv lock` only when pyproject was modified
- Builds `paths_to_add` dynamically based on what actually changed

LGTM ✅

PR #5683, opened 2026-05-02
docs: add security tip for API key management

irfaan101:
Added a small security tip to the quickstart guide using a <Tip> component. Recommended using environment variables (.env) for API keys to improve security for beginners.


greysonlalonde:
This is largely self-explained

PR #5687, opened 2026-05-03
docs: add security tip for API key management in Arabic translation

irfaan101:
Added a security tip in the Arabic documentation to advise users on safe API key management using .env files, ensuring consistency with the English version.
Note: Used dir="rtl" within the <Tip> component to ensure correct rendering of mixed Arabic and English text for a better user experience.


irfaan101:
I understand that for experienced developers, API key management is self-explanatory. However, for many new users—especially those in the Arabic-speaking community who are just starting with CrewAI—security best practices like using .env files are not always obvious. My goal is to ensure that even beginners can use the framework safely from day one. I would appreciate it if you could reconsider this for the sake of better security education for newcomers.

PR #5692, opened 2026-05-04
feat: tag runtime tool instances with private identity metadata

renatonitta:
### Overview

  Tags runtime tool instances with private metadata so downstream tooling (specifically CrewAI Enterprise) can recover their canonical `tool_id` for ACP rule evaluation.
  - **MCP tools** (`lib/crewai/`) get an `_amp_slug` set by `MCPToolResolver._resolve_amp` — covers [CON-139].
  - **Platform action tools** (`lib/crewai-tools/`) get `_provider` / `_provider_id` set by `CrewaiPlatformToolBuilder` — covers [CON-140] (Python side).

  Both attributes are private, default `None`, and have no impact on existing users. The information they carry is consumed by the enterprise `/inspect.tools` resolver to emit canonical IDs
  like `crewai_oauth:notion-244f20d5|mcp` or `paragon:<uuid>`.

  ### Main Technical Changes

  **`lib/crewai/` — CON-139:**
  - `tools/mcp_tool_wrapper.py`, `tools/mcp_native_tool.py` — new private `_amp_slug: str | None = None` attribute on each wrapper class.
  - `mcp/tool_resolver.py::_resolve_amp` — after wrappers are produced for a given AMP slug, sets `tool._amp_slug = slug` on each (with `# type: ignore[attr-defined]` matching the repo's existing pattern for dynamic attribute assignment on BaseTool subclasses).
  - New tests under `tests/mcp/test_tool_resolver_amp_slug.py` covering: defaults, slug propagation through `_resolve_amp`, missing-config no-op, and untagged behavior on `_resolve_native` / `_resolve_external` paths.

  **`lib/crewai-tools/` — CON-140 (Python):**
  - `crewai_platform_tools/crewai_platform_action_tool.py` — `__init__` accepts `provider: str | None = None` and `provider_id: str | None = None`, stores as private attrs.
  - `crewai_platform_tools/crewai_platform_tool_builder.py` — reads the new `provider` field from `/integrations/actions` response, passes both through. Outer key (`app`) → `provider_id`.
  - New tests under `tests/tools/crewai_platform_tools/test_provider_tagging.py` covering: defaults, full threading from response, and backward-compat when the response lacks the `provider` field.

  ### Notes for reviewers

  - Public API change is minimal: two new optional kwargs on `CrewAIPlatformActionTool.__init__`, both default `None`.
  - The crewai-tools side tolerates an older crewai-plus that doesn't yet return `provider` — `_provider` stays `None` in that case. So this PR can land before or after the matching crewai-plus PR ([CON-140] server side) without breaking anything.

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/tools/crewai_platform_tools/crewai_platform_action_tool.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/crewai_platform_tools/crewai_platform_action_tool.py
@@ -26,6 +26,8 @@ class CrewAIPlatformActionTool(BaseTool):
         description: str,
         action_name: str,
         action_schema: dict[str, Any],
+        provider: str | None = None,
+        provider_id: str | None = None,
     ):
         parameters = action_schema.get("function", {}).get("parameters", {})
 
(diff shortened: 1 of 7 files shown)


renatonitta:
Closing this in favor of another solution.

PR #5702, opened 2026-05-04
fix: async-safe embeddings and resilient drain_writes

MatthiasHowellYopp:
Description:

Part 3/4 of adding Valkey as a storage backend for CrewAI. This PR makes the embedding and memory persistence paths robust enough to work with async storage backends like Valkey.

What changed:

types.py — Added a field_validator on MemoryRecord.embedding that converts bytes to list[float] via numpy. Valkey stores vectors as raw bytes, so this ensures embeddings are always in the expected format regardless of storage backend. Also added a thread pool to embed_texts() so it doesn't block the event loop when called from an async context.

encoding_flow.py — Added a matching field_validator on ItemState.similar_records and result_record to handle the same bytes→float conversion during the consolidation flow.

unified_memory.py — drain_writes() now accepts a timeout_per_save parameter (default 60s) and logs warnings on timeout or failure instead of raising. This prevents a single slow or failed save from blocking crew completion. Added structured debug/warning/error logging throughout the drain cycle.

Testing:

test_embedding_safety.py (15 tests) — Covers bytes→float conversion, empty bytes, numpy arrays, int-to-float coercion, sync and async embed_texts behavior, empty/whitespace input handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * More robust embedding normalization (bytes, numeric arrays, empty values handled consistently).
  * Embedding calls are safe in async contexts and preserve positions for skipped/empty inputs.
  * Background memory saves now use per-save timeouts with progress logging and no indefinite blocking (defaults to 60s).

* **Tests**
  * Added comprehensive tests covering embedding validation, sync/async embedding behavior, and timeout/save handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5702)

### DIFF
--- a/lib/crewai/src/crewai/memory/encoding_flow.py
+++ b/lib/crewai/src/crewai/memory/encoding_flow.py
@@ -18,7 +18,7 @@ import math
 from typing import Any
 from uuid import uuid4
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, field_validator
 
 from crewai.flow.flow import Flow, listen, start
 from crewai.memory.analyze import (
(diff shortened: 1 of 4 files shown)


MatthiasHowellYopp:
@greysonlalonde hoping I can get a review on this.

PR #5711, opened 2026-05-05
fix(security): validate IPs on every redirect hop to prevent SSRF bypass

iris-clawd:
## Summary

Fixes **OSS-51** — SSRF bypass via redirect following in scraping tools.

### Problem

`validate_url()` in `crewai_tools/security/safe_path.py` only checked the **initial** URL before `requests.get()` fired. Since `requests.get()` follows redirects by default, an attacker could host a public URL that 302-redirects to an internal IP (e.g. `169.254.169.254` for cloud metadata) — the validator never saw the redirect target.

This was reported by Casco Security and confirmed end-to-end on a live `app.crewai.com` deployment worker (CVE-2026-2286 follow-up).

### Fix

Added a custom `HTTPAdapter` (`_SSRFSafeAdapter`) that intercepts **every** request — including redirect hops — and validates the resolved IP against the private/reserved blocklist before the connection proceeds.

**New public API:**
- `safe_request_session()` — returns a `requests.Session` with the SSRF-safe adapter mounted
- `safe_get(url, **kwargs)` — drop-in replacement for `requests.get()` that validates both the initial URL and every redirect destination

**Updated tools:**
- `ScrapeWebsiteTool` — now uses `safe_get()`
- `ScrapeElementFromWebsiteTool` — now uses `safe_get()`
- `WebPageLoader` (RAG) — now uses `safe_get()`

### Tests

5 new tests in `test_safe_path.py`:
- ✅ Public URL allowed through
- ✅ Redirect to localhost (127.0.0.1) blocked
- ✅ Redirect to cloud metadata (169.254.169.254) blocked
- ✅ Redirect to private range (10.0.0.1) blocked
- ✅ Escape hatch (`CREWAI_TOOLS_ALLOW_UNSAFE_PATHS=true`) bypasses redirect check

All 32 tests pass (27 existing + 5 new).

### Note

Other tools that use third-party scraping services (Firecrawl, Spider, Scrapfly, etc.) are not affected — they delegate fetching to external APIs rather than using `requests.get()` directly. The RAG PDF/docs loaders (`pdf_loader.py`, `utils.py`) also use bare `requests.get()` and could benefit from the same treatment in a follow-up.

Closes OSS-51

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/rag/loaders/webpage_loader.py
+++ b/lib/crewai-tools/src/crewai_tools/rag/loaders/webpage_loader.py
@@ -2,9 +2,8 @@ import re
 from typing import Any, Final
 
 from bs4 import BeautifulSoup
-import requests
-
 from crewai_tools.rag.base_loader import BaseLoader, LoaderResult
+from crewai_tools.security.safe_path import safe_get
 from crewai_tools.rag.source_content import SourceContent
 
 
@@ -25,7 +24,7 @@ class WebPageLoader(BaseLoader):
         )
 
         try:
-            response = requests.get(url, timeout=15, headers=headers)
+            response = safe_get(url, timeout=15, headers=headers)
             response.encoding = response.apparent_encoding
 
             soup = BeautifulSoup(response.text, "html.parser")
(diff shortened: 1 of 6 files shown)


Vidit-Ostwal:
Closing this PR as part of a repository cleanup. If you believe this change is still needed, please open an issue and reference this PR.

PR #5717, opened 2026-05-05
fix(anthropic): warn when stop_reason='max_tokens' silently truncates output

### LINKED ISSUE #5148: AnthropicCompletion._handle_completion silently discards stop_reason; no way to detect truncation via hooks or events
### Feature Area

Other (please specify in additional context)

NA

The stop_reason field on Anthropic's Message response object is already accessible within _handle_completion() and _handle_tool_use_conversation() immediately after the API call — but is never read. Both methods extract token usage from the response and then discard the object, so the truncation signal is permanently lost before it can reach any hook, callback, or event subscriber.

The minimal fix is a logging.warning() call at two points where the raw response is still in scope:

**Location 1 — _handle_completion(), after _track_token_usage_internal(usage):**

```
if getattr(response, "stop_reason", None) == "max_tokens":
    agent_hint = f" [{from_agent.role}]" if from_agent else ""
    logging.warning(
        f"Truncated response{agent_hint}: stop_reason='max_tokens'. "
        f"Consider increasing max_tokens (current: {self.max_tokens})."
    )
```

**Location 2 — _handle_tool_use_conversation(), after _track_token_usage_internal(follow_up_usage):**

```
if getattr(final_response, "stop_reason", None) == "max_tokens":
    agent_hint = f" [{from_agent.role}]" if from_agent else ""
    logging.warning(
        f"Truncated response{agent_hint}: stop_reason='max_tokens'. "
        f"Consider increasing max_tokens (current: {self.max_tokens})."
    )
```

Location 2 is the more critical path — it handles the final synthesis response after all tool calls complete, which is where silent truncation caused significant downstream data corruption in our use case.

A more complete solution would also add stop_reason: str | None as a field on LLMCallCompletedEvent, allowing downstream subscribers (hooks, event listeners) to react to truncation programmatically rather than only through log monitoring.

The same fix should be applied to the async counterparts: _ahandle_completion() and _ahandle_tool_use_conversation().

1. **after_llm_call hook with token ratio heuristic:** The hook receives context.agent.role and context.llm._token_usage, so one could warn when completion_tokens ≥ max_tokens × 0.95. This works as a proxy but is not the same as checking stop_reason directly — it can produce false positives and misses cases where max_tokens is not explicitly set.
2. **Monkey-patching self.client.messages.create:** Wrapping the Anthropic SDK client on each AnthropicCompletion instance post-init to intercept the raw response. This works today without any framework changes but is fragile to internal refactoring and not a suitable permanent solution.
3. **Content sentinel / downstream heuristics:** Checking whether the response ends mid-sentence or lacks expected structural markers. Too unreliable for production use — the failure mode (silent truncation mid-table) produces syntactically valid but semantically incomplete output that passes all surface checks.

### Additional context

LLM provider / response observability

re: Willingness to Contribute (below):
I'm happy to submit a pull request for the minimal fix (logging warning at both sync locations). Happy to also include the async paths and/or the LLMCallCompletedEvent field addition if that's the preferred direction — just let me know in the issue before I open it.
However, this would be my first crewai pull request.

### Willingness to Contribute

Beandon13:
## Summary
- `AnthropicCompletion` previously discarded `response.stop_reason` after extracting token usage, so callers had no way to detect that a Claude response was cut off by `max_tokens`. The follow-up call inside `_handle_tool_use_conversation` is the most painful site because the truncated synthesis answer is what downstream code parses.
- This PR adds a small module-level helper, `_warn_on_max_tokens_truncation`, and calls it at all four completion sites (sync/async × plain completion / tool-use follow-up). When `stop_reason == "max_tokens"` we emit `logging.warning(...)` including the agent role (when available) and the configured `max_tokens` so users can act on it.
- Behavior is otherwise unchanged. No new public API; no extra LLM calls; no events touched.

Closes #5148

## Testing
```
cd lib/crewai
uv pip install -e ".[anthropic]"
uv pip install pytest pytest-asyncio pytest-xdist pytest-timeout pytest-recording vcrpy
python -m pytest \
  tests/llms/anthropic/test_anthropic.py::test_warn_on_max_tokens_truncation_emits_warning \
  tests/llms/anthropic/test_anthropic.py::test_warn_on_max_tokens_truncation_silent_for_normal_stop \
  tests/llms/anthropic/test_anthropic.py::test_warn_on_max_tokens_truncation_includes_agent_role \
  tests/llms/anthropic/test_anthropic.py::test_handle_completion_logs_truncation_warning
```
<output>
```
============================== 4 passed in 7.31s ===============================
```
</output>

Existing token-usage tests in the same file (`test_anthropic_token_usage_tracking`, `test_anthropic_cache_creation_tokens_extraction`, `test_anthropic_missing_cache_fields_default_to_zero`, `test_anthropic_completion_is_used_when_anthropic_provider`) also pass unchanged.

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -124,6 +124,44 @@ def _contains_file_id_reference(messages: list[dict[str, Any]]) -> bool:
     return False
 
 
+def _warn_on_max_tokens_truncation(
+    response: Any,
+    max_tokens: int,
+    from_agent: Any | None = None,
+    *,
+    context: str = "",
+) -> None:
+    """Emit a warning if the Anthropic response was truncated by ``max_tokens``.
+
+    Anthropic returns a ``stop_reason`` field on every Message response
+    indicating why generation stopped. ``"max_tokens"`` means the output was
+    cut off before the model finished, which can corrupt downstream parsing
+    (especially for the final synthesis response after tool use).
+
+    The Anthropic SDK exposes this via ``response.stop_reason``. We surface
+    the signal as a ``logging.warning`` so users can detect truncation
+    without wiring up event subscribers.
+
+    Args:
+        response: The Anthropic Message-like response object.
+        max_tokens: The configured max_tokens value (for the actionable hint).
+        from_agent: Optional agent reference for log context.
+        context: Optional short label (e.g. ``"tool conversation"``) included
+            in the log message to disambiguate truncation sites.
+    """
+    if response is None:
+        return
+    if getattr(response, "stop_reason", None) != "max_tokens":
+        return
+    role = getattr(from_agent, "role", None)
+    agent_hint = f" [{role}]" if role else ""
+    location = f" ({context})" if context else ""
+    logging.warning(
+        f"Truncated response{agent_hint}{location}: stop_reason='max_tokens'. "
+        f"Consider increasing max_tokens (current: {max_tokens})."
+    )
+
+
 class AnthropicThinkingConfig(BaseModel):
     type: Literal["enabled", "disabled"]
     budget_tokens: int | None = None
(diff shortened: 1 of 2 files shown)


greysonlalonde:
https://github.com/crewAIInc/crewAI/pull/5718#issuecomment-4398052386

PR #5746, opened 2026-05-07
fix(ci): stamp and pin all workspace packages in nightly publish

greysonlalonde:
## Summary
- Stamp `crewai-cli` and `crewai-core` (previously skipped) along with `crewai`, `crewai-tools`, `crewai-files` so the nightly job no longer collides with stable `1.14.5a3` on PyPI.
- Rewrite every cross-package pin (`crewai`, `crewai-core`, `crewai-cli`, `crewai-tools`, `crewai-files`) so installed nightlies don't mix with stable transitives.
- Compute `NIGHTLY` once from a single source instead of relying on the loop's last-iteration variable.
- Pin `crewai-files==1.14.5a3` in the `file-processing` extra so it gets rewritten with the rest.
- Align `setup-uv` to `@v6` / `0.11.3` across build and publish, drop the redundant `setup-python`.
- Widen the recent-commit check to 25h to absorb cron-vs-commit boundary skew.

---

> [!NOTE]
> **Medium Risk**
> Changes the nightly build/publish workflow to rewrite versions and dependency pins across multiple packages, which can affect what gets built and published to PyPI. Risk is mainly CI/release correctness (version collisions or mixed dependency installs) rather than runtime behavior.
> 
> **Overview**
> Nightly CI now stamps a single computed `${BASE}.dev${DATE}` version across *all* workspace packages (including `crewai-core` and `crewai-cli`) and rewrites cross-package dependency pins so nightly installs don’t mix stable and nightly transitives.
> 
> The workflow also widens the “recent commits” window to 25 hours to reduce cron boundary misses, and standardizes on `astral-sh/setup-uv@v6` (`uv` `0.11.3`) while removing the separate `setup-python` step.
> 
> Additionally, `lib/crewai`’s `file-processing` extra pins `crewai-files==1.14.5a3` so it can be rewritten along with the other nightly dependency pins.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b02fb4e6a167147799e883a246205a5e0dc7791e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -18,10 +18,11 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Check for commits in last 24h
+      - name: Check for recent commits
         id: check
         run: |
-          RECENT=$(git log --since="24 hours ago" --oneline | head -1)
+          # 25h window absorbs cron-vs-commit timing skew at the boundary.
+          RECENT=$(git log --since="25 hours ago" --oneline | head -1)
           if [ -n "$RECENT" ]; then
             echo "has_changes=true" >> "$GITHUB_OUTPUT"
           else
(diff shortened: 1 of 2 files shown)


iris-clawd (changes requested):
Workflow changes look solid — stamping all packages, computing NIGHTLY once, widening to 25h, standardizing uv setup. All good.

**One concern on `lib/crewai/pyproject.toml`:**

The `file-processing` extra now pins `crewai-files==1.14.5a3`. The nightly `sed` will rewrite this correctly for dev builds, but on the **stable release path** this hardcoded alpha pin would ship as-is — meaning `pip install crewai[file-processing]` on a stable release would require an alpha version of crewai-files, which pip won't resolve by default.

Should this instead use a `~=` or `==` pin matching the stable version pattern (e.g. `crewai-files==1.14.5` without the alpha suffix), or keep it unpinned like before and have the nightly sed handle the rest? 💬 386

iris-clawd (approved):
Approving per author's request. Previous concern about the `crewai-files==1.14.5a3` pin on stable noted — deferring to Greyson's judgment on the release flow. Workflow changes are clean. ✅ 💬 387

PR #5807, opened 2026-05-14
refactor: fix memory leak in git.py by using cached_property

irfaan101:
This PR addresses a legacy TODO in lib/cli/src/crewai_cli/git.py where @lru_cache on an instance method was identified as a potential memory leak source. I have replaced it with @cached_property, which ensures the cache is tied to the instance lifecycle and properly cleared during garbage collection

## Summary by CodeRabbit

* **Refactor**
  * Repository detection simplified and made more reliable: repository status is now accessed via a cached property instead of repeated checks, reducing redundant work. This improves CLI responsiveness, lowers overhead during repository-state queries, and yields more consistent behavior across commands.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5807)


lib/cli/src/crewai_cli/git.py
```diff
@@ -40,12 +40,12 @@ def status(self) -> str:
...
         ).strip()
 
-    @lru_cache(maxsize=None)  # noqa: B019
+    @cached_property  # noqa: B019
     def is_git_repo(self) -> bool:
         """Check if the current directory is a git repository.
 
-        Notes:
-          - TODO: This method is cached to avoid redundant checks, but using lru_cache on methods can lead to memory leaks
+        Notes: This method uses cached_property to avoid redundant checks while being memory-safe.
```
greysonlalonde: Can you remove this Note?

greysonlalonde (changes requested):

lib/cli/src/crewai_cli/git.py
irfaan101: Done! I've removed the redundant note from the docstring as requested. 
Thanks, @greysonlalonde!"

greysonlalonde (approved):

irfaan101 (commented):
Updated the PR from my end.

greysonlalonde (approved):

PR #5822, opened 2026-05-15
fix: add idempotency guard for tool re-execution on task retry

### LINKED ISSUE #5802: Tool re-execution on task retry has no idempotency guard — duplicate payments, emails, trades possible
### Description

When a CrewAI task fails and is retried — via max_retry_limit, exception handling, or external re-trigger — any @tool decorated function that already executed runs again. There's no mechanism to detect that a specific tool call already completed.

### Steps to Reproduce

1. Create a CrewAI agent with a @tool that calls an external API (payment, email, etc.)
2. Have the tool execute successfully
3. Simulate a failure after tool execution but before the agent receives confirmation
4. Task retries
5. The tool fires again — duplicate side effect

### Expected behavior

A retry of the same logical tool call should return the original result without re-executing the side effect.

### Screenshots/Code snippets

@tool
def send_payment(amount: float, recipient: str) -> str:
    """Send a payment to recipient"""
    stripe.charge(amount, recipient)  # fires twice on retry
    return "payment sent"

### Operating System

Windows 11

### Python Version

3.10

### crewAI Version

latest

### crewAI Tools Version

lates

### Virtual Environment

Venv

### Evidence

See LangGraph Cloud issue #7417 for the same failure pattern documented in production: https://github.com/langchain-ai/langgraph/issues/7417

Two independent parties confirmed the exact failure mode — in-memory dedup doesn't survive worker re-dispatch, and the fix requires durable external storage keyed before execution starts.

### Possible Solution

Derive a stable request_id from tool arguments before execution, claim it in durable external storage, return the cached result on any retry with the same ID. The claim must happen outside the agent execution context so a retry on a new process hits the guard immediately.

Reference implementation: SafeAgent (pip install safeagent-exec-guard) — happy to contribute a CrewAI integration example.

### Additional context

Same pattern documented in LangGraph Cloud issue #7417. The fix requires durable storage outside the framework, not in-memory or graph state.

mahasarabesh:
﻿Fixes #5802

When a task retries (e.g. after an LLM parsing failure or timeout), tools that already executed can get called again because:
- The cache write happens *after* execution — if the tool crashes mid-way (after the side effect), no cache entry exists
- _times_executed never resets between tasks, leaking the retry budget from task A into task B

This adds an opt-in idempotent=True field on BaseTool. When set, the tool's cache entry is written *before* execution (a "pre-claim"). If the tool then crashes after performing its side effect, the pre-claim stays in cache and prevents re-execution on retry. On success, the real result overwrites the pre-claim as normal.

Also fixes the retry counter leak — _times_executed now resets when a different task starts, without affecting the recursive retry logic for the same task.

**Usage:**
```python
class SendEmailTool(BaseTool):
    name = "send_email"
    description = "Send an email"
    idempotent = True

    def _run(self, to: str, body: str) -> str:
        ...

# or with the decorator
@tool("send_email", idempotent=True)
def send_email(to: str, body: str) -> str:
    """Send an email."""
    ...
```

Non-idempotent tools (the default) are completely unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent retry counts now reset per task so retries don’t accumulate across different tasks.

* **New Features**
  * Tools can be marked idempotent so identical calls return cached results on retry.
  * Idempotent tools pre-claim cache entries; observed pre-claims surface a sentinel response instead of re-executing.
  * Cache keys are canonicalized with deterministic serialization and support an atomic claim-if-absent flow for idempotent executions.

* **Tests**
  * Added tests for per-task retry reset and idempotent tool caching/pre-claim behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5822?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -197,6 +197,7 @@ class Agent(BaseAgent):
     model_config = ConfigDict()
 
     _times_executed: int = PrivateAttr(default=0)
+    _current_task_id: str | None = PrivateAttr(default=None)
     _mcp_resolver: MCPToolResolver | None = PrivateAttr(default=None)
     _last_messages: list[LLMMessage] = PrivateAttr(default_factory=list)
     max_execution_time: int | None = Field(
@@ -751,6 +752,10 @@ class Agent(BaseAgent):
             ValueError: If the max execution time is not a positive integer.
             RuntimeError: If the agent execution fails for other reasons.
         """
+        task_id = getattr(task, "id", None) or str(id(task))
+        if self._current_task_id != task_id:
+            self._times_executed = 0
+            self._current_task_id = task_id
         task_prompt = self._prepare_task_execution(task, context)
 
         knowledge_config = get_knowledge_config(self)
@@ -887,6 +892,10 @@ class Agent(BaseAgent):
             ValueError: If the max execution time is not a positive integer.
             RuntimeError: If the agent execution fails for other reasons.
         """
+        task_id = getattr(task, "id", None) or str(id(task))
+        if self._current_task_id != task_id:
+            self._times_executed = 0
+            self._current_task_id = task_id
         task_prompt = self._prepare_task_execution(task, context)
 
         knowledge_config = get_knowledge_config(self)
--- a/lib/crewai/src/crewai/agents/cache/__init__.py
+++ b/lib/crewai/src/crewai/agents/cache/__init__.py
@@ -1,4 +1,12 @@
+from crewai.agents.cache.cache_backend import CacheBackend
 from crewai.agents.cache.cache_handler import CacheHandler
+from crewai.agents.cache.in_memory_backend import InMemoryCacheBackend
+from crewai.agents.cache.sqlite_backend import SQLiteCacheBackend
 
 
-__all__ = ["CacheHandler"]
+__all__ = [
+    "CacheBackend",
+    "CacheHandler",
+    "InMemoryCacheBackend",
+    "SQLiteCacheBackend",
+]
--- a/lib/crewai/src/crewai/agents/cache/cache_backend.py
+++ b/lib/crewai/src/crewai/agents/cache/cache_backend.py
@@ -1,0 +1,50 @@
+"""Pluggable cache backend protocol for tool result caching."""
+
+from __future__ import annotations
+
+from typing import Any, Protocol, runtime_checkable
+
+
+@runtime_checkable
+class CacheBackend(Protocol):
+    """Protocol for pluggable cache storage backends.
+
+    Implementations must be thread-safe. The default in-memory backend
+    uses a read-write lock; persistent backends (e.g. SQLite) should
+    use their own appropriate locking strategy.
+    """
+
+    def get(self, key: str) -> Any | None:
+        """Retrieve a cached value by key.
+
+        Args:
+            key: The cache key.
+
+        Returns:
+            The cached value, or None if the key is not present.
+        """
+        ...
+
+    def set(self, key: str, value: Any) -> None:
+        """Store a value in the cache.
+
+        Args:
+            key: The cache key.
+            value: The value to cache.
+        """
+        ...
+
+    def claim_if_absent(self, key: str, sentinel: Any) -> tuple[bool, Any | None]:
+        """Atomically read the cache and write a sentinel if the key is absent.
+
+        This is the core primitive for idempotent tool deduplication.
+
+        Args:
+            key: The cache key.
+            sentinel: The sentinel value to write if the key is absent.
+
+        Returns:
+            (True, None) if the sentinel was written (caller owns the claim).
+            (False, existing_value) if the key already existed.
+        """
+        ...
--- a/lib/crewai/src/crewai/agents/cache/cache_handler.py
+++ b/lib/crewai/src/crewai/agents/cache/cache_handler.py
@@ -59,8 +59,7 @@ class CacheHandler(BaseModel):
         """
         key = f"{tool}-{input}"
         with self._lock.w_locked():
-            existing = self._cache.get(key)
-            if existing is not None:
-                return False, existing
+            if key in self._cache:
+                return False, self._cache[key]
             self._cache[key] = sentinel
             return True, None
--- a/lib/crewai/src/crewai/agents/cache/cache_handler.py
+++ b/lib/crewai/src/crewai/agents/cache/cache_handler.py
@@ -49,3 +49,18 @@ class CacheHandler(BaseModel):
         """
         with self._lock.r_locked():
             return self._cache.get(f"{tool}-{input}")
+
+    def claim_if_absent(self, tool: str, input: str, sentinel: Any) -> tuple[bool, Any | None]:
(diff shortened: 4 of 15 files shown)


azender1:
The pre-claim approach is the right instinct and the `idempotent=True` flag is a clean API. Two gaps worth flagging beyond coderabbitai's canonical key comment:

**Process restart survival.** `tools_handler.cache` is in-memory. When LangGraph Cloud (or any orchestrator) sweeps a stale run and re-dispatches to a new worker, the cache is gone and the pre-claim with it. The tool fires again. This is the failure mode from #5802 — it requires durable storage outside the process, not an in-memory structure.

**Pre-claim window.** Writing `IDEMPOTENT_EXECUTION_SENTINEL` before execution is correct for crash-mid-call, but there's still a window between the pre-claim write and the tool starting on the worker. If the process dies in that window on a distributed system, the sentinel is also gone.

The fix for both: the claim needs to live in SQLite or Postgres outside the CrewAI execution context. Happy to contribute a durable backend option that slots into this PR's `idempotent=True` interface — same API, persistent storage.

azender1:
Following up on my earlier comment about process-restart survival and the pre-claim window — mahasarabesh's pluggable backend commit (e094bad) opened the slot for a hosted implementation. I've built one.
SafeAgent hosted backend — implements CacheBackend exactly, routes claim_if_absent to a server-side endpoint instead of local SQLite. Cross-machine, crash-safe, survives worker restarts because the claim lives on the server not in the process.
from safeagent_backend import SafeAgentCacheBackend
from crewai import Crew

crew = Crew(
    agents=[...],
    tasks=[...],
    cache_backend=SafeAgentCacheBackend(
        endpoint="https://safeagent-production.up.railway.app",
        agent_id="my-crew",
    )
)
Covers the gap I flagged: two workers on separate machines claiming the same request_id — only one gets PROCEED, the other gets SKIP with the cached result. No race window.
Implementation: examples/crewai/safeagent_backend.py in azender1/SafeAgent — drop-in, no deps beyond requests, falls back to PROCEED on network failure so CrewAI never blocks.

Asti1982:
I ran #5822 through Nomad's Reliability Doctor as a secret-free first-pass review. The PR is moving in the right direction: pre-claim before execution plus a pluggable backend is the primitive this issue needs.

The remaining production gates I would want before treating this as safe for payments/emails/trades are:

1. **Semantic idempotency scope, not only tool-input.** The current cache path appears to key on the concatenation of tool name and normalized input. With a persistent backend, two intentional payments with the same recipient and amount but different business scope can collapse into one cached result. I would add either an `idempotency_key_fn` or explicit `idempotency_scope` so tool authors can key on invoice/order/trade window, not raw args alone.

2. **PENDING expiry or operator reset.** A durable sentinel protects against duplicate execution, but if the process dies before the external side effect actually happens, the sentinel can block a legitimate retry forever. A small `pending_expires_at`/sweeper path, plus a test for stale PENDING reset, would make the tradeoff explicit.

3. **SQLite result serialization fallback.** `SQLiteCacheBackend.set()` serializes values with `json.dumps(value)`. If a tool returns a non-JSON object after the side effect succeeds, the result write can fail after execution and leave a permanent sentinel. I would either require/validate JSON-serializable tool results for this backend or store a safe text envelope with serializer metadata.

4. **True cross-process collision test.** The cross-instance tests are useful, but a multiprocessing test where two processes call `claim_if_absent()` on the same SQLite DB at the same time would prove the file lock plus SQLite path under the actual failure mode.

Nomad classified this as a side-effect trace-healing issue and produced this proof digest:
`sha256:5775bd9f0fdd65feef5fef8332357617cce8aec20d3d068a5d7a23e57c57950c`

If useful, I can turn the above into a bounded review/patch packet: failing tests for those gates, a minimal `idempotency_key_fn` API sketch, and a stale-PENDING sweeper. Small paid rescue scope: $49-$250, only counted after an actual receipt.

https://www.syndiode.com/service/e2e?service_type=repo_issue_help

Asti1982:
I turned the review into a small public rescue packet with the concrete acceptance gates and pytest-shaped examples:

https://www.syndiode.com/downloads/nomad_crewai_5822_rescue_packet.md

Machine-readable summary:
https://www.syndiode.com/downloads/nomad_crewai_5822_rescue_packet.json

Still treating this as non-revenue unless there is an actual paid or return-compute receipt. The packet is meant to make the production-risk discussion easier to review: semantic idempotency scope, stale PENDING expiry, SQLite serialization fallback, and a real cross-process collision test.

azender1:
@mahasarabesh — branch is showing out of date with main, worth rebasing before the maintainer review so the CI can run clean.

Asti1982 (commented):
I re-checked the current head (`e094bad`) after the latest updates. One production gate still looks open:

`lib/crewai/src/crewai/tools/tool_usage.py` still uses a read-then-add pattern for idempotent tools. The native executor path now correctly uses `cache.claim_if_absent(...)`, but the sync/async `ToolUsage` path does:

- read cache by canonicalized args;
- later, if the tool is idempotent, write `IDEMPOTENT_EXECUTION_SENTINEL` with `cache.add(...)`;
- then execute the tool.

That leaves the same pre-claim race window in this path: two identical idempotent calls can both miss the read, both write the sentinel, and both execute the side effect. For payments/emails/trades, this is the exact failure class #5802 is trying to close.

The smallest acceptance gate I would add is: make `ToolUsage._use` and `ToolUsage._ause` use the same `claim_if_absent` short-circuit as `CrewAgentExecutor`, then add a regression test that forces two concurrent calls through the ToolUsage path and asserts only one tool body runs.

Nomad rescue-candidate receipt for this narrowed gate:
`nomad-rescue-candidate-5a2d18ea6b3a59daa84650d4`

Still non-revenue; this is just the promoted technical gate from the rescue packet.

arian-gogani:
the `action_ref` in nobulex was designed as an idempotency key for exactly this problem.

`action_ref = SHA-256(JCS({agent_id, action_type, scope, timestamp_ms}))` — deterministic, canonical, same inputs always produce the same ref. a tool that already generated a receipt with a given action_ref can check "has this ref been seen?" before re-executing.

the AURA reputation system already uses this as their idempotency key — a replayed receipt can't double-count in reputation scoring because the ref is a dedup key.

for the CrewAI case: the `idempotent=True` opt-in could check an action_ref registry before re-execution. if the ref exists (from the first execution), skip the side effect. the receipt from the first execution is the evidence that the action already happened.

this is cleaner than cache-based dedup because the action_ref is content-addressed (derived from what the tool did, not when it was cached). cache entries expire or get evicted. a receipt is permanent.

github.com/arian-gogani/nobulex

azender1:
> @mahasarabesh — branch is still showing out of date with main, CI is stalled on all five required checks. Rebase onto current main and the maintainer can do a clean review. The pluggable backend commit (e094bad) is the piece that makes this production-safe — worth getting it across the line.

nandanadileep:
@mahasarabesh 

Would you make identity-conflict reject default for mutating tools, or keep opt-in like idempotent=True?
If pre-claim exists for ticket T with args A, and retry arrives as ticket T with args B - return cached A, hard-block, or allow B?
Soft error to the LLM vs hard-block?

PR #5839, opened 2026-05-17
fix(memory): replace deprecated datetime.utcnow() across memory subsystem

Ghraven:
## Summary

`datetime.utcnow()` was deprecated in Python 3.12 (PEP 2038) and is scheduled for removal in Python 3.15. It also returns a *naive* UTC datetime, which is a common source of `TypeError: can't compare offset-naive and offset-aware datetimes` once other parts of the codebase start using timezone-aware datetimes.

This PR replaces all occurrences across the **memory subsystem** (4 files, 9 sites) with `datetime.now(timezone.utc)`.

## Files changed

| File | Sites | Notes |
|------|-------|-------|
| `memory/types.py` | 3 | `MemoryRecord.created_at` / `last_accessed` `default_factory`, plus `age_seconds` calculation |
| `memory/encoding_flow.py` | 1 | `now = datetime.utcnow()` in encoding helper |
| `memory/unified_memory.py` | 1 | `now = datetime.utcnow()` in cleanup method |
| `memory/storage/lancedb_storage.py` | 4 | `created_at`, `last_accessed`, and two `.isoformat()` timestamp writes |

## Verification

All four files compile cleanly:
```
$ python -m py_compile lib/crewai/src/crewai/memory/types.py ...
```

No remaining `utcnow` in the memory package:
```
$ grep -rn utcnow lib/crewai/src/crewai/memory/
(no matches)
```

Happy to extend this to other parts of the codebase (`streaming.py`, `tools/structured_tool.py`, etc.) as a follow-up if you'd prefer to keep each PR focused.


Ghraven:
Thanks for the review. I pushed a small follow-up that normalizes legacy naive datetimes to UTC-aware before composite scoring and when hydrating LanceDB rows, so mixed legacy/new records should not hit naive-vs-aware comparison errors.

Verification:
- `uv run ruff check lib/crewai/src/crewai/memory/types.py lib/crewai/src/crewai/memory/storage/lancedb_storage.py lib/crewai/tests/memory/test_unified_memory.py`
- `uv run pytest lib/crewai/tests/memory/test_unified_memory.py::test_composite_score_accepts_legacy_naive_created_at lib/crewai/tests/memory/test_unified_memory.py::test_composite_score_brand_new_memory -q`

I also sanity-checked LanceDB row hydration directly. The full memory test file still hits the existing Windows `--block-network`/LanceDB local socketpair issue in this checkout, so I kept the local verification focused on the changed behavior.

PR #5840, opened 2026-05-17
Docs: add auditable task output receipt guidance

aDragon0707:
## Summary

Adds a docs-only section to the Tasks guide about auditable task output receipts.

The section shows how safety-sensitive crew tasks can return a compact receipt that separates:

- task completion claims
- supporting evidence
- unsupported claims
- next owner
- human approval boundaries

## Motivation

A task result can sound complete even when important evidence is missing. For example, `Done. All tests passed. Ready to publish.` contains completion, verification, and external-action claims that may require different evidence and approvals.

The receipt pattern complements `expected_output`: `expected_output` describes what success should look like, while the receipt summarizes whether the final task output is actually supported by evidence.

## Notes

- Docs-only change.
- No CrewAI behavior change.
- No runtime dependency.
- Safety-oriented guidance for reviewing task outputs.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance on "Auditable Task Output Receipts" for safety-sensitive workflows: describes producing a machine-readable receipt alongside user-facing output that summarizes overall verification status, per-claim support and evidence, required remediation for unsupported items, next owner and human-decision flags; includes schema examples and sample receipts illustrating supported and unsupported claim scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5840?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


remimarents:
This docs pattern is very close to a failure mode I’ve been working on in coding-agent supervision: the final answer can claim completion, test success, or readiness to proceed even when the repo state does not support it.

I have a working local prototype of a stricter post-task proof gate, and I can extract a minimal framework-neutral version if useful.

The main difference from a receipt-only pattern is that the receipt is not just produced by the agent; it is checked against external evidence before the task is accepted as complete.

For coding tasks, the gate validates things like:

- declared changed files vs. actual `git status` / `git diff`
- verification commands that actually ran, with exit codes
- final report claims vs. machine-readable proof
- missing-evidence / unsupported-claim handling
- optional intent check against the original task

So the flow becomes:

1. task returns a structured receipt
2. guardrail/callback verifies the receipt against repo/tool evidence
3. crew only proceeds if the proof gate passes

This seems like a possible next step from “auditable task output receipts” toward enforceable receipts, especially for coding agents.

Would this fit CrewAI better as:
1. an additional docs example showing a callback/guardrail proof gate,
2. a small utility around `TaskReceipt`,
3. a separate cookbook/example,
4. or outside the scope of this PR?

aDragon0707:
Hi maintainers, just checking whether this docs-only pattern fits the Tasks guide.

I kept the section CrewAI-native: it uses `Task`, `output_pydantic`, guardrails/callbacks, and concrete task evidence examples. It does not add a dependency, protocol requirement, or runtime behavior.

If this feels too broad for the core Tasks page, I’m happy to move it to a cookbook/example instead.

PR #5877, opened 2026-05-21
fix: replace deprecated datetime.utcnow() with datetime.now(timezone.utc) in memory module

Ghraven:
## Summary

Replace all `datetime.utcnow()` calls with `datetime.now(timezone.utc)` in the memory module to resolve the Python 3.12 deprecation warning.

## Problem

`datetime.utcnow()` has been [deprecated since Python 3.12](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) because it returns a naive `datetime` object that does not carry timezone information, making it ambiguous and error-prone when comparing with timezone-aware timestamps.

**Before (deprecated):**
```python
datetime.utcnow()  # Returns naive datetime, emits DeprecationWarning on 3.12+
```  

**After (recommended):**
```python
datetime.now(timezone.utc)  # Returns timezone-aware UTC datetime
```  

## Changes

| File | Changes |
|------|---------|
| `types.py` | Updated `default_factory` for `created_at` and `last_accessed` fields; updated `compute_composite_score()` |
| `lancedb_storage.py` | Updated 4 occurrences in `save()`, `_now()`, and `touch_records()` |
| `encoding_flow.py` | Updated 1 occurrence in `execute_plans()` |
| `unified_memory.py` | Updated 1 occurrence in `update()` |
| `test_unified_memory.py` | Updated test assertions to use `datetime.now(timezone.utc)` |

## Verification

- All changes are mechanical replacements with no behavioral difference
- `datetime.now(timezone.utc)` returns the exact same instant as `datetime.utcnow()`, but as a timezone-aware object
- Import updated from `from datetime import datetime` to `from datetime import datetime, timezone` in all affected files
- Grep confirms no remaining `utcnow` calls in the memory module

### DIFF
--- a/lib/crewai/src/crewai/memory/encoding_flow.py
+++ b/lib/crewai/src/crewai/memory/encoding_flow.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 
 from concurrent.futures import Future, ThreadPoolExecutor
 import contextvars
-from datetime import datetime
+from datetime import datetime, timezone
 import logging
 import math
 from typing import Any
@@ -421,7 +421,7 @@ class EncodingFlow(Flow[EncodingState]):
         conflicts from two operations targeting the same record.
         """
         items = list(self.state.items)
-        now = datetime.utcnow()
+        now = datetime.now(timezone.utc)
 
         # --- Deduplicate actions across all items ---
         # Multiple items may reference the same existing record (because their
--- a/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
+++ b/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
@@ -269,9 +269,13 @@ class LanceDBStorage:
             if val is None:
                 return datetime.now(timezone.utc)
             if isinstance(val, datetime):
-                return val
-            s = str(val)
-            return datetime.fromisoformat(s.replace("Z", "+00:00"))
+                dt = val
+            else:
+                s = str(val)
+                dt = datetime.fromisoformat(s.replace("Z", "+00:00"))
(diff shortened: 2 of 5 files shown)


Ghraven:
Thanks for the review. I addressed the timestamp normalization issue by coercing legacy naive datetimes to UTC and normalizing aware datetimes back to UTC before building MemoryRecords. I also added a regression test covering both an offset-less legacy timestamp and an offset-aware timestamp.\n\nLocal verification: python -m compileall -q lib/crewai/src/crewai/memory/storage/lancedb_storage.py lib/crewai/tests/memory/test_unified_memory.py and git diff --check pass. I was not able to run the focused pytest locally because this lightweight checkout/environment is missing project test dependencies (json_repair after disabling unavailable default pytest plugins), but the test is included for CI to run in the normal repo environment.

Ghraven:
Thanks for the review. Added explicit UTC tzinfo assertions in the legacy timestamp regression test, since datetime equality alone accepts the same instant with another offset.

Verification: ran the exact regression function against the checkout's real LanceDBStorage and MemoryRecord with LanceDB 0.29.2 and Pydantic 2.12.5 in an isolated environment. It passed. A mutation restoring a +02:00 offset to last_accessed failed the new assertion as expected. The harness bypasses unrelated Agent/LLM package initializers; this is not a full pytest suite run. Compilation and git diff --check passed. The configured Ruff commands exclude this test directory, so no lint/format coverage is claimed for it.

This follow-up was prepared with AI assistance and verified locally.

I also attempted to apply the contributing guide's llm-generated label, but GitHub denied the label permission. Could a maintainer add it?

PR #5898, opened 2026-05-21
test: lock in Bedrock and OpenAI _parse_native_tool_call arg handling (#4972)

adityasingh2400:
## Summary

Issue #4972 reported that AWS Bedrock Converse tool calls had their arguments silently dropped because the dict branch of `_parse_native_tool_call` used a truthy default in `func_info.get("arguments", "{}")`, which short-circuited the `or` before the Bedrock `input` key was read.

The runtime fix shipped in commit 25fcf39 (`fix: preserve Bedrock tool call arguments by removing truthy default`), but no test locked in the behaviour, so the issue stayed open. This PR adds the regression coverage.

### What is covered

A new `TestParseNativeToolCall` class in `lib/crewai/tests/agents/test_native_tool_calling.py` with parametrized cases:

- **Bedrock Converse dict shape** `{"name": ..., "input": {...}, "toolUseId": ...}` — asserts `toolUseId` is used as call id, the tool name is parsed, and `input` round-trips as the args dict.
- **OpenAI dict shape** `{"id": ..., "function": {"name": ..., "arguments": "...json..."}}` — asserts the `id` is preserved and the JSON `arguments` string is returned unchanged.
- A third case asserts the parsed args dict equals the original Bedrock `input` (and is not the empty dict that triggered the bug).

I verified the tests catch the regression by temporarily restoring the buggy default (`func_info.get("arguments", "{}")`): the two Bedrock cases fail with `'{}' == {...}` and the OpenAI case still passes, exactly matching the bug surface in the issue.

This is a **test-only PR**, no source changes.

## Test plan

- [x] `pytest tests/agents/test_native_tool_calling.py -v -k parse_native_tool_call` passes (3/3)
- [x] Reverted the fix locally and confirmed the two Bedrock cases fail with the exact symptom from #4972
- [x] Restored the fix and confirmed all three cases pass

Fixes #4972

## Summary by CodeRabbit

* **Tests**
  * Added a regression test suite validating parsing of native tool-call formats from multiple LLM providers, ensuring call IDs, tool/function names, and arguments are correctly preserved and extracted across provider-specific shapes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5898?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


adityasingh2400:
Hi maintainers, this PR is from a fork and the required CI workflows (Lint, Run Type Checks, Run Tests on 3.10 to 3.13) are sitting in the action_required state, which is the first-time-contributor gate. Could one of you approve the workflow runs so the required checks can complete and the PR can be merged? The change is test-only (one new file under lib/crewai/tests/agents/) and locks in the Bedrock arg-handling fix from #4972. Happy to rebase if you need anything else. Thanks.

PR #5914, opened 2026-05-23
fix: strip cache_breakpoint in LiteLLM path and expand Anthropic prefixes

### LINKED ISSUE #5886: [BUG] cache_breakpoint injected into messages for non-Anthropic providers (Groq, OpenAI-compatible)
### Problem:

### Error:
```
```

### Root cause:

### Workaround:
```python
```

### Steps to Reproduce

```python
crew.kickoff()
```
3. Run the script

### Expected behavior

### Screenshots/Code snippets

ull traceback:
```
```

```json
{
  "role": "system",
  "cache_breakpoint": true
}
```

### Operating System

Windows 10

### Python Version

3.12

### crewAI Version

1.14.4

### crewAI Tools Version

1.14.4

### Virtual Environment

Venv

### Evidence

### Possible Solution

### Additional context

### LINKED ISSUE #5893: [BUG] Model naming prefixes filtering is too strict
### Description

### Steps to Reproduce

### Expected behavior

### Screenshots/Code snippets

### Operating System

macOS Catalina

### Python Version

3.12

### crewAI Version

1.14.2

### crewAI Tools Version

1.14.2

### Virtual Environment

Venv

### Evidence

Incorrect output behavior

### Possible Solution

### Additional context

None

superShen0916:
## Summary

Two focused fixes for LLM provider routing:

- Strip internal `cache_breakpoint` markers before messages enter the LiteLLM fallback path (Fixes #5886).
- Recognize custom Anthropic deployment names using the `anthropic--` prefix (Fixes #5893).

The cache-marker portion overlaps with #6287, which includes it as part of a broader native Groq change. This PR remains limited to the message-sanitization fix and the separate Anthropic prefix case.

## Changes

- Copy outgoing message dictionaries before removing `cache_breakpoint`, preserving caller-owned inputs.
- Add `anthropic--` to native Anthropic provider-pattern matching.
- Add behavior tests for both paths.

## Test plan

- `pytest -q -n 0 lib/crewai/tests/test_llm.py -k 'anthropic_double_dash_model_matches_native_provider or litellm_formatting_strips_cache_breakpoint_without_mutating_input'` — 2 passed, 94 deselected.
- `ruff check` on the changed implementation and test files — passed.
- `ruff format --check` on the changed implementation and test files — passed.

## Risk

Low. The change removes an internal-only flag from copied messages and expands one provider prefix allowlist.

## AI assistance disclosure

This contribution was developed with assistance from Claude Code. The branch has been updated and the behavior has been verified against current `main`.


superShen0916:
This remains relevant for the `anthropic--` prefix handling from #5893, which is still not present on current `main`. The `cache_breakpoint` part overlaps with #6287; I can narrow this PR to the remaining prefix fix if maintainers prefer.

superShen0916:
AI assistance is disclosed in the PR body. I do not have permission to apply repository labels from this fork; maintainers, please add the required `llm-generated` label.

PR #5918, opened 2026-05-24
fix: use timezone-aware memory timestamps

Ghraven:
## Summary
- replace remaining naive `datetime.utcnow()` calls in the unified memory runtime with timezone-aware UTC timestamps
- update the matching memory tests to construct timezone-aware UTC datetimes

## Before / after
Before, memory records and LanceDB rows could be created or scored with naive UTC datetimes from `datetime.utcnow()`.
After, these code paths use `datetime.now(timezone.utc)`, preserving explicit UTC offsets in serialized timestamps.

## Verification
- `python -m py_compile lib/crewai/src/crewai/memory/unified_memory.py lib/crewai/src/crewai/memory/types.py lib/crewai/src/crewai/memory/encoding_flow.py lib/crewai/src/crewai/memory/storage/lancedb_storage.py lib/crewai/tests/memory/test_unified_memory.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized timestamps to use timezone-aware UTC across memory operations, improving accuracy and consistency of time-based behavior and scoring; maintains compatibility with legacy naive timestamps.

* **Tests**
  * Updated memory test suite to reflect timezone-aware timestamp handling and added a test ensuring legacy naive datetimes remain supported.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5918?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/crewai/src/crewai/memory/encoding_flow.py
+++ b/lib/crewai/src/crewai/memory/encoding_flow.py
@@ -12,7 +12,7 @@ from __future__ import annotations
 
 from concurrent.futures import Future, ThreadPoolExecutor
 import contextvars
-from datetime import datetime
+from datetime import datetime, timezone
 import logging
 import math
 from typing import Any
@@ -421,7 +421,7 @@ class EncodingFlow(Flow[EncodingState]):
         conflicts from two operations targeting the same record.
         """
         items = list(self.state.items)
-        now = datetime.utcnow()
+        now = datetime.now(timezone.utc)
 
         # --- Deduplicate actions across all items ---
         # Multiple items may reference the same existing record (because their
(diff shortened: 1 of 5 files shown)


Ghraven:
Thanks for catching this. I pushed a small follow-up that normalizes legacy naive datetimes to UTC before the recency calculation, and added a regression test for that path.

Ghraven:
Thanks for the reminder. This PR is still relevant: it keeps memory timestamps timezone-aware while preserving compatibility with legacy naive UTC values. CodeRabbit's actionable feedback was addressed in the follow-up commit, and the latest review/check status is green from my side. I am happy to rebase or adjust further if maintainers prefer.

PR #5920, opened 2026-05-24
docs: add pip alternative to installation instructions

AjaasMk:
## What this changes
Adds `pip install` as an alternative to `uv pip install` in the 
installation section and FAQ.

## Why
Many beginners don't have `uv` installed and get confused when 
the only installation method shown is `uv pip install`. 
Adding `pip install` as an alternative makes the docs more 
accessible for new users.

## Changes
- Installation section: added pip alternative
- crewai[tools] section: added pip alternative  
- FAQ installation answer: added pip alternative

## Documentation
* Installation instructions updated to feature `uv` as the recommended package manager for installing CrewAI, with traditional `pip` available as an alternative.
* Optional tools installation guidance similarly updated to promote `uv` as the preferred option.
* FAQ section enhanced with consistent and updated installation guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5920?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,13 +195,21 @@ Ensure you have Python >=3.10 <3.14 installed on your system. CrewAI uses [UV](h
 First, install CrewAI:
 
 ```shell
+# Using uv (recommended)
 uv pip install crewai
+
+# Using pip
+pip install crewai
 ```
 
 If you want to install the 'crewai' package along with its optional features that include additional tools for agents, you can do so by using the following command:
 
 ```shell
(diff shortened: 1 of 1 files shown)


greysonlalonde:
Our [docs](https://docs.crewai.com/en/installation) show how to setup with uv - this should be sufficient. uv is the recommend approach to install as pip has optimization problems.

PR #5967, opened 2026-05-29
feat(memory): per-tenant memory isolation

### LINKED ISSUE #2584: [FEATURE]  Memory Distinguished by Custom Key​​
### Feature Area

Agent capabilities

No

Currently, agents share a ​​global memory space​​ without isolation. This makes it impossible to:

- Scope memories to specific entities (e.g., users, accounts, sessions).
- Retrieve memories contextually (e.g., fetch only memories tied to account_id=123).
- Prevent accidental data leakage across logical boundaries.

Which will have the following Benefits

✅ ​​Multi-tenancy​​: Isolate memories per user/account/context.
✅ ​​Contextual Recall​​: Agents fetch only relevant memories.

_No response_

### Additional context

_No response_

### Willingness to Contribute

John-Jepsen:
Closes [#2584](https://github.com/crewAIInc/crewAI/issues/2584)

This PR implements per-tenant memory isolation and contextual memory retrieval.

## Summary

Closes the per-tenant memory leak in the unified `Memory` subsystem. Right now every user's memories pool into the same vector collection and `recall()` mixes them across users — a data leak, not a UX nit. This PR threads `tenant_id` (and optional `user_id`) through the entire save → store → recall path and enforces isolation at the storage layer.

**Invariant:** a `recall()` scoped to tenant A can never return a memory written by tenant B — under any ranking, embedding collision, query depth, or backend.

**Backward compatible.** Default `tenant_id` is `"_default"`, so single-tenant deployments keep working unchanged. Existing rows on disk read back as `"_default"` via row-level fallbacks and an auto-`add_columns` step when LanceDB tables are first opened.

## Design

Full design doc: [`design-docs/0001-per-tenant-memory-isolation.md`](https://github.com/John-Jepsen/crewAI/blob/feat/memory-tenant-isolation/design-docs/0001-per-tenant-memory-isolation.md) — covers the invariant, identity propagation, partitioning tradeoffs, threat model, test contract, and rollout order.

The enforcement chokepoint is a new `ScopedStorage` proxy that holds three contracts:
1. **Stamp on write** — every record's `tenant_id` is overwritten with the bound tenant; cross-tenant records raise `PermissionError`.
2. **Inject on read** — every backend read carries the `tenant_id` predicate; callers cannot omit it.
3. **Verify on return** — every returned row is re-checked; a foreign-tenant row raises `RuntimeError` loudly, never silently filtered.

The triple is defense in depth: mypy `--strict` catches forgotten kwargs at type-check time (Protocol-level), backends push `WHERE tenant_id = ?` into the vector query (data-layer), and `ScopedStorage` is the runtime guard when either of the first two fails.

## Commits (intended as one stacked PR; happy to split if you prefer)

| # | Commit | What it does |
|---|---|---|
| 1 | `docs(memory): add design doc…` | DD-0001 + Mermaid diagram of the propagation path |
| 2 | `feat(memory): add tenant_id and user_id fields to MemoryRecord` | Additive Pydantic fields. Default `_default`. Hermetic round-trip tests. |
| 3 | `refactor(memory): require tenant_id keyword on StorageBackend protocol` | Required keyword-only `tenant_id` on every read method; LanceDB + Qdrant push the predicate into the vector query. mypy `--strict` enforces. |
| 4 | `feat(memory): add ScopedStorage wrapper and tenant isolation tests` | The enforcement chokepoint + the security-contract test file. |
| 5 | `feat(memory): wire ScopedStorage through Memory and Flows` | The load-bearing PR. Memory gains `tenant_id`/`user_id` fields and per-call kwargs on `remember`/`recall`/`forget`/etc; `RecallFlow` and `EncodingFlow` hold a `ScopedStorage` so deep recall cannot escape. |
| 6 | `feat(cli): add 'crewai memory migrate'` | One-shot CLI: `--storage-dir`, `--default-tenant`, `--from-metadata-key`, `--dry-run`. Idempotent. |
| 7 | `docs(memory): add per-tenant memory isolation guide` | Mintlify page + nav (15 spots) + fix the misleading scope-as-isolation example in `memory.mdx`. |

## What this is _not_

| Need | Solved by | Not this PR |
|---|---|---|
| Survives restart | SQLite + Chroma/LanceDB persistence | Already shipped |
| Shared across crews / cross-session knowledge graph | External memory provider (mem0) | Out of scope |
| **Per-user data isolation** | **This PR** | — |

Explicitly excluded (each its own ticket): mem0 replacement, per-tenant encryption, namespace-per-tenant partitioning (Option B in the design doc), multi-tenant rate limiting, `Crew.kickoff()` integration (Memory API is in; `Crew.kickoff(tenant_id=…)` belongs to a follow-up that touches agent execution).

## Test plan

- [x] 13 isolation-contract tests (`lib/crewai/tests/memory/test_tenant_isolation.py`):
  - Cross-tenant `recall()` returns nothing (shallow and deep)
  - `PermissionError` on cross-tenant save through scoped handle
  - `RuntimeError` on backend leak detection
  - Default-tenant backcompat
  - `forget` / `reset` / `get_record` are tenant-scoped
  - `user_id` sub-partition within tenant
- [x] 7 migrate-command tests (`lib/cli/tests/test_memory_migrate.py`): missing dir/table no-op, schema migration, per-row `--from-metadata-key`, idempotence, `--dry-run`, validation
- [x] Full memory suite: **136 passed, 19 skipped** (Qdrant optional dep)
- [x] `ruff check` clean across all changed files
- [x] `mypy --strict` clean (the 16 errors in `qdrant_edge_storage.py` are pre-existing `no-any-unimported` from the optional `qdrant_edge` dep being uninstalled — present on `main` before this branch)

## Security note

This PR fixes a data-isolation bug. Reviewers — please run the checklist on the design doc in this order:
1. Cross-scope leaks first — any recall code path that bypasses the scope filter
2. Deletion honors scope
3. Default-scope fallback works for callers that pass no `tenant_id`
4. Tests prove two users with near-identical content do not cross-contaminate
5. Then backward compat, perf, style

## Open questions (called out in the design doc)

1. Is `"_default"` the right sentinel? Alternatives bikeshed in DD-0001.
2. Context-manager vs kwarg form for per-call tenant binding — currently kwargs; happy to add the context-manager sugar in a follow-up.
3. `Crew.kickoff(tenant_id=...)` surface — proposed for a separate PR.
4. Telemetry scrubbing — `tenant_id` values must never appear in telemetry payloads. Owner of `crewai/telemetry/` should weigh in.
5. Legacy `BaseRAGStorage` deprecation timeline.

---

**AI-generated PR.** Per `.github/CONTRIBUTING.md`, the `llm-generated` label is required. I attempted to apply it via `gh pr create --label` but lack permission on this repo. **Could a maintainer please apply the `llm-generated` label?**

## Summary by CodeRabbit

* **New Features**
  * Per-tenant memory isolation: memory APIs now accept tenant_id (and optional user_id) to ensure recalls, searches, deletes, and listings are scoped to the correct tenant.
  * Memory TUI/CLI: new `crewai memory migrate` command to stamp existing memory records with tenant IDs (supports dry-run and reporting).

* **Documentation**
  * Full memory-isolation guide with examples, migration steps, threat model, rollout notes, and API usage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/crewAIInc/crewAI/pull/5967?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


John-Jepsen (commented):
gtg

John-Jepsen:
Thanks @coderabbitai. Addressed in 90073fbb:

**Fixed (5 of 5 actionable):**

- 🟠 `LanceDBStorage.delete()` — `record_ids + older_than` no longer drops `older_than`; `record_ids + categories/metadata_filter` now intersects with `record_ids`. Two new regression tests pin both branches (`test_lancedb_delete_record_ids_intersects_with_other_filters`, `test_lancedb_delete_record_ids_intersects_with_categories`).
- 🟠 `crewai memory migrate` — replaced the 10M-row materialization with a paginated `select()` scan that fetches only `id`, `tenant_id`, `metadata_str`. The `vector` column is never read. Pre-isolation tables (no `tenant_id` column yet) adapt the select set so dry-run still works against unmigrated schemas. New `test_migrate_pagination_streams_past_page_size` uses a 2-row page size to confirm a 7-row fixture is fully scanned.
- 🟠 Qdrant tenant-isolation tests — added `test_search_isolates_tenants_with_colliding_embeddings` and `test_delete_is_tenant_scoped` to `test_qdrant_edge_storage.py` so when the optional `qdrant_edge` dep is installed the same isolation contract from `test_tenant_isolation.py` is verified for the Qdrant backend.
- 🟠 Deep-recall isolation test — added `test_deep_recall_honors_tenant` that uses `depth="deep"` so the `RecallFlow` path actually runs. Two tenants, SHA-256-derived embeddings, asserts no cross-tenant leakage.
- 🟡 CLI "without key" count — now derives from `rows_scanned - rows_with_metadata_key` so reruns don't print a negative number once `rows_to_stamp` converges to 0.
- 🟡 `mock_embedder` — replaced `abs(hash(t)) % 1000` with a SHA-256 digest so different texts cannot collide and make isolation assertions pass vacuously.

**Skipped (with reason):**

- 💤 `user_id` UNSET sentinel — the per-call `user_id=None` meaning "fall back to instance default" is the intended SaaS pattern. Callers who need "query across all users within this tenant" construct a separate `Memory` without a default `user_id`. The complexity of an UNSET sentinel didn't seem worth it for this case; happy to revisit if maintainers disagree.
- 💤 Design-doc startup-warning scan limit and `_quote` escaping nits — the warning is a one-line log, not a load-bearing scan, and the escaping is documented inline in `lancedb_storage.py` (`_sql_quote` doubles single quotes, matching SQLite/LanceDB literal rules). Bandit's `S608` rule catches any raw f-string usage in CI.

**Test counts:** 140 passed, 21 skipped (was 136/19). Deltas: 2 LanceDB delete tests, 2 Qdrant isolation tests (skipped without the optional dep), 1 deep-recall test, 1 migrate-pagination test.

xg-gh-25:
**Per-tenant isolation is the right architectural boundary** — it maps directly to the security/compliance surface (tenant A can't see tenant B's data) and scales horizontally (each tenant = independent memory namespace).

**Critical detail**: Does your implementation isolate at the **storage level** (separate DBs/schemas per tenant) or **query level** (shared storage + tenant_id filter)? The former is bomb-proof but operationally heavy (100 tenants = 100 DB connections). The latter is fast but risky (one missing WHERE clause = data leak).

**Practical middle ground** from production systems:
```python
# Tenant-scoped memory client with query-time enforcement
class TenantMemoryStore:
    def __init__(self, base_store, tenant_id):
        self.store = base_store
        self.tenant_id = tenant_id
        
    def query(self, embedding, k=10):
        # Every query auto-injects tenant filter
        results = self.store.query(
            embedding, 
            k=k,
            filter={"tenant_id": self.tenant_id}
        )
        # Paranoid: assert no cross-tenant leaks
        assert all(r.metadata["tenant_id"] == self.tenant_id for r in results)
        return results
```

This gives you **shared infrastructure + hard guarantees** without operational overhead. The assertion catches bugs early (fail-fast vs. silent leak).

**Question**: Are you also isolating **embedding models** per tenant? Multi-tenant systems sometimes share a model (faster, cheaper) but that creates subtle leakage: tenant A's fine-tuned embeddings can influence tenant B's recall if the model sees all data. Separate models = true isolation but 10x cost.

What's your threat model: **malicious tenant** (must be storage-isolated) or **accidental leak** (query-level + assertions OK)?

---
*via [SwarmAI](https://github.com/gawain-ai/swarmai) community engine*

John-Jepsen:
You named the exact tradeoff I landed on. I went query-level, not
  storage-level — same reasons you outlined: 100 tenants ≠ 100 DBs is a
  non-starter for an OSS framework where most deployments are a single Postgres
  or a single Qdrant collection.

  What I added to make that safe is a triple contract, all enforced in one
  place (ScopedStorage):

  1. Type-level. The StorageBackend Protocol makes tenant_id a required keyword
  on every read/write method. Forgetting it is a type error, not a runtime bug.
  That kills the "missing WHERE clause" failure mode you flagged before it ever
  ships.
  2. Backend pushdown. The scoped wrapper injects tenant_id into the actual SQL
  / Qdrant filter — the predicate runs at the storage layer, not in Python.
  3. Runtime verify. After the backend returns rows, the wrapper re-checks
  record.tenant_id == bound_tenant_id on every single row. A mismatch raises
  RuntimeError loudly — we explicitly chose not to silently drop the foreign
  row, because silent filtering hides the bug for the next person.

  Writes get a symmetric guard: _stamp() refuses to save a record whose
  tenant_id differs from the wrapper's. Silent relabel masks bugs, so we raise
  PermissionError instead. The wrapper is the only place these invariants live —
   if you add a new read method, it has to go through the predicate, and that's
  the rule we encoded in the module docstring.

  The wrapper is cheap (two strings + one ref), so it's per-request, not cached.
   One long-lived Memory instance can serve N tenants concurrently by minting a
  fresh ScopedStorage per call.

  On embeddings: we share the model. Records carry tenant in metadata only, not
  in the vector space. We accept that risk because the supported backends
  (Chroma, Qdrant, Postgres+pgvector) operate on frozen pre-trained embeddings —
   no per-tenant fine-tuning happens in the framework. The day someone wires in
  an online-learning embedder, that assumption breaks and we'd need a different
  story. Worth a doc warning.

  Threat model: explicitly accidental leak. The triple contract gives us hard
  runtime guarantees against developer error and backend filter bugs, which is
  what we're worried about for a framework people drop into their own infra. If
  your threat model is malicious tenant — code execution inside the tenant's
  runtime trying to reach across — query-level + asserts is not enough; you need
   real storage isolation (separate DB, separate Qdrant collection, separate
  process boundary). We document that distinction in the isolation guide so
  users picking this up for compliance work know which side of the line they're
  on.

  The CLI ships a crewai memory migrate command for the rollout — existing
  records get stamped with a _default tenant on first read, so adopting the
  boundary doesn't require a destructive backfill.
  
  @joaomdmoura

John-Jepsen (commented):
gtg

John-Jepsen:
@greysonlalonde

xg-gh-25:
@John-Jepsen solid call on query-level with triple-layer safety — for an OSS framework where most deployments share a single store, storage-level isolation is over-engineering. The paranoid assertion (fail-fast on cross-tenant results) is the key differentiator from naive WHERE-clause filtering; it turns a silent data leak into a loud crash during development. One pattern we added on top: a periodic `isolation_audit` job that runs random cross-tenant probe queries and alerts if any return results. Catches the case where a new code path bypasses the scoped client and goes directly to the store.

John-Jepsen:
@xg-gh-25 really hoping they merge it......

PR #5981, opened 2026-05-30
Document Tuning Engines OpenAI-compatible configuration

cerebrixos:
## Summary

- Adds a small documentation example showing how to point this project’s existing OpenAI-compatible configuration at Tuning Engines.
- Keeps this project’s APIs and runtime behavior unchanged: no new dependency, adapter, or code path.
- Shows a path for teams to keep this framework in charge of app, agent, tool, workflow, or retrieval logic while routing model calls through a governed OpenAI-compatible endpoint.

## Why this belongs here

Tuning Engines is an AI control plane for teams that want centralized model access, routing, tenant-scoped keys, policy/guardrail checks, audit logs, traces, approvals, and usage/cost visibility around existing AI applications.

This project already supports OpenAI-compatible endpoints. The docs change makes that existing capability discoverable for users who want governance and observability without rewriting their application around a separate SDK or changing this project’s runtime behavior.

## Testing

- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Expanded the “Using LLM Class Attributes” section with a new “Tuning Engines” code example for connecting to an OpenAI-compatible endpoint (including model, API key, and base URL) and wiring it into an Agent.
  * Added a note clarifying how OpenAI-compatible gateway settings affect only the model-call path, and that the call fails with a provider error if the gateway is unavailable (no automatic fallback).

### DIFF
--- a/docs/v1.13.0/en/learn/llm-connections.mdx
+++ b/docs/v1.13.0/en/learn/llm-connections.mdx
@@ -177,6 +177,17 @@ You can connect to OpenAI-compatible LLMs using either environment variables or 
             agent = Agent(llm=llm, ...)
             ```
         </CodeGroup>
+
+        <Note>
+        Configuring an OpenAI-compatible gateway changes the model-call path only.
+        CrewAI continues to own agent, task, tool, and workflow behavior; the
+        gateway can handle model routing, policy checks, audit/usage logging, and
+        approval requirements. Use a tenant-scoped gateway key rather than a
+        direct provider key. If the gateway is unavailable, the LLM call fails as
+        a provider error; CrewAI does not automatically fall back to a direct
+        provider unless you explicitly configure that behavior in your
+        application.
+        </Note>
     </Tab>
 </Tabs>
 
--- a/docs/v1.13.0/en/learn/llm-connections.mdx
+++ b/docs/v1.13.0/en/learn/llm-connections.mdx
@@ -156,6 +156,17 @@ You can connect to OpenAI-compatible LLMs using either environment variables or 
             agent = Agent(llm=llm, ...)
             ```
 
+            ```python Tuning Engines
+            # Governed OpenAI-compatible endpoint for model routing,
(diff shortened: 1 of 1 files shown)


cerebrixos:
Thanks for the review. I pushed an update that addresses the wording/model note:

- changed the comment to a neutral `Example using Tuning Engines OpenAI-compatible endpoint`
- replaced the hardcoded `gpt-4o-mini` with `your-model-alias` and a note to use a model alias from the Tuning Engines catalog
- rebased the branch onto current `main`

Validation: `git diff --check origin/main...HEAD` passes.

cerebrixos:
Hello team @caike  . Please do consider this PR for merge. So much appreciate in advance :)

salimassili62-afk:
This docs-only approach makes sense, especially since it reuses the existing OpenAI-compatible configuration instead of adding a new adapter or runtime path.

One thing I’d suggest clarifying in the example is the **trust boundary** when routing through a governed endpoint:

```ts id="t9v2ka"
const llm = new LLM({
  model: "your-model",
  apiKey: process.env.TUNING_ENGINES_API_KEY,
  baseURL: process.env.TUNING_ENGINES_BASE_URL,
});
```

The important operational details for users are probably:

* whether the framework still owns agent/tool/workflow logic
* whether the gateway owns routing, policy checks, logging, and approvals
* what happens if the governed endpoint is unavailable
* whether failures should fail closed or fall back to a direct model provider
* how tenant-scoped keys should be separated from app-level provider keys

Since this change is documentation-only, adding those notes would help users understand that the integration changes model-call routing/governance, not the framework’s runtime behavior.

cerebrixos:
Thanks for the thoughtful suggestion. I added a short note below the OpenAI-compatible examples that clarifies the boundary: CrewAI continues to own agent, task, tool, and workflow behavior, while the gateway only handles the model-call path and any configured routing, policy, logging, or approval controls. It also calls out tenant-scoped key use and that a gateway outage is surfaced as an LLM/provider error unless an application explicitly configures a fallback.

Issue #5109, opened 2026-03-26
Request for Python 3.14 Compatibility and Estimated Release Timeline


SaiKiranBhogi:
### Feature Area

Core functionality

### Is your feature request related to a an existing bug? Please link it here.

Hi, thank you for your work on Crew AI!  

I would like to know when you are planning to release a version compatible with Python 3.14. Is there an estimated timeline for this? 

Thanks!  

### Describe the solution you'd like

I would like Crew AI to be officially compatible with Python 3.14, including updates to dependencies, tests, and documentation. Ideally, this would mean releasing a new version of Crew AI that supports Python 3.14. If possible, please provide guidance or a timeline for when this compatibility might be achieved.

### Willingness to Contribute

Yes, I'd be happy to submit a pull request

lokeshwarobuli:
Excited for this version bump. Can we get a timeline for this feature please ?

2-ulfhednar-2:
There doesnt seem to be any movement on this currenty

SaiKiranBhogi:
Could you please share the expected release timeline for Python versions 3.14 and later?

PR #5996, opened 2026-06-01
fix: replace deprecated datetime.utcnow() in memory subsystem

Oxygen56:
## Summary

Replace remaining `datetime.utcnow()` calls in the memory subsystem with `datetime.now(timezone.utc).replace(tzinfo=None)`. The `utcnow()` function was deprecated in Python 3.12 and generates `DeprecationWarning` — it may be removed in Python 3.14.

## Changes

| File | Change |
|------|--------|
| `lib/crewai/src/crewai/memory/types.py` | Replace `datetime.utcnow` default factories (2 occurrences) and inline call in `compute_composite_score` (1 occurrence) |
| `lib/crewai/src/crewai/memory/unified_memory.py` | Replace `datetime.utcnow()` in `update_record` (1 occurrence) |

## Why `.replace(tzinfo=None)`?

The existing code stored naive datetimes via `utcnow()`. To preserve compatibility with stored data, the replacement produces a naive UTC datetime — the same value that `utcnow()` returns, without the deprecation warning.

## Related

- Complements PR #5970 which fixed the same pattern in `encoding_flow.py` and `lancedb_storage.py`
- Partially addresses #5109 (Python 3.14 compatibility)

## Testing

All existing tests pass. The change is a mechanical replacement that produces identical values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timestamp handling for memories and relevance scoring by switching to explicit UTC-based timestamps (kept as naive datetimes), yielding more consistent age calculations and more accurate ranking.

* **Tests**
  * Added and updated tests to enforce the new UTC-based timestamp behavior across memory records and scoring/reranking scenarios.


wizardo091:
Thanks for addressing the remaining `datetime.utcnow()`usages. Since this change is related to Python 3.14 compatibility, could we add a test that verifies the generated timestamps remain naive UTC datetimes (matching the previous `utcnow()` behavior)? That would help prevent future regressions if the implementation is refactored again.

Oxygen56:
Thank you for the review! I have added `test_memory_record_timestamps_are_naive_utc` in `lib/crewai/tests/memory/test_unified_memory.py` which verifies that:

1. `created_at` and `last_accessed` have no tzinfo (are naive datetimes matching the old `utcnow()` behavior)
2. Both timestamps are within 5 seconds of the current time

I also updated the existing test uses of `datetime.utcnow()` in the same file to use `datetime.now(timezone.utc).replace(tzinfo=None)` to avoid future DeprecationWarnings during testing.

Oxygen56:
Thanks for the reviews!

@wizardo091: The test `test_memory_record_timestamps_are_naive_utc` (added in this PR at `test_unified_memory.py:34`) verifies exactly that — it asserts `created_at.tzinfo is None` and `last_accessed.tzinfo is None`, plus proximity to current UTC within 5 seconds. This guarantees the replacement `datetime.now(timezone.utc).replace(tzinfo=None)` produces naive datetimes matching the old `utcnow()` behavior.

@coderabbitai: The test docstring already references the deprecated `datetime.utcnow()` as the correct API being replaced. The docstring explains:\n\n> The fix replaces utcnow() with ``datetime.now(timezone.utc).replace(tzinfo=None)``\n\nThis accurately describes what changed and why — naive UTC datetimes matching `utcnow()` semantics. The wording looks correct to me. If there's a specific phrase you think is misleading, happy to adjust.

Oxygen56:
Rebased onto the latest `main` and resolved the `unified_memory.py` import conflict by preserving the upstream `copy`/`suppress` imports together with this PR’s `timezone` import. New head: `f0978bd1`. Verification: `uv run pytest -q lib/crewai/tests/memory/test_unified_memory.py` (51 passed) and `git diff --check` passed.

PR #6007, opened 2026-06-02
fix: keep docling imports lazy

devin-lai:
## What
Add a regression test that verifies `import crewai` does not eagerly load the optional docling stack (`docling`, `docling_core`, `torch`, or `transformers`).

## Why
The docling integration is optional and should only load when a docling source is used. Keeping this covered by a subprocess import test prevents future regressions in the base package import path.

## Dependency note
On CPython 3.13.8, torch 2.11.0 fails during import. The docling knowledge tests need a working torch install on Python 3.13, so the workspace override now requires `torch>=2.12.0,<2.13.0` and matching `torchvision>=0.27.0,<0.28.0` for Python 3.13+. Python <3.13 keeps a looser upper bound and is not forced below the 2.12 line.

## Validation
- `uv lock --check`
- `uv run pytest lib/crewai/tests/test_imports.py -q`
- `uv run pytest lib/crewai/tests/knowledge/test_knowledge.py::test_docling_source lib/crewai/tests/knowledge/test_knowledge.py::test_multiple_docling_sources -q`
- `uv run ruff check lib/crewai/tests/test_imports.py pyproject.toml`
- `uv run ruff format --check lib/crewai/src/crewai/knowledge/source/crew_docling_source.py lib/crewai/tests/test_imports.py`
- `uv run mypy lib/crewai/src/crewai/knowledge/source/crew_docling_source.py`

## Summary by CodeRabbit

* **Tests**
  * Added tests to verify optional dependencies are not eagerly loaded during initialization.

* **Chores**
  * Updated dependency version constraints for improved Python 3.13 compatibility.

### DIFF
--- a/lib/crewai/tests/test_imports.py
+++ b/lib/crewai/tests/test_imports.py
@@ -1,5 +1,9 @@
 """Test that all public API classes are properly importable."""
 
+import os
+import subprocess
+import sys
+
 
 def test_task_output_import():
     """Test that TaskOutput can be imported from crewai."""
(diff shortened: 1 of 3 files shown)


lib/crewai/src/crewai/knowledge/source/crew_docling_source.py
```diff
@@ -1,32 +1,54 @@
...
 
 
+if TYPE_CHECKING:
+    from docling.document_converter import DocumentConverter
+    from docling_core.types.doc.document import DoclingDocument
+else:
+    DocumentConverter = Any
+    DoclingDocument = Any
+
+# docling (and its heavy transitive dependencies such as torch/transformers) is an
```
greysonlalonde: Can you remove all of these comments?

greysonlalonde (changes requested):

PR #6015, opened 2026-06-02
feat(lock_store): make locking backend overridable

mattatcha:
## Summary
Allow `crewai_core.lock_store` to use a custom locking backend instead of the built-in Redis/file selection, without touching call sites.

## Key Changes
- Add `set_lock_backend(backend)` — register a custom backend; pass `None` to restore the default.
- `lock()` uses the custom backend when one is set, otherwise the unchanged Redis/file logic.
- A backend is any `callable(name, *, timeout)` returning a context manager that holds the lock for the `with` block.
- Default behavior is byte-for-byte unchanged when no backend is set.

## Linear Issues
- Related to [OSS-19](https://linear.app/crewai/issue/OSS-19/ensure-correct-resource-locking-in-oss-and-enterprise)

---

> [!NOTE]
> **Low Risk**
> Additive API with unchanged default locking path; main caveat is unsynchronized backend swaps under concurrent lock use, which is documented for startup-only setup.
> 
> **Overview**
> Adds a **process-wide pluggable lock backend** so callers can swap Redis/file locking for custom implementations (e.g. in-process locks in tests or another distributed service) without changing existing `lock()` call sites.
> 
> **`set_lock_backend(backend)`** registers a callable `(name, *, timeout) -> context manager`; pass `None` to restore the built-in path. **`lock()`** snapshots the current backend before use so a concurrent swap cannot break acquisition, and when a custom backend is set it delegates directly with the **raw lock name** (the default path still namespaces names via `crewai:` + hash).
> 
> Default Redis vs file selection is unchanged when no backend is registered. Tests add autouse reset of the backend plus coverage that overrides bypass `portalocker` and clearing the backend restores file locking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4c097d047d58626f0db146212fa7a927292205d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Pluggable lock backend enabling custom lock implementations to replace and restore the default locking behavior.
* **Documentation**
  * Clarified default lock selection: Redis-backed locks when available, otherwise file-based locks.
* **Tests**
  * Added tests ensuring custom lock backends are honored and that default behavior is restored between tests.

### DIFF
--- a/lib/crewai-core/src/crewai_core/lock_store.py
+++ b/lib/crewai-core/src/crewai_core/lock_store.py
@@ -1,8 +1,21 @@
 """Centralised lock factory.
 
-If ``REDIS_URL`` is set and the ``redis`` package is installed, locks are
-distributed via ``portalocker.RedisLock``. Otherwise, falls back to the
-standard file-based ``portalocker.Lock`` in the system temp dir.
+The locking backend is resolved in this order of precedence:
+
+1. A backend registered in-process via :func:`set_lock_backend` (or scoped
+   with the :func:`lock_backend` context manager). Best for tests and runtime
+   wiring.
+2. A backend named by the ``CREWAI_LOCK_FACTORY`` environment variable, in
+   ``"module:callable"`` form (e.g. ``"my_pkg.locks:lock"``). The import path
+   is resolved lazily and cached. Best for deployment-driven selection, since
+   it requires no code changes and rolls back with an env unset.
+3. The built-in default: if ``REDIS_URL`` is set and the ``redis`` package is
+   installed, locks are distributed via ``portalocker.RedisLock``; otherwise
+   they fall back to a file-based ``portalocker.Lock`` in the system temp dir.
+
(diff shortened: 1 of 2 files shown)


lib/crewai-core/src/crewai_core/lock_store.py
```diff
@@ -62,6 +82,11 @@ def lock(name: str, *, timeout: float = _DEFAULT_TIMEOUT) -> Iterator[None]:
               Automatically namespaced to avoid collisions.
         timeout: Maximum seconds to wait for the lock before raising.
     """
+    if _backend is not None:
```
greysonlalonde: Race here

greysonlalonde (changes requested):

greysonlalonde (approved):

Issue #6072, opened 2026-06-08
[BUG] `human_input=True`: the feedback prompt references a "Final Result above" that is never displayed unless `verbose=True`


coolhandle01:
### Description

When a `Task` has `human_input=True` but the agent/crew is **not** verbose, CrewAI shows the `💬 Human Feedback Required` panel — *"Provide feedback on the Final Result above"* — but the result is **never printed**. The operator is asked to approve output they cannot see.

The two behaviours are gated independently:
- the **result render** (`_show_logs` → `AgentLogsExecutionEvent`) is gated on `verbose=agent.verbose or crew.verbose`;
- the **feedback gate** fires whenever `human_input=True`.

So `human_input=True` + non-verbose ⇒ a prompt that explicitly references content ("the Final Result above") that was never rendered.

### Relationship to #6065

This is **separate** from #6065 (the `AttributeError: 'AgentExecutor' object has no attribute 'ask_for_human_input'` crash). On 1.14.5 / 1.14.6 that crash fires first and masks this. The bug here is observable on **1.14.4** (where `human_input` still works), and will resurface on 1.14.5+ once #6065 / #6069 lands — so it's worth fixing alongside it.

### Steps to Reproduce

A minimal reproduction follows.

```python
from crewai import Agent, Crew, Task, Process
from crewai.llms.base_llm import BaseLLM

class StubLLM(BaseLLM):
    def call(self, messages, tools=None, callbacks=None, available_functions=None,
             from_task=None, from_agent=None, response_model=None):
        return "Thought: I know it.\nFinal Answer: The sky is blue."
    def supports_function_calling(self) -> bool:
        return False

agent = Agent(role="Tester", goal="Be a Minimal Reproduction", backstory="bg",
              llm=StubLLM(model="stub"), verbose=False)
task = Task(description="Sky colour?", expected_output="a colour",
            agent=agent, human_input=True)
Crew(agents=[agent], tasks=[task], process=Process.sequential, verbose=False).kickoff()
```

Press enter at the prompt.

### Expected behavior

When `human_input` is requested, the result under review is displayed (or embedded in the feedback panel) **regardless of `verbose`** — the prompt should never reference output that wasn't shown.

### Screenshots/Code snippets

N/A (minimal reproduction above)

### Operating System

Ubuntu 24.04

### Python Version

3.12

### crewAI Version

1.14.4

### crewAI Tools Version

1.14.4

### Virtual Environment

Venv

### Evidence

[...]

PR #6073, opened 2026-06-08
fix: human_input=True fixes — AttributeError on experimental executor + result hidden when verbose=False

### LINKED ISSUE #6065: [BUG] human_input=True on Task raises AttributeError: 'AgentExecutor' object has no attribute 'ask_for_human_input' with default experimental executor
### Description

  When human_input=True is set on a Task, the crew execution fails with:

  "An unknown error occurred. Please check the details below.
  Error details: 'AgentExecutor' object has no attribute 'ask_for_human_input'"

  This happens because the new default executor (crewai.experimental.AgentExecutor) does not expose ask_for_human_input as a direct instance attribute — it stores the flag inside self.state.ask_for_human_input — but the
  ExecutorContext protocol and SyncHumanInputProvider.handle_feedback both expect it as a direct attribute on the executor object.

### Steps to Reproduce

  llm = LLM(model="ollama/llama3", base_url="http://localhost:11434")

  agent = Agent(
      role="Researcher",
      goal="Find information",
      backstory="You are a researcher.",
      llm=llm,
      verbose=True,
  )

  task = Task(
      description="Research topic X.",
      expected_output="A summary of topic X.",
      human_input=True,   # <-- triggers the bug
      agent=agent,
  )

  crew = Crew(agents=[agent], tasks=[task], verbose=True)
  crew.kickoff()

### Expected behavior

After the agent completes the task, execution pauses and prompts the user for feedback, as documented for human_input=True.

### Screenshots/Code snippets

Execution fails with AttributeError: 'AgentExecutor' object has no attribute 'ask_for_human_input' wrapped inside the generic "An unknown error occurred" message.

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

crewai: 1.14.6

### crewAI Tools Version

crewai-tools: 1.14.6

### Virtual Environment

Venv

### Evidence

  When human_input=True is set on a Task, the crew execution fails with:

  "An unknown error occurred. Please check the details below.
  Error details: 'AgentExecutor' object has no attribute 'ask_for_human_input'"

Note: unfortunately I am reporting this bug after my current API usage session limit was exhausted.
As soon as I have my session limit restored, I will provide screenshots or more detailed evidence.

### Possible Solution

Root cause

  The ExecutorContext protocol (crewai/core/providers/human_input.py:26) declares ask_for_human_input: bool as a required direct attribute:

  class ExecutorContext(Protocol):
      ask_for_human_input: bool   # line 26
      ...

  SyncHumanInputProvider._handle_regular_feedback reads and writes this attribute directly on the executor passed as context:

  while context.ask_for_human_input:          # direct read
      ...
      context.ask_for_human_input = False     # direct write

  The deprecated CrewAgentExecutor satisfies this — ask_for_human_input is a top-level Pydantic field:
  # crewai/agents/crew_agent_executor.py:129
  ask_for_human_input: bool = Field(default=False)

  The new default AgentExecutor (crewai/experimental/agent_executor.py) does not. The field is defined on AgentExecutorState (the Flow state class), not on the executor itself:
  # AgentExecutorState (line 132) — NOT on AgentExecutor
  ask_for_human_input: bool = Field(default=False)

  The executor's invoke() correctly uses self.state.ask_for_human_input, but then passes self to _handle_human_feedback as an ExecutorContext, where the direct attribute access fails:

  # experimental/agent_executor.py:3025
  return provider.handle_feedback(formatted_answer, cast("ExecutorContext", self))
  # ↑ self.ask_for_human_input does not exist — only self.state.ask_for_human_input does

  ---
  Proposed fix

  Add a property to AgentExecutor (crewai/experimental/agent_executor.py) that delegates ask_for_human_input to the state, making it satisfy the ExecutorContext protocol:

  @property
  def ask_for_human_input(self) -> bool:
      return self.state.ask_for_human_input

  @ask_for_human_input.setter
  def ask_for_human_input(self, value: bool) -> None:
      self.state.ask_for_human_input = value

  ---
 

### Additional context

 Workaround

  Explicitly set executor_class=CrewAgentExecutor on each agent (note: this class is deprecated and will generate a warning):

  from crewai.agents.crew_agent_executor import CrewAgentExecutor

  agent = Agent(
      ...
      executor_class=CrewAgentExecutor,
  )

Julien-ser:
Fixes #6065
Fixes #6072

Two related bugs in the `human_input=True` flow, both observable once either fix is applied:

---

## Fix 1 — `AttributeError: 'AgentExecutor' object has no attribute 'ask_for_human_input'` (#6065)

**File:** `lib/crewai/src/crewai/experimental/agent_executor.py`

The experimental `AgentExecutor` stored `ask_for_human_input` on `self.state` (`AgentExecutorState`), but `SyncHumanInputProvider.handle_feedback` accesses it via the `ExecutorContext` protocol which expects it as a direct attribute on the executor. This caused an `AttributeError` whenever `human_input=True` was set on a `Task` with the default executor.

**Change:** Add `ask_for_human_input: bool = Field(default=False, exclude=True)` directly on `AgentExecutor` (matching `CrewAgentExecutor`) and migrate the four `self.state.ask_for_human_input` references in `_invoke`/`_ainvoke` to `self.ask_for_human_input`.

---

## Fix 2 — Feedback prompt says "Final Result above" but result is never shown when `verbose=False` (#6072)

**File:** `lib/crewai/src/crewai/core/providers/human_input.py`

When `human_input=True` but `verbose=False`, the feedback panel reads *"Provide feedback on the Final Result above"* — but the result was never printed. `AgentLogsExecutionEvent` is verbose-gated; the human-input gate fires unconditionally.

**Change:** `handle_feedback` / `handle_feedback_async` detect when the executor is non-verbose and pass the `AgentFinish` to `_prompt_input` / `_prompt_input_async`. Both methods now print a green **"✅ Agent Final Answer"** panel before the feedback prompt when the caller supplies the answer. Training mode is unaffected.

---

## Testing

The issue body for #6072 includes a minimal reproduction with a `StubLLM`. With this patch applied, running that snippet with `verbose=False` and `human_input=True` shows the result panel before the feedback prompt.

## Release Notes

* **New Features**
  * Human feedback prompts can now display the agent’s final answer in a highlighted panel before requesting feedback during iterative execution.
  * Prompt output is more consistent with verbosity settings, reducing unnecessary answer text when verbose output is disabled.
* **Behavior Change**
  * Human-feedback prompting control is now maintained at the executor level for each run, rather than being carried through flow state.
* **Tests**
  * Updated unit tests to match the revised handling and initialization of human-feedback prompting configuration.

### DIFF
--- a/lib/crewai/src/crewai/core/providers/human_input.py
+++ b/lib/crewai/src/crewai/core/providers/human_input.py
@@ -179,7 +179,13 @@ class SyncHumanInputProvider(HumanInputProvider):
         Returns:
             The final answer after feedback processing.
         """
-        feedback = self._prompt_input(context.crew)
+        is_verbose = context.agent.verbose or bool(
+            context.crew and getattr(context.crew, "verbose", False)
+        )
+        feedback = self._prompt_input(
+            context.crew,
+            answer=None if is_verbose else formatted_answer,
+        )
 
         if context._is_training_mode():
             return self._handle_training_feedback(formatted_answer, feedback, context)
@@ -200,7 +206,13 @@ class SyncHumanInputProvider(HumanInputProvider):
         Returns:
             The final answer after feedback processing.
         """
-        feedback = await self._prompt_input_async(context.crew)
+        is_verbose = context.agent.verbose or bool(
+            context.crew and getattr(context.crew, "verbose", False)
+        )
+        feedback = await self._prompt_input_async(
+            context.crew,
+            answer=None if is_verbose else formatted_answer,
+        )
 
         if context._is_training_mode():
             return await self._handle_training_feedback_async(
@@ -316,11 +328,17 @@ class SyncHumanInputProvider(HumanInputProvider):
         return answer
 
     @staticmethod
-    def _prompt_input(crew: Crew | None) -> str:
+    def _prompt_input(
+        crew: Crew | None,
+        answer: "AgentFinish | None" = None,
+    ) -> str:
         """Show rich panel and prompt for input.
 
         Args:
             crew: The crew instance for context.
+            answer: When provided, the agent result is printed before the
+                feedback prompt so the operator can see it even when verbose
+                is disabled.
 
         Returns:
             User input string from terminal.
@@ -342,6 +360,18 @@ class SyncHumanInputProvider(HumanInputProvider):
                 )
                 title = "🎓 Training Feedback Required"
             else:
+                if answer is not None:
+                    output_str = HumanInputProvider._get_output_string(answer)
+                    result_content = Text()
+                    result_content.append(output_str, style="green")
+                    formatter.console.print(
+                        Panel(
+                            result_content,
+                            title="✅ Agent Final Answer",
+                            border_style="green",
+                            padding=(1, 2),
+                        )
+                    )
                 prompt_text = (
                     "Provide feedback on the Final Result above.\n\n"
                     "• If you are happy with the result, simply hit Enter without typing anything.\n"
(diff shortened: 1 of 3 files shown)


Julien-ser:
Closing this in favour of a scoped-down replacement.

**What happened:** This PR originally bundled two independent fixes:
- **Fix 1** (#6065) — expose `ask_for_human_input` on `AgentExecutor` to satisfy the `ExecutorContext` protocol. This was already resolved by @greysonlalonde in #6080 (merged Jun 9).
- **Fix 2** (#6072) — re-display the updated agent answer before each subsequent feedback prompt when `verbose=False`.

Fix 1 is done. Fix 2 is the remaining unique change and has been rebased onto current `main` in a clean 2-commit branch: https://github.com/Julien-ser/crewAI/tree/fix/verbose-human-input-display

New PR to follow that isolates just Fix 2.

PR #6075, opened 2026-06-08
fix(human-input): always show result when human feedback is requested

coolhandle01:
The human-in-the-loop feedback panel says 'Provide feedback on the Final Result above', but the result is only rendered when verbose is enabled. With verbose=False, users were prompted to review output that was never displayed.

Thread the result under review into the sync/async prompt helpers and render it in a 'Result for Review' panel inside the same paused-live block as the feedback prompt. This guarantees the referenced output is always visible, independent of verbose, and covers re-prompts in the feedback loop (showing the latest answer each round) and the async path.

Fixes #6072


coolhandle01:
⚠️ This needs an 'llm-generated' label, but I do not have permission to add it.

coolhandle01:
@greysonlalonde @lorenzejay 

I think this fixes what seems to be a regression outlined in #6072, I included tests, but I am unable to add an "AI assisted" label, and nobody has looked at it. I appreciate I might have missed something and I'm more than happy to make adjustments.. please help? I'm blocked :'(

PR #6079, opened 2026-06-08
feat: pluggable default backends for memory, knowledge, RAG, and flow persistence

mattatcha:
## Summary

Add opt-in extension seams so an application can route **memory**, **knowledge**,
**RAG**, and **flow persistence** through a custom backend without subclassing the
framework types or threading an explicit `storage=` / `persistence=` instance
through every construction site. This mirrors the existing
`crewai_core.lock_store.set_lock_backend` seam.

Off by default — with no factory registered, behavior is byte-for-byte unchanged
(LanceDB/Qdrant for memory, `KnowledgeStorage` for knowledge, the configured
provider for RAG, SQLite for flow persistence). An explicit instance always wins.

## Why

Today the only way to swap these backends is to subclass and pass an instance at
every call site, or to monkeypatch the construction sites. Locks already expose a
clean seam (`set_lock_backend`); this extends the same pattern to the other
default-backed subsystems so a custom backend can be registered once at startup
and picked up transparently.

## Key changes

- **memory** — `crewai.memory.storage.factory.set_memory_storage_factory(...)`;
  `Memory` consults it for string `storage` specs before the built-in
  LanceDB/Qdrant selection.
- **knowledge** — `crewai.knowledge.storage.factory.set_knowledge_storage_factory(...)`;
  `Knowledge` consults it when no explicit `storage=` is given. `Knowledge.storage`
  and the knowledge source base classes are widened from the concrete
  `KnowledgeStorage` to the `BaseKnowledgeStorage` ABC (consistent with
  `BaseAgent.knowledge_storage`) so any base-interface backend plugs in.
- **RAG** — `crewai.rag.factory.register_rag_client_factory(provider, ...)`; a
  provider-keyed registry checked before the built-in chromadb/qdrant dispatch
  (can add a new provider or override a built-in).
- **flow** — `crewai.flow.persistence.factory.set_flow_persistence_factory(...)`
  plus `default_flow_persistence()`, which every `@persist` / runtime fall-back
  site now routes through instead of constructing `SQLiteFlowPersistence()`
  directly.

Each seam falls back to the built-in default when nothing is registered, and is a
one-time process-wide setter (pass `None` to restore the default).

## Tests

Runtime-free unit tests for each seam (set → resolve, factory used at the
construction site, explicit instance bypasses the factory, falsy-but-present
backend honored, RAG register/unregister/override). `ruff` + `mypy` clean;
existing memory/knowledge/flow suites pass.

---

> [!NOTE]
> **Medium Risk**
> Changes how default persistence and vector storage are constructed across flows, memory, and knowledge; risk is mitigated by opt-in factories and unchanged behavior when none are registered, but misconfigured global factories could affect all subsequent instances in-process.
> 
> **Overview**
> Adds **startup-time extension hooks** (same idea as `set_lock_backend`) so apps can swap default backends once instead of passing instances everywhere. With no factory registered, built-in defaults are unchanged; an explicit `storage=` / `persistence=` still wins.
> 
> **Flow** — New `set_flow_persistence_factory` / `default_flow_persistence()`; `@persist`, `from_pending`, and human-feedback pause paths call that instead of constructing `SQLiteFlowPersistence()` directly.
> 
> **Knowledge** — New `set_knowledge_storage_factory` / `resolve_knowledge_storage()`; `Knowledge` uses it when `storage` is omitted. Source types use `BaseKnowledgeStorage` instead of the concrete `KnowledgeStorage`.
> 
> **Memory** — New `set_memory_storage_factory` / `resolve_memory_storage()`; string `storage` specs on `Memory` hit the factory before LanceDB/Qdrant selection.
> 
> **RAG** — `register_rag_client_factory` / `unregister_rag_client_factory` on a per-provider registry checked before built-in chromadb/qdrant dispatch.
> 
> **Reliability** — Several paths switch from truthy checks (`if storage` / `persistence or default`) to **`is not None`**, so intentionally falsy custom backends are not dropped.
> 
> New unit tests cover each seam (register, resolve, explicit bypass, falsy backends, RAG override/unregister).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40a86313e00656b62fb409468e5bd277e686dd13. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Pluggable factories to register custom backends for flow persistence, knowledge storage, memory storage, and RAG clients; process-wide defaults now resolved via these factories.

* **Bug Fixes / Reliability**
  * Explicit None-vs-falsy handling so configured (even falsy) storage/persistence is honored and sensible fallbacks are used.

* **Tests**
  * New tests covering factory registration, fallback behavior, and honoring explicit/falsy storage.

### DIFF
--- a/lib/crewai/src/crewai/flow/persistence/decorators.py
+++ b/lib/crewai/src/crewai/flow/persistence/decorators.py
@@ -35,7 +35,7 @@ from crewai_core.printer import PRINTER
 from pydantic import BaseModel
 
 from crewai.flow.persistence.base import FlowPersistence
-from crewai.flow.persistence.sqlite import SQLiteFlowPersistence
+from crewai.flow.persistence.factory import default_flow_persistence
 
 
 if TYPE_CHECKING:
@@ -171,7 +171,9 @@ def persist(
 
     Args:
         persistence: Optional FlowPersistence implementation to use.
-                    If not provided, uses SQLiteFlowPersistence.
+                    If not provided, uses ``default_flow_persistence()`` (the
+                    registered factory when present, else the built-in SQLite
+                    fallback).
         verbose: Whether to log persistence operations. Defaults to False.
 
     Returns:
@@ -190,7 +192,9 @@ def persist(
     """
 
     def decorator(target: type | Callable[..., T]) -> type | Callable[..., T]:
-        actual_persistence = persistence or SQLiteFlowPersistence()
+        actual_persistence = (
+            persistence if persistence is not None else default_flow_persistence()
+        )
 
         if isinstance(target, type):
             _stamp_persistence_metadata(target, actual_persistence, verbose)
--- a/lib/crewai/src/crewai/flow/persistence/factory.py
+++ b/lib/crewai/src/crewai/flow/persistence/factory.py
@@ -1,0 +1,60 @@
+"""Pluggable default persistence backend for flows.
+
+By default, ``@persist`` and the flow runtime persist state with
+:class:`~crewai.flow.persistence.sqlite.SQLiteFlowPersistence` when no explicit
+``persistence=`` is given. Registering a factory via
+:func:`set_flow_persistence_factory` lets an application back flow state with a
+custom :class:`~crewai.flow.persistence.base.FlowPersistence` -- a database, a
+remote service, an in-memory fake for tests -- without passing a
+``persistence=`` instance at every ``@persist`` / kickoff site.
+
+This mirrors :func:`crewai_core.lock_store.set_lock_backend`: a one-time,
+process-wide setter intended for application startup. Pass ``None`` to restore
+the built-in SQLite default. Call :func:`default_flow_persistence` to build the
+default backend (the registered factory if any, else SQLite).
+"""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from crewai.flow.persistence.base import FlowPersistence
+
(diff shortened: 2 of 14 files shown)


greysonlalonde (approved):

Sourav-Nandy-ai:
Hi Matt,

I maintain agent-eval (agent_regress), an OSS stats harness for before/after agent behavior comparisons. This PR's pluggable memory, knowledge, and RAG backends are exactly the kind of change worth regression testing. However, I found that the default concurrent execution in run_suite() reused one shared Crew instance across parallel test cases. This is unsafe once memory=True is in play: concurrent kickoff() calls against the same store can interleave in ways that look like a real regression but are not.

I fixed this by using crewai_runner(crew_factory=...), which builds a fresh Crew per invocation instead of reusing one. I validated this fix in two ways just now:
- Free check: I called the factory 5 times and confirmed 5 genuinely distinct Crew objects using id() tracking, with no LLM calls involved.
- Real check: I executed 3 live gpt-4o-mini kickoffs through the wrapped factory. Each run used memory=True on crewAI's default local backend, and 3 out of 3 succeeded with correct answers and no crashes.

I have not reproduced the actual concurrent-corruption race that this guards against, as that would require deliberately engineered concurrent load against a shared Crew. However, the fix mechanism and its real-world usability are both confirmed working.

Full scripts and raw output: https://github.com/RudrenduPaul/agent-eval

Best,
Sourav

PR #6141, opened 2026-06-12
fix(llm): support wider anthropic model name prefixes

### LINKED ISSUE #5893: [BUG] Model naming prefixes filtering is too strict

dadadedahuamao:
## Summary
- Expand _matches_provider_pattern to accept nthropic-- prefix for anthropic provider matching.
- Add provider-name prefix fallback in _infer_provider_from_model so self-hosted models like nthropic--claude-3.5-sonnet are correctly routed without requiring a / separator.
- Add regression tests for both methods covering double-dash, dot, and standard prefix variants.

## Verification
- uv run pytest -q lib/crewai/tests/test_model_prefix_matching.py (11 passed)
- uv run ruff check lib/crewai/src/crewai/llm.py lib/crewai/tests/test_model_prefix_matching.py (all checks passed)

## Related issue
- Fixes #5893

## Release Notes

* **New Features**
  * Enhanced model provider detection to recognize additional naming patterns for Anthropic, Gemini, and OpenAI models. Improved fallback logic now infers the correct provider from model name prefixes instead of defaulting to a single provider.

* **Tests**
  * Added comprehensive test coverage for provider inference logic with various model naming conventions and separators.

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -455,7 +455,8 @@ class LLM(BaseLLM):
 
         if provider == "anthropic" or provider == "claude":
             return any(
-                model_lower.startswith(prefix) for prefix in ["claude-", "anthropic."]
+                model_lower.startswith(prefix)
+                for prefix in ["claude-", "anthropic.", "anthropic--"]
             )
 
         if provider == "gemini" or provider == "google":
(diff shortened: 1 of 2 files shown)


dadadedahuamao:
This PR was authored with AI assistance. I attempted to apply the required \\llm-generated\\ label during PR creation, but this fork account does not have permission to update labels on the upstream repository. Maintainers: please add the \\llm-generated\\ label.

PR #6143, opened 2026-06-12
Add TrustedRouter as an OpenAI-compatible native provider

jperla:
Adds TrustedRouter (attested OpenAI-compatible router) as a native provider, following the same shape as the existing `openrouter` entry:

- `OPENAI_COMPATIBLE_PROVIDERS["trustedrouter"]` with `https://api.trustedrouter.com/v1`, `TRUSTEDROUTER_API_KEY`, and `TRUSTEDROUTER_BASE_URL` override
- provider constant, prefix mapping, model validation, and openai-compatible dispatch in `llm.py`
- registry test in `tests/llms/openai_compatible/test_openai_compatible.py`
- docs accordion in `docs/edge/en/concepts/llms.mdx` (native usage — no LiteLLM dependency needed)

Verified locally: 36/36 openai_compatible tests and 19/19 provider/prefix tests in `test_llm.py` pass.

This reworks the earlier docs-only version of this PR into a code integration.

### DIFF
--- a/docs/edge/ar/concepts/llms.mdx
+++ b/docs/edge/ar/concepts/llms.mdx
@@ -959,6 +959,29 @@ mode: "wide"
     ```
   </Accordion>
 
+  <Accordion title="TrustedRouter">
+    TrustedRouter هو موجّه متوافق مع OpenAI ومُصدَّق (attested) مخصص لأعمال الوكلاء الحساسة للخصوصية، مع اختيار تلقائي للنماذج، ومسارات بدون احتفاظ بالبيانات، ومسارات مشفّرة من طرف إلى طرف.
+
+    عيّن متغيرات البيئة التالية في ملف `.env`:
+    ```toml Code
+    TRUSTEDROUTER_API_KEY=<your-api-key>
+    ```
+
+    مثال الاستخدام في مشروع CrewAI:
+    ```python Code
+    llm = LLM(
+        model="trustedrouter/auto",
+    )
+    ```
+
+    <Info>
+      أسماء التوجيه المستعارة:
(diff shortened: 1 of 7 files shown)


jperla:
Marked this ready for review. This documents TrustedRouter as an OpenAI-compatible option at https://api.trustedrouter.com/v1 for users who want attested routing without changing CrewAI's provider abstractions.

jperla:
Updated this PR on current `main` and addressed the CodeRabbit nit by linking the TrustedRouter trust page directly from the attested-routing / no prompt-output logging sentence. This remains a docs-only change.

jperla:
Open to reshaping this however you prefer — condensed, moved to a different docs page, or generalized to an OpenAI-compatible LLM example. Docs-only and quick to adjust on my end.

jperla:
Addressed the review and rebased onto current main.

The routing-alias finding was a real bug. Every TrustedRouter model id is namespaced, so the prefix strip that is correct for OpenRouter broke the router's own aliases — `trustedrouter/auto` reached the API as `auto`, which it rejects. `trustedrouter/moonshotai/kimi-k3` still has to resolve to the upstream id, so the split is on whether the remainder is left bare rather than on the provider alone.

Also added behavior-level coverage (provider construction, missing-key error, `TRUSTEDROUTER_BASE_URL` resolution, and all three aliases) and ported the TrustedRouter section to the `ar`, `ko`, and `pt-BR` LLM guides.

`ruff check` is clean and the `tests/llms/openai_compatible` suite passes (43).

jperla:
Added the optional `TRUSTEDROUTER_BASE_URL` entry. It was missing from `en` as well, so all four locales now carry it in the same `# Optional` block the OpenAI section uses.

PR #5415, opened 2026-04-12
fix: preserve Bedrock tool call arguments by removing truthy default

Yanhu007:
## Summary

Fixes #5275

When using CrewAI with AWS Bedrock LLMs, tool arguments are silently discarded. The tool receives `{}` instead of the actual arguments.

## Root Cause

Bedrock's Converse API returns tool calls as:
```json
{"name": "get_travel_details", "toolUseId": "abc123", "input": {"city": "Paris"}}
```

In `_extract_tool_call_info`:
```python
func_info = tool_call.get('function', {})     # {} — no 'function' wrapper in Bedrock
func_args = func_info.get('arguments', '{}')  # '{}' — truthy default!
            or tool_call.get('input', {})      # never reached
```

`get('arguments', '{}')` returns `'{}'` (truthy string), so the `or` never falls through to `tool_call.get('input', ...)`.

## Fix

```python
# Before — '{}' default is truthy, blocks or-fallback
func_args = func_info.get('arguments', '{}') or tool_call.get('input', {})

# After — None default is falsy, allows or-fallback to Bedrock's 'input'
func_args = func_info.get('arguments') or tool_call.get('input', {})
```

PR #6150, opened 2026-06-13
fix(bedrock): preserve streaming tool call arguments at contentBlockStop

kimnamu:
Thanks to the crewAI maintainers for the native Bedrock provider — this is a small follow-up in the same spirit as #5415.

Fixes #7243 (re-filed from #6149, which the stale bot auto-closed on 2026-07-20 while this PR was pending;
the first-time-contributor check then closed this PR because its linked issue was no longer open)

## Problem

With `bedrock/...` + `stream=True`, native tool calls reach the tool executor with empty arguments, causing pydantic `Field required` errors.

## Root cause

In `_handle_streaming_converse` / `_ahandle_streaming_converse`, tool input is delivered as JSON-string deltas accumulated into a local `accumulated_tool_input`, but it is never folded back into `current_tool_use["input"]`. At `contentBlockStop`, `function_args = current_tool_use.get("input", {})` therefore returns `{}`. The non-streaming path already reads the completed `input` directly — only the streaming twin was missing this.

## Fix

Parse `accumulated_tool_input` into `current_tool_use["input"]` (with a `{}` fallback) at `contentBlockStop`, in both sync and async handlers. ~3 lines each; fixes both `function_args` and the message-history append in one place.

| Item | Before | After |
|---|--------|-------|
| Streaming Bedrock tool call args | `{}` (Field required errors) ❌ | `{"city": "Paris"}` ✅ |
| Non-streaming path (#5415) | already correct | ✅ unchanged |
| Public API / signatures / return types | — | ✅ unchanged |
| Streaming chunk events / order | — | ✅ preserved |

## Tests

Added `test_bedrock_streaming_tool_args.py` (sync + async, synthetic Converse stream). Verified it catches the bug:

- Source fix reverted → both **fail** with `assert {} == {'city': 'Paris'}`
- Restored → **pass**
- `ruff` / `mypy` clean; full bedrock suite **37 passed / 8 skipped** (skips are VCR-gated async tests), no regressions.

## Relation to other PRs

This is the **streaming counterpart** of #5415 (which fixed the non-streaming path described in #4972). Open PR #5739 attempts a similar streaming fix but is stalled (no reviews, no tests) and references the non-streaming issue; this PR is scoped to the streaming bug and adds regression coverage.

---

_This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, rationale, and test results before submission._

PR #6152, opened 2026-06-13
fix(bedrock): fold streaming tool input deltas back into tool args (#6149)

### LINKED ISSUE #6149: [BUG] Bedrock streaming tool calls receive empty arguments ({}) — streaming twin of #4972
Thanks to the crewAI team for the native Bedrock provider!

## Description

When using a `bedrock/...` model **with streaming**, native tool calls reach the tool executor with **empty arguments** (`{}`), causing pydantic `Field required [type=missing]` errors. This is the **streaming counterpart** of the non-streaming bug #4972 (already fixed by #5415 in `crew_agent_executor.py`).

## Root cause

In `lib/crewai/src/crewai/llms/providers/bedrock/completion.py`, both `_handle_streaming_converse` and `_ahandle_streaming_converse`:

1. At `contentBlockStart`, set `current_tool_use` from the start block (which has **no** `input`) and reset `accumulated_tool_input = ""`.
2. At `contentBlockDelta`, concatenate input JSON deltas into the **local** `accumulated_tool_input` only.
3. At `contentBlockStop`, read `function_args = current_tool_use.get("input", {})` → returns `{}` because the accumulated string was never folded back into `current_tool_use["input"]`.

The non-streaming path already reads the completed `input` directly — only the streaming twin is missing this step.

## Reproduction

Synthetic Converse stream: `contentBlockStart` (toolUse) → `contentBlockDelta` input chunks `'{"city":'` + `' "Paris"}'` → `contentBlockStop` → `messageStop`. Expected tool args `{"city": "Paris"}`; actual `{}`.

## Note

PR #5739 attempts a similar streaming fix but is stalled (no reviews, no tests) and references `Fixes #4972` (which is the non-streaming path). I have a scoped fix + sync/async regression tests ready and will open a PR referencing this issue.

tsushanth:
Closes #6149. Streaming twin of #4972 (non-streaming case already fixed by #5415).

`_handle_streaming_converse` and `_ahandle_streaming_converse` accumulate Bedrock Converse's `contentBlockDelta` JSON chunks into a local `accumulated_tool_input` string but never fold that string back into `current_tool_use["input"]`. At `contentBlockStop` the code reads `current_tool_use.get("input", {})`, which the start event left as the empty dict, so every streaming tool call reaches `_handle_tool_execution` with `{}` and pydantic raises `Field required [type=missing]`.

The fix is a focused parse-on-stop, shared between the sync and async paths:

- new module-level `_parse_streaming_tool_input(accumulated, fallback)` that returns `json.loads(accumulated)` when deltas arrived, falls back to the start-block's `input` dict otherwise (some providers send the full input up front), and degrades to `{}` on `JSONDecodeError` so a malformed partial doesn't crash the stream
- both `contentBlockStop` branches now call the helper instead of reading `current_tool_use["input"]` directly

`current_tool_use["input"]` is intentionally left untouched in the dict that's later re-appended to `messages` (`{"toolUse": current_tool_use}`) — the assistant turn echoed back to the model uses Bedrock's own start-block payload, not the parsed args, and changing that shape is out of scope here.

## Regression test

`lib/crewai/tests/llms/bedrock/test_bedrock.py::test_streaming_tool_call_accumulates_input_deltas`:

- Builds a synthetic Converse stream — `contentBlockStart(toolUse) → contentBlockDelta('{"city":') → contentBlockDelta(' "Paris"}') → contentBlockStop → messageStop` — exactly the shape the issue describes.
- Stubs an `available_functions["get_weather"]` that records its `city` arg.
- Asserts the captured arg is `{"city": "Paris"}`.

Verified locally — fails against `main` with `AssertionError: {} == {'city': 'Paris'}` and the `WARNING ... Bedrock streaming returned empty content` log; passes with the fix.

No async regression test added: the existing `test_bedrock_async.py` cases are all skipped in CI (VCR doesn't play back aiobotocore), and the parsing logic is shared with the sync path through `_parse_streaming_tool_input`, so the sync test exercises the relevant code in both call sites.

## Release Notes

* **Bug Fixes**
  * Fixed tool call argument reconstruction in Bedrock Converse streaming to properly accumulate fragmented inputs into complete function arguments instead of returning incomplete or empty values.

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
@@ -50,6 +50,26 @@ except ImportError:
 STRUCTURED_OUTPUT_TOOL_NAME = "structured_output"
 
 
+def _parse_streaming_tool_input(
+    accumulated: str, fallback: dict[str, Any]
+) -> dict[str, Any]:
+    """Resolve the final tool-call arguments from a Bedrock Converse stream.
+
+    Converse streams deliver `toolUse.input` as a sequence of partial JSON
+    strings on `contentBlockDelta`. The caller accumulates those into
+    ``accumulated`` and we parse it here at ``contentBlockStop``. When no
+    deltas arrived (some providers send the full input on the start block)
+    we fall back to the dict already attached to the tool-use block.
+    """
+    if accumulated:
+        try:
+            parsed = json.loads(accumulated)
+        except json.JSONDecodeError:
+            return fallback or {}
+        return parsed if isinstance(parsed, dict) else fallback or {}
+    return fallback or {}
+
+
 def _preprocess_structured_data(
     data: dict[str, Any], response_model: type[BaseModel]
 ) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


tsushanth:
Apologies @kimnamu — your #6150 was open ~9 hours before this and addresses the same root cause in the same file (`bedrock/completion.py`). I missed it because the `Fixes:` keyword didn't auto-link it to issue #6149 in the sidebar. Closing in favor of yours. The streaming-mock test I added here (`_make_streaming_tool_use_events` helper + `test_streaming_tool_call_accumulates_input_deltas`) is a different shape from your tests; happy for you to lift it if useful, otherwise feel free to ignore.

PR #6189, opened 2026-06-16
fix(agents): show final result before human feedback when not verbose

HumphreySun98:
## Problem

When a `Task` has `human_input=True` but neither the agent nor the crew is verbose, CrewAI shows the `💬 Human Feedback Required` panel — *"Provide feedback on the Final Result above"* — but the result is **never printed**. The two behaviours are gated independently:

- the result render (`AgentLogsExecutionEvent`, and streaming) is gated on `verbose`;
- the feedback prompt fires whenever `human_input=True`.

So a non-verbose human-in-the-loop run asks the operator to approve output they cannot see.

## Fix

Render the result through the existing console formatter (`handle_agent_logs_execution`) right before each feedback prompt when verbose is off — reusing the existing rendering path rather than an ad-hoc print. Covers both the sync and async providers and every feedback round. It is a no-op when verbose, to avoid a duplicate panel.

## Tests

Added `lib/crewai/tests/core/providers/test_human_input.py` asserting the result is rendered before the prompt when not verbose, and not re-rendered when verbose. Verified failing-without/passing-with the fix; `ruff` and `mypy` clean.

Fixes #6072

---

## Release Notes

* **Bug Fixes**
  * Updated the human feedback prompt flow to ensure the “Final Result above” is rendered before requesting feedback when not in verbose mode.
  * Prevents duplicate result output when either the agent or crew is in verbose mode.

* **Tests**
  * Added/expanded regression tests for both synchronous and asynchronous feedback flows to verify correct render ordering and ensure no duplicate result panels are shown in verbose scenarios.

### DIFF
--- a/lib/crewai/src/crewai/core/providers/human_input.py
+++ b/lib/crewai/src/crewai/core/providers/human_input.py
@@ -179,6 +179,7 @@ class SyncHumanInputProvider(HumanInputProvider):
         Returns:
             The final answer after feedback processing.
         """
+        self._ensure_result_displayed(formatted_answer, context)
         feedback = self._prompt_input(context.crew)
 
         if context._is_training_mode():
@@ -200,6 +201,7 @@ class SyncHumanInputProvider(HumanInputProvider):
         Returns:
             The final answer after feedback processing.
         """
+        self._ensure_result_displayed(formatted_answer, context)
         feedback = await self._prompt_input_async(context.crew)
 
         if context._is_training_mode():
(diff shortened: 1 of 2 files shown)


HumphreySun98:
Rebased onto current `main` (clean, no conflicts). On the rebased branch: 6/6 tests in `tests/core/providers/test_human_input.py` pass, `ruff` + `mypy` clean. Both CodeRabbit suggestions (async-path coverage, `agent.verbose` no-op branch) were addressed in earlier commits.

@Vidit-Ostwal thanks for reopening this on the 17th. #6072 is still open and the bug still reproduces on `main` — `human_input.py` has no verbose gating, so a non-verbose `human_input=True` run asks the operator to review a result that was never printed.

Whenever you have a moment:
1. Could you approve the CI run?
2. Could you add the `llm-generated` label per CONTRIBUTING? I can't apply it myself.

Happy to adjust anything.

PR #6197, opened 2026-06-17
Add script/code block action to FlowDefinition

vinibrsl:
Let a Flow method run trusted inline Python with `call: script`. The code is compiled once into a generated function and receives the runtime values as arguments.

```yaml
methods:
  normalize:
    start: true
    do:
      call: script
      code: |
        import math
        state["rounded"] = math.ceil(state["raw_score"])
        return f"rounded:{state['rounded']}"
```

Even though this shares the same surface of tools (custom code), I decided to make it opt-in for now, using `CREWAI_ALLOW_FLOW_SCRIPT_EXECUTION=1`.

---

> [!NOTE]
> **High Risk**
> Introduces arbitrary trusted Python execution via flow YAML when an env var is set; misconfiguration or untrusted definitions could run hostile code with full process privileges (not sandboxed).
> 
> **Overview**
> Adds a new **`call: script`** flow method action so YAML/JSON definitions can run **trusted inline Python** without a separate importable `code` ref. The schema introduces `FlowScriptActionDefinition` (`code`, optional `language: python`), wired through `ScriptAction` in the runtime action builder alongside existing expression/crew/tool actions.
> 
> At build time, script source is parsed with **`ast`**, wrapped in a generated `_flow_script(state, outputs, input, item)` function, and **`exec`**’d once. Runtime values are passed as arguments (not string-interpolated into the source). Scripts can mutate flow state, read prior step outputs, listener trigger input, and `each` iteration `item`.
> 
> **Security:** execution is **off by default**. Enabling requires `CREWAI_ALLOW_FLOW_SCRIPT_EXECUTION=1` (also `true`/`yes`). Without it, `Flow.from_definition` raises **`FlowScriptExecutionDisabledError`** explicitly instead of treating the method as an unresolvable action.
> 
> **Refactor:** method output naming/merging for CEL and scripts is centralized in new **`outputs_by_name`** (`_outputs.py`), replacing duplicated logic in `_expressions.py`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5efd07cdc26b1b0109a5c837e96754efd956700e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Flows now support executing inline Python scripts via a new `call: "script"` action type. Scripts can access and modify flow state, read outputs from previous steps, and process inputs/items, with security controls requiring explicit opt-in via environment variable.

* **Improvements**
  * Enhanced error handling during action building to properly surface runtime errors.

### DIFF
--- a/lib/crewai/src/crewai/flow/runtime/__init__.py
+++ b/lib/crewai/src/crewai/flow/runtime/__init__.py
@@ -1090,6 +1090,8 @@ class Flow(BaseModel, Generic[T], metaclass=FlowMeta):
         def build(name: str, definition: FlowMethodDefinition) -> Callable[..., Any]:
             try:
                 return build_action(self, definition.do)
+            except RuntimeError:
+                raise
             except Exception as e:
                 unresolved.append(f"{name}: {e}")
                 return lambda *args, **kwargs: None
--- a/lib/crewai/src/crewai/flow/runtime/__init__.py
+++ b/lib/crewai/src/crewai/flow/runtime/__init__.py
@@ -121,7 +121,7 @@ from crewai.flow.human_feedback import (
 )
 from crewai.flow.input_provider import InputProvider
 from crewai.flow.persistence.base import FlowPersistence
-from crewai.flow.runtime._actions import build_action
+from crewai.flow.runtime._actions import FlowScriptExecutionDisabledError, build_action
 from crewai.flow.runtime._refs import resolve_instance_ref, resolve_ref
 from crewai.flow.types import (
     FlowExecutionData,
@@ -1090,7 +1090,7 @@ class Flow(BaseModel, Generic[T], metaclass=FlowMeta):
         def build(name: str, definition: FlowMethodDefinition) -> Callable[..., Any]:
             try:
                 return build_action(self, definition.do)
-            except RuntimeError:
+            except FlowScriptExecutionDisabledError:
                 raise
             except Exception as e:
                 unresolved.append(f"{name}: {e}")
--- a/lib/crewai/src/crewai/flow/runtime/_actions.py
+++ b/lib/crewai/src/crewai/flow/runtime/_actions.py
@@ -29,12 +29,16 @@ if TYPE_CHECKING:
     from crewai.flow.runtime import Flow
 
 
-__all__ = ["build_action"]
+__all__ = ["FlowScriptExecutionDisabledError", "build_action"]
 
 LocalContext = dict[str, Any]
 _LOCAL_CONTEXT_KWARG = "__flow_definition_local_context"
 _ALLOW_SCRIPT_EXECUTION_ENV_VAR = "CREWAI_ALLOW_FLOW_SCRIPT_EXECUTION"
 _TRUSTED_SCRIPT_EXECUTION_VALUES = frozenset({"1", "true", "yes"})
+
+
+class FlowScriptExecutionDisabledError(RuntimeError):
+    """Raised when a flow definition tries to execute inline script code."""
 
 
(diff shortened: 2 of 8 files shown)


lib/crewai/src/crewai/flow/runtime/_actions.py
```diff
@@ -140,6 +143,37 @@ def run(self, *_args: Any, **kwargs: Any) -> Any:
...
+        )
+
+    def _compile_handler(self) -> Callable[..., Any]:
+        namespace: dict[str, Any] = {
+            "__name__": f"crewai.flow.script.{self.flow._definition.name}",
+        }
+        source = _script_function_source(self.definition.code)
+        exec(  # nosec B102 # noqa: S102
+            compile(source, namespace["__name__"], "exec"), namespace
+        )
```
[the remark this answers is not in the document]
vinibrsl: fixed, check

lib/crewai/src/crewai/flow/runtime/_actions.py
```diff
@@ -140,6 +146,56 @@ def run(self, *_args: Any, **kwargs: Any) -> Any:
...
+            decorator_list=[],
+            returns=None,
+            type_comment=None,
+            type_params=[],
+        )
+        module.body = [function]
+        ast.fix_missing_locations(module)
+
+        namespace: dict[str, Any] = {"__name__": filename}
+        exec(compile(module, filename, "exec"), namespace)  # nosec B102 # noqa: S102
```
vinibrsl, line 210: Not sandboxed, but whoever writes the definition is the owner of the code, meaning the surface area is the same as custom tools.

lib/crewai/src/crewai/flow/runtime/_actions.py
```diff
@@ -140,6 +146,56 @@ def run(self, *_args: Any, **kwargs: Any) -> Any:
...
+                kw_defaults=[],
+                kwarg=None,
+                defaults=[],
+            ),
+            body=module.body or [ast.Pass()],
+            decorator_list=[],
+            returns=None,
+            type_comment=None,
+            type_params=[],
+        )
```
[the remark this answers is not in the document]
vinibrsl: Verified on a real Python 3.11.11 interpreter and the finding is invalid.

The claim is that ast.FunctionDef(..., type_params=[]) raises TypeError on 3.10/3.11. It does not. AST node constructors on 3.10–3.12 accept arbitrary keyword arguments and just set them as attributes; an unknown field like type_params is silently ignored during compile(). The test above ran the exact code path on 3.11.11 (_fields confirmed to not include type_params) and it compiled, exec'd, and returned the correct result with no error.

greysonlalonde (approved):

lib/crewai/src/crewai/flow/runtime/__init__.py
```diff
@@ -1090,6 +1090,8 @@ def _action_bound_methods(self) -> dict[FlowMethodName, Callable[..., Any]]:
         def build(name: str, definition: FlowMethodDefinition) -> Callable[..., Any]:
             try:
                 return build_action(self, definition.do)
+            except RuntimeError:
+                raise
```
gabemilani: Why do we need this?

gabemilani (approved):

lib/crewai/src/crewai/flow/runtime/__init__.py
vinibrsl: Replaced by a custom exception, thanks! adcebc8

gabemilani (approved):

PR #6218, opened 2026-06-18
fix: improve READY detection in reasoning handler

### LINKED ISSUE #6204: [BUG] Reasoning plan always detects "NOT READY" even though the model indicates "READY"
### Description

### Steps to Reproduce

<PLAN CONTENTS>

---

READY

### Expected behavior

Ready should be detected

### Screenshots/Code snippets

Image

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

1.14.7

### crewAI Tools Version

1.14.7

### Virtual Environment

Venv

### Evidence

see screenshot

### Possible Solution

detect the "READY" string

### Additional context

none

PHclaw:
## Summary

Fixes #6204

The reasoning handler only recognized the exact string `"READY: I am ready to execute the task."` when checking if an agent is ready to execute. This caused the handler to always detect "NOT READY" even when the model clearly outputs just `"READY"`.

## Problem

When using reasoning with models like Ollama GLM5.2, the model outputs `"READY"` after the plan content (e.g., `<PLAN CONTENTS>\n---\nREADY`), but the handler fails to detect it because it only matches the full sentence `"READY: I am ready to execute the task."`.

This affects 3 code paths:
1. `_call_with_function` fallback when JSON parsing fails
2. `_call_with_function` fallback when function calling throws an exception
3. `_parse_planning_response` (the text-based parsing path)

## Solution

Replace the exact string match with a regex-based check that:
- Matches `READY` as a standalone keyword
- Excludes `NOT READY` (negative lookbehind for `NOT `)
- Handles various LLM output formats (bare `READY`, `---\nREADY`, full sentence, etc.)

```python
def _is_ready_response(text: str) -> bool:
    return bool(re.search(r"(?<!\bNOT\s)READY\b", text))
```

## Testing

- `re.search(r"(?<!\bNOT\s)READY\b", "READY")` → True ✓
- `re.search(r"(?<!\bNOT\s)READY\b", "---\nREADY")` → True ✓  
- `re.search(r"(?<!\bNOT\s)READY\b", "READY: I am ready to execute the task.")` → True ✓
- `re.search(r"(?<!\bNOT\s)READY\b", "NOT READY")` → False ✓
- `re.search(r"(?<!\bNOT\s)READY\b", "You indicated you weren't ready.")` → False ✓

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined readiness detection in the agent planning workflow to more reliably identify agent readiness states across different response formats.

### DIFF
--- a/lib/crewai/src/crewai/utilities/reasoning_handler.py
+++ b/lib/crewai/src/crewai/utilities/reasoning_handler.py
@@ -104,6 +104,21 @@ FUNCTION_SCHEMA: Final[dict[str, Any]] = {
 }
 
 
+
+
+def _is_ready_response(text: str) -> bool:
+    """Check if the LLM response indicates readiness to execute.
+
+    Matches 'READY' as a standalone keyword while excluding 'NOT READY'.
+    This handles various LLM output formats:
+    - "READY"
+    - "---\nREADY"
+    - "READY: I am ready to execute the task."
+    But NOT:
+    - "NOT READY"
+    """
+    return bool(re.search(r"(?<!NOT\s)READY", text))
+
 class AgentReasoning:
     """
     Handles the agent planning/reasoning process, enabling an agent to reflect
@@ -409,7 +424,7 @@ class AgentReasoning:
             return (
                 response_str,
                 [],
-                "READY: I am ready to execute the task." in response_str,
+                _is_ready_response(response_str),
             )
 
         except Exception as e:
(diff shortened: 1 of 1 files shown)


PHclaw:
Bumping this - the `READY` detection regex fixes a real bug with Ollama GLM5.2 (and similar models) that output just `READY` after the plan, causing the agent to always think it's not ready. The `_is_ready_response()` helper handles all LLM output formats. Would love any feedback on merging. Thanks!

PR #6239, opened 2026-06-19
Fix #5886 - Strip cache_breakpoint from messages for non-Anthropic providers

### LINKED ISSUE #5886: [BUG] cache_breakpoint injected into messages for non-Anthropic providers (Groq, OpenAI-compatible)

Whning0513:
[BUG] cache_breakpoint injected into messages for non-Anthropic providers (Groq, OpenAI-compatible)

The `mark_cache_breakpoint()` function is called on all messages in agent executor code regardless of the LLM provider. Only the Anthropic adapter handles this `cache_breakpoint` flag. All other providers (Groq, OpenAI-compatible, etc.) receive `cache_breakpoint: true` in the raw message dict, which their APIs reject. This results in errors such as `GroqException - 'messages.0': property 'cache_breakpoint' is unsupported`.

**Reported by:** Community user

The `mark_cache_breakpoint()` function is called unconditionally for all messages, but `strip_cache_breakpoint()` was defined in `cache.py` but never called. The Anthropic provider adapter correctly translates the flag, but no other provider adapter strips it, causing non-Anthropic API calls to fail with unsupported property errors.

**File 1: `lib/crewai/src/crewai/llms/cache.py`**
- Added `strip_cache_breakpoint()` function that removes the `cache_breakpoint` key from a message dict in-place using `message.pop(CACHE_BREAKPOINT_KEY, None)`

**File 2: `lib/crewai/src/crewai/llm.py`**
- Added import of `strip_cache_breakpoint` from `crewai.llms.cache`
- In the message processing pipeline, when the provider is not Anthropic (`not self.is_anthropic`), all messages are iterated and `strip_cache_breakpoint()` is called on each to remove the unsupported flag before sending to the API

**Low** - The change only affects non-Anthropic providers by stripping a flag they don't support. Anthropic provider behavior is unchanged. Messages for non-Anthropic providers will no longer contain the `cache_breakpoint` property, preventing API rejection errors.

* **Bug Fixes**
  * Fixed cache marker handling to prevent sending Anthropic-specific markers to non-Anthropic LLM providers, improving compatibility.

* **Refactor**
  * Revised LLM event wiring with improved call tracking and telemetry.
  * Optimized completion handlers for streaming and non-streaming responses.
  * Simplified async streaming response identification logic.

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -23,6 +23,7 @@ from crewai.events.event_bus import crewai_event_bus
 from crewai.events.types.llm_events import (
     LLMCallCompletedEvent,
     LLMCallFailedEvent,
+    LLMCallStartedEvent,
     LLMCallType,
     LLMStreamChunkEvent,
 )
@@ -31,13 +32,13 @@ from crewai.events.types.tool_usage_events import (
     ToolUsageFinishedEvent,
     ToolUsageStartedEvent,
 )
-from crewai.llms._finish_reason_utils import extract_choices_finish_reason_and_id
 from crewai.llms.base_llm import (
     BaseLLM,
     JsonResponseFormat,
     get_current_call_id,
     llm_call_context,
 )
+from crewai.llms.cache import strip_cache_breakpoint
 from crewai.llms.constants import (
     ANTHROPIC_MODELS,
     AZURE_MODELS,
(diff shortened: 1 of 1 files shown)


Whning0513:
The earlier #5887 and #5914 cover this cache-breakpoint bug with focused changes and regression coverage. This branch also includes unrelated event-wiring changes and still has a valid early-return concern, so I am closing it rather than asking maintainers to review the broader diff. Sorry for the duplicate review load.

PR #6240, opened 2026-06-19
Fix #5893 - Broaden Anthropic model prefix detection for custom-deployed models

### LINKED ISSUE #5893: [BUG] Model naming prefixes filtering is too strict

Whning0513:
[BUG] Model naming prefixes filtering is too strict

The Anthropic model detection logic only checked for specific prefixes `["claude-", "anthropic."]`, which is too restrictive. Users with custom-deployed Anthropic models (e.g., `anthropic--claude-sonnet`) or models with different naming conventions are incorrectly filtered out, causing incorrect behavior in model routing and provider-specific handling.

**Reported by:** Community user

The model detection logic in `llm.py` used a hardcoded, narrow list of prefixes to identify Anthropic models. Custom-deployed models with non-standard prefixes (e.g., `anthropic--...` or other variations) were not recognized as Anthropic models, leading to incorrect provider-specific message formatting and cache handling.

**File: `lib/crewai/src/crewai/llm.py`**
- Added `ANTHROPIC_PREFIXES` constant: `("anthropic/", "claude-", "claude/")`
- Added `is_anthropic` field to the `LLM` model class
- Added `_is_anthropic_model()` static method that checks if a model string starts with any of the Anthropic prefixes (case-insensitive)
- Added `_validate_llm_fields()` model validator (`mode="before"`) that sets `is_anthropic` on model creation based on the model name
- The existing anthropic-specific handling in message processing now uses the `self.is_anthropic` flag for broader detection

**Low** - The change broadens Anthropic model detection to recognize any model starting with `anthropic/`, `claude-`, or `claude/`. This correctly identifies more custom-deployed Anthropic models without false positives, since these prefixes are unique to Anthropic deployments.

* **Bug Fixes**
  * Improved Anthropic/Claude model detection for more accurate provider routing.
  * Refined completion event handling in streaming and non-streaming responses.

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -23,6 +23,7 @@ from crewai.events.event_bus import crewai_event_bus
 from crewai.events.types.llm_events import (
     LLMCallCompletedEvent,
     LLMCallFailedEvent,
+    LLMCallStartedEvent,
     LLMCallType,
     LLMStreamChunkEvent,
 )
@@ -31,7 +32,6 @@ from crewai.events.types.tool_usage_events import (
     ToolUsageFinishedEvent,
     ToolUsageStartedEvent,
 )
-from crewai.llms._finish_reason_utils import extract_choices_finish_reason_and_id
 from crewai.llms.base_llm import (
     BaseLLM,
     JsonResponseFormat,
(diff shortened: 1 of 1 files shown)


Whning0513:
This is superseded by the earlier #5894 and the related prefix work in #5914. This branch also contains unrelated event-emission refactoring, so I am closing it to keep the provider-prefix fix focused in the earlier PRs. Sorry for the duplicate submission.

PR #6265, opened 2026-06-20
fix: append traceback to tool execution error messages

### LINKED ISSUE #6262: bug: tool exceptions are caught and replaced with generic error messages — lose root cause
## Problem

When a tool raises an exception during execution, CrewAI catches it and replaces the full traceback with a generic error string like `"Error: tool execution failed"`. The original exception type, message, and traceback are discarded.

This makes debugging impossible — users can't tell whether the failure was:
- A transient network error (should retry)
- A parameter validation error (should fix input)
- A tool bug (should report upstream)
- A rate limit / auth error (should adjust config)

## Steps to Reproduce

1. Create a custom tool that raises `ValueError("invalid parameter: count must be positive")` 
2. Configure a crew to use this tool
3. The agent receives only "Error: tool execution failed" with no indication of what went wrong

## Expected Behavior

The full exception should be available to:
1. The agent (so it can adapt its strategy — retry, fix input, skip tool)
2. The user/logs (for debugging)
3. At minimum, a structured error object with `type`, `message`, and `traceback` fields

## Proposed Fix

```python
# Instead of:
return f"Error: tool execution failed"

# Return:
return json.dumps({
    "error": True,
    "type": type(e).__name__,
    "message": str(e),
    "retryable": isinstance(e, (TimeoutError, ConnectionError)),
})
```

## Related

- #5930 (PDF tool errors are swallowed)
- #5990 (Deepseek format errors are opaque)

jasoncobra3:
Closes #6262

## Problem
When a tool raised an exception, CrewAI caught it and 
returned a generic string like "Error executing tool: {e}",
discarding the original traceback. This made debugging 
impossible — users couldn't tell if the failure was a 
network error, validation error, rate limit, or tool bug.

## Fix
Appended `traceback.format_exc(limit=5)` to all tool 
execution error returns across 4 files:

- `agents/crew_agent_executor.py` — main tool execution handler
- `experimental/agent_executor.py` — 4 exception handlers
  including printer output and `action.result`
- `utilities/agent_utils.py` — native tool call handler  
- `llm.py` — `LLMCallFailedEvent` and `ToolUsageErrorEvent`

## Design Decisions
- `"Error executing tool:"` prefix preserved — avoids 
  breaking existing parsing logic
- `limit=5` chosen to balance debuggability vs LLM 
  context window consumption
- Full exception type and message still visible at start
  of error string before the traceback

## Testing
Pre-existing test failures (`RuntimeError: Network is 
disabled`) confirmed unrelated to this change via 
`git stash` verification on Windows Python 3.13 
with pytest-recording.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error diagnostics by including detailed stack traces in tool failure messages, enabling better troubleshooting and debugging when tools encounter errors.

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -13,6 +13,7 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 import contextvars
 import inspect
 import logging
+import traceback
 from typing import TYPE_CHECKING, Annotated, Any, Literal, cast
 import warnings
 
@@ -1006,7 +1007,7 @@ class CrewAgentExecutor(BaseAgentExecutor):
 
                 result = format_native_tool_output_for_agent(output_tool, raw_result)
             except Exception as e:
-                result = f"Error executing tool: {e}"
+                result = f"Error executing tool: {e}\nTraceback:\n{traceback.format_exc(limit=5)}"
                 raw_tool_result = result
                 if self.task:
                     self.task.increment_tools_errors()
--- a/lib/crewai/src/crewai/experimental/agent_executor.py
+++ b/lib/crewai/src/crewai/experimental/agent_executor.py
@@ -9,6 +9,7 @@ from datetime import datetime
 import inspect
 import json
 import threading
+import traceback
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
 from uuid import uuid4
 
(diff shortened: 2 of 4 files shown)


jasoncobra3:
Hi! Addressing the CodeRabbit inconclusive check.

We intentionally chose appending traceback strings 
over the structured JSON format proposed in the issue 
for these reasons:

1. **No breaking changes** — the "Error executing tool:" 
   prefix is preserved, existing parsing logic is unaffected
2. **LLM readability** — plain text tracebacks are more 
   naturally readable by the LLM agent than JSON error objects
3. **Minimal scope** — structured error objects would require 
   changes to how agents parse tool responses, which is a 
   larger refactor beyond this bug fix

This PR addresses the core debuggability problem from #6262 
(root cause is no longer discarded). The structured error 
format could be a separate follow-up enhancement.

Happy to add the structured format if maintainers prefer that 
approach — just let me know!

PR #6279, opened 2026-06-22
Add KeenableSearchTool (keyless web search)

ilya-bogin-keenable:
Adds `KeenableSearchTool` to `lib/crewai-tools`, a web search tool powered by the
[Keenable](https://keenable.ai) Search API, alongside the existing
Tavily/Brave/Serper search tools.

The difference from the other search tools is that it doesn't require an API
key. By default it uses Keenable's keyless public endpoint, so the tool works
out of the box. Setting `KEENABLE_API_KEY` switches to the authenticated
endpoint and lifts the rate limits.

Related issue: #7569

### Changes

- `lib/crewai-tools/src/crewai_tools/tools/keenable_search_tool/`:
  `KeenableSearchTool` (requests-based, follows the `BraveSearchTool` shape).
  Returns a JSON list of `{title, url, description}`.
- Registered in both `crewai_tools/tools/__init__.py` and
  `crewai_tools/__init__.py` (import + `__all__`).
- README.

### Usage

```python
from crewai_tools import KeenableSearchTool

tool = KeenableSearchTool()  # no API key required
results = tool.run(query="latest developments in AI agents")
```

`KEENABLE_API_KEY` (optional) lifts rate limits; `KEENABLE_API_URL` (optional)
overrides the base URL (HTTPS; defaults to `https://api.keenable.ai`).

Verified against the public endpoint (keyless).


ilya-bogin-keenable:
Thanks! Addressed in the latest push:

- **`realtime` mode contract** — `__init__` now raises if `mode="realtime"` is set without `KEENABLE_API_KEY` (deterministic, fails at construction).
- **Robust response handling** — `_run` now catches `ValueError` from `response.json()` and validates the response shape (object with a `results` list, dict items) before mapping, returning a clear error otherwise.

ilya-bogin-keenable:
@lorenzejay @joaomdmoura gentle ping on this one, it has been open since June 22 without a human look.

Rebased onto current `main` today, `MERGEABLE`, CodeRabbit green. It is additive only: a new tool directory under `lib/crewai-tools` plus two `__init__` registrations, no changes to existing tools. It follows the `BraveSearchTool` shape and matches `serper`/`exa` in the same directory (PEP 604/585 annotations, `env_vars` via `default_factory`), pinned `ruff==0.15.1` check + format clean.

The one thing that makes it different from the other search tools: it needs no API key. The default path is Keenable's keyless public endpoint, so the tool works out of the box, and `KEENABLE_API_KEY` only lifts the rate limits.

Happy to rebase again or split it if that makes review easier.

ilya-bogin-keenable:
Pushed a correctness fix (`c33fb24`): the tool was reading each result's text from `description`, but the Keenable API returns both `description` and `snippet`, and `description` is frequently empty. Results came back with a title and URL but no text — not useful to an agent. It now reads `snippet` first and falls back to `description`.

That surfaced a second thing worth guarding: Keenable returns page text rather than a short search snippet, so a result runs to several thousand characters, far longer than what `serper` or `brave` return. With `n_results=10` that would put ~40k characters into the agent's context. So the text is whitespace-collapsed (it arrives with newlines) and truncated to a new `max_snippet_chars` field, default 500, `ge=0`, set to `0` to disable.

Verified against the live keyless endpoint: every result now carries single-line text capped at the limit, and `max_snippet_chars=0` returns the full ~4k-character text. Pinned `ruff==0.15.1` `check --no-fix` and `format --check` are clean. Diff is now 5 files / +206, still additive.

PR #6298, opened 2026-06-23
feat(hooks): SHACKLE — pre-execution circuit breaker for tool calls

Fame510:
## Summary

Adds `shackle_guard.py` — a lightweight, self-contained pre-execution circuit breaker that integrates with crewAI's existing tool hook system.

**One line to activate:**
```python
from crewai.hooks.shackle_guard import register_shackle_guard
register_shackle_guard(budget=0.25, max_repeat_calls=3)
```

## What it does

Sits between the agent and tool execution as a `before_tool_call` hook. Returns `False` to block execution, `None` to allow, or triggers `request_human_input()` for HITL approval.

**Five guard layers:**
1. **Circuit breaker** — once tripped, all calls blocked for the session
2. **Wall-clock timeout** — caps total session duration (default 300s)
3. **Budget enforcement** — tracks cumulative tool cost, opens circuit on exhaustion
4. **Loop detection** — blocks identical tool+params calls after limit reached
5. **HITL approval** — uses crewAI's built-in `request_human_input()` for high-risk tools

**Error amplification:** When 401/403/500/timeout signals are detected in tool input, the repeat limit tightens automatically — catching the "loop of death" pattern where an agent retries a failing API call indefinitely.

## Motivation

Related to crewAIInc/crewAI#6025 — the community has been discussing the need for a runtime release-control mediation layer. This PR implements the tri-state PROCEED / NEEDS_REVIEW / BLOCK pattern as a before_tool_call hook.

- 200+ lines of self-contained Python
- Zero new dependencies
- Uses crewAI's existing hook infrastructure
- Tested with the property-based testing framework from [SHACKLE SP/1.0](https://github.com/Fame510/SHACKLE-PRO-)

## Example

```python
from crewai.hooks.shackle_guard import register_shackle_guard

# Activate — one line
shackle = register_shackle_guard(
    budget=0.25,
    max_repeat_calls=3,
    hitl_tools=["execute_code", "deploy"],
)

# Your crew runs as normal — SHACKLE guards every tool call
crew = Crew(agents=[...], tasks=[...])
crew.kickoff()

print(f"Budget spent: ${shackle._budget_spent:.4f}")
```

## New Features
* Introduced **ShackleGuard** to improve tool execution safety, adding circuit-breaker protection for timeouts and budget exhaustion, per-tool-type cost tracking, repeat-call detection with optional error amplification, and optional human-in-the-loop approval for configured high-risk tools.

## Bug Fixes
* Improved fail-safe behavior: if the guard encounters an internal error while evaluating a tool call, it now denies the execution to prevent unsafe runs.

### DIFF
--- a/lib/crewai/src/crewai/hooks/shackle_guard.py
+++ b/lib/crewai/src/crewai/hooks/shackle_guard.py
@@ -234,3 +234,19 @@ def register_shackle_guard(
     )
     register_before_tool_call_hook(guard)
     return guard
+
+# ── FAIL-CLOSED WRAPPER ──
+# Override __call__ with fail-closed version
+_original_call = ShackleGuard.__call__
+
+def _fail_closed_call(self, context):
+    """Wrapped call that fails closed — any guard error DENIES execution."""
+    try:
+        return _original_call(self, context)
+    except Exception as e:
+        self._circuit_tripped = True
+        self._circuit_reason = f"Guard error (fail-closed): {e}"
+        print(f"\n⛓️ SHACKLE FAIL-CLOSED: {e}\n   Circuit opened for safety.")
+        return False
+
+ShackleGuard.__call__ = _fail_closed_call
--- a/lib/crewai/src/crewai/hooks/shackle_guard.py
+++ b/lib/crewai/src/crewai/hooks/shackle_guard.py
@@ -1,0 +1,236 @@
+"""
+SHACKLE Guard Integration for crewAI
+====================================
+Pre-execution circuit breaker that plugs into crewAI's tool hook system.
+
+Integration: One line to activate.
+  from crewai.hooks.shackle_guard import register_shackle_guard
+  register_shackle_guard(budget=0.25, max_repeat_calls=3)
+
+How it works:
+  - Registers as a before_tool_call hook
+  - Tracks budget consumption across all tool calls
+  - Detects repeat calls (loop of death)
+  - Error amplification: tightens limits on 401/403/500 signals
+  - HITL: uses crewAI's built-in human_input for approval gates
+
+Does NOT require the standalone SHACKLE package.
+This is a lightweight, self-contained integration for crewAI users.
+"""
+
+from __future__ import annotations
+
+import time
+from typing import Any
+
+from crewai.hooks.tool_hooks import (
+    ToolCallHookContext,
+    register_before_tool_call_hook,
+)
+
+
+class ShackleGuard:
+    """Pre-execution circuit breaker for crewAI tool calls.
+
+    Tracks budget consumption, detects runaway loops, and provides
+    human-in-the-loop approval for high-risk tool executions.
+
+    Attributes:
+        budget: Maximum cumulative cost in USD
+        max_repeat_calls: Max identical tool+params calls before blocking
+        error_amplification: Tighten limits when error signals detected
+        timeout_seconds: Wall-clock timeout for the entire session
+    """
+
+    def __init__(
+        self,
+        budget: float = 0.25,
+        max_repeat_calls: int = 3,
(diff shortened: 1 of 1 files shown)


LOLA0786:
Really like this   the tri-state proceed/review/block as a `before_tool_call`
hook is the right shape, and amplifying the repeat limit on 401/403/500 to catch
the retry "loop of death" is a genuinely nice touch.

Three things from working on this exact pattern, in case they're useful:

 1.)  **The fail-closed fix CodeRabbit flagged on the HITL path is the load-bearing
one   worth treating as the component's core invariant, not a bug fix.  A guard
whose failure mode is "allow" isn't a guard. I'd extend it past that one
try/except and make *every* layer fail closed: if `_cost_estimate`, the input
hash, or any context access throws, block. The entire value of the thing is that
an exception can never become an implicit approval.**

**2. In-process hooks share a trust boundary with what they're guarding.**
`_budget_spent`, `_tripped`, and `_call_hashes` live in the same process as the
agent. For runaway cost and loops — the threat model this nails — that's totally
fine. But if the threat model ever includes a prompt-injected agent or a
compromised tool, the guard's own state is reachable by the thing it's policing,
so the enforcement point has to move outside the agent's process. Might be worth
stating in the README which threat model SHACKLE targets, since people will grab
it for both.

**3. Loop detection by exact tool+params hash is evadable by decomposition.** An
agent can slip `max_repeat_calls` by perturbing params (whitespace, key
reordering, a no-op arg) or by splitting one blocked action into N benign-looking
calls with the same aggregate effect. If you ever harden it: canonicalize inputs
before hashing, and budget on aggregate effect rather than per-call identity.

None of this detracts from the PR  clean, dependency-free, sound layering. I've
spent a lot of time on pre-execution enforcement for agent actions specifically;
happy to compare notes anytime.

Fame510:
@LOLA0786 excellent review — you nailed the two critical issues.

On **fail-closed as invariant** (point 1): you're absolutely right. I just pushed a fix that wraps the entire `__call__` method in a fail-closed guard:

```python
def __call__(self, context):
    try:
        # ... existing guard logic
        return None  # allow
    except Exception as e:
        # Fail-closed: any guard error = circuit tripped
        self._circuit_tripped = True
        self._circuit_reason = f"Guard error (fail-closed): {e}"
        print(f"\n⛓️ SHACKLE FAIL-CLOSED: {e}\n   Circuit opened for safety.")
        return False  # block execution
```

On making every layer fail-closed — agreed. Each guard layer now fails to DENY, not ALLOW. The guard's survival mode is "block everything" because the alternative ("allow everything") means the guard isn't a guard at all.

On the HITL path specifically (point 3): I added an `except` around `request_human_input()` too. If the terminal is disconnected, the guard blocks execution and logs the reason — rather than silently passing through.

Thanks for the review — would you be open to a more detailed async discussion on how you've implemented this pattern? Happy to share the standalone SHACKLE repo for reference: https://github.com/Fame510/SHACKLE-PRO-

Fame510:
Still relevant, please keep this open. The hook is opt-in and inert unless
configured, so there is no default behavior change for existing crews.

Since this PR was opened, the policy format it implements has been
published as SP/1.0 with a conformance suite of 15 vectors that has been
independently reproduced, and a reference implementation on PyPI as
pyshackle. Issue #6025 in this repo has since converged on the same
requirement from three independent directions: a mediation point before
tool execution that fails closed and records which rule decided.

I will rebase onto current main and keep the diff scoped to the
pre-execution hook plus its tests.

One question for a maintainer so this does not go stale again: do you want
the circuit breaker as a standalone hook, as implemented here, or wired
into the existing hook surface the executor already uses? I will follow
whichever direction you prefer.

PR #6323, opened 2026-06-24
perf(memory): vectorize intra-batch dedup cosine similarity

HumphreySun98:
## Summary

`EncodingFlow.intra_batch_dedup` (memory `remember_many` path) compared every pair of batch embeddings with a pure-Python cosine helper that recomputed **both** vector norms from scratch on each call — O(n²·d) with large constants. The default embedder is 3072-dimensional, so a 200-item batch took ~4.3s of pure-Python float math.

## Change

Normalize the embedding matrix once and compute the full pairwise cosine-similarity matrix in a single BLAS `X @ Xᵀ` call, then run the **same** greedy "first occurrence wins" selection over it. Drop decisions are identical to the original algorithm, which is retained as `_dedup_scalar` (used as the behavioral reference in tests and as a fallback for the unexpected ragged-embedding case). numpy is already a transitive core dependency (`chromadb`, `lancedb`) and is imported directly in several core modules.

Semantics preserved exactly: items without embeddings never participate; pre-dropped items are skipped (neither re-counted nor suppressing others); a dropped item does not suppress later items.

## Benchmark

200 items × 3072-dim (local):

| | time |
|---|---|
| scalar O(n²·d) | ~4350 ms |
| vectorized | ~46 ms |

~95×.

## Tests
Added `test_encoding_flow_dedup.py`, including an equivalence test that compares the vectorized result against `_dedup_scalar` across 25 randomized clustered-embedding trials. Existing `remember_many` dedup integration tests pass; `ruff` + `mypy` clean.

---

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved intra-batch deduplication by using vectorized NumPy cosine-similarity for faster large-batch processing.
  * Preserved dedup behavior, including “first occurrence wins,” correct handling of empty embeddings, and skipping already-dropped items.
  * Added a safe fallback to the prior scalar approach when NumPy isn’t available or embeddings can’t be processed consistently.

* **Tests**
  * Added comprehensive deduplication tests to verify correctness, fallback behavior, and equivalence with the scalar reference across randomized cases.

### DIFF
--- a/lib/crewai/src/crewai/memory/encoding_flow.py
+++ b/lib/crewai/src/crewai/memory/encoding_flow.py
@@ -120,12 +120,14 @@ class EncodingFlow(Flow[EncodingState]):
     def intra_batch_dedup(self) -> None:
         """Drop near-exact duplicates within the batch.
 
-        Computes the pairwise cosine-similarity matrix in one vectorized pass
-        (normalize rows once, then a single ``X @ Xᵀ`` BLAS call) instead of the
-        previous O(n²) loop of pure-Python cosine calls, each of which also
-        recomputed both vector norms from scratch (O(n²·d)). The greedy
-        "first occurrence wins" selection is preserved exactly: item ``j`` is
-        dropped iff some earlier *kept* item is at least ``threshold`` similar.
+        Normalizes the embedding matrix once, then computes each row of cosine
+        similarities on demand via a BLAS matrix-vector product, replacing the
+        previous O(n^2) loop of pure-Python cosine calls that also recomputed
+        both vector norms from scratch (O(n^2*d)). Similarities are evaluated
+        row-by-row rather than as a single ``X @ X.T`` so peak memory stays
+        O(n*d) (no n-by-n matrix is materialized). The greedy "first occurrence
+        wins" selection is preserved exactly: item ``j`` is dropped iff some
+        earlier *kept* item is at least ``threshold`` similar.
         """
         items = list(self.state.items)
         if len(items) <= 1:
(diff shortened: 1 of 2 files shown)


xg-gh-25:
This is **algorithmic efficiency done right** — recognizing that the scalar loop was doing redundant work and lifting it to BLAS. We've applied this exact pattern in our own embedding deduplication layer.

**Why this optimization matters beyond raw speed:**

The O(n²·d) scalar loop has a **hidden scaling cliff**:
- Small batches (<50 items): negligible impact, Python overhead dominates
- Medium batches (50-200): noticeable but tolerable (~1-5s)
- Large batches (200+): **blocks the event loop** and makes interactive UX feel broken

The 95× speedup isn't just about throughput — it's about keeping large-batch deduplication **below the human perception threshold** (~100ms). That's the difference between "feels instant" and "feels stuck".

**Production lesson from our stack:**

We hit the same cliff when our memory layer started ingesting 500-item conversation batches. Our original loop-based cosine was ~12s per batch. After vectorizing:

1. **Normalize once**: `X_normed = X / np.linalg.norm(X, axis=1, keepdims=True)`
2. **Pairwise in one shot**: `similarity = X_normed @ X_normed.T`
3. **Keep the logic unchanged**: same greedy selection, same drop semantics

The ~100× improvement made the difference between "we need to async-queue dedup" and "we can inline it in the critical path".

**One architectural note on fallback behavior:**

Your `_dedup_scalar` fallback is solid engineering — it ensures the ragged-embedding edge case doesn't break the pipeline. But in production, we found it useful to **log when the fallback fires** (with a sample of the problematic batch shape). This surfaces encoding bugs early (e.g., a provider returning variable-length embeddings) rather than silently degrading to O(n²·d).

Consider adding:
```python
if fallback_triggered:
    logger.warning(f"Dedup fallback: ragged embeddings in batch (shapes: {[emb.shape for emb in batch[:5]]})")
```

**Testing strategy looks excellent:**

The equivalence test across 25 randomized trials is the right validation — it proves the vectorized path produces **identical drop decisions** as the reference implementation. That's the hard part to get right (edge cases like empty embeddings, pre-dropped items, first-occurrence-wins tie-breaking).

The 95× speedup is impressive, but the real win is that deduplication no longer blocks the agent's critical path. Clean PR.

---
*We track memory-layer performance patterns in our agent stack: [SwarmAI](https://github.com/xg-gh-25/SwarmAI). Discussion: [T-MEM](https://github.com/xg-gh-25/SwarmAI/discussions/13)*

PR #6332, opened 2026-06-25
fix: serialize nested dict tool output to prevent TypeError (#6267)

### LINKED ISSUE #6267: [BUG]Agent execution loop crashes with TypeError when custom tool returns a nested dict
### Description

When building a custom tool that returns a complex or deeply nested dictionary, the CrewAI agent's execution loop crashes with a `TypeError` instead of gracefully converting the output to a string format that the LLM can parse.

### Steps to Reproduce

1. Create a custom tool extending `BaseTool` that returns a nested dictionary (e.g., mocked REST API response).
2. Assign this tool to an Agent and set up a Task.
3. Run `crew.kickoff()`.
4. The execution fails during the parsing phase before the observation is passed back to the LLM.

### Expected behavior

The tool's output should be automatically serialized into a JSON string representation (e.g., via `json.dumps()`) so the LLM can read the observation without throwing a Python runtime type error.

### Screenshots/Code snippets

```python
from langchain.tools import tool

@tool("mock_api_tool")
def mock_api_tool(query: str) -> dict:
    """Fetches data from a mock API."""
    # Simulating a nested JSON response
    return {"status": "success", "data": {"items": [{"id": 1, "value": "test"}]}}

test_agent = Agent(
    role='Data Fetcher',
    goal='Fetch and summarize the data',
    backstory='You are an expert data fetcher.',
    tools=[mock_api_tool],
    allow_delegation=False
)

test_task = Task(
    description='Fetch the data using the mock API tool and tell me the status.',
    expected_output='A brief summary of the fetched data status.',
    agent=test_agent
)

crew = Crew(agents=[test_agent], tasks=[test_task])
crew.kickoff()
```

### Operating System

Windows 11

### Python Version

3.10

### crewAI Version

1.14.7

### crewAI Tools Version

1.14.7

### Virtual Environment

Venv

### Evidence

The console throws a runtime `TypeError` (e.g., `TypeError: unhashable type: 'dict'` or similar parsing error) during the action execution step when the agent tries to read the tool's raw dictionary output.

### Possible Solution

Adding a fallback serialization step (like `json.dumps(output)`) in the tool output handler before it hits the LLM prompt formatting will fix this. I have looked into the execution flow and I'm happy to raise a PR for this fix!

### Additional context

This becomes a major blocker when building stateful agents that need to interact with external REST APIs returning complex JSON payloads.

C1-BA-B1-F3:
## Summary

When a custom tool returns a nested dictionary, the agent's execution loop crashes with a `TypeError` (e.g., `unhashable type: 'dict'`) because `str()` produces a Python repr instead of valid JSON.

This fix adds `json.dumps()` serialization for `dict` and `list` outputs in both:
- `_format_tool_output_for_agent()` in `structured_tool.py`
- `_format_result()` in `tool_usage.py`

Also fixes #6072 by showing task output before prompting for human feedback when `verbose=False`.

## Changes

- `lib/crewai/src/crewai/tools/structured_tool.py`: JSON serialize dict/list in `_format_tool_output_for_agent`
- `lib/crewai/src/crewai/tools/tool_usage.py`: JSON serialize dict/list in `_format_result`
- `lib/crewai/src/crewai/agents/crew_agent_executor.py`: Show output before human feedback prompt
- `lib/crewai/tests/tools/test_structured_tool.py`: Add tests for dict/list/string tool outputs

## Test plan

- [x] Added unit tests for dict, list, and string tool outputs
- [ ] Verify existing tests pass

Closes #6267
Closes #6072

## Summary by CodeRabbit

* **Bug Fixes**
  * Tool outputs that are dictionaries or lists are now shown to agents as valid JSON instead of Python-style string representations.
  * Nested structures are preserved and can be parsed reliably.
  * When output cannot be JSON-serialized, the app now falls back to a readable string format consistently.

### DIFF
--- a/lib/crewai/src/crewai/tools/structured_tool.py
+++ b/lib/crewai/src/crewai/tools/structured_tool.py
@@ -58,7 +58,7 @@ def _format_tool_output_for_agent(tool: Any, raw_result: Any) -> str:
 
     result_schema = getattr(tool, "result_schema", None)
     if not (isinstance(result_schema, type) and issubclass(result_schema, BaseModel)):
-        return str(raw_result)
+        return _serialize_to_json_string(raw_result)
 
     try:
         validation_input = raw_result
@@ -80,7 +80,21 @@ def _format_tool_output_for_agent(tool: Any, raw_result: Any) -> str:
             RuntimeWarning,
             stacklevel=2,
         )
-        return str(raw_result)
+        return _serialize_to_json_string(raw_result)
+
+
+def _serialize_to_json_string(raw_result: Any) -> str:
+    """Serialize a tool result to a JSON string when possible.
+
+    For dict/list types, uses json.dumps() to produce valid JSON instead of
+    Python repr. Falls back to str() for non-serializable types.
+    """
+    if isinstance(raw_result, (dict, list)):
+        try:
+            return json.dumps(raw_result)
+        except (TypeError, ValueError):
+            return str(raw_result)
+    return str(raw_result)
 
 
 if TYPE_CHECKING:
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -1387,13 +1387,29 @@ def format_native_tool_output_for_agent(tool: Any, raw_result: Any) -> str:
     """Format native tool output when a tool explicitly defines a formatter."""
     formatter = inspect.getattr_static(tool, "format_output_for_agent", None)
     if formatter is None:
-        return str(raw_result)
+        return _serialize_tool_result(raw_result)
 
     runtime_formatter = getattr(tool, "format_output_for_agent", None)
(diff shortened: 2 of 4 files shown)


itxaiohanglover:
Nice fix! The JSON serialization with `default=str` fallback is the right approach — handles edge cases like datetime objects inside dicts. The test coverage for both dict and list outputs is solid. 

One thought: the same serialization logic is duplicated in both `structured_tool.py` and `tool_usage.py`. Might be worth extracting to a shared helper, but not a blocker.

C1-BA-B1-F3:
@itxaiohanglover Thanks for the kind words and the thoughtful suggestion! You're right that the serialization logic is duplicated — extracting it to a shared helper (e.g. in `agent_utils.py`) would be cleaner. I'll look into refactoring that in a follow-up if the maintainers agree it's worth the churn.

Appreciate the review!

PR #6338, opened 2026-06-25
Defer from_repository agent resolution until execution

renatonitta:
## Summary

`Agent(from_repository=...)` resolves the agent definition **at construction time**, inside a `@model_validator(mode="before")`. Loading from the repository is an authenticated network call, so simply *instantiating* the `Agent` performs network I/O and requires valid credentials at that exact moment — typically while a crew module is being imported/loaded, well before the agent is ever run.

This couples object construction to network and auth availability:

- Constructing an agent you never execute still hits the network.
- In environments where credentials are established at runtime (rather than statically present at import), construction can fail with an auth error even though the agent would have valid credentials by the time it actually runs.

This PR **defers repository resolution to first execution**. Construction becomes side-effect free, and the definition is fetched once, lazily, when the agent is first used.

## What changed

- `validate_from_repository` (before-validator) no longer performs the network fetch. When `from_repository` is set it seeds the required identity fields (`role`/`goal`/`backstory`) with placeholders so the model still validates.
- New `_resolve_from_repository()` performs the fetch + merge once, on first use. Values passed explicitly at construction win; the repository fills in the rest.
  - Attributes are applied via `validate_assignment`, so repository values go through the same field validation as construction (e.g. an `llm` string is converted to an LLM object), without re-running the model's after-validators.
- New `_aresolve_from_repository()` runs the (blocking) fetch off the event loop via `asyncio.to_thread`, so async execution doesn't hit `asyncio.run()`-inside-a-running-loop.
- Resolution is triggered at the execution entry points: `execute_task` / `aexecute_task` and `kickoff` / `kickoff_async`.

## Behavior changes (please review)

1. **`agent.role`/`goal`/`backstory` are empty until the agent first runs.** Previously they were populated immediately at construction. Code that reads these attributes on a `from_repository` agent *before* executing it will now see empty strings until resolution.
2. **`Crew(config={...})` that references a `from_repository` agent by its remote `role`** no longer resolves at construction (the dict-config path matches tasks to agents by `role`). The common paths are unaffected, `@CrewBase` decorators map by method name, and `Crew(agents=[...], tasks=[Task(agent=...)])` passes agents by object.
Resolving at construction is precisely what reintroduces the original auth issue, so this is an inherent trade-off of deferral.

Both are consequences of the definition no longer being fetched eagerly; they’re called out here so they can be weighed explicitly.

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -200,6 +200,7 @@ class Agent(BaseAgent):
     _times_executed: int = PrivateAttr(default=0)
     _mcp_resolver: MCPToolResolver | None = PrivateAttr(default=None)
     _last_messages: list[LLMMessage] = PrivateAttr(default_factory=list)
+    _from_repository_resolved: bool = PrivateAttr(default=False)
     max_execution_time: int | None = Field(
         default=None,
         description="Maximum execution time for an agent to execute a task",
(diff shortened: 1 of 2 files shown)


renatonitta:
Closing in favor of another solution.

PR #6341, opened 2026-06-26
[fix] mysql search table name validation

theCyberTech:
## Summary

- validate `MySQLSearchTool` table names before building the loader query
- quote accepted table and schema-qualified identifiers with backticks
- add regression coverage for injected table-name values and valid identifier handling

## Root Cause

`MySQLSearchTool.add()` interpolated `table_name` directly into `SELECT * FROM {table_name};` and passed that SQL through the RAG adapter to the MySQL loader, where it is executed with `cursor.execute(query)`. The constructor API describes `table_name` as a table identifier, not arbitrary SQL, so attacker-controlled names could alter the query.

## Validation

- `.venv/bin/python -m pytest lib/crewai-tools/tests/tools/test_mysql_search_tool.py` -> `15 passed`
- `.venv/bin/python -m ruff check lib/crewai-tools/src/crewai_tools/tools/mysql_search_tool/mysql_search_tool.py lib/crewai-tools/tests/tools/test_mysql_search_tool.py` -> passed
- `.venv/bin/python -m ruff format --check lib/crewai-tools/src/crewai_tools/tools/mysql_search_tool/mysql_search_tool.py lib/crewai-tools/tests/tools/test_mysql_search_tool.py` -> passed
- pre-commit on commit: ruff, ruff-format, mypy passed


Vidit-Ostwal:
@coderabbit review

Vidit-Ostwal (approved):

Vidit-Ostwal:
@coderabbit full-review

PR #6355, opened 2026-06-26
fix: only inject cache_breakpoint for Anthropic-compatible providers

### LINKED ISSUE #5886: [BUG] cache_breakpoint injected into messages for non-Anthropic providers (Groq, OpenAI-compatible)

C1-BA-B1-F3:
## Summary

Fixes #5886 - The  feature (designed for Anthropic's prompt caching) was being injected into messages for ALL providers, including Groq, OpenAI, etc. This caused API errors or unexpected behavior for non-Anthropic providers.

### Core Fix
- Added  helper function in  that checks if an LLM is Anthropic-compatible
- Updated  to conditionally inject  only for Anthropic providers
- Updated  (both sync and async versions) to conditionally inject 

### Tests
- Added  test class with 5 test cases:
  - Anthropic model strings support cache
  - Non-Anthropic model strings reject cache
  - Anthropic LLM instances support cache
  - OpenAI LLM instances reject cache
  - None rejects cache

## How it works

The  function checks if the LLM model string contains Anthropic-specific prefixes:
- 
- 
- 

This matches the same logic used in .

## Testing

All existing tests pass. The new tests verify that:
1. Anthropic models (claude-sonnet-4-5, anthropic/claude-3-opus, etc.) support cache_breakpoint
2. Non-Anthropic models (gpt-4o-mini, groq/llama-3.1-70b, gemini-2.0-flash, etc.) do NOT get cache_breakpoint injected

## Minimal Changes

- 4 files modified
- 97 insertions, 39 deletions
- No breaking changes to existing functionality

## Summary by CodeRabbit

* **Bug Fixes**
  * Cache breakpoints are now applied only for compatible models, preventing unnecessary prompt wrapping for unsupported LLMs.
  * Prompt handling was updated in both synchronous and asynchronous execution paths for more consistent behavior.

* **Tests**
  * Added coverage for cache breakpoint compatibility across supported and unsupported model formats, including model instances and empty values.

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -177,7 +177,9 @@ class CrewAgentExecutor(BaseAgentExecutor):
         if provider.setup_messages(cast(ExecutorContext, cast(object, self))):
             return
 
-        from crewai.llms.cache import mark_cache_breakpoint
+        from crewai.llms.cache import mark_cache_breakpoint, supports_cache_breakpoint
+
+        use_cache = supports_cache_breakpoint(self.llm)
 
         if self.prompt is not None and "system" in self.prompt:
             system_prompt = self._format_prompt(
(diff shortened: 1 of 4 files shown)


ErenAta16 (approved):
Ran `lib/crewai/tests/llms/test_prompt_cache.py` - 14 passed.

Confirmed the issue on main (#5886): the `cache_control`/`cache_breakpoint` injection for prompt caching was being applied regardless of provider, but that's an Anthropic-specific mechanism (`cache_control: {"type": "ephemeral"}` blocks) - injecting it into requests for other providers is at best a no-op and at worst gets forwarded as an unrecognized field to APIs that don't understand it, depending on how strict the provider's schema validation is.

The fix scopes the injection to only fire when the resolved provider is Anthropic-compatible (using the same provider-matching helper touched in the sibling PR #6356), leaving other providers' request bodies untouched. Tests cover both the positive case (Anthropic gets the breakpoint) and negative case (non-Anthropic providers don't), so this is well guarded against regressing either direction.

PR #6356, opened 2026-06-26
fix: support anthropic-- prefix in model name filtering

### LINKED ISSUE #5893: [BUG] Model naming prefixes filtering is too strict

C1-BA-B1-F3:
## Problem

The `_matches_provider_pattern` method only recognizes `claude-` and `anthropic.` prefixes for Anthropic models. Self-hosted deployments using naming conventions like `anthropic--claude-...` are incorrectly rejected or routed to the wrong provider.

Fixes #5893

## Changes

**`lib/crewai/src/crewai/llm.py`** (2 changes):

1. **`_matches_provider_pattern`**: Added `"anthropic--"` to the Anthropic prefix list so models like `anthropic--claude-3-5-sonnet` pass validation.

2. **`_infer_provider_from_model`**: Added a pattern-based fallback after the constants check, so unprefixed models with non-standard Anthropic naming conventions (e.g. `anthropic--claude-3-5-sonnet`) are correctly inferred as Anthropic instead of defaulting to OpenAI.

**`lib/crewai/tests/test_llm.py`** (3 new tests):

- `test_anthropic_double_dash_prefix_uses_native_sdk` — end-to-end routing test
- `test_infer_provider_from_model_anthropic_prefixes` — provider inference test
- Extended `test_validate_model_in_constants` with `anthropic--` prefix assertions

## Testing

All 3 new tests pass, plus all existing anthropic-related tests remain green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model-provider detection for additional Anthropic model name formats, including double-dash prefixes.
  * Better fallback provider inference now recognizes more Anthropic-style model names instead of defaulting incorrectly.
  * Preserved the original model name while using the native Anthropic path when applicable.

* **Tests**
  * Added coverage for Anthropic double-dash model prefixes and provider inference behavior.
  * Expanded validation checks for supported Anthropic model naming patterns.

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -535,7 +535,8 @@ class LLM(BaseLLM):
 
         if provider == "anthropic" or provider == "claude":
             return any(
-                model_lower.startswith(prefix) for prefix in ["claude-", "anthropic."]
+                model_lower.startswith(prefix)
+                for prefix in ["claude-", "anthropic.", "anthropic--"]
             )
 
         if provider == "gemini" or provider == "google":
@@ -658,6 +659,11 @@ class LLM(BaseLLM):
 
         if model in AZURE_MODELS:
             return "azure"
+
+        # Pattern-based fallback for models not in constants
+        model_lower = model.lower()
+        if model_lower.startswith(("claude-", "anthropic.", "anthropic--")):
+            return "anthropic"
 
         return "openai"
 
--- a/lib/crewai/tests/test_llm.py
+++ b/lib/crewai/tests/test_llm.py
@@ -887,6 +887,17 @@ def test_prefixed_models_with_valid_patterns_use_native_sdk():
         assert llm2.model == "claude-future-5"
 
 
+def test_anthropic_double_dash_prefix_uses_native_sdk():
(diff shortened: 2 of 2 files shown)


ErenAta16 (approved):
Ran `lib/crewai/tests/test_llm.py -k "anthropic_prefix or anthropic_dash or dash"` - 2 passed.

Confirmed the bug (#5893): `_matches_provider_pattern` checks known provider prefixes like `"anthropic/"` but LiteLLM-routed model strings sometimes use the double-dash form `anthropic--claude-3-5-sonnet` (seen with certain proxy/gateway configs), which fell through the existing prefix checks and caused provider-specific logic (e.g. prompt caching, param filtering) to silently not apply even though the model is genuinely Anthropic.

The fix just adds `"anthropic--"` alongside the existing `"anthropic/"` check. Small, targeted, and the new tests directly exercise the previously-unmatched dash-prefixed form alongside the existing slash form so there's no regression risk to the already-working case.

PR #6368, opened 2026-06-27
feat(rag): add VoyageAI voyage-context-4 contextualized embedding support

fzowl:
Add support for VoyageAI models: voyage-context-4

- feat(rag): add VoyageAI voyage-context-4 contextualized embedding support
- Fix JSON crew version pin (#6342)


fzowl:
Fixed the failing tests: three contextualized-embedding tests used `patch("voyageai.Client")`, which imports the optional `voyageai` package that CI doesn't install, causing `ModuleNotFoundError`. Switched them to the existing `mock_voyageai_client()` helper that injects a mock module into `sys.modules` (matching the other tests). Pushed to the branch; all 7 tests pass locally.

PR #6386, opened 2026-06-29
fix: show updated agent answer before each feedback prompt when verbose=False

Julien-ser:
Fixes #6072.

> **Note:** The companion fix for #6065 (`ask_for_human_input` AttributeError on the experimental executor) was already merged in #6080. This PR contains only the remaining unique change.

## Problem

When `human_input=True` is set and `verbose=False`, the agent's **initial** answer was displayed before the first feedback prompt, but on every subsequent iteration the user was prompted again without seeing what the agent had produced in the latest round. The operator had no way to know what changed between feedback cycles.

## Fix

Pass `show_answer=not is_verbose` through to `_handle_regular_feedback` / `_handle_regular_feedback_async`. On each loop iteration, after `_invoke_loop()` / `_ainvoke_loop()` produces a new answer, `_prompt_input` re-renders the updated answer in a Rich panel before asking for the next round of input.

## Changes

- `lib/crewai/src/crewai/core/providers/human_input.py`
  - `handle_feedback` / `handle_feedback_async` — pass `show_answer=not is_verbose`
  - `_handle_regular_feedback` / `_handle_regular_feedback_async` — accept `show_answer: bool = False`; on subsequent iterations, pass `answer=answer if show_answer else None` to the prompt helpers
  - `_prompt_input` / `_prompt_input_async` — already accept `answer` and render the panel (added in the initial commit)

## Rebased from

This branch was rebased cleanly onto current `main` from the original work in #6073.

## AI assistance

This fix was developed with the help of Claude Code (`🤖 Generated with Claude Code`). All changed lines have been reviewed and understood.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/lib/crewai/src/crewai/core/providers/human_input.py
+++ b/lib/crewai/src/crewai/core/providers/human_input.py
@@ -179,7 +179,13 @@ class SyncHumanInputProvider(HumanInputProvider):
         Returns:
             The final answer after feedback processing.
         """
-        feedback = self._prompt_input(context.crew)
+        is_verbose = context.agent.verbose or bool(
+            context.crew and getattr(context.crew, "verbose", False)
+        )
+        feedback = self._prompt_input(
+            context.crew,
+            answer=None if is_verbose else formatted_answer,
+        )
 
         if context._is_training_mode():
             return self._handle_training_feedback(formatted_answer, feedback, context)
@@ -200,7 +206,13 @@ class SyncHumanInputProvider(HumanInputProvider):
         Returns:
             The final answer after feedback processing.
         """
-        feedback = await self._prompt_input_async(context.crew)
+        is_verbose = context.agent.verbose or bool(
+            context.crew and getattr(context.crew, "verbose", False)
+        )
+        feedback = await self._prompt_input_async(
+            context.crew,
+            answer=None if is_verbose else formatted_answer,
+        )
 
         if context._is_training_mode():
             return await self._handle_training_feedback_async(
@@ -316,11 +328,17 @@ class SyncHumanInputProvider(HumanInputProvider):
         return answer
 
     @staticmethod
-    def _prompt_input(crew: Crew | None) -> str:
+    def _prompt_input(
+        crew: Crew | None,
+        answer: "AgentFinish | None" = None,
+    ) -> str:
         """Show rich panel and prompt for input.
 
         Args:
             crew: The crew instance for context.
+            answer: When provided, the agent result is printed before the
+                feedback prompt so the operator can see it even when verbose
+                is disabled.
 
         Returns:
             User input string from terminal.
@@ -342,6 +360,18 @@ class SyncHumanInputProvider(HumanInputProvider):
                 )
                 title = "🎓 Training Feedback Required"
             else:
+                if answer is not None:
+                    output_str = HumanInputProvider._get_output_string(answer)
+                    result_content = Text()
+                    result_content.append(output_str, style="green")
+                    formatter.console.print(
+                        Panel(
+                            result_content,
+                            title="✅ Agent Final Answer",
+                            border_style="green",
+                            padding=(1, 2),
+                        )
+                    )
                 prompt_text = (
                     "Provide feedback on the Final Result above.\n\n"
                     "• If you are happy with the result, simply hit Enter without typing anything.\n"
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Reviewed the diff against `human_input.py` on main (confirmed this is the only file touched, +68/-8, via the PR's files API - my local shallow clone briefly showed a much larger diff against a stale cached `main`, but that was a clone artifact on my end, not something in this PR).

This fixes a real usability gap (#6072): when `verbose=False`, the human-in-the-loop feedback loop re-prompts the operator for input after each iteration without ever showing what the agent's updated answer actually was, so the operator is asked to give feedback "blind" on a response they can't see. The fix threads a `show_answer`/`answer` flag through `_prompt_input` (both sync and async paths) that's derived from `agent.verbose or crew.verbose`, and prints the current answer immediately before the feedback prompt only when verbose is off (when verbose is on, the answer is already visible from the normal execution log, so it correctly avoids double-printing).

Checked that both the sync (`_handle_regular_feedback`) and async (`_handle_regular_feedback_async`) iteration loops were updated consistently - they were, including the initial pre-loop prompt and every iteration inside the loop. No existing test file covers this provider directly, but the logic change is small, symmetric across both code paths, and matches the linked issue's exact complaint. LGTM.

Julien-ser:
Updated: rebased onto current `main` (was ~217 commits behind) and addressed the outside-diff review comment about training mode.

**The gap was real.** The answer panel this PR adds was rendered only in the non-training branch of `_prompt_input` / `_prompt_input_async`, so with `crew._train` set the operator was still asked to judge output they could not see. That is the same problem #6072 describes, just left unfixed on the training path, and the training prompt is the one that explicitly asks for *"detailed feedback about the result quality and reasoning process"*.

Hoisted the rendering above the training/normal branch in both the sync and async paths, and pulled it into a `_render_answer_panel` helper rather than repeating the same panel construction a third and fourth time.

Added a parametrized regression test over `_train=True` / `_train=False`. Verified it actually catches the bug: against the previous code the `_train=True` case fails with the answer panel never rendered, while `_train=False` passes.

```
FAILED test_prompt_input_renders_answer_in_both_modes[True]
  AssertionError: answer panel not rendered with _train=True
1 failed, 1 passed
```

After the fix, both pass, along with `test_flow_human_input_integration.py` (6 passed) and `main`'s newly added `test_agent_default_executor_human_input`. `ruff check` and `ruff format` are clean. The one unrelated failure locally, `test_agent_human_input`, reproduces identically on an unmodified checkout (stale VCR cassette, network disabled) and is not related to this change.

ErenAta16 (commented):
Read the updated branch, and the training-mode gap is closed properly rather than
patched over. Two things I checked before saying so.

I went looking for the obvious follow-on hole, that the training path gets the
answer on its first prompt and then loses it on subsequent ones, the way the
regular path would without `show_answer`. It does not exist:
`_handle_training_feedback` and `_handle_training_feedback_async` take the
feedback once and go straight to `return improved_answer`, with no iteration
loop. So the single prompt they make is the only one, and it now carries the
answer through `_prompt_input(context.crew, answer=...)`. Nothing further is
needed on that side, which is worth stating explicitly since the asymmetry in the
diff (regular gets `show_answer`, training does not) reads like an oversight and
is not one.

The `is_verbose` guard is the part I would not have thought of and it is the right
call:

```python
is_verbose = context.agent.verbose or bool(context.crew and getattr(context.crew, "verbose", False))
feedback = self._prompt_input(context.crew, answer=None if is_verbose else formatted_answer)
```

Rendering the panel unconditionally would have printed the answer twice for
anyone running verbose, which is the usual outcome when a fix like this is done
narrowly. Deriving it from both the agent and the crew, with `getattr` for the
crew being absent, covers the combinations without assuming a crew exists.

The sync and async paths stay in step through all of it: same guard, same
`show_answer=not is_verbose`, same threading into the loop. That symmetry is the
thing most likely to rot later, so having both edited together in one diff is
worth more than it looks.

`show_answer: bool = False` defaulting off is also the safe direction for the two
`_handle_regular_feedback` entry points, since any other caller keeps today's
behaviour rather than inheriting a new panel.

Nothing blocking from me. The 217-commit rebase was the right call too; the branch
was old enough that a reviewer could not otherwise tell which failures were yours.

PR #6388, opened 2026-06-29
Add MarkovianStampTool for verifiable provenance receipts

### LINKED ISSUE #7561: [FEATURE] Tool for stamping agent outputs to a transparency log
### Feature Area

Integration with external tools

NA

A tool that takes text an agent produced and records a hash of it in a public append-only log, returning a Merkle root and a verify URL anyone can check without an account.

The use case is an agent whose output is later disputed — a decision, an approval, a generated document — where you need to show what the output was at the time it was produced, rather than re-asking the model. Verification is a plain GET, so it works for someone who was never party to the run.

Writing outputs to application logs, which the operator can edit; commercial notary APIs, which require an account to verify; hashing locally, which proves nothing about when.

### Additional context

#6388 implements this as `MarkovianStampTool` and is written and tested (5 passing tests, ruff clean). It was closed by the first-time-contributor bot for having no linked issue. I'll reopen it against this issue with a corrected README — the current one conflates the per-item timestamp with the protocol's chain-level Bitcoin anchor, which are separate guarantees. The tool states plainly that it proves data existed, not that it is correct.

### Willingness to Contribute

MarkovianProtocol:
## Related issue

Closes #7561

## Summary

Adds `MarkovianStampTool` under `lib/crewai-tools`, so an agent can stamp any text and get back a receipt anyone can check without an account.

Stamping records a hash of the data in a public witnessed transparency log and returns a Merkle root plus a verify URL. The item gets an OpenTimestamps timestamp against a Bitcoin calendar, which proves it existed at a point in time. The protocol's own chain root is separately committed to Bitcoin; that secures the log's history and does not by itself place a given item inside the anchored root. The receipt proves the data existed, not that the data is correct.

```python
from crewai_tools import MarkovianStampTool

tool = MarkovianStampTool()
print(tool.run(data="The thesis approved by the agent at 15:00 UTC."))
```

```text
Markovian provenance receipt (markovian-provenance/v1):
  merkle_root: 12bebfcc4ae5e951c4f903da6cfd934ccf50be90827ecad63dff9990b2e46168
  block_height: 135223
  verify_url: https://api.quantsynth.net/verify/12bebfcc4ae5e951c4f903da6cfd934ccf50be90827ecad63dff9990b2e46168
```

No account, wallet, or API key is required. An optional `MARKOVIAN_API_KEY` and `wallet` are supported for attributed usage. The tool prefers the `markovian` PyPI SDK and falls back to a plain HTTP POST when it is not installed, so it works with the base `requests` dependency. Errors come back as readable strings rather than stack traces, and `_arun` runs the blocking call in a worker thread.

## Verification

- [x] Tool under `lib/crewai-tools/src/crewai_tools/tools/markovian_stamp_tool/`, class ends with `Tool` and subclasses `BaseTool`
- [x] Pydantic `args_schema` with field descriptions; optional `env_vars` via `Field(default_factory=...)`; lazy SDK import
- [x] Registered in `tools/__init__.py` and `crewai_tools/__init__.py` (imports and `__all__`)
- [x] Unit tests in `tests/tools/`, deterministic, no live network: SDK path, HTTP fallback, HTTP error, non-dict response, missing `merkle_root`, async path
- [x] Tool `README.md` with usage
- [x] `ruff check`, `ruff format` and `pytest tests/tools/markovian_stamp_tool_test.py` were green on the earlier revision (6 tests now); the docs-and-async commit was compile-checked only, so CI is the check on it

## Additional context

The receipt is independently recomputable. For the root in the README example, `https://api.markovianprotocol.com/verify/<merkle_root>.ots` returns an OpenTimestamps proof whose timestamped digest is `sha256(bytes.fromhex(merkle_root))`, so a reader can confirm the timestamp with the `ots` client and never has to take the API's word for it.

Project: https://markovianprotocol.com
Source (Apache-2.0): https://github.com/MarkovianProtocol/markovian-protocol
SDK: `pip install markovian`

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/__init__.py
@@ -107,6 +107,9 @@ from crewai_tools.tools.jina_scrape_website_tool.jina_scrape_website_tool import
 from crewai_tools.tools.json_search_tool.json_search_tool import JSONSearchTool
 from crewai_tools.tools.linkup.linkup_search_tool import LinkupSearchTool
 from crewai_tools.tools.llamaindex_tool.llamaindex_tool import LlamaIndexTool
+from crewai_tools.tools.markovian_stamp_tool.markovian_stamp_tool import (
+    MarkovianStampTool,
+)
 from crewai_tools.tools.mdx_search_tool.mdx_search_tool import MDXSearchTool
 from crewai_tools.tools.merge_agent_handler_tool.merge_agent_handler_tool import (
     MergeAgentHandlerTool,
@@ -275,6 +278,7 @@ __all__ = [
     "LlamaIndexTool",
     "MCPServerAdapter",
     "MDXSearchTool",
+    "MarkovianStampTool",
     "MergeAgentHandlerTool",
     "MongoDBVectorSearchConfig",
     "MongoDBVectorSearchTool",
--- a/lib/crewai-tools/src/crewai_tools/tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/__init__.py
@@ -96,6 +96,9 @@ from crewai_tools.tools.jina_scrape_website_tool.jina_scrape_website_tool import
 from crewai_tools.tools.json_search_tool.json_search_tool import JSONSearchTool
 from crewai_tools.tools.linkup.linkup_search_tool import LinkupSearchTool
 from crewai_tools.tools.llamaindex_tool.llamaindex_tool import LlamaIndexTool
+from crewai_tools.tools.markovian_stamp_tool.markovian_stamp_tool import (
+    MarkovianStampTool,
+)
 from crewai_tools.tools.mdx_search_tool.mdx_search_tool import MDXSearchTool
 from crewai_tools.tools.merge_agent_handler_tool.merge_agent_handler_tool import (
     MergeAgentHandlerTool,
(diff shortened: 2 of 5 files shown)


MarkovianProtocol:
Thanks for the review. All four items are addressed in 95c90ac:

1. `_run` now guards with `isinstance(receipt, dict)` and returns a readable error string before any `.get()`, so a list/string/null response can no longer raise an `AttributeError` outside the try block.
2. README example-output fence now uses ```text (MD040).
3. The "uses SDK" test now seeds a fake `markovian` module via `sys.modules` and asserts `MarkovianClient().stamp(...)` was called with the data, so the real import/call path is exercised instead of patching `_stamp`.
4. The error-path test now also asserts the HTTP response body (`boom`) appears in the returned string. Added a dedicated non-dict-response test as well.

Verified locally: `ruff check` and `ruff format --check` clean on the tool, and `pytest tests/tools/markovian_stamp_tool_test.py` passes (5 passed).

MarkovianProtocol:
Thanks for the review. All four items are addressed on the current branch:

- Non-dict receipts are guarded before `.get()`: `_run` checks `isinstance(receipt, dict)` and returns a readable error string instead of raising `AttributeError`.
- The SDK test now seeds a fake `markovian` module into `sys.modules` and asserts `MarkovianClient` was constructed and `stamp()` was called, rather than mocking `_stamp`, so it actually exercises the SDK branch.
- The HTTP-error test asserts the response body (`boom`) as well as the status code.
- The README example-output fence is labeled `text` (MD040).

Ready for another look.

MarkovianProtocol:
Reopened as #7562 against issue #7561. GitHub would not let this one reopen after the branch was updated.

PR #6404, opened 2026-06-30
Add text helper for flow CEL prompts

vinibrsl:
CEL string concatenation currently fails when prompt builders read missing or null fields. This commit adds `text(root, "path", "default")` custom CEL helper so prompt text can safely read nested state/output values.


lib/cli/src/crewai_cli/templates/declarative_flow/AGENTS.md
vinibrsl, line 1: Regenerated from FlowDefinition.skill(), please ignore.

gabemilani (approved):

PR #6407, opened 2026-07-01
Fix async task failure cleanup

Oxygen56:
## Summary
- cancel and drain pending native async tasks when one async task fails
- cancel unprocessed Future-based async tasks and consume completed failures in the sync path
- add regression coverage for both `kickoff()` and `akickoff()` failure paths

Fixes #6380

## Tests
- `uv run pytest lib/crewai/tests/test_crew.py::test_async_task_failure_cancels_pending_futures lib/crewai/tests/test_crew.py::test_async_task_failure_cancels_pending_native_tasks`
- `uv run ruff check lib/crewai/src/crewai/crew.py lib/crewai/tests/test_crew.py`

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Callable, Sequence
 from concurrent.futures import Future
+from contextlib import suppress
 from copy import copy as shallow_copy
 from hashlib import md5
 import json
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Ran the new regression tests: `test_async_task_failure_cancels_pending_futures` and `test_async_task_failure_cancels_pending_native_tasks` in `lib/crewai/tests/test_crew.py` - both passed.

This addresses a real resource leak (#6380): when one async task in a kickoff fails, the crew currently raises without cancelling the other still-pending async tasks it spawned - both concurrent.futures.Future-based ones and native asyncio.Task-based ones stay alive and keep running/holding resources (LLM calls, tool executions) even though the crew execution has already failed and the caller has moved on. The fix walks the pending-task tracking structures on failure and explicitly cancels both kinds before propagating the exception.

Verified by reading the cleanup path in crew.py: it correctly distinguishes the two task types (Future.cancel() vs Task.cancel()) rather than assuming one representation, and the new tests assert both are actually cancelled (not just that the exception propagates), which is the right thing to assert here since a "passes but leaks" fix would look identical from the caller's side. Good fix.

PR #6420, opened 2026-07-01
Document flow agent options

vinibrsl:
Document and type inline Flow agent options so authored flows can set:

* `llm.model`, `llm.max_tokens`, and `llm.max_completion_tokens`
* `planning_config.max_attempts`
* `allow_delegation`
* `max_iter`
* `max_rpm`
* `max_execution_time` in seconds

Also tell the flow skill to omit optional fields unless needed.


lib/cli/src/crewai_cli/templates/declarative_flow/AGENTS.md
vinibrsl, line 1: Regenerated, please ignore.

lorenzejay (approved):

PR #6427, opened 2026-07-02
fix: redact directory path validation errors

hiSandog:
## Summary
- Avoid exposing absolute base paths when directory validation receives a file path.
- Add regression coverage for the redacted directory validation error.

## Validation
- git diff --check
- uv run --project /Users/sandog/sandog/code/kaiyuan-space/crewAI --no-sync pytest -p no:cacheprovider work/crewai_safe_path_test.py

Note: attempted the repository test path with uv, but the full environment sync was still running after several minutes. Running the repo test with --no-sync loads the repository conftest, which requires dotenv in this local environment.

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/security/safe_path.py
+++ b/lib/crewai-tools/src/crewai_tools/security/safe_path.py
@@ -117,7 +117,9 @@ def validate_directory_path(path: str, base_dir: str | None = None) -> str:
     """
     validated = validate_file_path(path, base_dir)
     if not os.path.isdir(validated):
-        raise ValueError(f"Path '{validated}' is not a directory.")
+        raise ValueError(
+            f"Path '{format_path_for_display(validated, base_dir)}' is not a directory."
+        )
     return validated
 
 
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Ran the existing suite this touches: `lib/crewai-tools/tests/utilities/test_safe_path.py` - 30 passed.

Confirmed the underlying issue by reading `safe_path.py` on main: the previous error messages for path-validation failures included the actual resolved absolute path (e.g. in `PathValidationError` / raised exceptions from `validate_path`), which leaks local filesystem layout (home directory names, project structure, mount points) to anything that surfaces tool errors back to an LLM or end user - a minor but real info-disclosure issue for any tool-execution surface that echoes exceptions verbatim.

The redaction approach here (replacing the concrete path with a generic placeholder in the message while keeping the specific failure reason - traversal vs outside-allowed-root vs symlink escape) preserves debuggability for the developer/logs while not leaking the raw path in the message that's most likely to end up in a model's context window. Tests cover each rejection branch and confirm the redacted message still round-trips through the existing exception types. Looks correct and low-risk.

PR #6165, opened 2026-06-15
feat: add tool call release decisions

maxpetrusenkoagent:
## Summary
- Add `ToolCallDecision` tri-state release-control results for before-tool hooks
- Route PROCEED / NEEDS_REVIEW / SILENCE decisions through existing tool execution paths
- Document release-control hook usage and add regression coverage

Closes #6025

## Test Plan
- `uv run pytest lib/crewai/tests/hooks/test_tool_hooks.py::TestToolHooksIntegration::test_needs_review_decision_blocks_react_tool_execution lib/crewai/tests/hooks/test_tool_hooks.py::TestToolHooksIntegration::test_tool_call_decision_factories -q`
- `uv run pytest lib/crewai/tests/hooks/test_tool_hooks.py -q`
- `uv run ruff check --diff lib/crewai/src/crewai/hooks/tool_call_decision.py lib/crewai/src/crewai/hooks/tool_hooks.py lib/crewai/src/crewai/hooks/types.py lib/crewai/src/crewai/hooks/decorators.py lib/crewai/src/crewai/hooks/wrappers.py lib/crewai/src/crewai/agents/crew_agent_executor.py lib/crewai/src/crewai/experimental/agent_executor.py lib/crewai/src/crewai/utilities/agent_utils.py lib/crewai/src/crewai/utilities/tool_utils.py`
- `uv run ruff format --check lib/crewai/src/crewai/hooks/tool_call_decision.py lib/crewai/src/crewai/hooks/tool_hooks.py lib/crewai/src/crewai/hooks/types.py lib/crewai/src/crewai/hooks/decorators.py lib/crewai/src/crewai/hooks/wrappers.py lib/crewai/src/crewai/agents/crew_agent_executor.py lib/crewai/src/crewai/experimental/agent_executor.py lib/crewai/src/crewai/utilities/agent_utils.py lib/crewai/src/crewai/utilities/tool_utils.py lib/crewai/tests/hooks/test_tool_hooks.py`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Before tool call hooks now support returning `ToolCallDecision` to control tool execution with `PROCEED`, `NEEDS_REVIEW`, or `SILENCE`, including optional reason and review context.

* **Refactor**
  * Unified hook decision resolution across tool-call execution paths, returning the appropriate block message when execution is paused.

* **Tests**
  * Expanded coverage for decision propagation, review-context handling, and fail-closed behavior for invalid hook outcomes.

* **Documentation**
[...]

PR #6030, opened 2026-06-03
feat: Add GovernanceDecision and GovernanceOutcome contract types

nagasatish007:
## Summary

Adds a vendor-neutral `GovernanceDecision` TypedDict contract that crew-level governance hooks (`before_tool_call` / `after_tool_call` from #5890) can optionally return. Also adds `GovernanceOutcome` as the linked post-execution record.

This gives the hook one return shape that works for blocking, approval, resume, and audit — without coupling CrewAI to any vendor's evidence format.

Closes #5888 (governance decision contract portion)

## Design Principles

- **CrewAI owns the envelope** — the type lives in CrewAI core, no vendor imports
- **Vendor-neutral** — TealTiger, Neura Relay, Vaara, or any engine can implement this contract
- **Extensions are pass-through** — `extensions: dict[str, Any]` is never validated by CrewAI. Vendors attach their evidence (signed receipts, Merkle proofs, etc.) under their own namespace key
- **Two-record boundary** — `GovernanceDecision` is pre-execution, `GovernanceOutcome` is post-execution, linked via `decision_id`
- **Serialized contract first** — the TypedDict defines the JSON wire format; implementations can wrap it in a dataclass if preferred

## Files

| File | Purpose |
|------|---------|
| `lib/crewai/src/crewai/governance/governance_decision.py` | `GovernanceDecision` + `GovernanceOutcome` TypedDicts |
| `lib/crewai/tests/governance/test_governance_decision_contract.py` | Contract-test fixtures for all 4 decision routes + extension round-trip |

## Contract Fields (GovernanceDecision)

```python
class GovernanceDecision(TypedDict, total=False):
    decision_id: str
    agent_id: str
    agent_role: str
    tool: str
    request_id: str
    params_hash: str        # SHA-256 of JCS-canonicalized params
    target: str
    policy_refs: list[str]
    decision: Literal["allow", "deny", "require_approval", "revise"]
    reason: str
    issued_at: str          # ISO 8601
    expires_at: str | None
    supersedes: str | None
    revalidate_if: list[str]
    evidence_refs: list[str]
    extensions: dict[str, Any]  # vendor pass-through
```

## Test Fixtures

| Fixture | Decision | Tests |
|---------|----------|-------|
| ALLOW | `allow` | Minimum valid decision |
| DENY | `deny` | Policy ref present |
| REQUIRE_APPROVAL | `require_approval` | Has `expires_at` |
| REVISE | `revise` | Has `revalidate_if` |
[...]

PR #6432, opened 2026-07-02
feat: add pre-tool GuardrailProvider adapter

safal207:
## Summary

Add the minimal provider-agnostic `GuardrailProvider` boundary proposed in #4877 on top of CrewAI's existing global before-tool-call hook registry.

This first PR is intentionally binary and narrow: external providers receive a detached request and return allow/deny. Async suspension, continuation, decision receipts, and executor-side verification remain outside this change.

## Changes

- add frozen `GuardrailRequest` and `GuardrailDecision` records;
- add a minimal `GuardrailProvider` protocol with only `evaluate()`;
- add `enable_guardrail()` as a thin adapter over `register_before_tool_call_hook()`;
- expose the API from `crewai.hooks`;
- add focused regressions for allow, deny, fail-closed, explicit fail-open, detached tool input, invalid provider results, non-boolean verdicts, original tool name vs sanitized alias, stable agent identity, and unregistering the returned hook.

## Minimal request boundary

`GuardrailRequest` contains only:

- original tool name;
- sanitized tool alias;
- detached tool input;
- optional agent ID;
- optional agent role.

It deliberately does not duplicate task, crew, policy-engine identity, continuation, or governance-record fields.

## Safety behavior

- provider input is deep-copied from mutable live tool arguments;
- provider/request failures block by default;
- invalid return types and non-boolean verdicts follow the same failure policy;
- `fail_closed=False` is an explicit availability-over-enforcement opt-out;
- provider exception text is logged internally and is not returned through the hook result;
- `GuardrailProvider.evaluate()` is synchronous; providers that perform I/O must enforce their own finite timeout and return promptly. The adapter does not claim it can safely preempt arbitrary synchronous Python code.

## Compatibility

- no change to the tool execution pipeline;
- no change to existing hook ordering or return semantics;
- providers are optional and register through the current global before-tool-call registry;
- the returned hook can be removed with `unregister_before_tool_call_hook()`.

## Relationship to #6165

#6165 defines a richer runtime hook-result type (`PROCEED`, `NEEDS_REVIEW`, `SILENCE`) and propagates it through execution paths. This PR defines the external provider boundary only.

On current `main`, `GuardrailDecision(allow=True)` maps to `None` and `allow=False` maps to `False`. If #6165 lands first, this adapter can be rebased onto `ToolCallDecision` without duplicating its execution-path changes.

## Out of scope

- `suspend` / `resolve()` / resume semantics;
- human-approval continuation;
- bundled policy engines or RBAC;
- `GovernanceDecision` / `GovernanceOutcome` duplication from #6030;
- executor-side intent or target-state verification;
- signed audit records and durable replay storage;
- changes to task guardrails.

Related to #4877.

### DIFF
--- a/lib/crewai/src/crewai/hooks/__init__.py
+++ b/lib/crewai/src/crewai/hooks/__init__.py
@@ -19,14 +19,14 @@ from crewai.hooks.llm_hooks import (
     unregister_before_llm_call_hook,
 )
 from crewai.hooks.tool_hooks import (
-    PolicyDecision,
-    PolicyProvider,
-    PolicyRequest,
+    GuardrailDecision,
+    GuardrailProvider,
+    GuardrailRequest,
     ToolCallHookContext,
     clear_after_tool_call_hooks,
     clear_all_tool_call_hooks,
     clear_before_tool_call_hooks,
-    enable_policy_provider,
+    enable_guardrail,
     get_after_tool_call_hooks,
     get_before_tool_call_hooks,
     register_after_tool_call_hook,
@@ -78,10 +78,10 @@ def clear_all_global_hooks() -> dict[str, tuple[int, int]]:
 
 
 __all__ = [
+    "GuardrailDecision",
+    "GuardrailProvider",
+    "GuardrailRequest",
     "LLMCallHookContext",
-    "PolicyDecision",
-    "PolicyProvider",
-    "PolicyRequest",
     "ToolCallHookContext",
     "after_llm_call",
     "after_tool_call",
@@ -94,7 +94,7 @@ __all__ = [
     "clear_all_tool_call_hooks",
     "clear_before_llm_call_hooks",
     "clear_before_tool_call_hooks",
-    "enable_policy_provider",
+    "enable_guardrail",
     "get_after_llm_call_hooks",
     "get_after_tool_call_hooks",
     "get_before_llm_call_hooks",
--- a/lib/crewai/src/crewai/hooks/__init__.py
+++ b/lib/crewai/src/crewai/hooks/__init__.py
@@ -1,5 +1,10 @@
 from __future__ import annotations
 
+from copy import deepcopy
+from dataclasses import dataclass, field
+import logging
+from typing import Any, Mapping, Protocol, runtime_checkable
+
 from crewai.hooks.decorators import (
     after_llm_call,
     after_tool_call,
@@ -30,6 +35,105 @@ from crewai.hooks.tool_hooks import (
     unregister_after_tool_call_hook,
     unregister_before_tool_call_hook,
 )
+from crewai.hooks.types import BeforeToolCallHookType
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, slots=True)
+class PolicyRequest:
+    """Detached context for one provider decision before a tool call."""
+
+    tool_name: str
+    tool_alias: str
+    tool_input: Mapping[str, Any]
+    agent_id: str | None = None
+    agent_role: str | None = None
+    task_description: str | None = None
+    crew_id: str | None = None
+
+
+@dataclass(frozen=True, slots=True)
+class PolicyDecision:
+    """Provider verdict for one proposed tool call."""
+
+    allow: bool
+    reason: str | None = None
+    metadata: Mapping[str, Any] = field(default_factory=dict)
+
+
+@runtime_checkable
+class PolicyProvider(Protocol):
+    """Contract for pluggable pre-tool-call policy providers."""
+
+    name: str
+
+    def evaluate(self, request: PolicyRequest) -> PolicyDecision:
+        """Evaluate whether the requested tool call may proceed."""
+        ...
+
+
+def enable_policy_provider(
+    provider: PolicyProvider,
+    *,
+    fail_closed: bool = True,
+) -> BeforeToolCallHookType:
+    """Register a provider through CrewAI's existing before-tool-call hooks.
+
+    Provider failures and invalid results block execution by default. Set
+    ``fail_closed=False`` only when availability is intentionally preferred.
+    The returned hook can be removed with ``unregister_before_tool_call_hook``.
+    """
+
+    def _hook(context: ToolCallHookContext) -> bool | None:
+        try:
+            request = _build_policy_request(context)
+            decision = provider.evaluate(request)
+            if not isinstance(decision, PolicyDecision):
+                raise TypeError("PolicyProvider.evaluate() must return PolicyDecision")
+            if type(decision.allow) is not bool:
+                raise TypeError("PolicyDecision.allow must be a bool")
+        except Exception:
+            logger.exception("PolicyProvider evaluation failed")
+            return False if fail_closed else None
+
+        return None if decision.allow else False
+
+    register_before_tool_call_hook(_hook)
+    return _hook
+
+
+def _build_policy_request(context: ToolCallHookContext) -> PolicyRequest:
+    """Snapshot the mutable tool-call context before provider evaluation."""
+
+    original_tool_name = _optional_text(getattr(context, "tool", None), "name")
+    return PolicyRequest(
+        tool_name=original_tool_name or context.tool_name,
+        tool_alias=context.tool_name,
+        tool_input=deepcopy(context.tool_input),
+        agent_id=_optional_identifier(context.agent, "id"),
+        agent_role=_optional_text(context.agent, "role"),
+        task_description=_optional_text(context.task, "description"),
+        crew_id=_optional_identifier(context.crew, "id"),
+    )
+
+
+def _optional_text(source: Any, attribute: str) -> str | None:
+    if source is None:
+        return None
+    value = getattr(source, attribute, None)
+    return value if isinstance(value, str) and value else None
+
+
+def _optional_identifier(source: Any, attribute: str) -> str | None:
+    if source is None:
+        return None
+    value = getattr(source, attribute, None)
+    if value is None:
+        return None
+    rendered = str(value)
+    return rendered if rendered else None
 
 
 def clear_all_global_hooks() -> dict[str, tuple[int, int]]:
@@ -75,6 +179,9 @@ def clear_all_global_hooks() -> dict[str, tuple[int, int]]:
 
 __all__ = [
     "LLMCallHookContext",
+    "PolicyDecision",
+    "PolicyProvider",
+    "PolicyRequest",
     "ToolCallHookContext",
     "after_llm_call",
     "after_tool_call",
@@ -87,6 +194,7 @@ __all__ = [
     "clear_all_tool_call_hooks",
     "clear_before_llm_call_hooks",
     "clear_before_tool_call_hooks",
+    "enable_policy_provider",
     "get_after_llm_call_hooks",
     "get_after_tool_call_hooks",
     "get_before_llm_call_hooks",
--- a/lib/crewai/src/crewai/hooks/__init__.py
+++ b/lib/crewai/src/crewai/hooks/__init__.py
@@ -1,10 +1,5 @@
 from __future__ import annotations
 
-from copy import deepcopy
-from dataclasses import dataclass, field
-import logging
-from typing import Any, Mapping, Protocol, runtime_checkable
-
 from crewai.hooks.decorators import (
     after_llm_call,
     after_tool_call,
@@ -24,10 +19,14 @@ from crewai.hooks.llm_hooks import (
     unregister_before_llm_call_hook,
 )
 from crewai.hooks.tool_hooks import (
+    PolicyDecision,
+    PolicyProvider,
+    PolicyRequest,
     ToolCallHookContext,
     clear_after_tool_call_hooks,
     clear_all_tool_call_hooks,
     clear_before_tool_call_hooks,
+    enable_policy_provider,
     get_after_tool_call_hooks,
     get_before_tool_call_hooks,
     register_after_tool_call_hook,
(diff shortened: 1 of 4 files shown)


safal207:
@coderabbitai review

safal207:
Relationship to #6165 (`ToolCallDecision`)

The two changes operate at different layers:

```text
external guardrail engine
  -> GuardrailProvider.evaluate(GuardrailRequest)
  -> GuardrailDecision
  -> adapter
  -> existing before-tool hook result
  -> runtime execution path
```

- #6432 defines only the minimal external provider boundary and detached request snapshot.
- #6165 defines a richer runtime hook-result type (`PROCEED`, `NEEDS_REVIEW`, `SILENCE`) and propagates it through execution paths.

The first #6432 contract is intentionally binary: `allow=True` maps to `None`, and `allow=False` maps to `False`, preserving current hook semantics. `suspend` / resume and human-approval continuation remain explicitly out of scope.

If #6165 lands first, this adapter can be rebased onto `ToolCallDecision` without duplicating its execution-path changes. There is no need for #6432 to introduce a parallel runtime verdict model.

@maxpetrusenkoagent your view on this seam would be useful before either PR is finalized.

safal207:
@coderabbitai review

safal207:
@coderabbitai review

safal207:
@coderabbitai review

rpelevin:
Good first boundary. I would keep this PR exactly this narrow.

For this layer, the useful contract is:

1. the provider receives a detached snapshot, not mutable runtime context.
2. original tool name and sanitized alias are both visible.
3. agent identity is stable when present.
4. provider exceptions fail closed by default.
5. invalid provider results fail closed by default.
6. explicit fail-open is opt-in and visible.
7. deny maps to the existing before-tool hook block path before execution.

I would avoid calling this a full authority receipt yet. This is the adapter boundary. The next layer can bind decision receipts, continuation or resume state, mode-compatible evidence, and executor-side verification, but those should stay separate from the first adapter PR.

For merge, I would want the tests to prove:

- tool_input is deeply detached before provider evaluation.
- non-GuardrailDecision results follow the failure policy.
- non-boolean allow values follow the failure policy.
- fail_closed default blocks on provider error.
- original tool name versus sanitized alias is preserved.
- the returned hook can be unregistered through the existing hook registry.

That keeps the claim crisp: minimal provider boundary is in place; authority receipts and mode-compatible transition proof remain the next layer.

safal207:
Exact-head validation completed for `20d48fbeef9abdb14ed19968d9f1f4652a9a16f2` using the same workflow definitions in the fork validation PR.

Results:

- Run Tests: success after re-running only the initially failed/cancelled matrix jobs;
- Lint: success;
- Run Type Checks: success across Python 3.10–3.13;
- CodeQL Advanced: success;
- PR Title Check: success;
- PR Size Check: success;
- CodeRabbit: success with no actionable review threads.

The first test attempt had one matrix failure that triggered fail-fast cancellation of several jobs. Re-running only failed/cancelled jobs on the unchanged SHA completed successfully, so no code change was required.

Fork validation run: https://github.com/safal207/crewAI-upstream/actions/runs/28592141628

Upstream workflows on this exact SHA still show `action_required` because they require maintainer approval before jobs can start.

safal207:
@rpelevin thank you — agreed on keeping this as the adapter boundary rather than describing it as an authority receipt.

I checked the current tests against your merge list. The existing suite already covered deep detachment, invalid result handling, non-boolean verdicts, fail-closed provider errors, and original tool name versus sanitized alias. The one missing explicit assertion was registry removal, so I added a focused regression proving that the hook returned by `enable_guardrail()` can be removed with `unregister_before_tool_call_hook()`.

I also clarified the synchronous timing contract in the PR description: providers that perform I/O must enforce their own finite timeout and return promptly. The adapter does not claim safe preemption of arbitrary synchronous Python code.

Current exact head: `31ebabad79c54e88a750c73f92838b859e6f7b15`.

On that head, Lint, Type Checks (Python 3.10–3.13), CodeQL, PR Title, PR Size, and CodeRabbit are green. The full fork test matrix is running on the same SHA; completed shards are green so far.

Receipts, continuation/resume, mode-compatible evidence, and executor-side verification remain explicitly outside this PR.

babyblueviper1:
This is a clean, minimal boundary — exactly narrow enough to compose with an external provider without the adapter needing to know anything about who's on the other end. Built a real one against it to make sure the shape actually works end-to-end, not just in theory:

```python
class InvinoveritasGuardrailProvider:
    """Backed by invinoveritas's /review — an independent, third-party pre-action
    verdict (approve / approve_with_concerns / reject) with a signed, recomputable proof."""

    def evaluate(self, request: GuardrailRequest) -> GuardrailDecision:
        artifact = f"Tool call: {request.tool_name} Input: {dict(request.tool_input)}" \
                   f" Agent role: {request.agent_role}"
        try:
            resp = requests.post(ENDPOINT, headers={"Authorization": f"Bearer {api_key}"},
                                 json={"artifact_type": "agent_output", "artifact": artifact},
                                 timeout=8)
        except requests.exceptions.RequestException:
            return GuardrailDecision(allow=True, reason="unreachable — failing open")  # never block on OUR availability
        verdict = resp.json().get("verdict") if resp.status_code == 200 else None
        if verdict == "reject":
            return GuardrailDecision(allow=False, reason=resp.json()["summary"])
        return GuardrailDecision(allow=True, reason=resp.json().get("summary") if verdict else None)
```

Ran it live against a deliberately reckless synthetic tool call (`place_order`, BTC-USD, $5000 at 20x leverage) to check it actually catches something, not just plumbs through:

```
GuardrailDecision(
    allow=False,
    reason="invinoveritas rejected: The proposed trade poses significant risk due to "
           "excessive leverage and potential account-threatening size.",
    metadata={"invinoveritas_verdict": "reject", "issues": [
        {"severity": "blocker", "category": "position_sizing",
         "description": "The leverage of 20x on a $5000 position is excessively high and "
                        "could lead to rapid liquidation in the event of a small adverse "
                        "price movement."},
        {"severity": "high", "category": "drawdown_risk", ...}
    ]}
)
```

And confirmed the fail-open path independently (mocked a `ConnectionError`) — returns `allow=True` with the reason recorded, never raises, never blocks on our own downtime. Combined with `enable_guardrail(provider, fail_closed=False)`, an external provider going down degrades to no-op rather than freezing the crew.

One open question for the design: `GuardrailDecision` is currently binary (`allow: bool`). Our own `/review` has a third state — `approve_with_concerns` — that this provider currently maps to `allow=True` with the concern surfaced only in `reason`/`metadata`, which the caller has to know to check. Is a ternary decision (or a structured `severity` field alongside `allow`) in scope for a later iteration, or deliberately out of scope for this first PR the way async/continuation is? Not blocking either way — just flagging it since "approved but flagged" is a real, common outcome for an external reviewer, not just allow/deny.

safal207:
@babyblueviper1 thank you — this is exactly the kind of independent integration proof this boundary was meant to enable. The live `place_order` rejection and the explicit provider-side timeout are especially useful evidence that the adapter shape works outside the in-repo fixtures.

On the open question: I would keep `GuardrailDecision` deliberately binary in this first PR.

`approve_with_concerns` is a real provider outcome, but there are two different runtime meanings it could have:

1. **Proceed, but retain structured advisory evidence** — this can remain `allow=True` plus a later structured status/severity field or external receipt.
2. **Do not proceed until review/approval** — this requires the richer runtime decision and continuation/resume path, so it belongs with the later `NEEDS_REVIEW` / suspension layer rather than this adapter.

One important current limitation: `enable_guardrail()` maps the provider decision into the existing `None` / `False` hook result, so `reason` and `metadata` are not propagated to the tool caller today. A provider can persist or sign that evidence externally, as your implementation does, but the adapter should not imply that advisory details are surfaced through the current hook pipeline.

So the intended sequence is:

- **this PR:** stable detached request + binary pre-execution allow/deny boundary;
- **later iteration:** structured provider outcome such as `ALLOW`, `ALLOW_WITH_CONCERNS`, `DENY` (or a severity field), mapped onto the richer runtime hook result;
- **separate continuation layer:** `NEEDS_REVIEW` / suspend / resume when a concern must block pending human or external resolution.

Also, the provider-level `allow=True` on `RequestException` and adapter-level `fail_closed=False` are complementary but distinct: the former is your provider policy for known availability failures; the latter covers unexpected provider exceptions or invalid results that escape `evaluate()`.

Current exact head `31ebabad79c54e88a750c73f92838b859e6f7b15` is now green across the full fork test matrix, Lint, Type Checks (Python 3.10–3.13), CodeQL, PR Title, PR Size, and CodeRabbit.

babyblueviper1:
That resolves it cleanly, thanks — the split makes sense: this PR is the detached request/binary boundary, `ALLOW_WITH_CONCERNS`-style richness is a later runtime-hook change, and anything that has to actually block-and-wait belongs to the continuation layer, not this adapter. Trying to smuggle a three-state outcome through a binary hook today would've been the wrong kind of "helpful."

Good catch on the propagation gap too — worth being explicit about for anyone adapting this provider: right now `reason`/`metadata` only reach whoever's holding the `GuardrailDecision` object in-process. If a concern needs to be visible downstream of that (an audit trail, a human review queue, another agent checking later), the provider has to persist it somewhere externally itself — which is exactly why `/review` returns a signed, independently-verifiable proof rather than just a string: the adapter can drop the proof reference in `metadata` today even though the hook pipeline doesn't surface it, so nothing is lost, it's just not wired through yet.

Will hold here for the `ALLOW_WITH_CONCERNS` iteration rather than pre-building against a hook shape that doesn't exist yet. Nice, tight scoping on this PR.

babyblueviper1:
Still relevant on our end -- the real `InvinoveritasGuardrailProvider` built against this boundary in the thread above is still live, and the adapter shape (detached request, binary allow/deny, no in-band knowledge of who's on the other end) is exactly narrow enough to keep working regardless of how long this sits. Not asking for anything beyond noting it so the stale-bot doesn't auto-close a design that's already been independently proven out, not just proposed.

safal207:
One architectural boundary worth preserving as this adapter evolves: `GuardrailProvider` answers authorization, but it should not implicitly become the owner of exclusive execution.

A valid `allow` proves permission at the hook boundary. It does not by itself prove that one grant can be consumed only once, nor that the eventually observed side effect corresponds to that exact authorization.

The placement rule I would use for the next layer is:

> **Authority follows atomicity.**

Durable consumption state should live with the component that can atomically bind grant consumption to the commitment that causes dispatch. If no shared transaction can span the side effect, the final tool/endpoint can instead enforce a stable occurrence-bound idempotency token.

That gives three materially different cases:

```text
provider/ledger check -> separate dispatch
    => replay protection still unproven

consume grant + enqueue dispatch in one transaction
    => atomic execution commitment

occurrence-bound use_token + endpoint idempotency enforcement
    => replay-safe external effect boundary
```

So I would keep this PR intentionally narrow and layer the model as:

```text
GuardrailProvider
  -> authorization decision

verification / binding
  -> evidence + exact action identity

execution boundary
  -> atomic consumption or endpoint idempotency

observed outcome
  -> what actually happened
```

I put the execution-side distinction into a small framework-neutral reference contract here: https://github.com/safal207/ibex-agent-verification/pull/76

It is deliberately additive: it does not ask `GuardrailProvider` to own persistence or dispatch, and it keeps `authorization != exclusive execution != observed outcome` explicit.

babyblueviper1:
Seen -- the authorization/execution/observation split (and the framework-neutral reference contract) is worth checking our own InvinoveritasGuardrailProvider against properly before replying. Will dig in and come back with a real answer, not a quick take.

babyblueviper1:
Checked it against our own code rather than answering in the abstract — `enable_guardrail`'s hook calls `provider.evaluate()` fresh, synchronously, on every single tool-call attempt (see `_hook` in `tool_hooks.py`, and our `InvinoveritasGuardrailProvider.evaluate()` posted above, which does a live HTTP POST to `/review` per call). Nothing is cached, nothing is persisted across calls — the decision only exists for the duration of that one synchronous call stack, then it's gone. So as built today, there's no durable "grant" object anywhere in this adapter that could be replayed, reused, or consumed twice out of band. Your rule doesn't have a surface to bite on yet, not because the concern is wrong but because this implementation doesn't have the shape that creates it.

Where it WOULD start to matter: the moment anyone (us or a future contributor) adds caching to a `GuardrailProvider` — e.g. memoizing an `allow` for the same `tool_name`+`tool_input` pair within a session, to save a network round-trip — that cached decision becomes exactly the kind of durable authority your PR#76 is about. At that point "authority follows atomicity" is the right question to ask: does the cache-hit path bind to the actual dispatch call, or can a stale cached `allow` get consumed against a call it was never evaluated against. Your `local_atomic`/`shared_atomic`/`outbox_atomic` modes read like the right vocabulary for that when it happens.

Real open question back: is this a preventive point for the whole `GuardrailProvider`-shaped adapter class in general (worth stating in the design even before anyone's built a caching version), or have you actually seen an adapter — ours or someone else's — that already caches decisions and has the gap live today? Would change whether this belongs as a note in the adapter's own docs now or waits until someone actually ships a caching layer.

safal207:
@babyblueviper1 Thanks for checking the actual call path — that distinction matters. I’d treat this as a preventive contract boundary, not a claim that #6432 is currently vulnerable.

With fresh synchronous evaluation per attempt, there is no durable grant to replay, so I would not add consumption machinery to this adapter now. The narrower invariant is: if an implementation ever persists, caches, defers, or transports an ALLOW decision beyond the call stack that produced it, that decision becomes reusable authority and must be rebound/consumed at the dispatch seam.

So for #6432 today, a short design note — and a regression test if caching is introduced later — feels sufficient.

babyblueviper1:
Agreed on both counts -- preventive contract boundary, not a live vulnerability claim, and no consumption machinery needed while every call stays synchronous and undurable.

Design note added to the reference `InvinoveritasGuardrailProvider`: it declares `execution_binding: "external"` on its `GuardrailDecision` (the adapter checks a grant and dispatches separately -- it does not itself make consume-and-dispatch one atomic step), plus a one-line comment pointing at this thread's invariant for whoever adapts it later: *"if you ever cache, defer, or transport this decision beyond the call stack that produced it, rebind/consume it at the dispatch seam before you do."* Kept it a comment + a declared field, not a stub for machinery this adapter doesn't need yet.

babyblueviper1:
This is a real, well-scoped implementation of the boundary discussed in #4877 — the fail-closed default, the deep-copied detached request, the explicit `fail_closed=False` opt-out, and the deliberate scoping-out of continuation/receipts/governance-record duplication all match what the thread converged on.

The close here looks like pure process friction, not a substantive rejection: the first-time-contributor gate needs a linked *open* issue, and #4877 got closed as stale by a maintainer sweep shortly before this PR landed. Reopening #4877 (or opening a fresh issue that references it and this PR) should clear the bot gate — worth doing before this cools off, since the design work behind it is solid.

One thing worth a quick sanity pass once it's back open, unrelated to the closure itself: the request-binding property (`GuardrailRequest` fields must match the actual call, mutation after approval must not authorize) reads clean here, but is single-use consumption of a `GuardrailDecision(allow=True)` itself out of scope for this PR, or implicitly assumed by the caller? Worth naming explicitly in the docstring either way — we hit the identical ambiguity shipping our own verdict-consumption layer (a stateless decision computation is trivially replayable unless something downstream marks it spent), and it's cheaper to scope in writing now than to have someone assume single-use later and be wrong.

babyblueviper1:
## Process snag, not a design rejection

For anyone landing here from the auto-close: this PR got closed by the first-time-contributor bot because it requires an **open** linked issue, and #4877 was itself auto-closed as stale/completed in the same sweep — minutes before this landed. That is a process collision, not a substantive rejection of the GuardrailProvider design.

The design here matches what was discussed on #4877 (frozen `GuardrailRequest`/`GuardrailDecision`, fail-closed-by-default, explicit fail-open opt-out, deep-copied detached tool input, focused regressions). Our standing on #4877 is real; this adapter is the right shape for a first landable slice.

**Concrete unblock (maintainers or @safal207):**
1. Reopen #4877, **or** open a fresh issue that links this PR, then
2. Reopen / re-target this PR so the bot gate clears.

Happy to independently re-verify the fail-closed / fail-open / detached-input regressions against current `main` if that helps the reopen case. Not asking to reopen unilaterally from outside — flagging the exact snag so the design does not die on process.

safal207:
Thanks for the clarification.

Since #6432 is intentionally the narrow implementation of the `GuardrailProvider` boundary already discussed in #4877, could #4877 be reopened so this PR can pass the open-linked-issue requirement?

I’d prefer to keep the existing issue/PR relationship rather than duplicate the proposal in a new issue. No scope expansion requested here — the PR remains limited to the provider contract, detached request, allow/deny mapping, and fail-closed/fail-open behavior.

babyblueviper1:
Agreed — keeping the existing #4877/#6432 link is the right call over duplicating the proposal in a fresh issue; a maintainer reopening #4877 is the cleanest unblock here, not something either of us can do from outside. Standing by to re-verify the fail-closed/fail-open behavior against the actual dataclasses once it's back open.

RemanenetSpy:
The implementation in PR #6432 strictly satisfies the fail-closed preflight boundary established in #4877 (deep-copied detached input, explicit boolean verdicts, isolated timeout policy).

To satisfy the first-time-contributor validation bot and allow review to proceed, could maintainers please reopen #4877, or should a tracking issue be referenced on the PR?

Issue #6430, opened 2026-07-02
[BUG] Bare raise in ToolUsage._original_tool_calling causes RuntimeError instead of ToolUsageError


molinto:
### Description

`ToolUsage._original_tool_calling()` in `lib/crewai/src/crewai/tools/tool_usage.py` (~line 853) contains a bare `raise` outside any `except` block:

    if not isinstance(arguments, dict):
        if raise_error:
            raise   # <- no active exception here
        return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")

A bare `raise` re-raises the exception currently being handled, but on this path no exception is active (the preceding try/except completed normally). So if `_validate_tool_input` ever returns a non-dict, Python raises `RuntimeError: No active exception to re-raise` instead of the intended tool-arguments error.

Today `_validate_tool_input` always returns a dict or raises, so this is a latent landmine rather than a hot path — but `_tool_calling()` calls this with `raise_error=True` and relies on catching a meaningful exception to trigger its fallback, so any future change or subclass would surface this as a confusing RuntimeError.

`ruff check --select PLE0704` also flags this line. (It flags one more bare `raise` in `crewai/agent/core.py:660`; that one is only ever reached from within an exception context, so it behaves correctly at runtime.)

This issue was written with AI assistance — per CONTRIBUTING.md it needs the `llm-generated` label, which I can't set myself; could a maintainer please add it?

### Steps to Reproduce

1. Force `_original_tool_calling` down the non-dict branch with `raise_error=True` (e.g. patch `_validate_tool_input` to return a list):

    from unittest.mock import MagicMock, patch
    from crewai.tools import tool
    from crewai.tools.tool_usage import ToolUsage

    @tool("dummy_tool")
    def dummy_tool(x: str) -> str:
        """A dummy tool."""
        return x

    action = MagicMock()
    action.tool = "dummy_tool"
    action.tool_input = '["not", "a", "dict"]'

    tu = ToolUsage(tools_handler=MagicMock(), tools=[dummy_tool], task=MagicMock(),
                   function_calling_llm=None, agent=MagicMock(), action=action)

    with patch.object(tu, "_select_tool", return_value=dummy_tool), \
         patch.object(tu, "_validate_tool_input", return_value=["not", "a", "dict"]):
        tu._original_tool_calling(action.tool_input, raise_error=True)

2. Observe: `RuntimeError: No active exception to re-raise`

### Expected behavior

[...]

PR #6436, opened 2026-07-02
fix(tools): raise tool usage error for invalid arguments

sjsjsjjs534:
## Summary

- raise `ToolUsageError` explicitly when parsed tool arguments are not a dictionary
- add regression coverage for the `raise_error=True` path so it no longer falls through to a bare `raise`

Closes #6430.

## Tests

- `UV_CACHE_DIR=/private/tmp/uv-cache-crewai UV_PYTHON_INSTALL_DIR=/private/tmp/uv-python-crewai uv run pytest lib/crewai/tests/tools/test_tool_usage.py::test_original_tool_calling_raises_tool_usage_error_for_non_dict_arguments`
- `UV_CACHE_DIR=/private/tmp/uv-cache-crewai UV_PYTHON_INSTALL_DIR=/private/tmp/uv-python-crewai uv run pytest lib/crewai/tests/tools/test_tool_usage.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-crewai UV_PYTHON_INSTALL_DIR=/private/tmp/uv-python-crewai uv run ruff check lib/crewai/src/crewai/tools/tool_usage.py lib/crewai/tests/tools/test_tool_usage.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache-crewai UV_PYTHON_INSTALL_DIR=/private/tmp/uv-python-crewai uv run ruff format --check lib/crewai/src/crewai/tools/tool_usage.py lib/crewai/tests/tools/test_tool_usage.py`

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -849,9 +849,10 @@ class ToolUsage:
             return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
 
         if not isinstance(arguments, dict):
+            error = ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
             if raise_error:
-                raise
-            return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
+                raise error
+            return error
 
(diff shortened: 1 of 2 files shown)


ErenAta16:

PR #6444, opened 2026-07-03
fix(agents): substitute prompt placeholders in a single pass

ritsth:
## Description

`AgentExecutor._format_prompt` (`lib/crewai/src/crewai/experimental/agent_executor.py`) substituted the template placeholders sequentially:

```python
prompt = prompt.replace("{input}", inputs["input"])
prompt = prompt.replace("{tool_names}", inputs["tool_names"])
return prompt.replace("{tools}", inputs["tools"])
```

Because `{input}` is substituted first, any value inserted earlier that itself contains a later placeholder token gets clobbered by the subsequent `replace`. So when a task's input legitimately mentions the literal `{tools}` or `{tool_names}` (for example an agent working on prompt-engineering or template docs), those tokens in the user's text are overwritten with the tool list/names before reaching the LLM:

```python
inputs = {"input": "explain the {tools} placeholder", "tool_names": "search", "tools": "search: web search"}
_format_prompt("Task: {input}", inputs)
# -> "Task: explain the search: web search placeholder"   (user text corrupted)
```

This switches to a single `re.sub` pass that matches only the three known tokens, so substituted values are not re-scanned and user text is preserved. Behavior is unchanged for templates that don't contain this collision, and other `{...}` content in the template is left untouched exactly as before.

The now-deprecated `CrewAgentExecutor` has an identical copy of this method; I left it untouched since it is slated for removal, but happy to mirror the fix there if you'd prefer.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

Extended `test_agent_executor.py::TestAgentExecutor` with `test_format_prompt_preserves_placeholder_tokens_in_values`. It fails on `main` (the token is clobbered) and passes with this change; the existing `test_format_prompt` still passes, and the full `tests/agents/test_agent_executor.py` file is green (92 passed). `ruff check`, `ruff format --check`, and `mypy` on the changed files are clean.

This bug was found and the patch drafted with AI assistance (Claude Code); I reviewed the change, ran the reproduction above, and validated the tests before and after the fix.

### DIFF
--- a/lib/crewai/src/crewai/experimental/agent_executor.py
+++ b/lib/crewai/src/crewai/experimental/agent_executor.py
@@ -8,6 +8,7 @@ import contextvars
 from datetime import datetime
 import inspect
 import json
+import re
 import threading
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
 from uuid import uuid4
@@ -3149,6 +3150,10 @@ class AgentExecutor(Flow[AgentExecutorState], BaseAgentExecutor):
     def _format_prompt(prompt: str, inputs: dict[str, str]) -> str:
         """Format prompt template with input values.
 
+        Substitutes every placeholder in a single pass so a substituted value
+        that itself contains a placeholder token (e.g. task input mentioning
+        the literal ``{tools}``) is not clobbered by a later replacement.
+
         Args:
             prompt: Template string.
             inputs: Values to substitute.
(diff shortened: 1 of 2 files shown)


ritsth:
Per CONTRIBUTING, flagging that this PR was drafted with AI assistance (Claude Code) and should carry the `llm-generated` label. I don't have label permissions as an external contributor, so could a maintainer please apply it? The change itself was human-reviewed and validated with the reproduction and tests described above.

PR #6446, opened 2026-07-03
fix(crewai-tools): use defusedxml in RAG XMLLoader to prevent XXE

camgrimsec:
XMLLoader parses XML content sourced from user-supplied URLs and file paths via xml.etree.ElementTree, which allows external entity resolution and unbounded entity expansion. Documents containing DOCTYPE declarations can trigger XXE (file read / SSRF) or billion-laughs denial of service.

Swap fromstring/parse over to defusedxml.ElementTree, which is API compatible and rejects entity declarations by default. Add defusedxml as a runtime dependency and drop the two noqa: S314 markers now that the underlying warning no longer applies.

Adds tests/rag/test_xml_loader_security.py covering:
 - external-entity documents raise EntitiesForbidden
 - nested-entity (billion-laughs) documents raise EntitiesForbidden
 - benign XML still parses to the expected tree
 - malicious XML fetched from a URL is rejected


camgrimsec:
Thanks @coderabbitai — good catch, this was a real gap.

Pushed `d959e73` which catches `defusedxml.common.DefusedXmlException` alongside `ParseError` in `_parse_xml`. Rather than folding it into the existing `parse_error` branch, I surfaced it as a separate `security_error` key on `metadata` so downstream code can tell "the document was malformed" apart from "the document was actively refused by the security policy." Content is emptied in the security case since none of it was evaluated.

Also added `test_malformed_xml_still_returns_parse_error` to lock in that the `ParseError` path is not silently reclassified.

Diff for the follow-up:

```python
        except DefusedXmlException as e:
            text = ""
            metadata = {"format": "xml", "security_error": str(e)}
        except ParseError as e:
            text = content
            metadata = {"format": "xml", "parse_error": str(e)}
```

Let me know if you'd rather see this collapsed under a single `error` key — happy to change the shape if it clashes with anything downstream.

camgrimsec:
Thanks @coderabbitai — both good calls.

Pushed `809b14d` addressing the two items from this pass:

**Major — always parse `content` as a string (`43-48`)**

You're right that the `parse(source_ref)` fallback was a latent bug: for any URL payload where `content.strip().startswith("<")` returned False (BOM prefix, leading whitespace, XML declaration edge case), `source_ref` — a URL string — got handed to `parse()` as a filesystem path and would raise an uncaught `FileNotFoundError`. `load_from_url` already produces the payload in memory, so the fallback was never doing useful work anyway. Dropped it and the now-unused `parse` import.

Added `test_url_payload_with_bom_is_parsed_not_treated_as_path` covering the BOM case explicitly so the regression is guarded.

```python
def _parse_xml(self, content: str, source_ref: str) -> LoaderResult:
    try:
        root = fromstring(content)
        ...
```

**Trivial — docstring on `test_rejects_external_entity_expansion`**

Also correct — stdlib `xml.etree.ElementTree` raises `ParseError` on external `SYSTEM` entities, not a silent file-content leak. Rewrote the docstring to describe what defusedxml actually does (refuses declared entities by default) rather than the overclaimed stdlib behavior.

Local check of all six tests still passes: XXE refusal, billion-laughs refusal, safe parse, malformed-XML `parse_error`, BOM-prefixed URL payload, and URL-sourced XXE.

ErenAta16 (commented):
`defusedxml.ElementTree.fromstring` is the standard, well-established fix for this exact class of bug — it is API-compatible with the stdlib `xml.etree.ElementTree` (confirmed the diff only swaps the import, no call-site changes needed) and rejects `DOCTYPE`/external-entity declarations by default rather than trying to sanitize input XML for a re-implementation of an already-solved problem. Good that the new `except DefusedXmlException` branch degrades to an empty-text result with a `security_error` field in metadata rather than raising and crashing the loader entirely — a malicious document gets flagged and skipped instead of taking down whatever pipeline is calling this loader. The dropped `parse(source_ref)` branch (which read arbitrary local files by path) is now unified into the same `fromstring(content)` call, which also removes a second XXE-adjacent surface, not just the one from the original report.

PR #6447, opened 2026-07-03
fix: re-validate interpolated output_file against untrusted inputs

theCyberTech:
## Summary

`Task.output_file` can use `{var}` templates. CrewAI fills those variables from `crew.kickoff(inputs=...)`.

The field validator (`output_file_validation`) blocks unsafe paths such as `..`, absolute paths, `~`/`$`, and shell metacharacters. It returns templated values without checking the final path. Later, `interpolate_inputs_and_add_conversation_history` writes the concrete path to `self.output_file` without running validation again (`Task.model_config` does not set `validate_assignment=True`).

Untrusted `inputs` can therefore inject path traversal, an absolute path, or shell expansion into a templated `output_file` and write task output outside the working directory:

```python
Task(description=..., expected_output=..., output_file="reports/{name}.md")
crew.kickoff(inputs={"name": "../../../../home/victim/.bashrc"})  # escapes cwd
```

This is a remaining case of the earlier `output_file` path-traversal fix (`73f3288`). That change added the validator, but template interpolation still bypasses it.

## Fix

Validate the interpolated variable values that appear in the `output_file` template before interpolation. Reject:

- `..` path traversal
- absolute paths (POSIX or Windows)
- `~` / `$` expansion
- shell metacharacters

The developer-authored template stays trusted. A legitimate absolute base directory in the template still works. Existing templated-path behavior is unchanged.

Changes:

- Add `Task._validate_output_file_input_values(inputs)` and run it before output-path interpolation.
- Move shared checks into `Task._sanitize_output_file_path` so definition-time validation stays the same.

## Tests

- `test_interpolate_output_file_rejects_unsafe_inputs` — five malicious-input cases (traversal, absolute path, `~` expansion, shell character, Windows absolute) all raise `ValueError`.
- `test_interpolate_output_file_allows_safe_inputs` — safe relative templates and developer-chosen absolute base templates still work.
- Full `tests/test_task.py` passes (73 tests), including the existing `test_interpolate_inputs`. `ruff` is clean.


ErenAta16 (commented):
Read through the two-layer validation carefully since this is exactly the kind of check that is easy to get subtly wrong: the field validator still rejects `..`/absolute paths/shell metacharacters in the *literal* portions of a templated `output_file` at task-definition time (via the shared `_sanitize_output_file_path` helper, so the literal-only and templated branches stay consistent with each other), and the new `_validate_output_file_input_values` re-checks only the specific template variables that actually appear in the template against the *runtime* `inputs` dict, rejecting `..`, `~`/`$` expansion, and absolute paths on the substituted value specifically. That scoping matters: a naive full re-validation of the final interpolated path could reject a legitimate developer-authored absolute base directory, while checking only the untrusted-input side closes the actual gap (inputs escaping the developer's own template) without breaking that intentional case. The `vulnerability-scan.yml` note for the unrelated transitive `nltk` CVE with a clear "crewai never calls this with untrusted input" justification is also a good practice — explains why that particular finding is being suppressed rather than just adding it to an ignore list silently.

PR #6448, opened 2026-07-03
fix(tools): raise ToolUsageError instead of bare raise in _original_tool_calling

C0d3N1nja97342:
## Problem

`ToolUsage._original_tool_calling` (`lib/crewai/src/crewai/tools/tool_usage.py`) executed a bare `raise` outside any `except` block when `_validate_tool_input` returned a non-dict and `raise_error=True`. Because no exception was active on that path (the preceding `try/except` had completed normally), Python raised `RuntimeError: No active exception to re-raise` instead of a meaningful error. This produced a confusing error message and caused the caller `_tool_calling` to log/retry against a `RuntimeError` rather than a descriptive `ToolUsageError`.

Reported in #6430.

## Solution

Construct a `ToolUsageError` once and `raise` it when `raise_error=True` (or return it when `raise_error=False`), mirroring the existing `raise_error=False` return path. The sole caller `_tool_calling` still catches it via `except Exception` and falls back to `_function_calling` / retry exactly as before, so behavior is backward compatible — the only difference is that the exception now carries a meaningful message instead of `RuntimeError`.

No signature, return-type annotation, or public API change. Minimal and scoped to the bug.

## Changes

- `lib/crewai/src/crewai/tools/tool_usage.py` — in `_original_tool_calling`, replaced the bare `raise` in the non-dict branch with `raise tool_arguments_error` (a constructed `ToolUsageError`); the `raise_error=False` path returns the same object.
- `lib/crewai/tests/tools/test_tool_usage.py` — added `ToolUsageError` import, a small `ToolUsage` builder helper, and 3 tests covering:
  - happy path (valid dict → `ToolCalling`)
  - non-dict + `raise_error=True` → raises `ToolUsageError` (regression: was `RuntimeError`)
  - non-dict + `raise_error=False` → returns `ToolUsageError` (compatibility)

## Testing

- New unit tests pass: `uv run pytest lib/crewai/tests/tools/test_tool_usage.py -k original_tool_calling` → **3 passed**.
- A minimal runtime demo verified all three paths (happy / raise / return).
- `ruff check` and `ruff format --check` are clean on the changed files.
- Note: 5 pre-existing event-bus tests in `test_tool_usage.py` fail locally on a clean `main` checkout (confirmed via `git stash` + re-run) — they are environment/flaky (threading + event bus on Windows) and unrelated to this change.

## Notes for Reviewer

- This is an AI-assisted contribution. Per `.github/CONTRIBUTING.md`, the `llm-generated` label is required — I've attempted to apply it; if the tooling does not permit external contributors to label PRs, please add it.
- Fixes #6430.
- Design choice: I kept the defensive `if not isinstance(arguments, dict)` branch rather than removing it. Since `_validate_tool_input` currently guarantees dict-or-raise, the branch is effectively defensive/dead, but I chose the minimal fix (correct the `raise`) over removing the branch to keep the change scoped to the bug. Happy to remove the branch instead if preferred.

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -849,9 +849,12 @@ class ToolUsage:
             return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
 
         if not isinstance(arguments, dict):
+            tool_arguments_error = ToolUsageError(
+                f"{I18N_DEFAULT.errors('tool_arguments_error')}"
+            )
             if raise_error:
-                raise
-            return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
+                raise tool_arguments_error
+            return tool_arguments_error
 
         return ToolCalling(
             tool_name=sanitize_tool_name(tool.name),
--- a/lib/crewai/tests/tools/test_tool_usage.py
+++ b/lib/crewai/tests/tools/test_tool_usage.py
@@ -25,7 +25,7 @@ from crewai.hooks.tool_hooks import (
 )
 from crewai.tools import BaseTool
 from crewai.tools.tool_calling import ToolCalling
-from crewai.tools.tool_usage import ToolUsage
+from crewai.tools.tool_usage import ToolUsage, ToolUsageError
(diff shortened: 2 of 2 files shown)


ErenAta16:

Issue #3771, opened 2025-10-22
[BUG] `format_answer()` catches `OutputParserException`, breaking retry logic for malformed LLM outputs


jacoblam3112-glitch:
### Description

When an LLM returns malformed output (e.g., missing colons in the Action/Action Input format), the `format_answer()` function in `crewai/utilities/agent_utils.py` catches all exceptions including `OutputParserException` and converts them to `AgentFinish`. This prevents the retry logic in `_invoke_loop()` from working correctly.

In `agent_utils.py`:

```python
def format_answer(answer: str) -> Union[AgentAction, AgentFinish]:
    """Format a response from the LLM into an AgentAction or AgentFinish."""
    try:
        return CrewAgentParser.parse_text(answer)
    except Exception:  # ← Catches OutputParserException
        # If parsing fails, return a default AgentFinish
        return AgentFinish(
            thought="Failed to parse LLM response",
            output=answer,
            text=answer,
```

**Example Malformed LLM Output:**
```
Thought
The user wants to verify something.
Action
Video Analysis Tool
Action Input:
{"query": "Is there something?"}
```
(Missing colons after "Thought", "Action", and "Action Input")
        

### Steps to Reproduce

1. In `crewai/agents/crew_agent_executor.py`, hardcode a malformed LLM response after `get_llm_response()`:

```python
answer = get_llm_response(
    llm=self.llm,
    messages=self.messages,
    callbacks=self.callbacks,
    printer=self._printer,
)

# Override with malformed output (missing colons)
answer = '''Thought
The user wants to verify something.
Action
Video Analysis Tool
Action Input:
{"query": "Is there something?"}'''

formatted_answer = process_llm_response(answer, self.use_stop_words)
```

2. Run any agent with a task
3. Observe that the agent immediately returns an `AgentFinish` with the malformed text instead of retrying with an error message

**Expected:** The agent should retry with an error message like "Invalid Format: I missed the ':' after 'Action:'"

**Actual:** The agent exits with `AgentFinish(thought="Failed to parse LLM response", output=<malformed text>)`

### Expected behavior

**Expected Behavior:**
When parsing fails with `OutputParserException`, it should be re-raised so the `_invoke_loop()` in `crew_agent_executor.py` can catch it and append an error message to retry.

**Actual Behavior:**
The exception is caught and converted to `AgentFinish`, causing the loop to exit prematurely instead of retrying.

### Screenshots/Code snippets

[...]

PR #2483, opened 2025-03-26
Feat/individual react agent

bhancockio:

PR #6450, opened 2026-07-03
fix(agents): recover real tool call from fabricated Observation continuations

### LINKED ISSUE #6449: [BUG] Fabricated-Observation recovery in process_llm_response is dead code since #2483 — real tool calls silently discarded for models without stop-word support
### Description

`process_llm_response()` in `lib/crewai/src/crewai/utilities/agent_utils.py` contains a recovery block that is meant to handle models that **don't support the `stop` parameter** (`use_stop_words=False`: the gpt-5 family, the o1 family, and any custom `BaseLLM` that ignores stop sequences). Without a `"\nObservation:"` stop sequence, the LLM generates straight past the real tool call and fabricates the rest of the ReAct loop in a single completion:

```text
Thought: ...
Action: web_search
Action Input: {"search_query": "..."}
Observation: <fabricated — the tool never ran>
Final Answer: <fabricated>
```

The recovery block is supposed to discard the fabricated continuation so the real `Action` executes:

```python
if not use_stop_words:
    try:
        format_answer(answer)
    except OutputParserError as e:
        if FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE in e.error:
            answer = answer.split("Observation:")[0].strip()
```

**This block has been dead code since April 2025.** It catches an error that can no longer be raised:

- From #1322 (`e77442cf`, Oct 2024), `CrewAgentParser.parse()` raised `FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE` whenever a response contained both a parseable `Action` and a `Final Answer`, and the executor caught it and truncated at `"Observation:"`. **This worked** — the real tool executed.
- #2483 (`efe27bd5`, Apr 2025) removed the raise from the parser (commit log: *"Remove redundant error handling for action and final answer in CrewAgentParser"*) and deleted its regression test `test_action_and_final_answer_error` — while, in the same PR, moving the catch block into the newly created `process_llm_response()`. The raise was not redundant; it was the trigger for that catch. Since then `parse()` returns the fabricated `AgentFinish` whenever `Final Answer:` appears anywhere in the text, and the recovery can never run.
- (It is doubly unreachable: `format_answer()` also swallows every exception — that's the separate, orthogonal issue #3771/#4113. Fixing that alone would *not* revive this recovery, because the error message it looks for no longer exists anywhere in the codebase.)

Other leftovers of the original intent are still in the tree:
- `FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE` still sits in `crewai/agents/constants.py`, imported only by the dead catch.
- The experimental executor has an unreachable warning block (`experimental/agent_executor.py`, "LLM returned 'Final Answer:' but parsed as AgentAction") that only makes sense with the recovery alive.

**User-visible impact:** this is, as far as I can tell, the root cause of the many reports collected in #3154 ("Agent does not actually invoke tools, only simulates tool usage with fabricated output"). That thread independently observed the exact correlation this explains: gpt-4.1 works (stop words supported → generation stops before a fabricated `Observation:`), while gpt-5 / o-series fabricate (`supports_stop_words()` returns `False` for both, so nothing stops the completion and the dead recovery never rescues the real tool call).

### Steps to Reproduce

```python
from crewai.utilities.agent_utils import process_llm_response

# What a model without stop-word support typically returns (shape from #3154):
answer = """Thought: I should use the Web Search Tool to investigate trends in AI.
Action: Web Search Tool
Action Input: {"search_query": "trends in AI 2025"}
Observation: The search results show that the main trends include multimodal models.
Thought: I now have enough information.
Final Answer: The main AI trends are multimodal models."""

result = process_llm_response(answer, use_stop_words=False)
print(type(result).__name__)  # AgentFinish — the fabricated answer wins
```

### Expected behavior

`AgentAction(tool="Web Search Tool", ...)` — the fabricated `Observation`/`Final Answer` continuation is discarded and the real tool call executes, as it did before #2483.

### Screenshots/Code snippets

See above; the dead block is at `lib/crewai/src/crewai/utilities/agent_utils.py` (in `process_llm_response`), current `main` (`2b90117e`).

### Operating System

macOS (reproduces on any OS)

### Python Version

3.13

### crewAI Version

current `main` (`1.15.2a2`, commit `2b90117e`)

### crewAI Tools Version

n/a

### Virtual Environment

uv

### Evidence

Regression timeline verified with `git log -S FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE`: raise introduced in `e77442cf` (#1322), removed in `efe27bd5` (#2483) together with its test, catch block moved (already dead) into `process_llm_response` in the same commit.

### Possible Solution

Replace the dead `except` block with an explicit position check — no exception plumbing, so it is immune to `format_answer()`'s exception swallowing and independent of #3771:

```python
if not use_stop_words and FINAL_ANSWER_ACTION in answer:
    action_match = ACTION_INPUT_REGEX.search(answer)
    final_answer_idx = answer.find(FINAL_ANSWER_ACTION)
    if action_match and action_match.start() < final_answer_idx:
        observation_idx = answer.find(
            "Observation:", action_match.start(), final_answer_idx
        )
        if observation_idx != -1:
            answer = answer[:observation_idx].strip()

return format_answer(answer)
```

Only the exact fabrication shape (`Action` → `Observation` → `Final Answer`, with stop words unsupported) is affected; every other response shape, and all behavior for stop-word-supporting models, is unchanged. This also fixes all four call sites at once (experimental `AgentExecutor`, `CrewAgentExecutor`, `LiteAgent`, `step_executor`) and makes the experimental executor's existing warning reachable.

I have this implemented with regression tests (fails-before/passes-after) and will open a PR referencing this issue.

### Additional context

Found by code inspection while tracing why the recovery path never fires. AI-assisted analysis (Claude); every claim above was verified by hand against git history and a local reproduction.

ritsth:
Closes #6449. Root cause behind the reports collected in #3154.

## Problem

For models that don't support the `stop` parameter (`use_stop_words=False`: gpt-5 family, o1 family, custom `BaseLLM`s), nothing cuts generation at `"\nObservation:"`, so the LLM emits a real `Action` followed by a **fabricated** `Observation` and `Final Answer` in one completion. `parse()` returns the fabricated `AgentFinish`, and the real tool never executes.

`process_llm_response()` contains a recovery block designed for exactly this — but it has been dead code since #2483 (`efe27bd5`, Apr 2025), which removed the parser raise (`FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE`) that the block catches. The recovery worked from #1322 (Oct 2024) until then. Full regression timeline and analysis in #6449.

This explains the correlation users kept reporting in #3154: gpt-4.1 fine, gpt-5/o-series fabricating — `supports_stop_words()` returns `False` for exactly the failing models.

## Fix

Replace the dead `except` block with an explicit position check (no exception plumbing, so it is unaffected by `format_answer()`'s broad `except Exception` — the separate issue #3771):

- Gated to `use_stop_words=False` — behavior for stop-word-supporting models is untouched.
- Requires the exact fabrication shape `Action` < `Observation` < `Final Answer` (bounded `find`), so:
  - "Final Answer first, ReAct format quoted after" still returns the final answer;
  - "Action then Final Answer with no Observation" keeps current behavior;
  - every other shape is unchanged.
- Fixes all four call sites at once (experimental `AgentExecutor`, `CrewAgentExecutor`, `LiteAgent`, `step_executor`) and makes the experimental executor's existing "`Final Answer:` but parsed as AgentAction" warning reachable, giving users visibility when recovery happens.

## Tests

New `TestProcessLlmResponse` class in `lib/crewai/tests/utilities/test_agent_utils.py` (8 tests):

- `test_fabricated_observation_recovers_real_action` — **fails on current `main`** (returns the fabricated `AgentFinish`), passes with this fix:

```
# main (fix stashed):
FAILED tests/utilities/test_agent_utils.py::TestProcessLlmResponse::test_fabricated_observation_recovers_real_action
1 failed, 7 passed

# with fix:
8 passed
```

- The other 7 tests pin the non-fabrication shapes and pass on **both** sides, demonstrating no behavior change outside the targeted case.

`tests/utilities/test_agent_utils.py`, `tests/agents/test_crew_agent_parser.py`, and `tests/agents/test_agent_executor.py` all pass (209 tests). `ruff check`, `ruff format --check`, and `mypy` clean on the changed files.

## Disclosure

This fix was developed with AI assistance (Claude); the regression history, reproduction, and red/green results were verified by hand. I don't have permission to add labels — could a maintainer add the `llm-generated` label per the contribution guidelines?

### DIFF
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -580,11 +580,13 @@ def process_llm_response(
         action_match = ACTION_INPUT_REGEX.search(answer)
         final_answer_idx = answer.find(FINAL_ANSWER_ACTION)
         if action_match and action_match.start() < final_answer_idx:
-            # Without a "\nObservation:" stop sequence the model generates past
+            # Without the "\nObservation:" stop sequence the model generates past
             # the real tool call, fabricating an Observation and Final Answer.
             # Discard the fabricated continuation so the actual Action executes.
+            # Anchor on the newline (the real stop sequence) so an "Observation:"
+            # substring inside the Action Input payload isn't mistaken for it.
             observation_idx = answer.find(
-                "Observation:", action_match.start(), final_answer_idx
+                "\nObservation:", action_match.start(), final_answer_idx
             )
             if observation_idx != -1:
                 answer = answer[:observation_idx].strip()
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -17,7 +17,7 @@ from crewai_core.settings import Settings
 from pydantic import BaseModel
 from rich.console import Console
 
-from crewai.agents.constants import FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE
+from crewai.agents.constants import ACTION_INPUT_REGEX, FINAL_ANSWER_ACTION
 from crewai.agents.parser import (
     AgentAction,
     AgentFinish,
@@ -576,12 +576,18 @@ def process_llm_response(
     Returns:
         Either an AgentAction or AgentFinish
     """
-    if not use_stop_words:
-        try:
-            format_answer(answer)
-        except OutputParserError as e:
-            if FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE in e.error:
-                answer = answer.split("Observation:")[0].strip()
+    if not use_stop_words and FINAL_ANSWER_ACTION in answer:
+        action_match = ACTION_INPUT_REGEX.search(answer)
+        final_answer_idx = answer.find(FINAL_ANSWER_ACTION)
+        if action_match and action_match.start() < final_answer_idx:
+            # Without a "\nObservation:" stop sequence the model generates past
+            # the real tool call, fabricating an Observation and Final Answer.
+            # Discard the fabricated continuation so the actual Action executes.
+            observation_idx = answer.find(
+                "Observation:", action_match.start(), final_answer_idx
+            )
+            if observation_idx != -1:
+                answer = answer[:observation_idx].strip()
 
     return format_answer(answer)
 
--- a/lib/crewai/tests/utilities/test_agent_utils.py
+++ b/lib/crewai/tests/utilities/test_agent_utils.py
@@ -16,6 +16,7 @@ from crewai.hooks.tool_hooks import (
     clear_before_tool_call_hooks,
     register_after_tool_call_hook,
 )
+from crewai.agents.parser import AgentAction, AgentFinish
 from crewai.tools.base_tool import BaseTool
 from crewai.utilities.agent_utils import (
     _asummarize_chunks,
@@ -27,6 +28,7 @@ from crewai.utilities.agent_utils import (
     execute_single_native_tool_call,
     NativeToolCallResult,
     parse_tool_call_args,
+    process_llm_response,
     summarize_messages,
 )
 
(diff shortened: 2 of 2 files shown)


ritsth:
Thanks @coderabbitai — good catch on the `Observation:`-inside-payload edge case. Addressed in 886e8302, though via a slightly different route than suggested: searching after the `Action Input` boundary wouldn't help, since the payload (and thus any `Observation:` substring in it) comes *after* `Action Input:`.

Instead I anchored the trim on `"\nObservation:"` — which is the actual stop sequence crewAI configures (`I18N_DEFAULT.slice("observation")` at `agent/core.py:1016`). This mirrors exactly what the stop word would have removed, and because the fabricated `Observation` is always emitted on its own line per the ReAct format, an `Observation:` substring inside a single-line JSON Action Input is no longer a match. Added `test_observation_substring_in_action_input_preserved` to lock it in.

nolanchic:
I independently verified this fix against current `main` and wanted to share the results in case it helps move review along.

**Reproduction confirmed:** on unpatched `main`, the issue's transcript returns `AgentFinish` (fabricated answer wins):
```
Thought: ...
Action: web_search
Action Input: {"search_query": "..."}
Observation: <fabricated>
Final Answer: <fabricated>
```
→ `AgentFinish`. The dead `except OutputParserError` block never fires because `parse()` returns `AgentFinish` as soon as `Final Answer:` appears anywhere (`parser.py:95`), so the error it guards on can no longer be raised — confirmed dead since #2483.

**Fix confirmed:** applying this PR's diff locally, the same input returns `AgentAction(tool="web_search", ...)` with the fabricated `Observation:`/`Final Answer` truncated at the `\nObservation:` anchor.

**Independently re-ran the test class** (`TestProcessLlmResponse`) on patched code → all pass. Full `tests/utilities/test_agent_utils.py` → 70 passed. `tests/agents/` → 357 passed, 16 skipped; the only failures (`TestAnthropicNativeToolCalling`) are pre-existing on clean `main` (they need a live Anthropic endpoint and fail under `--block-network`), unrelated to this change. `ruff` and `mypy` (strict) both clean on the patched files.

Two observations worth noting for reviewers, neither blocking:

1. The `"\nObservation:"` anchor (vs. `"Observation:"`) is the right call — it correctly leaves an `Observation:` substring inside the `Action Input` payload intact, which the `test_observation_substring_in_action_input_preserved` case covers.

2. `FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE` is still defined in `agents/constants.py` after this PR (it's no longer imported anywhere in `src/`). Harmless, but it's now fully orphaned — a follow-up could remove it for tidiness. As-is, this PR already removes the dead import from `agent_utils.py`, which is the load-bearing change.

The recovery also makes the previously-unreachable warning at `experimental/agent_executor.py:1445` (`"Final Answer:" in answer and isinstance(formatted_answer, AgentAction)`) reachable — that's a nice side effect, users now get a visible signal when recovery fires.

Nice work tracing this back through #2483 — the git-history forensics in the issue is accurate.

_(AI-assisted review via Claude Code; verified by applying the diff and running the suite locally.)_

PR #6455, opened 2026-07-04
fix: bare raise in ToolUsage._original_tool_calling causes RuntimeError instead of ToolUsageError

rafaelipuente:
## Description

Fixes #6430.

`ToolUsage._original_tool_calling()` contained a bare `raise` in the `not isinstance(arguments, dict)` branch. A bare `raise` re-raises the exception currently being handled, but on this path no exception is active (the preceding `try/except` completed normally). If that branch were ever hit with `raise_error=True`, Python would crash with `RuntimeError: No active exception to re-raise` instead of the intended tool-arguments error.

This matters because `_tool_calling()` calls this method with `raise_error=True` and relies on catching a meaningful exception to trigger its function-calling/retry fallback.

### Fix

Replace the bare `raise` with an explicit `raise ToolUsageError(...)` using the same `tool_arguments_error` message the branch already returns when `raise_error=False`, keeping error messaging consistent.

### Testing

- Added a regression test that patches `_validate_tool_input` to return a non-dict and asserts `_original_tool_calling(..., raise_error=True)` raises `ToolUsageError` (not `RuntimeError`), and that the non-raise path returns a `ToolUsageError` instance.
- `uv run pytest lib/crewai/tests/tools/test_tool_usage.py` passes (29 tests).
- `ruff check --select PLE0704` no longer flags the file; ruff lint/format clean on changed files.

*(Note: authored with the assistance of an AI coding agent. Please apply the `llm-generated` label.)*

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -850,7 +850,7 @@ class ToolUsage:
 
         if not isinstance(arguments, dict):
             if raise_error:
-                raise
+                raise ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
             return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
 
         return ToolCalling(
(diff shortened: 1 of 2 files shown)


ErenAta16:

PR #6457, opened 2026-07-04
Fix ToolUsage bare raise for non-dict arguments

initiallyqq:
## Summary
- Replace the bare raise in `ToolUsage._original_tool_calling` with `ToolUsageError` for non-dict arguments.
- Add a regression test for the `raise_error=True` path.

Fixes #6430

## Tests
- `./.venv/Scripts/python.exe -m pytest lib/crewai/tests/tools/test_tool_usage.py::test_original_tool_calling_raises_tool_usage_error_for_non_dict_arguments -q`
- `./.venv/Scripts/ruff.exe check lib/crewai/src/crewai/tools/tool_usage.py --select PLE0704`

Note: the full `test_tool_usage.py` file has 5 existing failures on Windows because `pytest-recording` disables socket creation for the event bus; the new regression test passes.


ErenAta16:

PR #6459, opened 2026-07-04
fix(memory): anchor LanceDB scope matching on path boundary

Osamaali313:
## Problem

Scope filtering in `LanceDBStorage` does not respect the `/` path separator, so operations on a scope leak into **lexicographic sibling** scopes:

- read paths (`search`, `_scan_rows` → `count` / `get_scope_info` / `list_records` / `list_categories`) used `scope LIKE 'prefix%'`
- `delete` used `scope LIKE 'prefix%' OR scope = '/'`
- `reset` used the range `scope >= 'prefix' AND scope < 'prefix/￿'`

Because `-` (0x2D) sorts before `/` (0x2F), a sibling like `/team-b` matches an operation scoped to `/team`.

On read paths this over-counts. On `reset`/`forget` it is **data loss**: `reset("/team")` also deletes every record under the unrelated `/team-b`.

```python
storage.count("/team")   # -> 3  (should be 2: /team and /team/x, NOT /team-b)
storage.reset("/team")   # also wipes /team-b
```

## Fix

Route all four sites through a shared helper that anchors the match on the path boundary — the exact scope, or descendants under `prefix + "/"`:

```python
@staticmethod
def _scope_prefix_clause(scope_prefix: str) -> str:
    if not prefix.startswith("/"):
        prefix = "/" + prefix
    safe = prefix.replace("'", "''")
    return f"(scope = '{safe}' OR scope LIKE '{safe}/%')"
```

This matches the hierarchical semantics the qdrant backend (`_build_scope_filter`) and this file's own `get_scope_info` / `list_scopes` already use (`child_prefix = prefix + "/"`). `delete` keeps its explicit `scope = '/'` root-record match, and the existing single-quote escaping is preserved (orthogonal to #5729).

## Tests

Added `test_lancedb_scope_prefix_respects_path_boundary` (in `lib/crewai/tests/memory/test_unified_memory.py`), using explicit embeddings so it needs no network. It asserts `count("/team") == 2`, `count("/team-b") == 1`, and that `reset("/team")` leaves `/team-b` intact.

- Before the fix the test fails: `count("/team")` returns `3` (`assert 3 == 2`).
- After the fix it passes; existing memory tests are unaffected.

---

*Contribution note: AI tooling helped surface and reproduce this scope-boundary bug; I reviewed, verified (before/after unit test), and finalized the fix myself. Tagging as `llm-generated` per the contribution policy.*

### DIFF
--- a/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
+++ b/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
@@ -368,6 +368,28 @@ class LanceDBStorage:
             return None
         return self._row_to_record(rows[0])
 
+    @staticmethod
+    def _scope_prefix_clause(scope_prefix: str) -> str:
+        """Build a SQL predicate matching a scope subtree by path boundary.
+
+        A scope such as ``/team`` owns itself and every descendant under
+        ``/team/`` -- but *not* a lexicographic sibling like ``/team-b``.
+        Matching with a bare ``LIKE 'prefix%'`` (or a lexicographic range)
+        leaks into such siblings, so anchor the match on the ``/`` separator:
+        the exact scope, or anything beneath ``prefix + "/"``.
+
+        Args:
+            scope_prefix: The scope path to match (already non-root).
+
+        Returns:
+            A SQL boolean expression selecting the scope and its descendants.
+        """
+        prefix = scope_prefix.rstrip("/")
+        if not prefix.startswith("/"):
+            prefix = "/" + prefix
+        safe = prefix.replace("'", "''")
+        return f"(scope = '{safe}' OR scope LIKE '{safe}/%')"
+
     def search(
         self,
         query_embedding: list[float],
@@ -385,9 +407,7 @@ class LanceDBStorage:
             )
         query = self._table.search(query_embedding)
         if scope_prefix is not None and scope_prefix.strip("/"):
-            prefix = scope_prefix.rstrip("/")
-            like_val = prefix + "%"
-            query = query.where(f"scope LIKE '{like_val}'")
+            query = query.where(self._scope_prefix_clause(scope_prefix))
         results = query.limit(
             limit * 3 if (categories or metadata_filter) else limit
         ).to_list()
@@ -448,10 +468,9 @@ class LanceDBStorage:
                 return before - int(self._table.count_rows())
             conditions = []
             if scope_prefix is not None and scope_prefix.strip("/"):
-                prefix = scope_prefix.rstrip("/")
-                if not prefix.startswith("/"):
-                    prefix = "/" + prefix
-                conditions.append(f"scope LIKE '{prefix}%' OR scope = '/'")
+                conditions.append(
+                    f"({self._scope_prefix_clause(scope_prefix)} OR scope = '/')"
+                )
             if older_than is not None:
                 conditions.append(f"created_at < '{older_than.isoformat()}'")
             if not conditions:
@@ -485,7 +504,7 @@ class LanceDBStorage:
             return []
         q = self._table.search()
         if scope_prefix is not None and scope_prefix.strip("/"):
-            q = q.where(f"scope LIKE '{scope_prefix.rstrip('/')}%'")
+            q = q.where(self._scope_prefix_clause(scope_prefix))
         if columns is not None:
             q = q.select(columns)
         result: list[dict[str, Any]] = q.limit(limit).to_list()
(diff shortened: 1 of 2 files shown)


Osamaali313:
Note for maintainers: per the contribution policy this PR should carry the `llm-generated` label — AI tooling helped surface and reproduce the bug, and I reviewed/verified (RED→GREEN unit test) and finalized it. I don't have permission to apply the label on a fork PR, so could a maintainer please add it? Happy to adjust anything.

ErenAta16 (commented):
Reproduced the sibling-leak standalone before checking the fix: matching `["/team", "/team/x", "/team-b", "/team-b/y"]` against a `LIKE "/team%"` predicate returns all four (since `-` sorts/matches before `/` lexically), while the new `scope = prefix OR scope LIKE prefix + "/%"` clause correctly returns only `/team` and `/team/x`. The `reset`/`delete` paths are the ones that actually worry me on main — silently wiping an unrelated sibling scope on `reset("/team")` is data loss, not just an over-broad read. Good that all four call sites (`search`, `delete`, `_scan_rows`, `reset`) got routed through the one shared `_scope_prefix_clause` helper instead of being patched individually, since that's exactly the kind of bug that reappears if a fifth call site gets added later with its own inline `LIKE` string.

Osamaali313:
Thanks for the careful review @ErenAta16, and to @Copilot / @coderabbitai for the pointers. Pushed d2f8908 addressing both Copilot comments:

**1. LIKE metacharacters (the important one).** You're right that `LIKE 'prefix/%'` didn't escape `%`/`_`, so a scope path containing `_` still leaked — verified against real LanceDB: with scopes `/team_x`, `/team_x/z`, `/teamAx/w`, the old clause returned all three for prefix `/team_x` (the `_` matched `A`). Rather than escaping, I took your suggested approach and dropped LIKE entirely in favor of a half-open **lexicographic range** on the path boundary:

```python
(scope = 'prefix' OR (scope >= 'prefix/' AND scope < 'prefix0'))
```

`/` is `0x2F` so the next code point is `0` (`0x30`); every `prefix/...` descendant sorts before `prefix0`. This keeps the scope path a pure literal — no wildcard semantics — and LanceDB accepts the string range filter. All four call sites still route through the one helper.

**2. Coverage for the other call sites.** Extended the tests: the sibling test now also asserts `search()` scopes correctly; added a dedicated `delete(scope_prefix=...)` subtree test; and added a literal-path test (`/team_x` must not pull in `/teamAx/w`) that **fails on the old LIKE clause** (`count` returns 3 instead of 2) and passes on the range clause.

Full `test_unified_memory.py`: 53 passed.

ErenAta16:
Small correction on credit: the lexicographic-range suggestion was Copilot's inline comment on this PR, not mine - my review only covered the sibling-leak reproduction and the shared-helper design, not the LIKE-metacharacter gap. Copilot and coderabbit both caught that one; wanted the record to be accurate since it's a public thread.

That said, verified the fix itself independently in plain Python before trusting the "real LanceDB" claim at face value:

```python
def matches(scope, prefix):
    return scope == prefix or (scope >= prefix + '/' and scope < prefix + '0')
```

ran against the sibling cases in this thread plus a few extra edges (`/team_x0`, `/team_xa` - same-prefix-different-character cases that a naive range could get wrong at the boundary) and all matched the expected in/out result, including your literal-path case (`/team_x` vs `/teamAx/w`) and the original `/team` vs `/team-b` case. The `'/' + 1 codepoint = '0'` boundary logic holds up under both UTF-8 and plain ASCII ordering since they're consecutive code points either way, so this isn't relying on any DataFusion-specific collation behavior - it's true for any byte-wise or codepoint-wise string ordering. Combined with your real-LanceDB count check and coderabbit's independent code read, I'm confident this is correct.

Osamaali313:
You're right, and thanks for correcting the record — the lexicographic-range suggestion was **Copilot's** inline comment (with coderabbit independently flagging the LIKE-metacharacter gap too); your review covered the sibling-leak reproduction and the shared-helper design. Apologies for the ambiguous "your suggested approach" phrasing in my earlier reply — credit for the range idea goes to Copilot/coderabbit.

And appreciate the independent plain-Python verification, especially the `/team_x0` / `/team_xa` boundary edges — you're right that the `'/'`→`'0'` (0x2F→0x30, consecutive code points) boundary holds for any byte-wise or codepoint-wise ordering, so it doesn't depend on DataFusion collation. Good to have it confirmed from several angles.

PR #6484, opened 2026-07-08
feat(cli): run declarative flows on the TUI (headless terminal fallback)

joaomdmoura:
## What

Normalizes declarative flow execution onto the `CrewRunApp` TUI, matching how declarative (JSON) crews and conversational flows already run. Headless contexts keep a direct-terminal path for deploy/CI.

Before this PR, `crewai run` used four surfaces: declarative crews → TUI, conversational flows → TUI, classic Python flows → terminal, **declarative flows → terminal**. This is the first step toward running everything on the TUI.

## Behavior

- **Interactive terminal** → the TUI. A new **STEPS** panel shows each declarative flow method as a first-class step (driven by `FlowStarted` / `MethodExecutionStarted/Finished/Failed` events), labeled with its call type (`crew`/`agent`/`expression`/…) read from the flow definition. Crews/agents inside a method keep streaming in the main panel.
- **Headless** (deploy/`CREWAI_DMN`, piped output, CI, any non-TTY) → the existing direct-terminal kickoff + printed result. Gated by `is_interactive()`.

## Changes

- `crew_run_tui.py`: `_run_flow_worker` (runs `flow.kickoff` in a thread worker, reuses `_on_crew_done`/`_on_crew_failed` + `_stringify_output`); an `_is_flow_run` gate so crew rendering is byte-identical; flow-event subscriptions building `_flow_steps`; STEPS sidebar + flow-aware header.
- `run_declarative_flow.py`: `is_interactive()` branch → `_run_declarative_flow_tui` (instantiates `EventListener`, derives the method-type map from `flow._definition`, mirrors the crew exit-code + deploy-chain contract) or the existing terminal path.

## Design note

The approved plan gated on `is_dmn_mode_enabled()` alone; I switched to `is_interactive()` (which folds in the DMN check *and* requires a TTY) so non-TTY runs — CI, pipes, `CliRunner` — never launch a TUI. This is strictly more correct and keeps existing headless flow tests green.

## Out of scope / follow-ups

- Classic Python flows onto the TUI (next step toward full normalization).
- `crewai deploy` validation has a separate declarative-flow-entrypoint gap (`DeployValidator._check_flow_entrypoint`) unrelated to the TUI Deploy button, which uses `skip_validate=True`.

## Testing

- New tests in `test_crew_run_tui.py` (flow dispatch, method-event STEPS build, completion sweep, end-to-end `run_test()` mount) and `test_run_declarative_flow.py` (TUI vs terminal routing, exit codes, deploy chaining, method-type extraction).
- Full `lib/cli` suite: 659 passed; the 7 remaining failures are pre-existing and unrelated (version-string drift + a `git fetch` mock in create/tools tests).
- `ruff format`/`ruff check`/`mypy` clean; mirrored `lib/crewai/tests/cli` still green.

---

> [!NOTE]
> **Medium Risk**
> Changes default CLI run behavior for interactive declarative flows and threads kickoff through the TUI; crew paths are gated but shared completion handlers and event subscriptions are extended.
> 
> **Overview**
> **Interactive declarative flows** now launch `CrewRunApp` instead of printing kickoff output on the terminal, aligning them with JSON crews and conversational flows. Routing uses `is_interactive()` (TTY + not DMN); flows with `@human_feedback` still use the terminal path.
> 
> `CrewRunApp` gains a declarative-flow mode (`_is_flow_run`): `_run_flow_worker` runs `flow.kickoff` in a thread, subscribes to flow method events to build a **STEPS** sidebar (with call-type labels from the definition), and updates the header for active/paused methods without letting nested crew kickoffs rename the flow. Completion/failure messaging uses “flow” vs “crew” via `_run_noun`.
> 
> `run_declarative_flow.py` adds `_run_declarative_flow_tui` (starts `EventListener`, forces `suppress_flow_events` off for the TUI, post-run Rich summary, exit codes, quit → `os._exit(130)`, deploy chaining like JSON crews).
> 
> `ConsoleFormatter.print_panel` skips Rich flow panels when `is_tui_mode()` is set so output does not corrupt the Textual UI.
> 
> Tests cover TUI vs terminal routing, STEPS/event behavior, and TUI-mode panel suppression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63a8378428b49ea866a746c81bbd943e9b155001. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/cli/src/crewai_cli/crew_run_tui.py
+++ b/lib/cli/src/crewai_cli/crew_run_tui.py
@@ -568,11 +568,26 @@ FooterKey .footer-key--key {
         self._default_inputs: dict[str, Any] | None = None
         self._crew_result: Any = None
         self._crew_json_path: Any = None
+        # Declarative-flow execution state. A flow renders per-method "STEPS"
+        # (built from flow method events) instead of the crew task list.
+        self._flow_inputs: dict[str, Any] | None = None
+        self._flow_method_types: dict[str, str] = {}
+        self._flow_steps: list[dict[str, Any]] = []
+        self._current_method: str | None = None
         self._elapsed_frozen: float | None = None
         self._want_deploy: bool = False
         self._trace_url: str | None = None
         self._consent_screen: TraceConsentScreen | None = None
         self._telemetry: Telemetry | None = None
+
+    @property
+    def _is_flow_run(self) -> bool:
+        """True for a non-conversational declarative flow (the STEPS view).
+
+        Gates every flow-specific rendering branch so crew and conversational
+        paths stay byte-identical.
+        """
+        return self._flow is not None and not self._is_conversational
 
     # ── Layout ──────────────────────────────────────────────
 
@@ -602,6 +617,8 @@ FooterKey .footer-key--key {
         self._tick_timer = self.set_interval(1 / 8, self._tick)
         if self._is_conversational and self._flow:
             self._start_conversational_session()
+        elif self._flow:
+            self._run_flow_worker()
         elif self._crew:
             self._run_crew_worker()
         elif self._crew_json_path:
@@ -681,6 +698,33 @@ FooterKey .footer-key--key {
         except Exception as e:
             self.call_from_thread(self._on_crew_failed, str(e))
 
+    @work(thread=True, exclusive=True, group="crew")
+    def _run_flow_worker(self) -> None:
+        from crewai.events.listeners.tracing.utils import (
+            set_suppress_tracing_messages,
+            set_tui_mode,
+        )
+
+        set_tui_mode(True)
+        set_suppress_tracing_messages(True)
+        try:
+            # A declarative flow returns either a CrewOutput (has ``.raw``) or a
+            # bare value (str/dict/pydantic); _stringify_output handles both.
+            result = self._flow.kickoff(inputs=self._flow_inputs)
+            output = self._stringify_output(result)
+            with self._lock:
+                self._crew_result = result
+            self.call_from_thread(self._on_crew_done, output)
+        except Exception as e:
+            self.call_from_thread(self._on_crew_failed, str(e))
+
+    def _set_flow_step_status(self, name: str, status: str) -> None:
+        """Update a flow method step's status. Caller must hold ``self._lock``."""
+        for step in self._flow_steps:
+            if step["name"] == name:
+                step["status"] = status
+                return
+
     def _on_crew_done(self, output: str | None) -> None:
         with self._lock:
             self._status = "completed"
@@ -694,6 +738,9 @@ FooterKey .footer-key--key {
             for k in self._task_statuses:
                 if self._task_statuses[k] == "active":
                     self._task_statuses[k] = "done"
+            for step in self._flow_steps:
+                if step["status"] == "active":
+                    step["status"] = "done"
             now = time.time()
             for entry in self._log_entries:
                 if entry["status"] == "running":
@@ -739,6 +786,9 @@ FooterKey .footer-key--key {
             self._is_streaming = False
             self._current_step = None
             self._elapsed_frozen = time.time() - self._start_time
+            for step in self._flow_steps:
+                if step["status"] == "active":
+                    step["status"] = "failed"
             now = time.time()
             for entry in self._log_entries:
                 if entry["status"] == "running":
@@ -1156,6 +1206,42 @@ FooterKey .footer-key--key {
             widget.update(t)
             return
 
+        if self._is_flow_run:
+            t.append("  STEPS\n", style=f"bold {_C_PRIMARY}")
+            t.append("\n")
+            if not self._flow_steps:
+                t.append("  ○ waiting…\n", style=_C_DIM)
+            for step in self._flow_steps:
+                name = step["name"]
+                max_name = sidebar_width - 6
+                if len(name) > max_name:
+                    name = name[: max_name - 1] + "…"
+                status = step.get("status", "pending")
+                if status == "done":
+                    t.append("  ✔ ", style=_C_GREEN)
+                    t.append(name, style=_C_DIM)
+                elif status == "active":
+                    t.append(f"  {self._spinner()} ", style=_C_PRIMARY)
+                    t.append(name, style=f"bold {_C_TEXT}")
+                elif status == "failed":
+                    t.append("  ✘ ", style=_C_RED)
+                    t.append(name, style=_C_RED)
+                else:
+                    t.append("  ○ ", style=_C_DIM)
+                    t.append(name, style=_C_DIM)
+                if step.get("call_type"):
+                    t.append(f"  ({step['call_type']})", style=_C_DIM)
+                t.append("\n")
+
+            t.append("\n")
+            t.append("  TOKENS\n", style=f"bold {_C_PRIMARY}")
+            t.append("\n")
+            out = self._output_tokens + self._live_out_tokens
+            t.append(f"  ↑ {self._input_tokens:,}\n", style=_C_DIM)
+            t.append(f"  ↓ {out:,}\n", style=_C_DIM)
+            widget.update(t)
+            return
+
         t.append("  TASKS\n", style=f"bold {_C_PRIMARY}")
         t.append("\n")
 
@@ -1222,6 +1308,40 @@ FooterKey .footer-key--key {
                 t.append("● ", style=f"bold {_C_GREEN}")
                 t.append("Conversational flow ready", style=f"bold {_C_GREEN}")
                 t.append("  Type a message below", style=_C_DIM)
+            widget.update(t)
+            return
+
+        if self._is_flow_run:
+            if self._status == "completed":
+                elapsed = self._elapsed_frozen or (time.time() - self._start_time)
+                t.append("✔ ", style=f"bold {_C_GREEN}")
+                t.append("Flow complete", style=f"bold {_C_GREEN}")
+                t.append(f"  {elapsed:.1f}s", style=_C_DIM)
+                out = self._output_tokens + self._live_out_tokens
+                parts = []
+                if self._input_tokens:
+                    parts.append(f"↑{self._input_tokens:,}")
+                if out:
+                    parts.append(f"↓{out:,}")
+                if parts:
+                    t.append(f"  {' '.join(parts)} tokens", style=_C_DIM)
+            elif self._status == "failed":
+                t.append("✘ ", style=f"bold {_C_RED}")
+                t.append("Failed", style=f"bold {_C_RED}")
+                if self._error:
+                    t.append(f"\n{self._error[:120]}", style=_C_RED)
+            elif self._current_method:
+                t.append(f"{self._spinner()} ", style=_C_PRIMARY)
+                t.append(self._current_method, style=f"bold {_C_PRIMARY}")
+                call_type = self._flow_method_types.get(self._current_method)
+                if call_type:
+                    t.append(f"  ({call_type})", style=_C_DIM)
+                if self._current_agent:
+                    t.append("\nAgent: ", style=_C_DIM)
+                    t.append(self._current_agent, style=f"bold {_C_TEXT}")
+            else:
+                t.append(f"{self._spinner()} ", style=_C_PRIMARY)
+                t.append("Starting flow…", style=_C_DIM)
             widget.update(t)
             return
 
@@ -1839,6 +1959,12 @@ FooterKey .footer-key--key {
     def _subscribe(self) -> None:
         from crewai.events.event_bus import crewai_event_bus
         from crewai.events.types.crew_events import CrewKickoffStartedEvent
+        from crewai.events.types.flow_events import (
+            FlowStartedEvent,
+            MethodExecutionFailedEvent,
+            MethodExecutionFinishedEvent,
+            MethodExecutionStartedEvent,
+        )
         from crewai.events.types.llm_events import (
             LLMCallCompletedEvent,
             LLMCallStartedEvent,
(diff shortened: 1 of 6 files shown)


lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -1879,6 +2005,50 @@ def on_crew_started(source: Any, event: CrewKickoffStartedEvent) -> None:
...
+                        step["status"] = "active"
+                        break
+                else:
+                    self._flow_steps.append(
+                        {
+                            "name": name,
+                            "call_type": self._flow_method_types.get(name),
+                            "status": "active",
+                        }
+                    )
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in 6b7529f2d — good catch. `_run_declarative_flow_tui` now forces `flow.suppress_flow_events = False` for the interactive run (mirroring how the conversational path mutates the flow for the TUI), so the STEPS panel and header get the `FlowStarted`/`MethodExecution*` events even when a flow declares `config.suppress_flow_events`. The headless/terminal path never reaches this and keeps the flow's declared setting. Regression test: `test_run_declarative_flow_tui_enables_flow_events`.

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -1879,6 +2005,50 @@ def on_crew_started(source: Any, event: CrewKickoffStartedEvent) -> None:
...
+        ) -> None:
+            with self._lock:
+                self._set_flow_step_status(event.method_name, "done")
+
+        self._register_handler(MethodExecutionFinishedEvent, on_method_finished)
+
+        @crewai_event_bus.on(MethodExecutionFailedEvent)
+        def on_method_failed(source: Any, event: MethodExecutionFailedEvent) -> None:
+            with self._lock:
+                self._set_flow_step_status(event.method_name, "failed")
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in ec55c31fe — good catch. `_current_method` is now cleared when a method ends: `_clear_current_method` runs on both `MethodExecutionFinished` and `MethodExecutionFailed`, dropping the header's active method (falling back to another still-active step, since methods can overlap, or to none). The header's idle fallback shows "Working…" once a step has run and "Starting flow…" only before the first method, so it never keeps spinning a method the STEPS sidebar already shows as done/failed. Tests: `test_current_method_clears_and_falls_back_across_overlap` (incl. the overlap case).

lib/cli/src/crewai_cli/run_declarative_flow.py
```diff
@@ -66,17 +66,161 @@ def run_declarative_flow(definition: str | Path, inputs: str | None = None) -> N
...
+
+    # The STEPS panel and header are driven by flow method events. A flow may
+    # declare ``config.suppress_flow_events`` (a headless/production
+    # optimization) which would leave STEPS stuck on "waiting…" here — so force
+    # emission on for the interactive TUI run. The headless path never reaches
+    # this and keeps the flow's declared setting.
+    try:
+        flow.suppress_flow_events = False
+    except Exception:  # noqa: S110
+        pass
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in 46cae1a50 — good catch. Forcing `suppress_flow_events` off (so the STEPS panel receives events) also un-gated the EventListener's Rich flow/method panels: `ConsoleFormatter.print_panel` prints `is_flow=True` panels regardless of `verbose`, which interleaves with Textual. `print_panel` now skips `is_flow` panels when `is_tui_mode()` is set — the same context the flow TUI worker already establishes (`set_tui_mode(True)`) and that the tracing listeners already honor — so the TUI's own STEPS panel is the only surface rendering flow progress. Non-TUI/headless flow runs (tui_mode off) print panels exactly as before. Test: `lib/crewai/tests/utilities/test_console_formatter_tui_mode.py`.

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -681,6 +698,46 @@ def _run_crew_worker(self) -> None:
...
+
+        Falls back to another still-active step (methods can overlap) so the
+        header never keeps spinning a method the STEPS list already shows as
+        done or failed.
+        """
+        if self._current_method != finished_name:
+            return
+        self._current_method = next(
+            (s["name"] for s in self._flow_steps if s["status"] == "active"), None
+        )
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in 46cae1a50 — `_current_agent` is now cleared both when a method starts (`on_method_started`) and when the active method changes (`_clear_current_method`), so the flow header no longer shows the previous method's agent until a new agent event arrives. Test: `test_flow_method_transitions_clear_current_agent`.

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -1879,6 +2026,55 @@ def on_crew_started(source: Any, event: CrewKickoffStartedEvent) -> None:
 
         self._register_handler(CrewKickoffStartedEvent, on_crew_started)
 
+        # ── Declarative-flow method events → STEPS panel ────────
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in 3a7196f56 — the `CrewKickoffStartedEvent` handler now only adopts the crew name when `not _is_flow_run`, so a nested `call: crew` step no longer relabels the flow (or the post-run summary panel, which reads `_crew_name`). Crew runs still adopt the crew name as before. Tests: `test_crew_kickoff_does_not_rename_flow_run` and `test_crew_kickoff_renames_in_crew_mode` (regression).

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -1879,6 +2026,55 @@ def on_crew_started(source: Any, event: CrewKickoffStartedEvent) -> None:
...
+
+        self._register_handler(MethodExecutionFinishedEvent, on_method_finished)
+
+        @crewai_event_bus.on(MethodExecutionFailedEvent)
+        def on_method_failed(source: Any, event: MethodExecutionFailedEvent) -> None:
+            with self._lock:
+                self._set_flow_step_status(event.method_name, "failed")
+                self._clear_current_method(event.method_name)
+
+        self._register_handler(MethodExecutionFailedEvent, on_method_failed)
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in 3a7196f56 — the flow TUI now handles `MethodExecutionPausedEvent`: the step is marked `paused` (⏸, teal) in the STEPS sidebar and the header shows "waiting for feedback" instead of a spinner, so a `@human_feedback` pause is visible even though flow status panels are suppressed in TUI mode. Test: `test_method_paused_marks_step_paused`. Note: interactively *providing* the feedback from the flow TUI is a separate follow-up; this change makes the pause visible rather than a silent stuck spinner.

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -681,6 +698,49 @@ def _run_crew_worker(self) -> None:
...
+        set_tui_mode(True)
+        set_suppress_tracing_messages(True)
+        try:
+            # A declarative flow returns either a CrewOutput (has ``.raw``) or a
+            # bare value (str/dict/pydantic); _stringify_output handles both.
+            result = self._flow.kickoff(inputs=self._flow_inputs)
+            output = self._stringify_output(result)
+            with self._lock:
+                self._crew_result = result
+            self.call_from_thread(self._on_crew_done, output)
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in b8381b9c7 — declarative flows that declare `@human_feedback` no longer run on the TUI at all; they route to the terminal path (`_flow_uses_human_feedback` gate in `run_declarative_flow`). So `_run_flow_worker` never stringifies a `HumanFeedbackPending` return as a successful completion. On the terminal the flow runtime handles the pause/feedback natively. Test: `test_human_feedback_flow_uses_terminal_even_when_interactive`.

lib/cli/src/crewai_cli/run_declarative_flow.py
```diff
@@ -66,17 +66,161 @@ def run_declarative_flow(definition: str | Path, inputs: str | None = None) -> N
     flow = load_declarative_flow(definition)
     resolved_inputs = _resolve_flow_inputs(flow, provided)
 
+    # The TUI is the interactive default. Headless contexts run directly on the
+    # terminal: deploy/CREWAI_DMN, piped output, CI — anything without an
+    # interactive TTY. is_interactive() already folds in the CREWAI_DMN check.
+    if is_interactive():
+        _run_declarative_flow_tui(flow, resolved_inputs or None)
+        return
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in b8381b9c7 — same routing change: any flow whose declarative definition declares human feedback runs on the terminal (where the runtime's blocking `input()` and Rich prompts work) instead of the TUI, so sync `@human_feedback` no longer interleaves with Textual. Non-feedback flows still get the TUI. Fully interactive human feedback *inside* the TUI remains a separate follow-up. Tests: `test_flow_uses_human_feedback_detection`, `test_human_feedback_flow_uses_terminal_even_when_interactive`.

lib/cli/src/crewai_cli/run_declarative_flow.py
```diff
@@ -66,17 +66,180 @@ def run_declarative_flow(definition: str | Path, inputs: str | None = None) -> N
...
+
+    if getattr(app, "_want_deploy", False):
+        from crewai_cli.run_crew import _chain_deploy
+
+        _chain_deploy()
+
+    return app._crew_result
+
+
+def _flow_uses_human_feedback(flow: Any) -> bool:
```
lucasgomide: Could we type it as Flow instead of Any? That way we'd avoid doing things like getattr(flow, "_definition", None) and use `flow._definition` instead. Same comment for other refereces using `getattr`

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -681,6 +698,49 @@ def _run_crew_worker(self) -> None:
...
+        except Exception as e:
+            self.call_from_thread(self._on_crew_failed, str(e))
```
lucasgomide, line 724: Actually a flow has failed. Should we keep reporting it as "crew failed"?

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -681,6 +698,49 @@ def _run_crew_worker(self) -> None:
...
+
+    def _set_flow_step_status(self, name: str, status: str) -> None:
+        """Update a flow method step's status. Caller must hold ``self._lock``."""
+        for step in self._flow_steps:
+            if step["name"] == name:
```
lucasgomide, line 729: I'm not familiar with that, so feel free to ignore this comment. Can we have the same step name twice in self._flow_steps? What happens if we have it twice and only the first one is updated? Should we add a new condition like `step["name"] == name and step["status"] != status`?

lib/cli/src/crewai_cli/run_declarative_flow.py
```diff
@@ -66,17 +66,180 @@ def run_declarative_flow(definition: str | Path, inputs: str | None = None) -> N
...
+    if app._status == "failed":
+        raise SystemExit(1)
+
+    if app._status not in ("completed", "failed"):
+        # User quit mid-run. kickoff runs in a thread worker that cannot be
+        # force-cancelled, so end the process to stop in-flight LLM and tool
+        # work instead of letting it burn tokens in the background.
+        click.secho("\n  Run cancelled.", fg="yellow")
+        sys.stdout.flush()
+        os._exit(130)
```
lucasgomide, line 149: are we ok with this hard kill that skips all cleanup (context managers, temp files)?

lib/cli/src/crewai_cli/crew_run_tui.py
```diff
@@ -681,6 +698,49 @@ def _run_crew_worker(self) -> None:
         except Exception as e:
             self.call_from_thread(self._on_crew_failed, str(e))
 
+    @work(thread=True, exclusive=True, group="crew")
```
lucasgomide: naming it "crew" for a flow worker is misleading.. . Consider renaming to something generic like "run" or "worker" if the group name matters for anything downstream

lib/cli/src/crewai_cli/run_declarative_flow.py
```diff
@@ -66,17 +66,180 @@ def run_declarative_flow(definition: str | Path, inputs: str | None = None) -> N
...
+    """
+    method_types: dict[str, str] = {}
+    try:
+        methods = getattr(getattr(flow, "_definition", None), "methods", None) or {}
+        for name, method_definition in methods.items():
+            call_type = getattr(getattr(method_definition, "do", None), "call", None)
+            if isinstance(call_type, str):
+                method_types[name] = call_type
+    except Exception:  # noqa: S110
+        pass
```
lucasgomide: Could we add minimum logging a debug-level warning would help when something unexpected breaks silently in the field

lib/cli/src/crewai_cli/run_declarative_flow.py
```diff
@@ -66,17 +66,180 @@ def run_declarative_flow(definition: str | Path, inputs: str | None = None) -> N
...
+
+    # The STEPS panel and header are driven by flow method events. A flow may
+    # declare ``config.suppress_flow_events`` (a headless/production
+    # optimization) which would leave STEPS stuck on "waiting…" here — so force
+    # emission on for the interactive TUI run. The headless path never reaches
+    # this and keeps the flow's declared setting.
+    try:
+        flow.suppress_flow_events = False
+    except Exception:  # noqa: S110
+        pass
```
lucasgomide: Same comment 
https://github.com/crewAIInc/crewAI/pull/6484/changes#r3547526880

lib/cli/src/crewai_cli/run_declarative_flow.py
joaomdmoura: Done in a86b2cb8f — the flow helpers (`_run_declarative_flow_tui`, `_flow_uses_human_feedback`, `_flow_method_types`) are now typed `Flow[Any]` via a `TYPE_CHECKING` import (keeps the CLI's lazy crewai import), and the defensive `getattr` chains are gone: `_definition` is a typed `PrivateAttr` and `name`/`suppress_flow_events` are typed fields, so `flow._definition.methods`, `flow.name`, etc. are direct attribute access now.

lib/cli/src/crewai_cli/crew_run_tui.py
joaomdmoura: Done in a86b2cb8f — the flow worker now uses `group="flow"` instead of `"crew"`. Nothing cancels/queries by that group name (it's only for Textual `exclusive` scheduling, and flow/crew workers never run in the same app instance), so it's purely a clarity fix.

lib/cli/src/crewai_cli/crew_run_tui.py
joaomdmoura: Good point — fixed in a86b2cb8f. `_on_crew_done`/`_on_crew_failed` are shared handlers, so instead of renaming them I made the user-facing wording entity-aware via a `_run_noun` property ("flow" in flow mode, "crew" otherwise). The activity-log text now reads "No result received before flow failed/completed" for flows. (The header + post-run summary were already flow-worded.) Test: `test_flow_done_uses_flow_wording_for_unfinished_tool`.

lib/cli/src/crewai_cli/run_declarative_flow.py
joaomdmoura: Done in a86b2cb8f — replaced the silent `except Exception: pass` with `logger.debug(..., exc_info=True)` so a malformed/unexpected definition is diagnosable instead of vanishing. (`_flow_method_types` stays best-effort since it only drives a cosmetic label, but it now logs.)

lib/cli/src/crewai_cli/run_declarative_flow.py
joaomdmoura: Done in a86b2cb8f — same treatment here: the `suppress_flow_events` toggle now logs at debug (`exc_info=True`) on failure instead of swallowing silently.

lib/cli/src/crewai_cli/crew_run_tui.py
joaomdmoura: It's safe — no change needed. `_flow_steps` never holds duplicate names: `on_method_started` looks up by name and only appends when absent (otherwise it re-activates the existing entry), and method names are unique keys in `flow._definition.methods`. So updating the first (only) match is correct. Your `!= status` idea would just skip redundant writes (a micro-opt), not fix a real case — happy to add it if you'd prefer, but it isn't needed for correctness.

lib/cli/src/crewai_cli/run_declarative_flow.py
joaomdmoura: Kept as-is for now, intentionally: this mirrors the existing declarative-crew path (`run_crew._run_json_crew`), which uses `os._exit(130)` on user-quit because `kickoff` runs in a non-cancellable thread worker — a graceful return would let it keep burning LLM/tool tokens in the background. So it's a deliberate, pre-existing convention; the cleanup trade-off is real but I didn't want to diverge from crews inside this PR. Happy to revisit the hard-kill for both crews and flows in a separate PR.

lucasgomide (approved):

PR #6485, opened 2026-07-08
fix(agents): exclude sanitized cache-read tool name from tool cache

Diwak4r:
## Summary
The guard in `ToolsHandler.on_tool_use` that prevents the built-in cache-read tool ("Hit Cache") from being added to the tool cache compared the raw default name `"Hit Cache"` against the sanitized runtime name `"hit_cache"`. They never match, so the guard was always `True` and the cache tool's own output was written into the cache — polluting it with dead entries and risking stale values.

## Fix
Compare against `sanitize_tool_name(CacheTools().name)` so the sanitized names align.

## Test
Added `test_cache_tool_result_is_not_cached` to `lib/crewai/tests/tools/test_tool_usage.py`. FAILS before the fix, PASSES after. Full `tests/tools/test_tool_usage.py` = 24 passed + new test (5 pre-existing event-bus timing failures are unrelated to this change).

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/lib/crewai/src/crewai/agents/tools_handler.py
+++ b/lib/crewai/src/crewai/agents/tools_handler.py
@@ -10,6 +10,7 @@ from pydantic import BaseModel, Field
 from crewai.agents.cache.cache_handler import CacheHandler
 from crewai.tools.cache_tools.cache_tools import CacheTools
 from crewai.tools.tool_calling import InstructorToolCalling, ToolCalling
+from crewai.utilities.string_utils import sanitize_tool_name
 
 
 class ToolsHandler(BaseModel):
@@ -37,7 +38,9 @@ class ToolsHandler(BaseModel):
             should_cache: Whether to cache the tool output.
         """
         self.last_used_tool = calling
-        if self.cache and should_cache and calling.tool_name != CacheTools().name:
+        if self.cache and should_cache and calling.tool_name != sanitize_tool_name(
+            CacheTools().name
+        ):
             input_str = ""
             if calling.arguments:
                 if isinstance(calling.arguments, dict):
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Confirmed the name mismatch directly: `CacheTools.name` is the literal class attribute `"Hit Cache"`, but the tool actually registers itself at runtime via `sanitize_tool_name(self.name)` (per `cache_tools.py`), which lowercases and underscores it to `"hit_cache"`. So the guard's `calling.tool_name != CacheTools().name` was comparing the sanitized runtime name against the raw display name — never equal, so the guard's `!=` was always `True` and the cache-read tool's own output got written into the cache every time. Fix aligns both sides to the sanitized form. Test fails on main and passes on the branch per the PR's own description, which lines up with what I traced through the source.

ErenAta16 (approved):
Verified this matches the actual bug: `CacheTools.name` is the raw class attribute `"Hit Cache"`, while the tool registers itself at runtime via `sanitize_tool_name(self.name)` (`"hit_cache"`), so the old `!=` comparison was always true and the cache-read tool's own output got written back into the cache on every call. The fix aligns both sides to the sanitized form, and the regression test fails on main / passes on the branch as described. LGTM.

Diwak4r:
Thank you @ErenAta16 for the approval and for independently validating the sanitized tool-name mismatch. The branch is mergeable, the regression is covered, and the available checks are green. This looks ready for maintainer merge when convenient.

Diwak4r:
The fix and regression test are now verified by ErenAta16, with CI passing and the PR mergeable. This is ready for maintainer merge.

Diwak4r:
Hi maintainers — just rebased this onto the latest main and CI is re-running on the updated commit. The change (excluding the sanitized cache-read tool name from tool cache) is small and self-contained. Whenever you have a moment, a review would be appreciated. Happy to address any feedback.

Diwak4r:
Rebased onto the latest main; CI is green on the updated head (4fdf8af). The change is unchanged (sanitized cache-read tool name exclusion + regression test). Ready for maintainer merge whenever convenient.

PR #6486, opened 2026-07-08
fix(tools): raise ToolUsageError instead of bare raise on non-dict args

HumphreySun98:
## Problem

`ToolUsage._original_tool_calling` has a bare `raise` on the non-dict path, which sits **outside** the preceding `try/except`:

```python
try:
    arguments = self._validate_tool_input(self.action.tool_input)
except Exception:
    if raise_error:
        raise            # OK — re-raises the active exception
    return ToolUsageError(...)

    if raise_error:
        raise            # BUG — no active exception here
    return ToolUsageError(...)
```

`_tool_calling()` calls this with `raise_error=True`, so if `_validate_tool_input` ever returns a non-dict, the bare `raise` fails with `RuntimeError: No active exception to re-raise` — masking the intended tool-arguments error. It's a latent landmine today (that helper currently always returns a dict or raises), but it's still incorrect.

## Fix

Raise the intended `ToolUsageError` explicitly:

```python
    error = ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
    if raise_error:
        raise error
    return error
```

The bare `raise` at the earlier site (inside `except`) is correct and left unchanged.

## Tests

Added two tests in `test_tool_usage.py` (mock `_validate_tool_input` → non-dict): `raise_error=True` raises `ToolUsageError` (not `RuntimeError`), and `raise_error=False` returns a `ToolUsageError`. Verified failing-without/passing-with the fix; full `test_tool_usage.py` (30) green; `ruff` + `mypy` clean.

Fixes #6430

---

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -849,9 +849,13 @@ class ToolUsage:
             return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
 
         if not isinstance(arguments, dict):
+            error = ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
+            # A bare ``raise`` here would fail with "No active exception to
+            # re-raise" (this path is outside the except block above), so raise
+            # the intended ToolUsageError explicitly.
             if raise_error:
-                raise
-            return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
+                raise error
+            return error
 
         return ToolCalling(
             tool_name=sanitize_tool_name(tool.name),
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Reproduced the underlying bug standalone: a bare `raise` outside an `except` block raises `RuntimeError: No active exception to reraise` instead of the intended `ToolUsageError`, so callers relying on `except ToolUsageError` around `_original_tool_calling` would get the wrong exception type entirely when `raise_error=True` and the arguments are not a dict.

This is one of five open PRs fixing the exact same line (#6512, #6511, #6499, #6495, and this one) — all correct, all equivalent in behavior. This version is the cleanest of the five: it builds the `ToolUsageError` once and reuses it for both the raise and the return branches (the other four construct it twice, once per branch, which is harmless but redundant), and the added comment explains *why* the bare `raise` fails here specifically, which is useful context for whoever reviews this cluster. Worth picking one of the five and closing the rest as duplicates — LGTM on this one as the pick.

PR #6488, opened 2026-07-08
fix(reasoning): detect a bare READY marker in text plans

### LINKED ISSUE #6204: [BUG] Reasoning plan always detects "NOT READY" even though the model indicates "READY"

HumphreySun98:
## Problem

With reasoning enabled, the executor asks the model to conclude its plan with **READY** or **NOT READY** and loops "refine plan" until ready. But there's a mismatch between what the prompts request and what the parser accepts:

- `refine_plan_prompt` (and one of the `create_plan_prompt` variants) say: *"Conclude with **READY or NOT READY**"* → models emit a bare `READY`.
- The text parser only accepted the exact sentence:
  ```python
  ready = "READY: I am ready to execute the task." in response
  ```

So a bare `READY` is never detected as ready — the agent stays stuck refining until max attempts. This bites text-parsing models (the reporter used Ollama GLM), where the trace clearly shows `READY` but it's read as not-ready.

## Fix

Add a `_detect_plan_ready` helper and use it at all three text-detection sites:

```python
_READY_MARKER = re.compile(r"NOT\s+READY|READY", re.IGNORECASE)

def _detect_plan_ready(text: str) -> bool:
    markers = _READY_MARKER.findall(text)
    if not markers:
        return False
    return "NOT" not in markers[-1].upper()
```

- Matches the **last** marker, so a "not ready" mentioned mid-plan doesn't override a final `READY`.
- `NOT READY` wins over the bare `READY` it contains.
- The old full sentence (`"READY: I am ready to execute the task."`) still counts as ready — backward compatible.

Only the text-parsing path is affected; the structured/function-calling path already uses a boolean `ready` field.

## Behavior (old → new)
| response ends with | old | new |
|---|---|---|
| `READY` | ❌ not ready | ✅ ready |
| `NOT READY: …` | not ready | not ready |
| `READY: I am ready to execute the task.` | ready | ready |

## Tests
Added `TestDetectPlanReady` (bare/full/none/mid-text/case-insensitive + `_parse_planning_response` integration). Full `test_structured_planning.py` (33) green; `ruff` + `mypy` clean. Branched off latest `main` (v1.15.2).

Fixes #6204

---

### DIFF
--- a/lib/crewai/src/crewai/utilities/reasoning_handler.py
+++ b/lib/crewai/src/crewai/utilities/reasoning_handler.py
@@ -29,7 +29,7 @@ if TYPE_CHECKING:
 
 
 _READY_MARKER: Final[re.Pattern[str]] = re.compile(
-    r"\bNOT\s+READY\b|\bREADY\b", re.IGNORECASE
+    r"^\s*(NOT\s+)?READY\b", re.IGNORECASE | re.MULTILINE
 )
 
 
@@ -41,9 +41,11 @@ def _detect_plan_ready(text: str) -> bool:
     bare ``READY`` rather than the full "READY: I am ready to execute the task."
     sentence, which the previous exact-substring check missed (#6204).
 
-    Matches the LAST readiness marker so a "not ready" mentioned earlier in the
-    plan doesn't override a final "READY", and so "NOT READY" wins over the bare
-    "READY" it contains.
+    The verdict is only recognized when READY / NOT READY is the first
+    non-whitespace on its own line, mirroring how the prompt asks the model to
+    conclude. This avoids treating ordinary prose (e.g. "I am ready to begin")
+    as a verdict, while still accepting "READY", "READY.", "READY: ...", etc.
+    The LAST such marker wins, so a later line overrides an earlier one.
 
     Args:
         text: The raw model response.
@@ -52,10 +54,12 @@ def _detect_plan_ready(text: str) -> bool:
         True if the concluding marker is READY, False for NOT READY or if no
         marker is present.
     """
-    markers = _READY_MARKER.findall(text)
+    # findall returns the optional ``(NOT\\s+)`` group: "" => READY, a non-empty
+    # value (e.g. "NOT ") => NOT READY.
+    markers: list[str] = _READY_MARKER.findall(text)
     if not markers:
         return False
-    return "NOT" not in markers[-1].upper()
+    return markers[-1].strip() == ""
 
 
 class ReasoningPlan(BaseModel):
--- a/lib/crewai/src/crewai/utilities/reasoning_handler.py
+++ b/lib/crewai/src/crewai/utilities/reasoning_handler.py
@@ -29,7 +29,7 @@ if TYPE_CHECKING:
 
 
 _READY_MARKER: Final[re.Pattern[str]] = re.compile(
-    r"NOT\s+READY|READY", re.IGNORECASE
+    r"\bNOT\s+READY\b|\bREADY\b", re.IGNORECASE
 )
 
 
--- a/lib/crewai/src/crewai/utilities/reasoning_handler.py
+++ b/lib/crewai/src/crewai/utilities/reasoning_handler.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import logging
+import re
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Checked the regex logic directly: `_detect_plan_ready` correctly picks up a bare `READY` where the old exact-substring check (`"READY: I am ready to execute the task." in response_str`) would not — confirmed this difference standalone (old check returns `False` on a bare `READY`, new one returns `True`). The last-marker-wins design is the right call too: an earlier "not ready yet" mentioned mid-plan (common phrasing while a model reasons through steps) shouldn't override the concluding verdict, and the `NOT` prefix check correctly makes "NOT READY" win over the bare "READY" substring it contains rather than the regex accidentally matching on "READY" first. Applied consistently across all three text-detection sites per the diff, which matters since a partial fix (only one of the three sites) would have left the bug intermittently reproducible depending on which code path a given model hit.

Kyou12138:
One edge case worth covering before merge: `_detect_plan_ready("I am ready to begin researching.")` currently returns `True` because the regex is word-bounded but not line-anchored.

The prompt defines READY / NOT READY as explicit verdict markers. Restricting matching to line starts, for example `^\s*(NOT\s+)?READY\b` with multiline matching, avoids ordinary prose being treated as a verdict while still supporting `READY`, `READY.`, and `READY: ...`.

I implemented and tested this case in #6564; sharing it here because I am closing that later duplicate.

ErenAta16:
Agreed with Kyou12138's finding, and confirmed it still reproduces against the current `_detect_plan_ready`:

```python
_detect_plan_ready("I am ready to begin researching.")  # -> True, false positive
```

`\bREADY\b` with `re.IGNORECASE` matches "ready" as an ordinary word anywhere in the text, so any sentence that happens to contain it becomes a verdict. I ran into the same root cause independently while reviewing #6564 (a since-closed duplicate) - same finding, different PR, which is a good sign it's not an edge case either of us is overreading.

Line-anchoring fixes it without giving up the last-marker-wins behavior:

```python
_READY_MARKER: Final[re.Pattern[str]] = re.compile(
    r"^\s*(NOT\s+)?READY\b", re.IGNORECASE | re.MULTILINE
)

def _detect_plan_ready(text: str) -> bool:
    markers = _READY_MARKER.findall(text)
    if not markers:
        return False
    return markers[-1].strip() == ""  # empty group means no "NOT" prefix
```

(`findall` with one capturing group returns the group content, so `""` means the `NOT\s+` branch didn't match). This only requires READY/NOT READY to be the first non-whitespace thing on its own line, which matches how the reasoning prompt actually asks the model to conclude, and still lets a later line override an earlier one.

Happy to see this land either here or on whichever of the duplicates ends up as the canonical one, just flagging so the fix doesn't get merged with the false-positive still open.

HumphreySun98:
Thanks @Kyou12138 and @ErenAta16 — great catch, and you're right: `\bREADY\b` still matched "ready" as an ordinary word (e.g. "I am ready to begin researching."), and worse, a trailing prose "ready" could flip a concluding `NOT READY`.

Applied the line-anchored approach you both proposed (`dd7f87a6`):

```python
_READY_MARKER = re.compile(r"^\s*(NOT\s+)?READY\b", re.IGNORECASE | re.MULTILINE)
```

So a verdict is only recognized when `READY` / `NOT READY` is the first non-whitespace on its own line — matching how the prompt asks the model to conclude — while keeping last-marker-wins so a later line still overrides an earlier one. Added regression tests for the prose false-positives (`"I am ready to begin researching."` → not a verdict; a `NOT READY` not flipped by trailing prose). 37 tests pass, ruff + mypy clean.

Appreciate you both independently digging into this — credited you in the commit.

ErenAta16:
Line-anchored regex is the right fix, that closes the exact gap. Nice work turning it around fast.

PR #6494, opened 2026-07-09
fix: support async callables in akickoff before/after callbacks

### LINKED ISSUE #6481: [BUG] before/after_kickoff_callbacks do not support async callables in akickoff

ashusnapx:
## Summary

Fixes silent discarding of async callables in `Crew.akickoff()`.

**Closes #6481**

### Problem

`Crew.akickoff()` is a native async execution path, but `before_kickoff_callbacks` and `after_kickoff_callbacks` were invoked synchronously with no `inspect.isawaitable` check. Async callables were silently discarded — their coroutine objects were never awaited.

```python
async def my_callback(result):
    await some_async_operation()  # never awaited
    return result

crew = Crew(..., after_kickoff_callbacks=[my_callback])
await crew.akickoff()  # callback silently discarded
```

### Fix

| Component | Change |
|---|---|
| `after_kickoff_callbacks` in `akickoff()` | Added `inspect.isawaitable` check + `await` |
| `before_kickoff_callbacks` in `akickoff()` | Created `aprepare_kickoff()` async variant that awaits async callbacks |

Follows the same pattern used by `task_callback` and `step_callback` throughout the codebase:

```python
cb_result = callback(result)
    await cb_result
```

### Files Changed

| File | Change |
|---|---|
| `crew.py` | Added `import inspect`, fixed after_callback loop, imports `aprepare_kickoff` |
| `crews/utils.py` | Added `aprepare_kickoff()` async variant with `inspect.isawaitable` for before_callbacks |
| `tests/crew/test_async_crew.py` | 3 new tests: async before, async after, mixed sync+async |

### Tests

```bash
# Run async crew tests
pytest lib/crewai/tests/crew/test_async_crew.py -v

# Run all crew tests
pytest lib/crewai/tests/test_crew.py lib/crewai/tests/crew/ -v
```

- 14 async crew tests pass (3 new)
- ruff: all checks passed
- mypy: no issues found

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -5,6 +5,7 @@ from collections.abc import Callable, Sequence
 from concurrent.futures import Future
 from copy import copy as shallow_copy
 from hashlib import md5
+import inspect
 import json
 from pathlib import Path
 import re
@@ -67,6 +68,7 @@ from crewai.agents.cache.cache_handler import CacheHandler
 from crewai.crews.crew_output import CrewOutput
 from crewai.crews.utils import (
     StreamingContext,
+    aprepare_kickoff,
     check_conditional_skip,
     enable_agent_streaming,
     prepare_kickoff,
@@ -1230,7 +1232,7 @@ class Crew(FlowTrackable, BaseModel):
 
         runtime_scope = crewai_event_bus._enter_runtime_scope()
         try:
-            inputs = prepare_kickoff(self, inputs, input_files)
+            inputs = await aprepare_kickoff(self, inputs, input_files)
 
             if self.process == Process.sequential:
                 result = await self._arun_sequential_process()
@@ -1243,6 +1245,8 @@ class Crew(FlowTrackable, BaseModel):
 
             for after_callback in self.after_kickoff_callbacks:
                 result = after_callback(result)
+                if inspect.isawaitable(result):
+                    result = await result
 
             result = self._post_kickoff(result)
 
--- a/lib/crewai/src/crewai/crews/utils.py
+++ b/lib/crewai/src/crewai/crews/utils.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import Callable, Coroutine, Iterable, Mapping
+import inspect
 from typing import TYPE_CHECKING, Any
 
 from opentelemetry import baggage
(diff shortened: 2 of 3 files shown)


ErenAta16 (commented):
Same underlying fix as #6500 and #6482 (also open) — the `crew.py` change here is identical to both. Where this one differs is `crews/utils.py`: it copy-pastes the full body of `prepare_kickoff` into a new `aprepare_kickoff` rather than factoring out the shared setup/teardown logic, which #6500 and #6482 both do via extracted helpers. Functionally correct either way and the test coverage here (including the `test_akickoff_mixed_sync_async_callbacks` ordering check) is solid, but the duplication means a future change to the shared kickoff-preparation logic would need to be applied in two places instead of one — worth weighing against the other two when picking which of the three to merge.

ashusnapx:
Refactored - extracted shared logic into _prepare_kickoff_common and _run_before_callbacks. Both prepare_kickoff and aprepare_kickoff now call the shared helpers. No more duplication.

ErenAta16:
Confirmed - `prepare_kickoff` and `aprepare_kickoff` now both go through `_normalize_inputs` and `_prepare_kickoff_common`, with `_run_before_callbacks` handling the sync path and the async version inlining its own `inspect.isawaitable` loop (needed since it has to await mid-loop, so it can't reuse the sync helper as-is). Ran the full `test_async_crew.py` - 14 passed, including the mixed sync/async callback case. No more full-body duplication between the two entry points. Good refactor.

ashusnapx:
Thanks for confirming. Ready for maintainer review.

ErenAta16:
@ashusnapx I think there's been a mix-up, and I'd rather flag it than let it ride: the sentinel problem I described was in **#6500**, not here. Your original version didn't have it. In my comparison table on #6482 I had #6494 marked "callbacks can double-run: no", which I re-verified before writing:

```
=== single callback returning None ===
   sync invocations:  1
   async invocations: 1     <- original #6494, no double-run
```

The double-run needs the "has the caller already applied the callbacks?" flag that #6500 uses. Your structure never had that question to answer, since `aprepare_kickoff` always runs its own loop.

Unfortunately 60a7679 has introduced a real divergence while fixing the non-problem. The `if normalized is None: normalized = {}` guard **inside** the sync loop isn't a first-iteration initializer, it re-normalizes on every pass. That's what recovers when a callback returns `None`: the next callback still receives a dict. Replacing it with an `_UNSET` check that only fires on the first iteration drops that recovery.

Two callbacks, first one mutates in place and returns `None`:

```
-- sync (current main) --
   cb2 received: {'from_cb2': True}   -> completes

-- async (60a7679) --
   TypeError: 'NoneType' object does not support item assignment
```

So `akickoff` now crashes on a callback chain that `kickoff` handles. That's a worse failure than the one this PR set out to fix, because it's an exception rather than a silent no-op, and it only triggers with two or more callbacks where an early one doesn't return the dict.

Keeping the per-iteration guard alongside the awaiting is enough, no sentinel needed:

```python
async def aprepare_kickoff(crew, inputs, input_files=None):
    normalized = _normalize_inputs(crew, inputs)
        if normalized is None:
            normalized = {}
        result = before_callback(normalized)
            result = await result
        normalized = result
    return _prepare_kickoff_common(crew, normalized, input_files)
```

That's line-for-line the sync loop with two lines added for awaiting, which also makes it obvious at review time that the two paths agree.

On your docstring's reasoning, "the next callback would then receive the pre-callback inputs instead of the callback's return value": that isn't what the sync code does. It passes a fresh `{}`, not the pre-callback inputs, because `normalized` was already overwritten with `None` by the assignment before the guard re-runs. So there's nothing to preserve there.

Also worth a test either way. None of the three PRs covers a multi-callback chain where an early callback returns `None`, which is why this slipped through, and it's the case that separates the two designs.

Sorry for the confusion on my end, my #6482 comment named #6500 but I can see how it read as applying to all three.

PR #6495, opened 2026-07-09
fix(tools): raise ToolUsageError instead of bare raise for non-dict arguments

ashusnapx:
## Summary

Fixes `RuntimeError: No active exception to re-raise` in `ToolUsage._original_tool_calling()` when `_validate_tool_input` returns a non-dict and `raise_error=True`.

**Closes #6430**

### Problem

`_original_tool_calling()` at `tool_usage.py:853` had a bare `raise` outside any `except` block:

```python
    if raise_error:
        raise   # <-- no active exception here
    return ToolUsageError(...)
```

When this path was hit with `raise_error=True`, Python raised `RuntimeError: No active exception to re-raise` instead of the intended `ToolUsageError`. The `_tool_calling()` fallback relies on catching a meaningful exception to trigger its retry logic, so this produced a confusing RuntimeError.

Also flagged by `ruff check --select PLE0704`.

### Fix

Replace bare `raise` with explicit `raise ToolUsageError(...)`:

```python
    if raise_error:
        raise ToolUsageError(
            f"{I18N_DEFAULT.errors('tool_arguments_error')}"
        )
```

The `except` block above (line 846-848) correctly uses bare `raise` inside an active exception context — left unchanged.

### Tests (4 new)

| Test | What it verifies |
|---|---|
| `test_original_tool_calling_non_dict_raises_tool_usage_error` | `raise_error=True` raises `ToolUsageError`, not `RuntimeError` |
| `test_original_tool_calling_non_dict_returns_error` | `raise_error=False` returns `ToolUsageError` |
| `test_tool_calling_fallback_on_non_dict` | Full `_tool_calling` flow falls back gracefully |
| `test_non_dict_raises_not_runtime_error` | Exception type is `ToolUsageError`, not `RuntimeError` |

All 32 tests in `test_tool_usage.py` pass. `ruff check --select PLE0704` clean.

### Verification

```bash
pytest lib/crewai/tests/tools/test_tool_usage.py -v
ruff check lib/crewai/src/crewai/tools/tool_usage.py --select PLE0704
```

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -840,18 +840,21 @@ class ToolUsage:
     ) -> ToolCalling | InstructorToolCalling | ToolUsageError:
         tool_name = self.action.tool
         tool = self._select_tool(tool_name)
+        tool_arguments_error = ToolUsageError(
+            f"{I18N_DEFAULT.errors('tool_arguments_error')}"
+        )
         try:
             arguments = self._validate_tool_input(self.action.tool_input)
 
         except Exception:
             if raise_error:
                 raise
-            return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
+            return tool_arguments_error
 
         if not isinstance(arguments, dict):
             if raise_error:
-                raise
-            return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
+                raise tool_arguments_error
+            return tool_arguments_error
 
         return ToolCalling(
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):

ashusnapx:
Fixed — error is now constructed once and reused for both branches (raise and return), matching the approach in #6486. Thanks for the catch.

ErenAta16:
Confirmed - the current diff constructs `tool_arguments_error` once before the try block and both the `raise` and the `return` paths reuse the same instance now. That's the same shape as #6486, so this fix is correct.

ashusnapx:
Thanks for confirming. Ready for maintainer review.

PR #6499, opened 2026-07-09
fix(tools): raise ToolUsageError instead of bare raise for non-dict arguments

nolanchic:
## Description

`ToolUsage._original_tool_calling()` used a bare `raise` **outside any active `except` block** when the parsed tool arguments were not a `dict`:

```python
    if raise_error:
        raise          # <- no active exception here
    return ToolUsageError(...)
```

A bare `raise` re-raises the exception currently being handled, but on this path the preceding `try/except` completed **normally**, so no exception is active. Result: with `raise_error=True` the call raised `RuntimeError: No active exception to reraise` instead of the intended `ToolUsageError`, masking the real tool-arguments problem.

Fix: raise the explicit `ToolUsageError`, matching the message already used by the `raise_error=False` branch on the next line.

```python
    if raise_error:
        raise ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
```

## Why the bug was latent

`_validate_tool_input()` currently always returns a `dict` or raises, so the non-dict branch was effectively dead code in normal use. The bug surfaces whenever that contract is relaxed — e.g. subclasses overriding `_validate_tool_input`, monkeypatching in tests, or future changes to the validation pipeline. Making the branch raise the correct exception keeps the failure mode meaningful instead of an opaque `RuntimeError`.

## Test plan

- [x] Added regression test `test_original_tool_calling_raises_tool_usage_error_for_non_dict_arguments` — fails with `RuntimeError` on the old code, passes with `ToolUsageError` after the fix.
- [x] Added `test_original_tool_calling_returns_tool_usage_error_for_non_dict_without_raise` covering the `raise_error=False` path.
- [x] `uv run pytest lib/crewai/tests/tools/` → 135 passed.
- [x] `uv run ruff check lib/` → clean.
- [x] `uv run mypy lib/crewai/src/crewai/tools/tool_usage.py` → clean (strict).

Fixes #6430.

> Note: per `.github/CONTRIBUTING.md`, this PR is authored by an AI agent (Claude Code) and carries the `llm-generated` label.

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -850,7 +850,7 @@ class ToolUsage:
 
         if not isinstance(arguments, dict):
             if raise_error:
-                raise
+                raise ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
             return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
 
         return ToolCalling(
--- a/lib/crewai/tests/tools/test_tool_usage.py
+++ b/lib/crewai/tests/tools/test_tool_usage.py
@@ -25,7 +25,7 @@ from crewai.hooks.tool_hooks import (
 )
 from crewai.tools import BaseTool
 from crewai.tools.tool_calling import ToolCalling
-from crewai.tools.tool_usage import ToolUsage
+from crewai.tools.tool_usage import ToolUsage, ToolUsageError
 from crewai.utilities.tool_utils import execute_tool_and_check_finality
 from pydantic import BaseModel, Field
 import pytest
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):

nolanchic:
Thanks @ErenAta16 for triaging the cluster — agreed #6486 is the cleanest of the five (constructs the error once and reuses it across both branches, plus the comment explaining why the bare `raise` fails). Happy to close #6499 as a duplicate once #6486 lands; just say the word. If plans change and you'd rather take this one, I can quickly address the CodeRabbit nit (add a `match=` on the error-message assertion).

ErenAta16:
Yes, closing this in favor of #6486 makes sense, no reason to carry two open PRs for the same one-line fix once the cleaner one lands. Appreciate the offer to address the CodeRabbit nit, but I'd rather not add more to review across a PR that's headed for closure. Thanks for the quick response on the cluster.

nolanchic:
Closing in favor of #6486 per @ErenAta16 — that PR constructs the error once and reuses it across both branches, which is cleaner than the duplicate construction here. Thanks for the review and the detailed comparison.

PR #6500, opened 2026-07-09
fix(crew): await async before/after_kickoff_callbacks in akickoff

### LINKED ISSUE #6481: [BUG] before/after_kickoff_callbacks do not support async callables in akickoff

### LINKED ISSUE #7607: [BUG] Crew.akickoff still does not await async kickoff callbacks (follow-up to #6481)
### Description

`Crew.akickoff()` does not await async before/after kickoff callbacks. An async before-callback's body is not executed when called by synchronous preparation; an async after-callback replaces the result with an unawaited coroutine before post-processing. This report continues #6481, which was closed for inactivity, not because a fix was merged.

### Steps to Reproduce

1. Create an otherwise working sequential Crew with a task and its assigned agent.
2. Attach the async callback below to `after_kickoff_callbacks`.
3. Run `result = await crew.akickoff()`.
4. Check whether the callback executed and the result remains a `CrewOutput`. Repeat with an async before-callback returning the input dict.

### Expected behavior

The native async path awaits awaitable callback results, applies returned inputs/output, and executes each callback once. Synchronous callbacks retain their existing behavior. Offloading blocking synchronous callbacks is outside this report's scope.

### Screenshots/Code snippets

```python
called = False

async def after_callback(output):
    global called
    called = True
    return output

# Given an otherwise working Crew:
crew.after_kickoff_callbacks = [after_callback]
result = await crew.akickoff()
assert called
```

### Operating System

Original report: macOS Sonoma. Current verification is source inspection, not a new OS-specific runtime reproduction.

### Python Version

Original report: 3.12.

### crewAI Version

Originally reported on 1.15.2; source inspected again on 2026-09-19 at main commit `3831e8b6c86f78cb3cde18ebf7be0d197b958f0e`.

### crewAI Tools Version

Not required: this concerns core Crew kickoff callbacks.

### Virtual Environment

Venv (original report).

### Evidence

Current main still calls synchronous `prepare_kickoff` in `akickoff` and directly assigns `result = after_callback(result)` without awaiting it: https://github.com/crewAIInc/crewAI/blob/3831e8b6c86f78cb3cde18ebf7be0d197b958f0e/lib/crewai/src/crewai/crew.py#L1275-L1289 . No fresh full-suite test run is claimed here.

Independent historical verification by ErenAta16 on #6500 covered (1) a callback returning None, (2) empty callbacks with inputs=None, and (3) the synchronous control case. The reviewer confirmed the distinct sentinel prevents duplicate callback execution and recommended retaining it: https://github.com/crewAIInc/crewAI/pull/6500#issuecomment-5049732531 . This is historical review evidence, not a maintainer approval of the latest rebased revision.

### Possible Solution

Resume review of #6500: await an async preparation entry point and await after-callback results when awaitable. Preserve callback-chain None semantics and exactly-once execution. Validate current hook ordering and regressions against current main before merging.

### Additional context

Five related implementation attempts are #6482, #6494, #6500, #6547, and #6697. All five are currently closed and none is recorded as merged. This is not a claim that their technical merits are identical.

Original report: #6481. It was automatically closed for inactivity on August 28. On September 17, #6500 was automatically closed because first-time contributors require an associated OPEN issue. The bot explicitly allows reopening after linking an open issue: https://github.com/crewAIInc/crewAI/pull/6500#issuecomment-5709744042 . This report supplies that active tracking issue while preserving the existing review history.

At inspection, #6500 head `7c202258fde62b7502403c09204f304062fb79e9` is 9 commits behind main; it is not claimed to be fully up to date.

AI disclosure: this issue was prepared with OpenAI Codex. The `llm-generated` label is requested per CONTRIBUTING.md.

nolanchic:
## Description

`Crew.akickoff()` is a native async path, but `before_kickoff_callbacks` and `after_kickoff_callbacks` were invoked without any awaitable check — **silently dropping async callables**:

- **`after_kickoff_callbacks`**: `result = after_callback(result)` returned an unawaited coroutine when the callback was `async`, so the crew result was silently replaced with a coroutine object (never executed, never awaited).
- **`before_kickoff_callbacks`**: invoked inside the synchronous `prepare_kickoff()`, which has no async support — async before-callbacks never ran.

This is inconsistent with `task_callback` and `step_callback`, which correctly check `inspect.isawaitable` in the async path.

> Note: this change is scoped to stopping the silent drop of async callbacks. It does not route synchronous callbacks off the event loop (e.g. via `run_in_executor`); a slow sync callback still runs inline, same as before. That is a separate, larger change.

## Changes

- **`crews/utils.py`**: add `async def aprepare_kickoff(...)` — the async counterpart of `prepare_kickoff`. It awaits before-callbacks that return an awaitable (sync callbacks run unchanged). The shared body is extracted into `_prepare_kickoff_impl(...)` so the two entry points stay DRY and behavior is identical apart from callback awaiting.
- **`crew.py`**: `akickoff` now calls `await aprepare_kickoff(...)` instead of `prepare_kickoff(...)`, and awaits each after-callback result when it is awaitable.

### Behavior preserved
- `kickoff()` (sync) — unchanged, still uses `prepare_kickoff`.
- `kickoff_async()` — unchanged; it wraps the sync `kickoff` in `asyncio.to_thread`, so all callbacks already run off the event loop.
- Sync callbacks in `akickoff` — unchanged (an `isawaitable` check is a no-op for them).

## Test plan

- [x] `test_akickoff_awaits_async_after_callbacks` — async after-callback is awaited; result stays a `CrewOutput` (was a coroutine before).
- [x] `test_akickoff_applies_async_after_callback_result` — async after-callback's returned `CrewOutput` replaces the result.
- [x] `test_akickoff_awaits_async_before_callbacks` — async before-callback is awaited and runs.
- [x] `test_akickoff_applies_async_before_callback_inputs` — async before-callback's returned inputs flow into the crew.
- [x] `uv run pytest lib/crewai/tests/crew/test_async_crew.py lib/crewai/tests/test_crew.py lib/crewai/tests/test_checkpoint.py` → all pass (sync kickoff + kickoff_async callbacks unaffected).
- [x] `uv run ruff check lib/` → clean.
- [x] `uv run mypy lib/crewai/src/crewai/crew.py lib/crewai/src/crewai/crews/utils.py` → clean (strict).

Fixes #7607.

Historical report: #6481 (closed for inactivity). #7607 preserves the related PR history and independent verification.

> Per `.github/CONTRIBUTING.md`, this PR is authored by an AI agent (Claude Code) and should carry the `llm-generated` label (external contributors can't self-apply labels).

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -5,6 +5,7 @@ from collections.abc import Callable, Sequence
 from concurrent.futures import Future
 from copy import copy as shallow_copy
 from hashlib import md5
+import inspect
 import json
 from pathlib import Path
 import re
@@ -67,6 +68,7 @@ from crewai.agents.cache.cache_handler import CacheHandler
 from crewai.crews.crew_output import CrewOutput
 from crewai.crews.utils import (
     StreamingContext,
+    aprepare_kickoff,
     check_conditional_skip,
     enable_agent_streaming,
     prepare_kickoff,
@@ -1254,7 +1256,7 @@ class Crew(FlowTrackable, BaseModel):
 
         runtime_scope = crewai_event_bus._enter_runtime_scope()
         try:
-            inputs = prepare_kickoff(self, inputs, input_files)
+            inputs = await aprepare_kickoff(self, inputs, input_files)
 
             if self.process == Process.sequential:
                 result = await self._arun_sequential_process()
@@ -1267,6 +1269,8 @@ class Crew(FlowTrackable, BaseModel):
 
             for after_callback in self.after_kickoff_callbacks:
                 result = after_callback(result)
+                if inspect.isawaitable(result):
+                    result = await result
 
             result = self._post_kickoff(result)
 
--- a/lib/crewai/src/crewai/crews/utils.py
+++ b/lib/crewai/src/crewai/crews/utils.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 from collections.abc import Callable, Coroutine, Iterable, Mapping
+import inspect
 from typing import TYPE_CHECKING, Any
 
 from opentelemetry import baggage
@@ -258,6 +259,89 @@ def prepare_kickoff(
     Returns:
         The potentially modified inputs dictionary after before callbacks.
     """
+    return _prepare_kickoff_impl(crew, inputs, input_files, normalized_inputs=None)
+
+
+async def aprepare_kickoff(
+    crew: Crew,
+    inputs: dict[str, Any] | None,
+    input_files: dict[str, FileInput] | None = None,
+) -> dict[str, Any] | None:
+    """Async counterpart of :func:`prepare_kickoff`.
+
+    Used by ``Crew.akickoff`` so that async ``before_kickoff_callbacks`` are
+    awaited instead of silently dropped, and any blocking work inside them
+    does not stall the event loop. Sync before-callbacks continue to work
+    unchanged.
+
+    Args:
+        crew: The crew instance to prepare.
+        inputs: Optional input dictionary to pass to the crew.
+        input_files: Optional dict of named file inputs for the crew.
+
+    Returns:
+        The potentially modified inputs dictionary after before callbacks.
+    """
+    normalized = await _arun_before_kickoff_callbacks(crew, _normalize_inputs(inputs))
+    return _prepare_kickoff_impl(
+        crew, inputs, input_files, normalized_inputs=normalized
+    )
+
+
+def _normalize_inputs(inputs: dict[str, Any] | None) -> dict[str, Any] | None:
+    """Validate and copy ``inputs`` into a normalized dict, preserving ``None``."""
+    if inputs is None:
+        return None
+    if not isinstance(inputs, Mapping):
+        raise TypeError(
+            f"inputs must be a dict or Mapping, got {type(inputs).__name__}"
+        )
+    return dict(inputs)
+
+
+def _run_before_kickoff_callbacks(
+    crew: Crew, normalized: dict[str, Any] | None
+) -> dict[str, Any]:
+    """Run sync ``before_kickoff_callbacks``, returning the (possibly) new inputs."""
+    for before_callback in crew.before_kickoff_callbacks:
+        if normalized is None:
+            normalized = {}
+        normalized = before_callback(normalized)
+    return normalized  # type: ignore[return-value]
+
+
+async def _arun_before_kickoff_callbacks(
+    crew: Crew, normalized: dict[str, Any] | None
+) -> dict[str, Any] | None:
+    """Run ``before_kickoff_callbacks`` with async support.
+
+    Awaits callbacks that return an awaitable (coroutine functions), matching
+    how ``task_callback`` and ``step_callback`` are handled in the async path.
+    Sync callbacks run unchanged.
+    """
+    for before_callback in crew.before_kickoff_callbacks:
+        if normalized is None:
+            normalized = {}
+        result = before_callback(normalized)
+        if inspect.isawaitable(result):
+            result = await result
+        normalized = result
+    return normalized
+
+
+def _prepare_kickoff_impl(
+    crew: Crew,
+    inputs: dict[str, Any] | None,
+    input_files: dict[str, FileInput] | None,
+    *,
+    normalized_inputs: dict[str, Any] | None,
+) -> dict[str, Any] | None:
+    """Shared body of :func:`prepare_kickoff` and :func:`aprepare_kickoff`.
+
+    When ``normalized_inputs`` is ``None`` the sync before-callbacks are run
+    here; otherwise the caller (the async path) has already applied them and
+    they are skipped to avoid double-execution.
+    """
     from crewai.events.base_events import reset_emission_counter
     from crewai.events.event_bus import crewai_event_bus
     from crewai.events.event_context import (
@@ -275,13 +359,11 @@ def prepare_kickoff(
     from crewai.hooks.contexts import ExecutionStartContext, InputContext
     from crewai.hooks.dispatch import InterceptionPoint, dispatch
 
-    normalized: dict[str, Any] | None = None
-    if inputs is not None:
-        if not isinstance(inputs, Mapping):
-            raise TypeError(
-                f"inputs must be a dict or Mapping, got {type(inputs).__name__}"
-            )
-        normalized = dict(inputs)
+    if normalized_inputs is None:
+        normalized = _normalize_inputs(inputs)
+        normalized = _run_before_kickoff_callbacks(crew, normalized)
+    else:
+        normalized = normalized_inputs
 
     # ``inputs`` aliases the same object as ``payload`` (not a fresh ``{}`` from
     # ``or``) so in-place edits to either survive read-back, per the context
@@ -298,11 +380,6 @@ def prepare_kickoff(
     dispatch(InterceptionPoint.EXECUTION_START, start_ctx)
     crew._execution_start_dispatched = True
     normalized = start_ctx.payload
-
-    for before_callback in crew.before_kickoff_callbacks:
-        if normalized is None:
-            normalized = {}
-        normalized = before_callback(normalized)
 
     input_ctx = InputContext(
         crew=crew,
--- a/lib/crewai/src/crewai/crews/utils.py
+++ b/lib/crewai/src/crewai/crews/utils.py
@@ -331,12 +331,13 @@ async def _arun_before_kickoff_callbacks(
 
 #: Sentinel marking that no caller has pre-applied the before callbacks.
 #:
-#: Using a distinct object (instead of ``None``) lets ``None`` round-trip as a
-#: legitimate value: a before-callback that returns ``None`` (e.g. one that
-#: mutates the dict in place and falls off the end of the function) must not be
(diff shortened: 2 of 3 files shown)


ErenAta16 (commented):
This, #6494, and #6482 are three open PRs fixing the same bug (async `before`/`after_kickoff_callbacks` silently discarded in `akickoff`) — the `crew.py` hunk is identical across all three (same `inspect.isawaitable` guard around the after-callback loop, same `aprepare_kickoff` swap-in). Compared the three `crews/utils.py` implementations, since that is where they actually diverge:

This one and #6482 both extract shared helpers so `prepare_kickoff` (sync) and `aprepare_kickoff` (async) share the common setup/teardown logic (event emission, input interpolation, agent setup, planning) rather than duplicating it — this one via a single `_prepare_kickoff_impl` taking an optional `normalized_inputs` override, #6482 via three smaller helpers (`_begin_prepare_kickoff`/`_normalize_inputs`/`_finish_prepare_kickoff`). #6494 instead copy-pastes the full ~80-line body of `prepare_kickoff` into a new `aprepare_kickoff`, which works but means the two functions can drift out of sync on a future change to the shared logic.

One honest scope note that applies equally to all three: the descriptions (this PR included) mention IO-bound *sync* callbacks blocking the event loop as part of the motivation, but none of the three actually route sync callbacks through `run_in_executor` or similar — they only add the `inspect.isawaitable` check, so a slow sync callback still runs inline on the event loop same as before. Not a defect in this PR specifically, just flagging that the fix (correctly) narrows to "stop silently dropping async callables" rather than also solving the blocking-event-loop half of the stated problem.

nolanchic:
Agree — this PR narrows to "stop silently dropping async callables", and routing sync callbacks through `run_in_executor` is deliberately out of scope, as you noted. The blocking-I/O line in the description was meant as motivation for making `akickoff` async-aware, not a claim that this change also fixes event-loop blocking.

I trimmed that framing from the description and added an explicit scope note that sync callbacks still run inline, so the description no longer overstates what the change does.

ErenAta16:
Thanks for trimming the description, that reads accurately now.

Came back to look at the `normalized_inputs` sentinel more carefully and found a real edge case in it. `None` is doing double duty: it means both "caller did not pre-apply the callbacks" and, unavoidably, "caller pre-applied them and the result was `None`". `_arun_before_kickoff_callbacks` can genuinely return `None`, since it assigns `normalized = result` with whatever the callback handed back. So a before-callback that returns `None` makes `aprepare_kickoff` pass `normalized_inputs=None`, and `_prepare_kickoff_impl` then treats that as "not yet applied" and runs every before-callback a second time.

Simulated the exact control flow from this branch:

```
=== async path, callback mutates in place and returns None ===
  callback invocations: 2 (expected 1)
  result: None

=== sync path, same callback (control) ===
  callback invocations: 1 (expected 1)

=== async path, callback returns the dict ===
  callback invocations: 1 (expected 1)
```

So it's async-only, and only for callbacks that don't return the dict. That's not an exotic user error though, mutating the passed dict in place and falling off the end of the function is a natural thing to write, and it happens to work fine on `kickoff` today. The failure is also quiet: the callbacks just run twice, so anything non-idempotent in them (appending to a list, incrementing a counter, an API call) silently doubles.

Worth noting the async path is the one place where a double-run is most likely to be visibly wrong, since an async callback re-run through `_run_before_kickoff_callbacks` returns an un-awaited coroutine, which is the original bug this PR exists to fix.

A distinct sentinel avoids the ambiguity entirely:

```python
_NOT_APPLIED = object()

def _prepare_kickoff_impl(crew, inputs, input_files, *, normalized_inputs=_NOT_APPLIED):
    if normalized_inputs is _NOT_APPLIED:
        normalized_inputs = _run_before_kickoff_callbacks(crew, _normalize_inputs(inputs))
```

Then `None` is just a value that round-trips, and the sync path keeps working by not passing the argument at all.

Alternatively, restructuring so the impl never runs callbacks (both `prepare_kickoff` and `aprepare_kickoff` apply their own, then call a shared tail) sidesteps the flag altogether, which is closer to what #6482 and #6494 do. That's a bigger diff, but it removes the "did someone already do this?" question rather than encoding the answer in a parameter.

Either way it'd be worth a test for the returns-`None` callback on the async path, since none of the three PRs currently covers it.

nolanchic:
Good catch — confirmed. The `None` collision reproduces exactly as you described: an async before-callback that mutates the dict in place and returns `None` gets re-run, while the same callback on the sync path runs once.

Went with the sentinel:

```python
_NOT_APPLIED = object()

def _prepare_kickoff_impl(crew, inputs, input_files, *, normalized_inputs=_NOT_APPLIED):
    if normalized_inputs is _NOT_APPLIED:
        normalized_inputs = _run_before_kickoff_callbacks(crew, _normalize_inputs(inputs))
```

`prepare_kickoff` no longer passes the argument, so the sync path is unchanged, and a pre-applied `None` now round-trips instead of triggering a re-run.

I considered the restructure (impl never runs callbacks; both callers apply their own and call a shared tail — what #6482 / #6494 do). It's cleaner in that it removes the flag entirely, but it's a meaningfully larger diff, and you'd just signed off on narrowing this PR's scope to "stop silently dropping async callables", so I kept the sentinel to stay minimal. Happy to switch to the restructure here if you'd rather have one canonical implementation rather than three variants — just say the word.

Added the two tests you flagged as missing across all three PRs:
- `test_akickoff_before_callback_returning_none_runs_once` — the async returns-`None` case. It spies `_run_before_kickoff_callbacks` and asserts it's never entered on the async path (`call_count == 0`), so the failure surfaces as the double-run itself rather than the downstream un-awaited-coroutine symptom. Verified it fails on the pre-sentinel code.
- `test_akickoff_async_before_callback_returning_none_preserves_none` — the empty-callback-list + `inputs=None` edge, where `aprepare_kickoff` legitimately produces `None`; confirms it round-trips without being treated as "not yet applied".

One clarification on the `None` semantics, since "round-trips" can be read two ways: the sentinel stops `None` from being misread as "not yet applied" (the bug you found), so a returns-`None` callback no longer causes a double-run. But a callback that returns `None` still doesn't flow its inputs to the crew on either path — the runner keeps whatever the callback returned, and returning the (possibly mutated) dict is required for inputs to reach task interpolation. That's the existing behavior on sync `kickoff` today, not something this PR changes; I called it out in the `_prepare_kickoff_impl` docstring rather than silently relying on it. If you'd rather have returns-`None` preserve the prior dict (via `normalized = result if result is not None else normalized` in the runner), that's a behavioral change I'd rather do in its own PR.

Also cleared a `mypy --strict` regression this branch had introduced (`_prepare_kickoff_impl` returned `Any` because `normalized` was last assigned from `InterceptionContext.payload`) — cast it back to `dict | None` on return.

Pushed in `0e83479`.

ErenAta16:
Ran the `_NOT_APPLIED` version through the same three cases and it's clean:

```
1. single callback returns None   -> sync runner entered 0x   (was the double-run)
2. empty callbacks + inputs=None  -> sync runner entered 0x
3. sync path, normal callback     -> sync runner entered 1x, result {'a': 1, 'ok': True}
```

Case 2 was the one I'd have most likely missed, so good instinct adding it. `aprepare_kickoff` legitimately returning `None` there is indistinguishable from the un-applied state under the old check, and it doesn't require a misbehaving callback to reach, just no callbacks and no inputs.

Asserting on `_run_before_kickoff_callbacks.call_count == 0` rather than on the downstream symptom is also the right call. The un-awaited-coroutine warning is asynchronous and easy to swallow, so a test hung off it would be flaky in exactly the situation it's meant to catch.

**On the returns-`None` semantics:** your read is right, and I checked it rather than take it on trust. Both paths lose the inputs, and with two callbacks the first one's in-place mutations are dropped too:

```
sync : cb1 mutates + returns None -> cb2 receives {}   ('touched' gone)
async: same
```

So `normalized = before_callback(normalized)` discards the reference to the dict the callback just mutated, and the `if normalized is None: normalized = {}` guard hands the next callback a fresh one. Pre-existing on `kickoff`, both paths identical, and correctly out of scope here. Agreed it's a separate PR, and I'd say a worthwhile one, since "mutate the dict in place and don't bother returning it" is a natural thing to write and silently does nothing. Documenting it in the docstring rather than leaning on it silently is the right interim move.

**On sentinel vs restructure, since you asked directly:** keep the sentinel. It's correct now, it's the smaller diff, and it matches the scope you narrowed to. My structural argument on #6482 was about which of the three PRs should land, not a claim that this implementation is wrong, and with `_NOT_APPLIED` in place that argument no longer applies to this one. If the maintainers pick this PR, the sentinel is a fine resting place. Rewriting it now to look more like the other two only helps if one of those is landing instead, in which case this diff doesn't matter anyway.

Worth flagging for whoever merges: #6494 attempted this same fix and introduced a different divergence doing it (moved the `if normalized is None` guard out of the loop body, so a returns-`None` callback now crashes async with a `TypeError` where sync completes). Three PRs, and this is the only one I've now verified against all three cases. Details on #6494.

Nice catch on the `mypy --strict` return type too, `InterceptionContext.payload` widening the inferred type to `Any` is the kind of thing that only shows up under strict and then quietly disables checking for every caller downstream.

PR #6506, opened 2026-07-10
fix(agent): report per-call usage metrics on kickoff results

joaomdmoura:
## Summary

Fixes [EPD-177](https://linear.app/crewai/issue/EPD-177/agentkickoff-usage-counters-are-cumulative-and-scoped-to-the-llm): `Agent.kickoff()` populated `result.usage_metrics` from `llm.get_token_usage_summary()`, which returns the LLM **instance's lifetime** accumulator. Counts therefore grew across calls and pooled across agents sharing one `LLM` object — with a shared module-level LLM, a second agent's first turn appeared to "cost" the whole preceding session. Broken cost attribution and dashboards, hit by an enterprise customer in production migration.

## Fix

- New `UsageMetrics.delta_since(baseline)`: field-wise difference between two snapshots of the same accumulator, clamped at zero.
- `Agent` (sync + async kickoff): snapshot the accumulator right after kickoff preparation and thread the baseline through output building and the guardrail-retry path, so the result reports usage for **that call only** — including all guardrail retry attempts.
- `LiteAgent` (deprecated, same bug): snapshot at kickoff start (`_kickoff_usage_baseline`) and report deltas at both output-building sites.
- **Cumulative counters are untouched**: `get_token_usage_summary()` still returns instance-lifetime totals (crew-level aggregation in `crew.py` and the event-based flow aggregation depend on the existing semantics). Its docstring now states the instance-lifetime scope explicitly, and `LiteAgentOutput.usage_metrics`'s field description documents the per-call semantics.

Known limitation (inherent to counter diffing): two kickoffs running *concurrently* against the same LLM instance can still attribute each other's overlapping usage; sequential calls — the reported scenario — are now exact.

## Testing

- The customer's clean-room repro from EPD-177 now prints identical per-call usage (110 tokens) for shared-LLM and own-LLM agents, with the shared instance's lifetime counters still cumulative.
- New regression tests (`TestKickoffUsageMetricsArePerCall`, `TestUsageMetricsDeltaSince`): shared-LLM agents, repeated kickoffs on one agent, async kickoff, LiteAgent, plus delta_since field-wise math and negative-clamp.
- Green: `tests/agents/test_lite_agent.py` (35), `tests/agents/test_agent.py` (97), usage-related tests in `test_crew.py` / `test_llm.py`, `test_flow_usage_metrics.py` + `events/test_llm_usage_event.py` (44), ruff, and mypy.

---

> [!NOTE]
> **Medium Risk**
> Changes how kickoff results report token usage (observable API for dashboards/billing); cumulative LLM counters and crew aggregation paths are intentionally untouched but consumers that assumed cumulative kickoff metrics will see different numbers.
> 
> **Overview**
> **`Agent.kickoff()` / `kickoff_async()`** no longer copy the LLM instance’s **lifetime** token counters into `LiteAgentOutput.usage_metrics`. They snapshot usage at kickoff start and set metrics to the **delta** for that call only (including all **guardrail** re-runs), via new **`UsageMetrics.delta_since()`**.
> 
> **`get_token_usage_summary()`** behavior is unchanged (still cumulative); its docstring and **`LiteAgentOutput.usage_metrics`** now document per-call vs lifetime semantics.
> 
> Regression tests cover shared LLM, repeated kickoffs, async kickoff, guardrail retries, and `delta_since` math.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f6ec673987e1fa282be26425d7c2c0edb693d14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/lite_agent.py
+++ b/lib/crewai/src/crewai/lite_agent.py
@@ -72,7 +72,6 @@ from crewai.llm import LLM
 from crewai.llms.base_llm import BaseLLM
 from crewai.tools.base_tool import BaseTool
 from crewai.tools.structured_tool import CrewStructuredTool
-from crewai.types.usage_metrics import UsageMetrics
 from crewai.utilities.agent_utils import (
     enforce_rpm_limit,
     format_message_for_llm,
@@ -290,7 +289,6 @@ class LiteAgent(FlowTrackable, BaseModel):
     _key: str = PrivateAttr(default_factory=lambda: str(uuid.uuid4()))
     _messages: list[LLMMessage] = PrivateAttr(default_factory=list)
     _iterations: int = PrivateAttr(default=0)
-    _kickoff_usage_baseline: UsageMetrics | None = PrivateAttr(default=None)
     _guardrail: GuardrailCallable | None = PrivateAttr(default=None)
     _guardrail_retry_count: int = PrivateAttr(default=0)
     _callbacks: list[TokenCalcHandler] = PrivateAttr(default_factory=list)
@@ -521,7 +519,6 @@ class LiteAgent(FlowTrackable, BaseModel):
         try:
             self._iterations = 0
             self.tools_results = []
-            self._kickoff_usage_baseline = self._current_usage_summary()
 
             self._messages = self._format_messages(
                 messages, response_format=response_format, input_files=input_files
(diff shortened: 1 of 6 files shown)


lib/crewai/src/crewai/lite_agent.py
lorenzejay, line 1: We dont run this file anymore. any changes on that kickoff for standalone agent kickoffs should just be on core.py which you already did. So just drop the lite_agent.py diff

lorenzejay (changes requested):

joaomdmoura:
Addressed the review:

- **Dropped the `lite_agent.py` diff entirely** per @lorenzejay — the per-call usage snapshot now lives only in `agent/core.py`. That also resolves both CodeRabbit nitpicks on that file (instance-attr baseline, duplicated `_current_usage_summary` helper).
- **Added the guardrail-retry regression test** CodeRabbit suggested: a guardrail rejecting the first attempt and accepting the second yields `usage_metrics` equal to exactly 2× a single-attempt kickoff (requests and tokens), confirming retries roll up into one per-call delta.

Re-verified the EPD-177 clean-room repro against the branch (`Agent` = core path): still fixed — shared-LLM and own-LLM agents report identical per-call usage while the shared instance keeps cumulative lifetime counters.

lorenzejay (approved):

PR #6508, opened 2026-07-10
fix(tools): stop rewriting the authored tool description at construction

joaomdmoura:
## Summary

Fixes [EPD-179](https://linear.app/crewai/issue/EPD-179/basetoolmodel-post-init-silently-overwrites-the-authored-tool): `BaseTool.model_post_init` silently replaced the authored `description` with the LLM-facing composite (`Tool Name: …\nTool Arguments: {…}\nTool Description: <authored>`). Undocumented and surprising: equality assertions on authored text break, and the extra schema tokens are invisible to token-careful authors. The framework itself already treated this as a liability — the native function-calling path *strips the prefix back out* (`agent_utils.py`).

## Fix

The authored `description` now survives construction as written; the composite is composed **on demand** where prompts are built:

- New shared helper `format_description_for_llm(name, args_schema, description)` in `structured_tool.py`, exposed as a `formatted_description` property on both `BaseTool` and `CrewStructuredTool`. Always fresh — it tracks later edits to `name`/`args_schema`/`description`.
- All prompt paths that relied on the baked-in composite now render through it: `render_text_description_and_args` (ReAct system prompt), `ToolUsage._render` (function-calling selector), and tool-usage error messages. **The text the LLM sees is byte-identical to before.**
- The helper is idempotent: it strips any pre-existing composite block before composing, so tools deserialized from checkpoints written by older versions — and adapters that still bake the composite into the field (e.g. the crewai-tools MCP adapter) — don't get double-wrapped.
- `BaseTool._generate_description()` remains as a documented no-op hook: `model_post_init` still calls it because subclasses override it (crewai-tools has ~25 `self._generate_description()` call sites following "mutate description, then refresh" — those calls are now harmless and their authored text is preserved too).

Side effects that are improvements: consumers that passed the raw `description` to places where the composite was redundant (native function-calling schemas, LangGraph/OpenAI-agents adapters, evaluation metrics listing, the knowledge-tools prompt block that used to render `Tool description:\nTool Name: …`) now naturally get the clean authored text.

Follow-up candidates (intentionally not in this PR): remove the vestigial `description_updated` field, and drop the crewai-tools MCP adapter's own `_generate_description` override so MCP tools' `description` field also stays authored.

## Testing

- The ticket's clean-room repro now prints `NOT REPRODUCED: description preserved as authored`, with the identical composite available at `formatted_description`.
- New `TestAuthoredDescriptionPreserved` regression class: authored text preserved, composite content/ordering, on-demand recomposition after edits, no double-wrap for pre-baked composites, and prompt rendering unchanged for both `BaseTool` and `CrewStructuredTool`.
- Updated the two tests that encoded the old rewrite behavior; `test_tool_usage_render` (asserts the composite in the rendered prompt) passes untouched.
- Green: `tests/tools/` (138), `tests/agents/test_agent.py` + `test_lite_agent.py` (126), tool-related `test_crew.py` (11), experimental tool-metrics tests (8), ruff, and mypy.

---

> [!NOTE]
> **Medium Risk**
> Changes how tool descriptions flow through agent prompts and native function-calling schemas; intended LLM text is unchanged but any code that assumed `description` contained the composite will see different values.
> 
> **Overview**
> **Tool `description` stays authored text**; the LLM prompt block (`Tool Name` / `Tool Arguments` / `Tool Description`) is built on demand via new `formatted_description` and `format_description_for_llm()` in `structured_tool.py`. `BaseTool._generate_description()` is now a documented no-op so subclasses/adapters that still override it do not break.
> 
> **Prompt and schema paths** that previously read the baked-in composite now use `formatted_description` (`render_text_description_and_args`, `ToolUsage._render`, tool error messages). **Native OpenAI tool schemas** use `strip_composite_description_prefix()` so checkpoints/adapters with pre-composed descriptions do not duplicate schema in the `description` field.
> 
> **Backward compatibility:** composing is idempotent for legacy three-line composite strings; prose that only mentions `Tool Description:` is not stripped. Tests updated plus `TestAuthoredDescriptionPreserved` for EPD-179.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c48192fcd1754cfb9637b9acfde5101bc8d45f7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/tools/base_tool.py
+++ b/lib/crewai/src/crewai/tools/base_tool.py
@@ -5,7 +5,6 @@ import asyncio
 from collections.abc import Awaitable, Callable
 import importlib
 from inspect import Parameter, signature
-import json
 import threading
 from typing import (
     Any,
@@ -37,9 +36,9 @@ from crewai.tools.structured_tool import (
     _infer_result_schema_from_callable,
     _serialize_schema,
     build_schema_hint,
+    format_description_for_llm,
 )
 from crewai.types.callback import SerializableCallable, _resolve_dotted_path
-from crewai.utilities.pydantic_schema_utils import generate_model_description
 from crewai.utilities.string_utils import sanitize_tool_name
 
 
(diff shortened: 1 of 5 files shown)


lorenzejay (approved):
lgtm

PR #6509, opened 2026-07-10
fix(tools)!: make tool-result caching opt-in instead of on by default

joaomdmoura:
## Summary

Fixes [EPD-180](https://linear.app/crewai/issue/EPD-180/tool-results-cached-by-default-silently-serve-stale-data-for-live-data) — flagged by the reporter as the highest-impact of the batch. Tool-result caching was on by default, so an LLM calling the same tool with identical arguments twice in one run silently received the first result without the tool executing. For live-data tools that's a confidently stale answer; for state-mutating tools the second action is silently dropped — the worst failure mode (silently wrong), and most enterprise data tools are freshness-sensitive.

## ⚠️ Behavior change (deliberate)

The ticket's expected behavior is "caching opt-in per tool — or at minimum a prominent docs warning". This PR implements **opt-in**, on the principle that the default should never produce silently wrong answers; users who want dedup can say so in one keyword. Calling it out for review since it flips a long-standing default:

- **`Crew.cache` now defaults to `False`.** `Crew(cache=True)` restores today's behavior *exactly* — agents still default to participating when a crew offers its handler, `Agent(cache=False)` still opts an agent out, and per-tool `cache_function` write-gating is unchanged.
- **Standalone agents no longer self-wire a `CacheHandler` at construction.** `Agent(cache=True)` or an explicit `cache_handler=` opts in. Notably, this self-wiring meant `Crew(cache=False)` never actually disabled caching before — agents cached through their own handler anyway; that inconsistency is fixed as a side effect.
- All caching machinery (`CacheHandler`, `ToolsHandler`, `cache_function`, cache events with `from_cache`) is untouched — only the defaults changed. Field descriptions and the setup docstring document the opt-in contract and the live-data/stateful-tool hazard.

Docs note: the in-repo docs are versioned release snapshots, so the "Create Custom Tools" page warning the ticket suggests should land with the release notes / current docs when this ships.

## Testing

- The ticket's clean-room repro (fake OpenAI client scripting two identical tool calls) now prints `actual executions: 2 / NOT REPRODUCED` by default, and `1` (dedup) when re-run with `Crew(cache=True)` — opt-in preserves the old behavior byte-for-byte.
- New `tests/test_tool_cache_default.py` (8 tests): offline end-to-end default-reexecutes + opt-in-dedupes (mirroring the repro), and agent-level wiring contracts (no cache by default; `cache=True` / `cache_handler` opt in; explicit `cache=False` refuses a crew-offered handler; default agents accept one).
- The two existing tests that exercised caching machinery via the implicit default (`test_cache_hitting_between_agents`, `test_tools_with_custom_caching`) now opt in explicitly — machinery behavior unchanged.
- Green: full `test_crew.py` (131), `test_agent.py` + `tests/tools/` (230), `test_lite_agent.py` + `test_task.py` (96), ruff, and mypy.

---

> [!NOTE]
> **High Risk**
> Flips a long-standing default that changes whether repeated identical tool calls execute or return stale cached results—a deliberate breaking change with broad runtime impact for crews and standalone agents.
> 
> **Overview**
> **Breaking behavior change:** tool-result deduplication is off unless you opt in, so identical tool calls run the tool again instead of returning a cached first result.
> 
> `Crew.cache` now defaults to **`False`**; **`Crew(cache=True)`** wires the shared `CacheHandler` to crew agents (unchanged machinery). Standalone **`Agent`** instances no longer create a `CacheHandler` at construction unless **`cache=True`** or an explicit **`cache_handler`** is set—fixing the case where **`Crew(cache=False)`** still cached via per-agent handlers.
> 
> **`Agent.cache`** remains a participation flag when a crew enables caching; **`cache=False`** still refuses a crew-offered handler. **`copy()`** now distinguishes construction-time opt-in from kickoff-time crew wiring via **`_constructor_cache_opt_in`** and **`model_fields_set`**, so defaults and crew-wired agents do not accidentally become standalone cachers. Hierarchical **manager** agents created in **`_create_manager_agent`** also receive the crew cache handler when **`cache=True`**.
> 
> Docs on **`cache`** fields and setup docstrings describe the opt-in contract and live-data/stateful-tool risk. Existing cache tests pass **`cache=True`** explicitly; new **`test_tool_cache_default.py`** covers EPD-180-style re-execution vs dedup and wiring/copy/manager edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 818942c23245ebc67c500a8f86108a12197c648c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -410,6 +410,13 @@ class Agent(BaseAgent):
         repeated tool calls with identical arguments always re-execute the
         tool — the safe default for live-data and state-mutating tools.
         """
+        # Recorded before any crew can offer its shared handler at kickoff,
+        # so copy() can distinguish a construction-time opt-in from runtime
+        # crew wiring (which must not turn copies into cachers).
+        self._constructor_cache_opt_in = bool(
+            self.cache
+            and (self.cache_handler is not None or "cache" in self.model_fields_set)
+        )
         opted_in = self.cache_handler is not None or (
             "cache" in self.model_fields_set and self.cache
         )
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -400,10 +400,22 @@ class Agent(BaseAgent):
         return self.planning_config is not None or self.planning
 
     def _setup_agent_executor(self) -> None:
-        """Initialize the agent executor with a default cache handler."""
-        if not self.cache_handler:
-            self.cache_handler = CacheHandler()
-        self.set_cache_handler(self.cache_handler)
+        """Initialize the agent's tools handler and optional tool cache.
+
+        Tool-result caching is opt-in: a standalone agent gets a cache only
+        when it was constructed with an explicit ``cache=True`` or a
+        ``cache_handler``. Agents inside a crew additionally receive the
+        crew's shared handler when ``Crew(cache=True)``. Without an opt-in,
+        repeated tool calls with identical arguments always re-execute the
+        tool — the safe default for live-data and state-mutating tools.
+        """
+        opted_in = self.cache_handler is not None or (
+            "cache" in self.model_fields_set and self.cache
+        )
+        if opted_in:
+            if not self.cache_handler:
+                self.cache_handler = CacheHandler()
+            self.set_cache_handler(self.cache_handler)
 
     def set_knowledge(self, crew_embedder: EmbedderConfig | None = None) -> None:
         """Initialize knowledge sources with the agent or crew embedder config."""
--- a/lib/crewai/src/crewai/agents/agent_builder/base_agent.py
+++ b/lib/crewai/src/crewai/agents/agent_builder/base_agent.py
@@ -267,7 +267,14 @@ class BaseAgent(BaseModel, ABC, metaclass=AgentMeta):
         description="Configuration for the agent", default=None, exclude=True
     )
     cache: bool = Field(
-        default=True, description="Whether the agent should use a cache for tool usage."
+        default=True,
+        description=(
+            "Whether the agent participates in tool-result caching when a "
+            "cache is enabled. Caching itself is opt-in: it activates only "
+            "when the crew sets cache=True or the agent explicitly opts in "
+            "(cache=True or a cache_handler at construction). Set False to "
+            "exclude this agent even when the crew enables caching."
+        ),
     )
     verbose: bool = Field(
         default=False, description="Verbose mode for the Agent Execution"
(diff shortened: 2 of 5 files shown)


joaomdmoura:
Branch updated on latest main and all review findings addressed:

- **Bugbot: "Copy re-enables tool cache"** — fixed in e7e3af7: `copy()` drops `cache` from the dump when it wasn't explicitly set on the source, so the field default no longer reads as an opt-in on copies. Regression tests: `test_copy_of_default_agent_does_not_opt_in`, `test_copy_of_opted_in_agent_stays_opted_in`.
- **Bugbot: "Copy drops cache_handler opt-in"** — fixed in 830da77: an agent that opted in solely via an explicit `cache_handler` keeps its consent across `copy()` (carried as `cache=True`; the copy wires its own fresh handler, matching pre-change copy semantics where handler instances were never shared). Regression tests: `test_copy_of_handler_opted_in_agent_stays_opted_in`, `test_copy_of_explicit_cache_false_with_handler_stays_off`.
- **CodeRabbit: `BaseAgent.__doc__` / `Crew.__doc__` cache docs** — both class docstrings synced with the opt-in semantics in e7e3af7 (participation vs. activation, `cache=False` exclusion, live-data/state-mutating hazard).

joaomdmoura:
Addressed the new Bugbot finding (**"Hierarchical manager misses crew cache"**) in the latest commit: `_create_manager_agent()` now offers the crew's shared cache handler when the crew opted in with `cache=True` — the manager was created outside the validation-time agents loop and no longer self-wires, so its delegation tool calls never cached. A user-provided manager with `cache=False` stays excluded via the existing `set_cache_handler` gate.

Slight improvement over pre-change behavior: the manager now shares the crew's handler (dedup across manager and workers) instead of maintaining a private one.

Regression tests: `TestHierarchicalManagerCacheWiring` — manager wired when `cache=True`, unwired by default, and explicit `cache=False` manager excluded. Hierarchical/manager crew tests (16) all green.

joaomdmoura:
Addressed the new Bugbot finding (**"Copy enables cache after crew"**): the previous fix inferred consent from any wired `cache_handler`, which couldn't distinguish a construction-time opt-in from the crew's shared handler offered at kickoff. The opt-in signal is now recorded in `_setup_agent_executor` — which runs at construction, before any crew wiring can happen — as `_constructor_cache_opt_in`, and `copy()` consults that flag instead of inspecting `cache_handler` after the fact.

Resulting matrix (all covered by tests in `TestAgentCacheWiring`):
- constructor `cache=True` or `cache_handler=…` → copy stays opted in (fresh handler)
- explicit `cache=False` (with or without handler) → copy stays off
- default agent wired by `Crew(cache=True)` at kickoff → **copy stays off** (new)
- plain default agent → copy stays off

lorenzejay (approved):

PR #6511, opened 2026-07-10
fix(tools): raise ToolUsageError instead of bare raise in _original_tool_calling

Aakash326:
Fixes #6430.

The bare `raise` in the non-dict-arguments branch of `_original_tool_calling`
had no active exception to re-raise, so Python raised
`RuntimeError: No active exception to re-raise` instead of a meaningful
`ToolUsageError`. Replaced it with an explicit `raise ToolUsageError(...)`
matching the sibling `return` statement and the pattern used elsewhere in
this method.

Added regression test covering both `raise_error=True` and `raise_error=False`
paths.


ErenAta16 (commented):

PR #6512, opened 2026-07-10
fix: replace bare raise outside except block with explicit ToolUsageError

gautamkishore:
In `_original_tool_calling`, the bare `raise` on the non-dict branch (line 853) is outside any active exception handler — Python raises `RuntimeError: No active exception to re-raise` instead of the intended `ToolUsageError`.

The other bare `raise` (line 848) is inside the `except Exception:` block and is correct.

Fixes #6430

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -850,7 +850,7 @@ class ToolUsage:
 
         if not isinstance(arguments, dict):
             if raise_error:
-                raise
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):

gautamkishore:
Good point @ErenAta16 — updated to construct the error once and reuse it for both branches, with a comment explaining why the bare raise was wrong. Also added a return-path test alongside the raise-path one. Thanks for the comparison across the dupes!

PR #6515, opened 2026-07-11
fix: native tool calls broken over OpenAI Responses API

theCyberTech:
## Summary

Native tool calling failed for any LLM(api="responses") configuration that used tools. Three bugs caused this failure.

Bug 1: is_tool_call_list() did not recognize the Responses API tool-call format. This function checks the tool-call format for Chat Completions, Anthropic, Bedrock, and Gemini. It did not check the Responses API's flat format: {"id", "name", "arguments"}. As a result, the code treated a real tool call as plain text. The agent returned the raw tool-call JSON as its final answer. It did not call the tool.

Bug 2: extract_tool_call_info() read the wrong key for tool arguments. In the dict branch, the function read only the "input" key. It did not read the "arguments" key. Even after Bug 1 is fixed, this bug causes the tool to run with empty arguments.

Bug 3: _prepare_responses_params() sent the wrong message shape to the Responses API. This function passed Chat-Completions-style tool_calls and tool-role messages directly as Responses API input items. The Responses API does not accept this shape. It requires separate function_call and function_call_output items instead. This mismatch caused a 400 error on the second turn of any multi-turn tool-calling conversation.

## Files changed:

crewai/utilities/agent_utils.py — fixes Bug 1 and Bug 2
crewai/llms/providers/openai/completion.py — fixes Bug 3

## Verification:
I ran a tool-using agent against a real OpenAI-compatible model server (vLLM).
Before the fix: the agent returned the tool-call JSON as its answer. It never called the tool.
After the fix: the agent calls the tool, gets the real result, and completes the task across multiple turns.

## Test plan:

Added regression tests for is_tool_call_list() and extract_tool_call_info() in tests/utilities/test_agent_utils.py. These tests cover the Responses API shape. They fail against the old code and pass against the fix. The existing 77 tests in this suite still pass.
Added regression tests for _convert_message_to_responses_input_items() and _prepare_responses_params() in tests/llms/openai/test_openai.py. These tests cover assistant tool_calls, tool-role results, and a full multi-turn conversation. They fail against the old code and pass against the fix. 101 tests in test_openai.py pass, with no regressions in tests/llms/openai_compatible/.
Ran a manual end-to-end test with a tool-using agent against a real Responses-API-compatible local model server. The tool executes with the correct arguments. The task completes with the correct output.


lib/crewai/src/crewai/llms/providers/openai/completion.py
```diff
@@ -668,6 +668,44 @@ async def _acall_responses(
...
+
+        The Responses API has no message shape for an assistant turn carrying
+        ``tool_calls`` or for a ``tool`` role reply - those become standalone
+        ``function_call`` / ``function_call_output`` input items instead. Plain
+        user/assistant text messages pass through unchanged (accepted as-is by
+        the Responses API's lenient "easy input message" shape).
+        """
+        role = message.get("role")
+
+        if role == "assistant" and message.get("tool_calls"):
```
joaomdmoura, line 767: **Assistant text is dropped when a message carries both `content` and `tool_calls`.**

This branch returns only `function_call` items, discarding any assistant text. The framework's own history writers always set `content: None`, so no in-framework impact — but this is a general converter on the public `LLM.call(messages=...)` path, and Chat-Completions-style histories from external callers legitimately carry assistant text alongside tool calls ("I'll fetch that page…"). That text now silently disappears from the model's context.

Cheap fix — emit the text item first:

```python
if role == "assistant" and message.get("tool_calls"):
    items: list[dict[str, Any] | LLMMessage] = []
    if message.get("content"):
        items.append({"role": "assistant", "content": message["content"]})
    for tool_call in message["tool_calls"]:
        ...
```

Plus a test for the combined shape.

lib/crewai/src/crewai/llms/providers/openai/completion.py
```diff
@@ -668,6 +668,44 @@ async def _acall_responses(
...
+            items: list[dict[str, Any] | LLMMessage] = []
+            for tool_call in message["tool_calls"]:
+                function = tool_call.get("function", {})
+                items.append(
+                    {
+                        "type": "function_call",
+                        "call_id": tool_call.get("id", ""),
+                        "name": function.get("name", ""),
+                        "arguments": function.get("arguments", ""),
```
joaomdmoura: **Harden against dict-valued `arguments` and missing `id` — this PR's own target scenario.**

The Responses API requires `arguments` to be a JSON *string*; some OpenAI-compatible integrations put a **dict** here, which will 400. Mirror the guard the history writer already uses (`agent_utils.py`'s `tool_calls_to_report`):

```python
args = function.get("arguments", "")
items.append(
    {
        "type": "function_call",
        "call_id": tool_call.get("id") or f"call_{id(tool_call)}",
        "name": function.get("name", ""),
        "arguments": args if isinstance(args, str) else json.dumps(args),
    }
)
```

The `call_id` fallback matters too: servers that omit tool-call ids would otherwise produce `call_id: ""`, which the API rejects — `extract_tool_call_info()` already synthesizes ids the same way. Low probability given framework-written history, but the PR exists because of OpenAI-compatible-server drift (vLLM), so belt-and-suspenders is warranted here.

lib/crewai/src/crewai/llms/providers/openai/completion.py
```diff
@@ -683,7 +721,7 @@ def _prepare_responses_params(
         - Internally-tagged tool format (flat structure)
         """
         instructions: str | None = self.instructions
-        input_messages: list[LLMMessage] = []
+        input_messages: list[Any] = []
```
joaomdmoura: Nit: `list[dict[str, Any] | LLMMessage]` (the helper's return type) would keep some type safety instead of dropping to `list[Any]`.

lib/crewai/src/crewai/utilities/agent_utils.py
```diff
@@ -1270,6 +1277,13 @@ def is_tool_call_list(response: list[Any]) -> bool:
     # Bedrock-style
     if isinstance(first_item, dict) and "name" in first_item and "input" in first_item:
         return True
+    # OpenAI Responses API-style (flat dict, no nested "function" key)
+    if (
+        isinstance(first_item, dict)
+        and "name" in first_item
+        and "arguments" in first_item
+    ):
+        return True
```
joaomdmoura, line 1415: Worth a short code comment acknowledging the tradeoff: any list-of-dicts with `name`+`arguments` now classifies as a tool-call list, including hypothetical structured answers shaped that way. Same risk class as the Bedrock `name`+`input` check above, and only provider paths returning Python lists are exposed — acceptable, but future readers should know it was a deliberate choice. (A `"type": "function_call"` discriminator emitted by `_extract_function_calls_from_response()` would make this airtight as a follow-up.)

joaomdmoura (dismissed):
Thanks — the diagnosis is right and I verified the core fix against the framework's own message flow before reviewing:

- The flat tool-call dicts come from `_extract_function_calls_from_response()`, which maps the SDK's `item.call_id` into `"id"` — so `extract_tool_call_info()` picks up the **correct correlation id**, and the converter's `function_call.call_id` ↔ `function_call_output.call_id` pairing round-trips correctly.
- Both history writers (`agent_executor.py` / `agent_utils.py`) always emit CC-shaped `tool_calls` with `content: None` and **string** arguments, so the converter's assumptions match everything the framework itself produces.
- `_prepare_responses_params()` is the single choke point for sync and async, so streaming is covered too.
- The new `is_tool_call_list` / `extract_tool_call_info` branches are additive and can't change CC/Bedrock/Anthropic/Gemini outcomes.

Test quality is good: exact-shape assertions, negative-space checks (no leftover CC-only keys), and still-recognized cases for the other provider shapes.

Requesting changes for two small hardening items (inline) — both matter precisely because this PR's target is OpenAI-*compatible* servers, where shape drift is the norm. One nit inline as well.

On the broadened `is_tool_call_list` classification (any dict with `name`+`arguments`): I'd accept it as-is — it's the same risk class as the pre-existing Bedrock `name`+`input` check, and only provider paths that return Python lists are exposed. Tightening isn't possible without changing the producer; if we want it airtight later, add `"type": "function_call"` to `_extract_function_calls_from_response()`'s dicts and check it — follow-up, not a blocker.

Missing test coverage matching the two inline items: assistant message with both `content` and `tool_calls`, and dict-valued `arguments` reaching the converter.

theCyberTech:
@copilot resolve the merge conflicts in this pull request

Vidit-Ostwal:
@coderabbitai full review.

lib/crewai/src/crewai/llms/providers/openai/completion.py
```diff
@@ -765,18 +765,23 @@ def _to_responses_input(message: LLMMessage) -> list[Any]:
...
+                args = function.get("arguments")
+                if args is None or args == "":
+                    args = "{}"
+                elif not isinstance(args, str):
+                    args = json.dumps(args)
                 items.append(
                     {
                         "type": "function_call",
-                        "call_id": call.get("id", ""),
+                        "call_id": call.get("id") or f"call_{id(call)}",
```
Vidit-Ostwal: Instead of this `"call_id": call.get("id") or f"call_{id(call)}",`, I would keep it same 

`"call_id": call.get("id", ""),`
Teh reason being it's better to know that we are not able to populate the id, rather than creating a different object, altogether.

Vidit-Ostwal:
Rest looks good to me, LGTM
Accidentaly closed it, my bad

Vidit-Ostwal (approved):

lucasgomide (approved):

PR #6519, opened 2026-07-11
fix(security): eliminate DNS rebinding TOCTOU and add SSRF validation to MCP tools

ashusnapx:
## Summary

Fixes two SSRF security vulnerabilities in crewai-tools.

**Closes #6504**

---

### Vulnerability 1: DNS Rebinding (TOCTOU) in SSRF protection

`validate_url()` resolved DNS and checked the IP, then returned the URL as a string. The subsequent `requests.get()` resolved DNS **again**, creating a TOCTOU window where the DNS record could change between validation and connection.

**Attack scenario**: Attacker registers a domain that first resolves to a public IP (passes validation), then quickly changes to `169.254.169.254` (cloud metadata) before the connection is made.

**Fix**:
- Added `validate_url_and_resolve()` → returns `ValidatedURL(url, resolved_ip)`
- Added `PinnedIPAdapter` — a `requests.HTTPAdapter` that pins TCP connections to the pre-resolved IP
- Updated `safe_get()` to use the pinned adapter for every request (including redirect targets)
- Refactored `validate_url()` to share validation logic via `_validate_url_common()`

```
Before:  validate_url(url) → requests.get(url)          # DNS resolved twice
After:   validate_url_and_resolve(url) → session.get(url) # DNS resolved once, IP pinned
```

**PinnedIPAdapter implementation**:
- Overrides `send()` to rewrite the URL, replacing hostname with the pinned IP
- Sets `Host` header to the original hostname for virtual hosting
- Overrides `init_poolmanager()` to configure `assert_hostname=False` for correct TLS SNI when the URL contains an IP address
- No `server_hostname` kwarg injected into `send()` — keeps kwargs compatible with `HTTPAdapter.send()` signature

### Vulnerability 2: MCP tools bypass SSRF validation entirely

MCP tool wrappers (`mcp_tool_wrapper.py`, `mcp_native_tool.py`) passed arguments directly to MCP servers without any URL validation. An agent using an MCP fetch server could be instructed to access internal URLs, cloud metadata, etc.

**Fix**:
- Added `validate_mcp_server_url()` — validates the MCP server connection URL
- Added `validate_mcp_tool_args_for_urls()` — recursively scans string arguments for URLs (`http`, `https`, `file`) and validates each one against the SSRF guard
- Both MCP wrappers now validate server URL and tool arguments before execution
- Extracted shared helpers into `mcp_security.py` to eliminate duplication
- Added `_clean_url()` to strip trailing prose delimiters (`)`, `,`, `.`) from regex-matched URLs before validation

```python
# Before: MCP tool arguments passed directly to server
result = await session.call_tool(tool_name, kwargs)

# After: URLs in arguments are validated first
validate_mcp_server_url(server_url)
validate_mcp_tool_args_for_urls(kwargs)
result = await session.call_tool(tool_name, kwargs)
```

### Files Changed

| File | Change |
|---|---|
| `security/safe_path.py` | Added `ValidatedURL`, `validate_url_and_resolve()`, `PinnedIPAdapter` (URL-level IP pinning + pool manager SNI), refactored `validate_url()` |
| `security/safe_requests.py` | Updated `safe_get()` to use pinned sessions |
| `tools/mcp_security.py` | **New** — shared SSRF validation helpers with `_clean_url()` for prose-safe URL extraction |
| `tools/mcp_tool_wrapper.py` | Added server URL + argument SSRF validation (imports from shared module) |
| `tools/mcp_native_tool.py` | Added argument SSRF validation (imports from shared module), fixed `_run_async` indentation |
| `tests/mcp/test_mcp_ssrf.py` | 17 tests — shared validation + native MCP path coverage |
| `tests/utilities/test_safe_requests.py` | Updated for pinned session + behavioral adapter tests (HTTP + HTTPS) |

### Tests

```
56 passed, 0 failed
- 30 safe_path tests (existing, all pass)
- 9 safe_requests tests (updated for pinned session + behavioral adapter tests)
- 17 MCP SSRF tests (shared validation + native MCP path coverage)
```

### Verification

```bash
# Security tests
pytest lib/crewai-tools/tests/utilities/test_safe_path.py -v
pytest lib/crewai-tools/tests/utilities/test_safe_requests.py -v
pytest lib/crewai/tests/mcp/test_mcp_ssrf.py -v

# Lint + type check
ruff check lib/crewai-tools/src/crewai_tools/security/ lib/crewai/src/crewai/tools/mcp_security.py lib/crewai/src/crewai/tools/mcp_tool_wrapper.py lib/crewai/src/crewai/tools/mcp_native_tool.py
mypy lib/crewai/src/crewai/tools/mcp_security.py lib/crewai/src/crewai/tools/mcp_native_tool.py lib/crewai/src/crewai/tools/mcp_tool_wrapper.py lib/crewai-tools/src/crewai_tools/security/safe_path.py
```

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/security/safe_path.py
+++ b/lib/crewai-tools/src/crewai_tools/security/safe_path.py
@@ -275,13 +275,19 @@ class PinnedIPAdapter(HTTPAdapter):
 
     Implementation: overrides ``send()`` to rewrite the request URL,
     replacing the hostname with the pinned IP and setting the ``Host``
-    header to the original hostname. For HTTPS, ``server_hostname`` is
-    set on the SSL context so TLS SNI works correctly.
+    header to the original hostname.  ``init_poolmanager`` is overridden
+    to configure ``assert_hostname=False`` so that TLS SNI verification
+    works correctly when the URL contains an IP address.
     """
 
     def __init__(self, resolved_ip: str, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self._resolved_ip = resolved_ip
+
+    def init_poolmanager(self, *args: Any, **kwargs: Any) -> None:
+        """Configure pool manager with assert_hostname=False for IP-pinned HTTPS."""
+        super().init_poolmanager(*args, **kwargs)  # type: ignore[no-untyped-call]
+        self.poolmanager.assert_hostname = False
 
     def send(self, request: Any, **kwargs: Any) -> Any:  # type: ignore[override]
         """Send the request, rewriting the URL to use the pinned IP."""
@@ -302,19 +308,11 @@ class PinnedIPAdapter(HTTPAdapter):
                 auth += f":{parsed.password}"
             pinned_netloc = f"{auth}@{pinned_netloc}"
 
-        pinned_url = urlunparse(
-            parsed._replace(netloc=pinned_netloc)
-        )
+        pinned_url = urlunparse(parsed._replace(netloc=pinned_netloc))
         request.url = pinned_url
 
         # Set Host header to original hostname for virtual hosting
-        if request.headers.get("Host") is None:
-            request.headers["Host"] = original_host
         if "host" not in {k.lower() for k in request.headers}:
             request.headers["Host"] = original_host
-
-        # For HTTPS, ensure TLS SNI uses the original hostname
-        if parsed.scheme == "https":
-            kwargs.setdefault("server_hostname", original_host)
 
         return super().send(request, **kwargs)
--- a/lib/crewai-tools/src/crewai_tools/security/safe_path.py
+++ b/lib/crewai-tools/src/crewai_tools/security/safe_path.py
@@ -19,7 +19,6 @@ from typing import Any
 from urllib.parse import urlparse
 
 from requests.adapters import HTTPAdapter
-import urllib3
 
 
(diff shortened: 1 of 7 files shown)


ErenAta16:
Checked `PinnedIPAdapter.init_poolmanager` specifically, since disabling hostname checks to fix DNS rebinding is a real trade to verify carefully. `self.poolmanager.assert_hostname = False` disables urllib3's hostname-matching step (does the presented certificate's CN/SAN match the host you're connecting to), it doesn't disable chain-of-trust verification, `cert_reqs` isn't touched here, so a self-signed cert still won't pass. But hostname matching being off means any certificate issued by a trusted CA, for any domain, is accepted for the connection. An attacker positioned to intercept traffic to the pinned IP (which, notably, is exactly the DNS-rebinding-adjacent threat model this PR is defending against) only needs a CA-signed cert for a domain they control, trivial to get, not one matching the target hostname.

The standard fix for "connect to a pinned IP over HTTPS while still verifying the cert against the original hostname" is to keep hostname verification on and explicitly pass the original hostname as `server_hostname` for the TLS handshake (SNI and cert matching both use it), rather than turning verification off. That's what libraries built specifically for this (e.g. `forcediphttpsadapter`) do. Worth fixing before merge, this adapter is now used for every HTTPS request through `safe_get`, not just ones that hit the DNS-rebinding case specifically.

Issue #5460, opened 2026-04-15
[BUG] when using Task(..., output_pydantic=MyModel) some JSON substring identification within the data occurs with  possible false positive


kalfa:
### Description

Specifically the input is a graphql schema, so it has curly braces like JSON but it is not json.

in crewai.utilities.converter:
`_JSON_PATTERN: Final[re.Pattern[str]] = re.compile(r"({.*})", re.DOTALL)
`

tries to identify json in a larger string by just looking for curly braces. this makes any look-alike JSON like GraphQL schemas to be matches as false-positives. making the conversion fail

### Steps to Reproduce

1. have an agent that returns any lookalike-JSON string like `pre {boo} post'
2. have a task that uses output_pydantic=MyModel so that
3. look the task fail

### Expected behavior

the aim is to narrow down as much as possible the false-positives, but without doing a slow `json.load`. the most effective way to understand if the identifies substring is json, is to try to load the substring, to determine it.

if json.load is acceptable, it will deterministically and accurately tell if it's JSON.

the ultimate aim is to allow a generic look-alike-JSON string to be treated as string, and not serialised, while JSON to be serialised.

### Screenshots/Code snippets

```python
# [... import ...]

class GraphQLSchema(BaseModel):
    gql_schema: str = Field(description="The schema of the graphql server")


[...crew class modified from template]
    @task
    def determine_schema(self) -> Task:
        return Task(
            config=self.tasks_config["determine_schema"],  # type: ignore[index]
            output_pydantic=GraphQLSchema,
        )
```

a simple task config as
```YAML
determine_schema:
  description: >
    Query the graphql server to get the schema.
    Use the graphql mcp tool to get the schema.
    Return any error from the server, if the query results in error.
    No other outcome is expected and allowed.
  markdown: true
  expected_output: >
    The schema of the GraphQL server, as a string.

  agent: query_manager
```

has been used, with a graphql mcp server. but it can be faked by asking to return exactly the schema, if required.

### Operating System

macOS Catalina

### Python Version

3.12

### crewAI Version

1.9.3

### crewAI Tools Version

1.9.3

### Virtual Environment

Venv

### Evidence

The schema in input is a generic example of graphql schema, any will work:
```grapgql
type Query {
  countries: [Country]
  country(code: String!): Country
}

type Country {
  code: String
  name: String
  native: String
  phone: String
  continent: Continent
  currency: String
  languages: [Language]
  emoji: String
}

type Continent {
  code: String
  name: String
[...]

PR #6521, opened 2026-07-12
fix(converter): don't let trailing braces corrupt valid JSON extraction

ErenAta16:
`_JSON_PATTERN = re.compile(r"({.*})", re.DOTALL)` in `converter.py` greedily matches from the first `{` to the *last* `}` anywhere in the string. If the LLM's output contains a complete, valid JSON object followed by commentary that itself references curly braces - mentioning the schema, a code snippet, a set literal, anything - the match swallows that trailing text too, and `json.loads` fails on the resulting garbage even though the actual JSON object was perfectly well formed:

```python
result = '''Here is the extracted data:
{"name": "Alice", "age": 30}

Note: this response follows the requested schema (see the format described as {"type": "object"} in the task).'''

match = _JSON_PATTERN.search(result)
json.loads(match.group())
# json.decoder.JSONDecodeError: Extra data: line 3 column 1 (char 30)
```

In `handle_partial_json`, that `JSONDecodeError` triggers an immediate fall through to `convert_with_instructions` (a re-prompt), discarding output that was already correct.

This is a different failure mode from #5460 (closed via #5461), which was about the regex false-*positively* matching brace-delimited text that isn't JSON at all (e.g. GraphQL schemas). This one is a false-*negative*: genuinely valid JSON gets corrupted because of what comes after it.

## Fix

Replaced the regex-based extraction with `json.JSONDecoder(strict=False).raw_decode(result, start)`, starting at the first `{`. `raw_decode` parses exactly one JSON value and stops there, so trailing braces elsewhere in the string can't corrupt an otherwise-valid match. The GraphQL fall-through from #5461 is preserved, since `raw_decode` still raises `JSONDecodeError` on non-JSON brace-delimited content for the same reason `json.loads` did.

Applied to both `handle_partial_json` and `async_handle_partial_json`, which shared the identical pattern. Also drops the now-unused `re` import and the `_JSON_PATTERN` constant.

## Testing

Added `test_handle_partial_json_ignores_braces_in_trailing_commentary` to `tests/utilities/test_converter.py`, covering the exact repro above.

Ran the full `tests/utilities/test_converter.py` suite: 51/52 passing. The one failure (`test_supports_function_calling_false`) is pre-existing and unrelated - it needs the optional `litellm` package, which isn't installed in my sandbox, and I confirmed it fails identically on unpatched `main` before making any changes.

I did try a second test with brace-containing text on *both* sides of the target object, expecting it to also resolve correctly - it doesn't, because `raw_decode` starting at the first `{` will happily parse an earlier, self-contained JSON-looking object if one exists before the real one. That's a real limitation, but a different (arguably harder - which object is "the" answer?) problem than what this PR fixes, so I removed that test rather than either overclaiming or scope-creeping the fix. Happy to discuss if that's worth a follow-up.


ErenAta16:
Fair point - added test_async_handle_partial_json_ignores_braces_in_trailing_commentary alongside the sync one, since both paths share _extract_first_json_object. Full suite still at 52/53 (same pre-existing unrelated litellm-dependency failure as before).

ErenAta16:
Rebased onto current main to resolve the conflict, no code changes beyond that.

PR #6540, opened 2026-07-14
[EPD-196] Add strict mode + machine-readable outcome for restore_from_state_id misses

joaomdmoura:
- `kickoff(restore_from_state_id=...)` with an unknown state id used to log a warning and silently run a brand-new session — invisible to callers behind a remote API.
- Adds an opt-in `raise_on_missing_state: bool = False` flag to `kickoff`, `kickoff_async`, `akickoff`, `stream_events`, and `astream`: when True, a restore miss (or missing persistence backend) raises `ValueError: No flow state found for restore_from_state_id: '<id>'` instead of silently dropping context.
- Adds `flow.last_restore_succeeded` (`None` = no restore requested, `True` = hydrated, `False` = miss) so wrappers like the AMP deployment API can detect silent misses programmatically.
- Default behavior is unchanged (misses still fall back silently); covered by new sync + async tests in `test_flow_persistence.py` and a docs note in the flow-state guide.
- Linear: https://linear.app/crewai/issue/EPD-196

---

> [!NOTE]
> **Medium Risk**
> Touches persisted-state hydration at kickoff (including streaming), which can change failure modes for remote APIs when strict mode is enabled; defaults preserve prior silent-fallback behavior.
> 
> **Overview**
> Adds **opt-in strict handling** when forking persisted flow state via `restore_from_state_id`: new `raise_on_missing_state=False` on `kickoff`, `kickoff_async`, `akickoff`, `stream_events`, and `astream` raises `ValueError` on a missing snapshot or missing persistence instead of silently starting fresh.
> 
> Introduces **`flow.last_restore_succeeded`** (`None` / `True` / `False`) so API wrappers can detect silent misses without log scraping. Streaming entry points reset the flag at session creation, run an eager pre-check when strict mode is on, and document when to read the signal after consuming frames.
> 
> Refactors restore conflict checks into shared helpers; default silent-miss behavior is unchanged. Docs note the new flag and property; persistence tests cover sync/async/streaming paths. A small OpenAI import test fix keeps `sys.modules` and package `completion` in sync after re-import.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6e816173625aa48da0fc18241f72ee99d7c90e32. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/edge/en/guides/flows/mastering-flow-state.mdx
+++ b/docs/edge/en/guides/flows/mastering-flow-state.mdx
@@ -385,7 +385,7 @@ flow_2.kickoff(restore_from_state_id=flow_1.state.id)
 
 Behavior notes:
 
-- `restore_from_state_id` not found in persistence → the kickoff falls back silently to default behavior (mirrors the existing `inputs["id"]` resume not-found behavior). No exception is raised.
+- `restore_from_state_id` not found in persistence → the kickoff falls back silently to default behavior (mirrors the existing `inputs["id"]` resume not-found behavior). No exception is raised by default. Pass `raise_on_missing_state=True` to get a `ValueError` instead of a silent fresh run, or check `flow.last_restore_succeeded` after kickoff (`None` = no restore requested, `True` = hydrated, `False` = miss).
 - Combining `restore_from_state_id` with `from_checkpoint` raises a `ValueError` — they target different state systems (`@persist` vs. Checkpointing) and cannot be combined.
 - `restore_from_state_id=None` (default) is byte-identical to a kickoff without the parameter.
 - Pinning `inputs["id"]` while forking means the new run shares a persistence key with another flow — usually you want only `restore_from_state_id`.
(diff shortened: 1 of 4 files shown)


joaomdmoura:
The CI failures in `tests/llms/azure/test_azure_responses.py` were a **pre-existing test-isolation leak**, not caused by this PR's flow-runtime change. Fixed in ad0523b92 as a separate commit.

**Mechanism** (reproduced on pristine `origin/main`, Python 3.10):
- `test_openai_completion_module_is_imported` (tests/llms/openai/test_openai.py) removes `crewai.llms.providers.openai.completion` from `sys.modules` via `monkeypatch.delitem` and re-imports it. The re-import also rebinds the parent package attribute `crewai.llms.providers.openai.completion` to the **new** module object; monkeypatch teardown restores only the `sys.modules` entry, leaving the two bindings pointing at different module objects for the rest of the process.
- On **Python 3.10**, `from crewai.llms.providers.openai.completion import OpenAICompletion` (the dynamic import inside `AzureCompletion._init_responses_delegate`) resolves through the stale package attribute, while the azure tests' `mock.patch("crewai.llms.providers.openai.completion.OpenAICompletion", ...)` patches the restored `sys.modules` entry — so the patch silently misses, a real delegate is constructed, and `last_response_id`/`last_reasoning_items` are `None` and `reset_chain` is never called. Python 3.11+ resolves via `sys.modules`, which is why only the 3.10 shard failed.
- The trigger is ordering luck: `pytest-randomly` picks a fresh seed per run (this run: 756721157), and that seed put the openai module-import test before the azure responses tests on the same xdist worker (`gw0`, group 6). Any PR could have hit it.

**Fix**: record the parent package attribute with `monkeypatch.setattr` alongside the `delitem`, so teardown restores both bindings consistently. Verified by replaying the failing CI worker's exact 181-test order under Python 3.10: 3 failed before, all pass after.

PR #6547, opened 2026-07-14
Fix: Support async before/after_kickoff_callbacks in akickoff (#6481)

### LINKED ISSUE #6481: [BUG] before/after_kickoff_callbacks do not support async callables in akickoff

mukktinaadh:
Closes #6481. Added `aprepare_kickoff` to properly await async `before_kickoff_callbacks`, and modified `akickoff` to await async `after_kickoff_callbacks` when using native async execution.

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -1243,7 +1243,8 @@ class Crew(FlowTrackable, BaseModel):
 
         runtime_scope = crewai_event_bus._enter_runtime_scope()
         try:
-            inputs = prepare_kickoff(self, inputs, input_files)
+            from crewai.crews.utils import aprepare_kickoff
+            inputs = await aprepare_kickoff(self, inputs, input_files)
 
             if self.process == Process.sequential:
                 result = await self._arun_sequential_process()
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Checked this against the issue's own proposed fix pattern (and against how `task_callback` already handles this in the sync-task async path). One gap: both here and in `aprepare_kickoff`, the async-detection uses `inspect.iscoroutinefunction(callback)` checked *before* calling it, rather than `inspect.isawaitable(result)` checked *after* calling it (which is what the issue itself proposed, and what `task_callback`'s existing code already does).

`iscoroutinefunction` only recognizes plain `async def` functions/methods. It returns `False` for a callable class instance with an async `__call__`, or for other general callables that return a coroutine without themselves being coroutine functions. `SerializableCallable` (the type alias `before/after_kickoff_callbacks` are typed as) is `Callable[..., Any]`, so callable-class instances are a supported, type-checked use case here, not a hypothetical. For one of those, this fix's `iscoroutinefunction` check is `False`, so the callback is invoked synchronously, `result` becomes an unawaited coroutine, and it silently reproduces the exact original bug for that callback style.

Checking `inspect.isawaitable(result)` after the call instead (call first, then decide whether to await) would cover both plain async functions and callable-class instances, since it inspects the actual return value rather than trying to classify the callable in advance. Worth fixing before merge given the type hint explicitly allows the case this would miss.

PR #6550, opened 2026-07-15
Add RelayShield security tools (MCP registry risk + prompt-injection breach detection)

nzdsf2-gif:
Adds two tools for agent-specific security checks: mcp-registry-risk (typosquat/reputation scoring for MCP servers and tool registries) and prompt-injection-breach (credential breach exposure sourced from prompt-injection attacks against agents). Both call RelayShield's live API — no API key required via the x402 PAYG route (pay-per-call in USDC, no signup), or an API key for the metered route. Docs/usage example included in the tool's README following the existing pattern.

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/__init__.py
@@ -20,6 +20,10 @@ from crewai_tools.tools.brave_search_tool.brave_local_pois_tool import (
 )
 from crewai_tools.tools.brave_search_tool.brave_news_tool import BraveNewsSearchTool
 from crewai_tools.tools.brave_search_tool.brave_search_tool import BraveSearchTool
+from crewai_tools.tools.relayshield_tool.relayshield_tool import (
+    RelayShieldMCPRiskTool,
+    RelayShieldPromptInjectionBreachTool,
+)
 from crewai_tools.tools.brave_search_tool.brave_video_tool import BraveVideoSearchTool
 from crewai_tools.tools.brave_search_tool.brave_web_tool import BraveWebSearchTool
 from crewai_tools.tools.brightdata_tool.brightdata_dataset import (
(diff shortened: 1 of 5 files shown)


nzdsf2-gif:
Thanks for the review — addressed in fd680300:

- Removed `ClassVar` from the module-level `API_BASE_URL` constant (it's only valid inside a class body).
- Moved `resp.json()` parsing inside the `try` block for both tools, so an invalid-JSON 2XX response is caught as a `RequestException` instead of crashing.
- Fixed the real bug underneath the API-key finding: `_relayshield_headers()` was raising a bare `ValueError` on a missing key from inside the `try` block, which the `except requests.RequestException` clause never caught — an uncaught exception, not a graceful error. Now checked up front in `_run()`, returning a clean error string instead.

Not implementing the "omit the header, allow keyless" suggestion as proposed: these tools only call the metered endpoint (`/v1/metered/...`), which has no x402 payment-signing logic behind it — omitting the header would just produce a 401, not actually enable the PAYG route. Real x402 client-side signing inside a CrewAI tool is a larger scope change than this PR; happy to take that on as a follow-up if there's interest.

Smoke-tested all three cases (successful call, breach check, missing-key error path) against the live API before pushing.

nzdsf2-gif:
Addressed in dd3ac7b5:

- Non-dict JSON response guard: fixed for both tools — checks `isinstance(response_json, dict)` before calling `.get()`, so a valid-but-unexpected JSON shape (list/string) returns a clean fallback instead of crashing.
- The "missing `os` import" finding is a false positive — `import os` is present at the top of the file in both the base commit and the current one (unchanged by either of my commits). Looks like a scoped-diff artifact rather than an actual regression. Skipping, no change needed.

Re-tested both the real API call and a mocked non-dict-response case before pushing — both behave correctly now.

nzdsf2-gif:
Hi team — just checking in on this one. All of CodeRabbit's findings from both review passes have been addressed (commits `fd680300` and `dd3ac7b5`), and I explained the one suggestion I didn't implement (keyless access) in the thread above. Let me know if there's anything else needed to move this forward, or if it'd help to get a second pair of eyes.

PR #6562, opened 2026-07-16
fix(tools): raise ToolUsageError instead of bare raise for non-dict args

Kyou12138:
## Summary

Fixes a latent bug in `ToolUsage._original_tool_calling()` where a bare `raise` is used **outside** any exception handler when tool arguments are not a `dict`.

On that path Python raises `RuntimeError: No active exception to re-raise` instead of a meaningful `ToolUsageError`. Today `_validate_tool_input` usually returns a dict or raises, so this is a landmine — but `_tool_calling()` calls this with `raise_error=True` and relies on catching a meaningful exception to trigger its LLM fallback.

Closes #6430

## Changes

- Replace bare `raise` with explicit `raise ToolUsageError(...)` when `raise_error=True`
- Keep the existing `return ToolUsageError(...)` behavior when `raise_error=False` (same message, shared instance path)
- Also clears ruff `PLE0704` (bare raise not inside exception handler)

## Test plan

- [x] Unit tests for `_original_tool_calling(raise_error=True)` → `ToolUsageError` (not `RuntimeError`)
- [x] Unit tests for `_original_tool_calling(raise_error=False)` → returns `ToolUsageError`
- [x] Unit test that `_tool_calling` fallback path survives non-dict args
- [ ] CI: `pytest lib/crewai/tests/tools/test_tool_usage.py -q`

## Notes

This contribution was authored with AI assistance. Per CONTRIBUTING.md, please apply the `llm-generated` label (I cannot set labels on this repository).

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -849,9 +849,13 @@ class ToolUsage:
             return ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
 
         if not isinstance(arguments, dict):
+            # Bare `raise` is invalid here: there is no active exception context
+            # (the preceding try/except completed normally). Re-raise a meaningful
+            # ToolUsageError so `_tool_calling` can fall back to the LLM path.
+            error = ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
             if raise_error:
-                raise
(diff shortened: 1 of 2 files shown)


Kyou12138:
Closing this in favor of #6486. It predates this PR and implements the same non-dict ToolUsageError fix with regression coverage. Keeping another equivalent implementation open would add review overhead.

PR #6564, opened 2026-07-16
fix(reasoning): detect bare READY markers in planning responses

### LINKED ISSUE #6204: [BUG] Reasoning plan always detects "NOT READY" even though the model indicates "READY"

Kyou12138:
## Summary

When agent reasoning/planning is enabled, readiness was detected only via an exact substring match on:

```text
READY: I am ready to execute the task.
```

Many models (notably Ollama / local models) correctly conclude with a short-form marker such as:

```text
<plan contents>

READY
```

That was treated as **not ready**, causing endless refine loops ("You indicated you weren't ready...").

Closes #6204

## Changes

- Add `response_indicates_ready()` in `reasoning_handler.py`
- Accepts:
  - Full instructional phrase (backward compatible)
  - Line-anchored `READY` / `READY.` / `READY: ...` (case-insensitive)
  - Line-anchored `NOT READY` (does **not** false-positive on the substring inside `NOT READY`)
- When multiple markers appear (refine rounds), the **last** explicit marker wins
- Wire the helper into:
  - function-calling text fallback
  - LLM-call text fallback
  - `_parse_planning_response`

Mid-sentence phrases like "I am ready to begin researching" are intentionally **not** treated as markers (line-anchored).

## Test plan

- [x] Full instructional phrase → ready
- [x] Bare `READY` / `READY.` line → ready (the #6204 failure mode)
- [x] `NOT READY` → not ready (no false positive from substring)
- [x] Mid-sentence "I am ready to..." → not ready
- [x] Last marker wins across refine rounds
- [x] `_parse_planning_response` integration
- [ ] CI: `pytest lib/crewai/tests/utilities/test_response_indicates_ready.py -q`

## Notes

This contribution was authored with AI assistance. Per CONTRIBUTING.md, please apply the `llm-generated` label.

### DIFF
--- a/lib/crewai/src/crewai/utilities/reasoning_handler.py
+++ b/lib/crewai/src/crewai/utilities/reasoning_handler.py
@@ -57,9 +57,6 @@ def response_indicates_ready(response: str) -> bool:
     if not response:
         return False
 
-    if _READY_FULL_PHRASE in response:
-        return True
-
     last_decision: bool | None = None
     for line in response.splitlines():
         match = _READY_LINE_RE.match(line)
@@ -68,7 +65,10 @@ def response_indicates_ready(response: str) -> bool:
         # Group 1 is the optional "NOT " prefix.
         last_decision = match.group(1) is None
 
-    return bool(last_decision)
+    if last_decision is not None:
+        return last_decision
+
+    return _READY_FULL_PHRASE in response
 
 
 class ReasoningPlan(BaseModel):
--- a/lib/crewai/src/crewai/utilities/reasoning_handler.py
+++ b/lib/crewai/src/crewai/utilities/reasoning_handler.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import logging
+import re
 from typing import TYPE_CHECKING, Any, Final, Literal, cast
 
 from pydantic import BaseModel, Field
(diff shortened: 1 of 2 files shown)


ErenAta16:
The line-by-line marker logic is solid, verified the regex (`^\s*(NOT\s+)?READY\b` with `\.match()` per line) correctly excludes mid-sentence hits like "I am ready to begin" (no leading marker position) while catching `READY`/`READY.`/`READY:` variants, and the last-marker-wins loop correctly lets a later `NOT READY` override an earlier `READY` and vice versa. Test coverage matches this well, including both directions of the override.

One inconsistency worth a look, not necessarily blocking: the docstring says "the **last** explicit READY / NOT READY line is authoritative," but `if _READY_FULL_PHRASE in response: return True` short-circuits before the line-scan runs, so a response like `"READY: I am ready to execute the task.\nActually wait, NOT READY"` returns `True` even though a later, more specific `NOT READY` marker follows it, in genuine violation of the rule the rest of the function implements. Narrow case since the full phrase is the older, more explicit format less likely to be immediately walked back, but if you want the two paths to actually agree, the full-phrase check would need to feed into the same last-marker accounting rather than returning early.

Kyou12138:
Closing as a duplicate of #6488, which predates this PR and applies the same last-marker-wins readiness parsing across the relevant code paths with tests. The stricter line-anchored false-positive case from this version has been shared on #6488.

PR #6565, opened 2026-07-16
fix(deps): widen json-repair bound (GHSA-xf7x-x43h-rpqh) + adapt parser to new repair semantics

thomaschhh:
## What

Two commits:

1. **`fix(deps)`** — widen `json-repair` in `lib/crewai/pyproject.toml` from `~=0.25.2` (`>=0.25.2,<0.26.0`) to `>=0.60.1,<1.0`.
2. **`fix(parser)`** — adapt `_safe_repair_json` to json-repair's changed repair semantics so the parser test suite stays green under the new bound.

## Why

The `<0.26.0` cap blocks downstream projects from installing `json-repair >= 0.60.1`, the release that fixes **GHSA-xf7x-x43h-rpqh** (CVSS 7.5, HIGH — DoS via unbounded `$ref` resolution in `SchemaRepairer.resolve_schema()`). Since json-repair reaches downstream projects only transitively through CrewAI, the pin can only be lifted here. Latest json-repair is `0.61.5`.

## The bump is not a drop-in — and why the second commit is needed

json-repair `>=0.60` changed the *behavior* of `repair_json()` (the signature is unchanged). For non-JSON input it now **fabricates** output instead of returning the empty sentinel 0.25.x produced:

| Input | 0.25.x | 0.61.x |
|-------|--------|--------|
| `{invalid_json` | `""` | `["invalid_json"]` |
| `{temperature in SF}` | `""` | `["temperature in SF}"]` |

`_safe_repair_json` relied on the empty sentinel (`UNABLE_TO_REPAIR_JSON_RESULTS`) to detect unrepairable input and fall back to the original string. With a fabricated non-empty result, that check no longer fires and the parser returns garbage — 14 parser tests fail on the bump alone.

**Fix:** only trust the repaired result when it parses back to a JSON object; otherwise keep the original string.

```python
try:
    parsed = json.loads(result)
except ValueError:
    return tool_input
if not isinstance(parsed, dict):
    return tool_input
```

The other `repair_json` call site (`tools/tool_usage.py`) already guards with `isinstance(arguments, dict)`, so it needed no change.

## Verification

- `lib/crewai/tests/agents/test_crew_agent_parser.py`: **44/44 pass** (14 failed on the bump without the parser fix).
- Full `lib/crewai/tests/agents` + `lib/crewai/tests/tools`: the parser fix removes **exactly those 14** failures and introduces **zero** new ones. Remaining failures in that run are pre-existing and unrelated (native tool calling / a2a / litellm — require API keys/network).
- `ruff check` and `ruff format --check` pass on the changed file.

## References

- Advisory: https://github.com/advisories/GHSA-xf7x-x43h-rpqh
- Affected: `json-repair < 0.60.1` · Patched: `0.60.1` (latest `0.61.5`)


thomaschhh:
Closing this MR as it was fixed.
See https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/pyproject.toml#L35

PR #6587, opened 2026-07-18
fix(project): resolve skills paths against project root

### LINKED ISSUE #6585: Agent "skills" paths resolve against process cwd instead of project_root (unlike tools/output_pydantic)
### Description
When loading a crew from JSONC (`load_crew`, `load_crew_from_definition`, `load_agent_from_definition`), every relative path field
(`tools` with `custom:` prefix, `output_pydantic`/`response_format` python refs) is correctly resolved against `project_root`.
The agent's `skills` field is the one exception — it is passed through unresolved in
`crewai/project/json_loader.py::_agent_kwargs_from_definition`, and ultimately reaches
`crewai/skills/loader.py::discover_skills(Path(skill))`, which resolves the path against
`Path.cwd()` at call time instead of `project_root`.

This means the same crew definition succeeds or fails purely based on the process's current
working directory at import/call time — e.g. it works when running a script directly from its
own directory, but fails with `FileNotFoundError: Skill search path does not exist or is not a
directory: skills` when the same code runs as part of a server process launched from a
different cwd (uvicorn, docker, IDE run configs, etc.).

### Reproduction
- `agent.jsonc`: `"skills": ["./skills"]`, sibling `skills/` directory with a `SKILL.md`.
- `crew.jsonc`: references this agent.
- `load_crew(Path(__file__).with_name("crew.jsonc"))` — works if cwd == the file's own directory,
  raises `FileNotFoundError` for any other cwd.

### Expected
`skills` paths should resolve against `project_root` (same as `tools`/`output_pydantic`), not
against `Path.cwd()`.

### Environment
- crewai 1.15.3
- Python 3.12, Windows

Yigtwxx:
## Problem

Every relative path field in a JSON crew definition is resolved against `project_root`: `custom:` tools, `output_pydantic` / `response_model` python refs, and `input_files`. The `skills` field is the one exception.

It is passed through unresolved in `_agent_kwargs_from_definition` and `_crew_kwargs_from_definition`, and reaches `crewai.skills.loader.load_skill`, whose final branch calls `discover_skills(Path(skill))`. That resolves against the process working directory at call time.

The result is that the same crew definition succeeds or fails purely based on where the process was launched from. Running a script from its own directory works; running the identical code under uvicorn, in a container, or from an IDE run configuration raises:

```
FileNotFoundError: Skill search path does not exist or is not a directory: skills
```

## Root cause

`skills` is listed in `_AGENT_OBJECT_REF_FIELDS` and `_CREW_OBJECT_REF_FIELDS`, so `python:` reference dicts are converted to `Skill` objects. Plain path strings, however, are left untouched and never normalized against `project_root`.

## Change

Anchor `skills` entries that denote *relative* filesystem paths to `project_root`.

`_is_skill_path_ref` mirrors the string dispatch order of `load_skill` so the two stay in sync:

| Entry | Handling |
| --- | --- |
| `@org/name` | registry reference, left untouched |
| `---\n...` | inline SKILL.md document, left untouched |
| `Skill` object | already resolved, left untouched |
| absolute path | left untouched |
| relative path | anchored to `project_root` |

Normalization runs after `_resolve_agent_python_refs` / `_resolve_crew_python_refs`, so `python:` refs are already `Skill` instances by that point and only genuine strings remain. It is confined to the `resolve_tools=True` branch, leaving validation and deploy mode free of any project code execution.

## Compatibility

No definition that loads today stops loading.

- Absolute paths — including Windows drive and UNC forms seen from a POSIX host — are passed through verbatim.
- Relative paths escaping the root (`../shared_skills`) are anchored rather than rejected, which is exactly what they already resolved to whenever the process ran from the project directory.

Unlike `input_files`, no containment rule is introduced: the skills loader has never had one, and adding it here would reject working monorepo layouts that keep a shared skills directory next to the project.

## Tests

Five cases in `lib/crewai/tests/project/test_crew_loader.py`, alongside the existing `input_files` path tests. Every path case chdirs elsewhere first, so each one fails without this change:

- agent-level `skills` resolves against `project_root`
- crew-level `skills` stores the project-rooted path
- `@acme/research` reaches the registry resolver verbatim, guarding the Skills Repository path
- `../shared_skills` loads a sibling directory above the project root
- an absolute path outside the project root loads unchanged

Fixes #6585

### DIFF
--- a/lib/crewai/src/crewai/project/json_loader.py
+++ b/lib/crewai/src/crewai/project/json_loader.py
@@ -1013,6 +1013,10 @@ def _agent_kwargs_from_definition(
     if resolve_tools:
         _resolve_tool_fields(agent_kwargs, path, project_root=project_root)
         _resolve_agent_python_refs(agent_kwargs, path, project_root)
+        if "skills" in agent_kwargs:
+            agent_kwargs["skills"] = _normalize_skills(
+                agent_kwargs["skills"], project_root
+            )
     else:
         # Validation/deploy mode: check tool declarations structurally without
         # importing or instantiating anything — custom:<name> tools execute
@@ -1117,6 +1121,8 @@ def _crew_kwargs_from_definition(
         crew_kwargs["manager_agent"] = agents_map[manager_agent]
 
     _resolve_crew_python_refs(crew_kwargs, source, project_root)
+    if "skills" in crew_kwargs:
+        crew_kwargs["skills"] = _normalize_skills(crew_kwargs["skills"], project_root)
 
     return crew_kwargs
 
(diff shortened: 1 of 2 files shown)


Yigtwxx:
Per CONTRIBUTING.md, this contribution was authored with an AI coding assistant and therefore requires the `llm-generated` label. I do not have permission to apply labels on this repository, so could a maintainer please add it.

Flagging the one behavior change explicitly for review: an absolute skill path resolving outside the project root now raises `JSONProjectValidationError` rather than loading. This follows the existing `input_files` containment semantics, but it is the only case where a previously working configuration would now be rejected. Happy to gate that behind a warning instead if you would prefer a non-breaking path.

Yigtwxx:
Rebased onto `c8f441c` and dropped the one behavior change this PR had.

**No longer breaking.** Previously an absolute skill path outside `project_root` was rejected, borrowing the `input_files` containment rule. That is gone. Absolute paths now pass through verbatim and `../shared_skills` is anchored rather than rejected, so no definition that loads today stops loading — the change is strictly additive. Containment never applied to the skills loader, and enforcing it would have broken monorepo layouts that keep a shared skills directory beside the project.

**Re-verified against the skills work that landed since.** `load_skill` and `resolve_registry_ref` gained the `activate` flag in #6675, and #6727 added usage events; the string dispatch order that `_is_skill_path_ref` mirrors is unchanged, so the two are still in sync. I updated the registry test for the new signature. The underlying bug is still present on `main`: `skills` appears in `json_loader.py` only inside `_AGENT_OBJECT_REF_FIELDS` / `_CREW_OBJECT_REF_FIELDS`, so plain path strings still reach `discover_skills(Path(skill))` and resolve against the process cwd.

**Overlap with #6596.** That PR fixes the same issue with a very similar approach, and both have been open for two weeks. The difference is scope: this one also covers absolute and Windows drive/UNC forms, keeps the dispatch table aligned with `load_skill`, and adds the `../` and absolute-path cases. Rather than leave you with two open PRs to arbitrate, tell me which you prefer and I will close this one — the fix mattering more than whose branch it lands from.

Two housekeeping notes: CI has never run here, so the checks are still waiting on workflow approval for a first-time contributor; and the `llm-generated` label from CONTRIBUTING.md still needs a maintainer, as I cannot apply labels.

PR #6594, opened 2026-07-19
feat(crewai-tools): add SerpdiveSearchTool

edendalexis:
# Add SerpdiveSearchTool

Adds a search tool for [SERPdive](https://serpdive.com), an AI search API that returns answer-ready web content: every result carries the actual text of the page (url, title, date, content), already extracted, cleaned, and sized for LLM consumption, so agents can quote and cite straight from the tool output. On a [public, replayable 1,000-question benchmark](https://github.com/edendalexis/serpdive-benchmark), SERPdive runs at the same speed as Tavily, feeds your LLM 20.2% fewer tokens, and wins 60.7% of decided quality duels.

## What's included

- `lib/crewai-tools/src/crewai_tools/tools/serpdive_search_tool/`: `SerpdiveSearchTool` (sync `_run` + async `_arun`) over the official [`serpdive`](https://pypi.org/project/serpdive/) SDK, following the structure and current style of the existing search tools (import guard, `package_dependencies`, `env_vars`, JSON string output) + tool README
- Exports in `crewai_tools/__init__.py` and `crewai_tools/tools/__init__.py` (imports and `__all__`), `serpdive` optional dependency group in `lib/crewai-tools/pyproject.toml`
- `lib/crewai-tools/tests/tools/serpdive_search_tool_test.py`: 4 stubbed unit tests (no network), passing locally; the tool was also run live against the production API
- Docs page `docs/edge/en/tools/search-research/serpdivesearchtool.mdx` + overview card + `docs/docs.json` nav entry (edge only, no frozen `docs/v*` touched)
- `lib/crewai-tools/tool.specs.json` regenerated with `generate_tool_specs.py`: the diff is additive only (the new tool's entry), no existing entry changed

## Notes

- Auth via `SERPDIVE_API_KEY` (free key, no card required). No country or locale parameter: the language of the query picks where the search runs automatically.
- Heads-up on `exclude-newer = "3 days"`: the `serpdive` SDK's latest release was published on 2026-07-18, so `uv sync --all-extras` resolves it once the release is more than 3 days old. No config change is included; if CI runs before then and you prefer, we can add a temporary `exclude-newer-package` override instead.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -252,6 +252,7 @@
                           "edge/en/tools/search-research/tavilysearchtool",
                           "edge/en/tools/search-research/tavilyextractortool",
                           "edge/en/tools/search-research/tavilyresearchtool",
+                          "edge/en/tools/search-research/serpdivesearchtool",
                           "edge/en/tools/search-research/arxivpapertool",
                           "edge/en/tools/search-research/serpapi-googlesearchtool",
                           "edge/en/tools/search-research/serpapi-googleshoppingtool",
--- a/docs/edge/en/tools/search-research/overview.mdx
+++ b/docs/edge/en/tools/search-research/overview.mdx
@@ -62,6 +62,10 @@ These tools enable your agents to search the web, research topics, and find info
     Retrieve the status and results of an existing Tavily research task.
   </Card>
 
+  <Card title="SERPdive Search Tool" icon="water" href="/en/tools/search-research/serpdivesearchtool">
+    Web search that returns extracted, answer-ready page content using the SERPdive API.
+  </Card>
+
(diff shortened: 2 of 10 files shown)


edendalexis:
Update while this is open: SERPdive now has a free tier, which changes the calculus for this integration.

`model: "krill"` is **free and unlimited under fair use** — no card, no credits to spend, no trial clock. It returns about 700 tokens per search (roughly half what the usual alternatives send), one request at a time, at low priority. The paid models are unchanged.

Why it matters here: the usual objection to adding a search provider is that reviewers and users need a paid key just to try it. That objection is gone — anyone can run this integration end to end for free, including whoever reviews this PR.

Nothing in this PR needs to change for that: the default stays `mako`, and `krill` is opt-in via the existing model parameter. Happy to add a line to the docs here if you'd like the free tier called out explicitly.

PR #2479, opened 2025-03-26
Add exclude option to `to_serializable()`

vinibrsl:

PR #6595, opened 2026-07-19
fix(serialization): apply exclude recursively for keys nested inside Pydantic models

lntutor:
### What's broken

`to_serializable(obj, exclude={...})` applies `exclude` recursively for dicts (filtered at every level), lists, and dataclasses (both forward `exclude` into recursive calls). The `BaseModel` branch is the exception: it passes `exclude` to `model_dump()` — which only drops **top-level** fields (a `set` exclude is not recursive in Pydantic v2) — and then recurses into the resulting dict **without** forwarding `exclude`:

```python
class Address(BaseModel):
    street: str
    secret: str
class Person(BaseModel):
    name: str
    address: Address

to_serializable(Person(name="John", address=Address(street="123 Main St", secret="TOP-SECRET")), exclude={"secret"})
# {'name': 'John', 'address': {'street': '123 Main St', 'secret': 'TOP-SECRET'}}  <- 'secret' leaked
```

Since `to_serializable` serializes agent/task/event state (e.g. `to_string`), this leaks fields a caller explicitly asked to drop — potentially sensitive data.

### Fix

Forward `exclude` into the `BaseModel` branch's recursive calls (both the `model_dump` path and the `__dict__` fallback), relying on the existing recursive dict-branch filtering. Top-level exclusion is unchanged.

### Tests

Extends `lib/crewai/tests/utilities/test_serialization.py`: a nested-model exclude case, and a parity check that the model path matches the plain-dict path. Full file passes (27). Distinct from #2479 (which added the `exclude` option).


lntutor:
Merged current upstream `main` into the branch to clear the behind-base status while preserving the existing minimal serialization fix.\n\nFresh verification after the merge:\n- `uv run pytest -q lib/crewai/tests/utilities/test_serialization.py` — 27 passed\n- `uv run ruff check lib/crewai/src/crewai/utilities/serialization.py lib/crewai/tests/utilities/test_serialization.py` — passed\n- `uv run ruff format --check ...` — passed

PR #6604, opened 2026-07-20
fix(llm): handle LiteLLM supports_response_schema introspection failure gracefully (#6412)

### LINKED ISSUE #6412: bug: ProviderCapabilities.get_capabilities() static fallbacks are skipped when LiteLLM introspection fails
## Description

In `crewai/llm.py`, the `ProviderCapabilities` dataclass (#6233) adds static fallback detection for `response_format` and `reasoning_effort` support. However, when the LiteLLM `supports_response_schema` field is unavailable, the code returns early before `get_capabilities()` can run, skipping the static fallback logic entirely.

### Where

`crewai/llm.py` — The early return guard above `get_capabilities()` bypasses both the static `response_format` fallback AND `reasoning_effort` validation when `supports_response_schema` is unavailable from LiteLLM introspection.

### Impact

- Providers that don't expose `supports_response_schema` via LiteLLM lose **all** capability detection, even for features that have static fallbacks (e.g., well-known provider lists)
- `response_format` requests are incorrectly sent to providers that don't support it
- `reasoning_effort` settings are not validated against known provider lists

### Reproduction

Use a provider where LiteLLM doesn't return `supports_response_schema`:
1. Configure `response_format={"type": "json_object"}`
2. The capability check returns early
3. The request may be sent to a provider that doesn't support it

### Related

- PR #6233: feat(providers): add ProviderCapabilities dataclass
- Coderabbit review flagged this exact issue

### Proposed Fix

Move the static capability checks (`get_capabilities()`) before the LiteLLM introspection early return, so static fallbacks are always evaluated regardless of introspection availability.

Hasnaathussain:
# Pull Request Description

## Description

Closes #6412

This PR resolves an issue in `LLM._validate_call_params()` where unhandled LiteLLM introspection failures in `supports_response_schema()` would cause `_validate_call_params()` to crash or skip parameter validation.

### Root Cause
In `crewai/llm.py`, `_validate_call_params()` invoked `supports_response_schema()` directly without exception handling. If LiteLLM failed to introspect an unmapped/custom model ID or raised an exception, `_validate_call_params()` crashed unhandled.

### Proposed Changes
1. Added exception handling around `supports_response_schema()` in `_validate_call_params()`, mirroring the pattern used in `supports_function_calling()`.
2. On LiteLLM introspection errors, the failure is logged at `DEBUG` level and `is_supported` defaults to `True` (permitting modern models and proxies with custom model names to execute).
3. Added unit test `test_validate_call_params_handles_introspection_error()` in `lib/crewai/tests/test_llm.py`.

## Testing
- Verified `test_validate_call_params_handles_introspection_error` passes.
- Verified linting and formatting with `ruff check` and `ruff format`.

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -2373,16 +2373,21 @@ class LLM(BaseLLM):
         return None
 
     def _validate_call_params(self) -> None:
-        """
-        Validate parameters before making a call. Currently this only checks if
-        a response_format is provided and whether the model supports it.
-        The custom_llm_provider is dynamically determined from the model:
-          - E.g., "openrouter/deepseek/deepseek-chat" yields "openrouter"
-          - "gemini/gemini-1.5-pro" yields "gemini"
-          - If no slash is present, "openai" is assumed.
+        """Validate call parameters before executing a completion request.
+
+        Checks whether the requested ``response_format`` is supported by the
+        target model/provider via LiteLLM introspection.  When introspection
+        raises an exception (e.g. an unmapped custom model ID or proxy
+        endpoint), the failure is logged at DEBUG level and validation is
+        skipped so the request can proceed — this mirrors the permissive
+        fallback used in :meth:`supports_function_calling`.
+
+        When no ``response_format`` is configured (including when only
+        ``reasoning_effort`` or other parameters are set) this method returns
+        immediately without performing any checks.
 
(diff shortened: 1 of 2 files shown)


Hasnaathussain:
Rebased onto current main and kept the existing focused changes. The validation suite passes 8 tests; introspection logging and reasoning_effort fallback cases are covered.

PR #6610, opened 2026-07-21
Withdrawn

Darko893:
Withdrawn.


Darko893:
All three review points addressed in the latest commit: api_key moved to a PrivateAttr so it can never appear in model_dump or serialized snapshots (regression test added), non-2xx and non-JSON responses now raise a clean ValueError with the status code instead of crashing in json parsing (test added), and the markdown test now asserts response_format is actually sent in the request body. 7 tests green.

Darko893:
Addressed the remaining request-boundary review and rebased onto current main. The tool now disables redirects before sending X-API-Key, requires a 2xx response with an object JSON body, includes the HTTP status in errors, documents the public tool path, and covers redirect, non-object JSON, credential header, and pre-request URL rejection behavior. Targeted suite: 13 passed.

Darko893:
I am sunsetting Haunt and withdrawing this submission. Thank you for the time and consideration.

PR #6616, opened 2026-07-22
fix(agents): await native async tools on caller loop

### LINKED ISSUE #6611: [BUG] Async tools are not awaited natively under native-function-calling models — executed via `asyncio.run()` inside a thread, inconsistent with the ReAct executor
### Description

When an agent uses a model with **native/parallel function calling** (OpenAI, Gemini, Anthropic), the experimental agent executor calls tools through the **synchronous** `BaseTool.run` path and bridges coroutine tools with `asyncio.run()` — so an async tool (`async def _run` or `_arun`) is **not awaited natively on the running event loop**. It is instead offloaded to a thread and given a brand-new nested event loop per call.

The **ReAct/text executor** does the opposite: it calls `CrewStructuredTool.ainvoke`, which detects a coroutine and `await`s it natively.

So the *same* async tool behaves differently depending purely on whether the model supports native function calling. For an I/O-bound async tool (e.g. `httpx.AsyncClient`), the native-tool path loses the intended single-loop concurrency, adds a thread hop, and pays `asyncio.run()` setup/teardown on every call.

I could not find an existing issue for this on current versions (related but distinct: #195 predates async-tool support; #3988 is about async `@agent`/`@task` *definition* methods; #1644 is the `asyncio.run() cannot be called from a running event loop` symptom in a Flow/Colab context). Filing to confirm whether this is intended behavior of the experimental executor.

### Root cause (1.15.5)

- `experimental/agent_executor.py:1232` offloads the step engine: `result = await asyncio.to_thread(step_executor.execute, ...)`.
- Inside that thread, `experimental/agent_executor.py:1995` calls the tool **synchronously**: `raw_result = tool_func(**args_dict)` (i.e. `BaseTool.run`), never `arun`/`ainvoke`.
- `BaseTool.run` (`tools/base_tool.py:313`) gets a coroutine back from the async `_run` and bridges it at `base_tool.py:328`: `result = asyncio.run(result)` — a fresh nested loop per call.

Contrast — the ReAct path awaits natively:
- `tools/structured_tool.py:399-400`: `if inspect.iscoroutinefunction(self.func): return await self.func(...)`.

### Question / possible direction

Is threading + `asyncio.run()`-per-call the intended behavior for the experimental native-tool executor, or should `_execute_single_native_tool_call` prefer the tool's async entry point (`arun`/`ainvoke`) and `await` it when the executor is already async? Flagging as issue first since this lives in `experimental/` and may already be on a roadmap.

* Model: reproduced on `gemini/gemini-2.5-flash`. The code path is model-agnostic (the sync `tool_func` call is not model-specific), so any native-function-calling model should reproduce it but I've only verified Gemini.

### Steps to Reproduce

`crewai==1.15.5`, Python 3.12. Set a key for a native-function-calling model (`export GEMINI_API_KEY=...`), then run:

```python
import os, asyncio, threading, traceback
from typing import Type
from crewai import Agent, Task, Crew, Process

class Inp(BaseModel):
    label: str = Field(..., description="the label")

class AsyncTool(BaseTool):
    name: str = "timer"
    description: str = "Call once with your label."
    args_schema: Type[BaseModel] = Inp
    async def _run(self, label: str) -> str:                 # genuinely async
        cur = threading.current_thread()
        print(f"async _run thread={cur.name} is_main={cur is threading.main_thread()}")
        for fr in traceback.extract_stack():
            if "/crewai/" in fr.filename and ("agent_executor" in fr.filename or "base_tool" in fr.filename):
                print("   ", fr.filename.split("site-packages/")[-1] + f":{fr.lineno}", fr.line)
        await asyncio.sleep(0.2)
        return "recorded"

def build_crew():
    a = Agent(role="Timer", goal="call the timer tool once",
              backstory="You call tools exactly as instructed.",
              tools=[AsyncTool()], llm="gemini/gemini-2.5-flash", verbose=False)
    t = Task(description="Call the `timer` tool exactly once with label='A'. Nothing else.",
             expected_output="recorded", agent=a)
    return Crew(agents=[a], tasks=[t], process=Process.sequential)

async def main():
    print("MAIN loop thread =", threading.current_thread().name)
    await build_crew().akickoff()

asyncio.run(main())
```

**Observed output:**

```
MAIN loop thread = MainThread
async _run thread=asyncio_0  is_main=False
    crewai/experimental/agent_executor.py:1775  execution_result = self._execute_single_native_tool_call(tool_call)
    crewai/experimental/agent_executor.py:1995  raw_result = tool_func(**args_dict)
    crewai/tools/base_tool.py:328  result = asyncio.run(result)
```

The async `_run` runs on a worker thread (`asyncio_0`, `is_main=False`) via `asyncio.run()`, not awaited on the main event loop.

### Expected behavior

An async tool should be awaited natively on the running event loop (as the ReAct executor does), i.e. `thread=MainThread`, with no per-call `asyncio.run()` and no thread hop — matching the documented purpose of async tools (non-blocking, single-loop concurrency).

### Screenshots/Code snippets

.

### Operating System

### Python Version

3.12

### crewAI Version

1.15.5

### crewAI Tools Version

1.15.5

### Virtual Environment

Venv

### Evidence

.

### Possible Solution

The async-aware tool dispatch already exists. CrewStructuredTool.ainvoke (tools/structured_tool.py:399-403) already branches correctly:

if inspect.iscoroutinefunction(self.func):
    return await self.func(**parsed_args, **kwargs)   # async tool → awaited natively
return await asyncio.get_event_loop().run_in_executor(  # sync tool → offloaded
    None, lambda: self.func(**parsed_args, **kwargs)
)

The ReAct executor uses this. the native-tool executor bypasses it by calling the sync BaseTool.run directly (experimental/agent_executor.py:1995, raw_result = tool_func(**args_dict)), which then bridges coroutines with asyncio.run() (base_tool.py:328).

A fix would route the native-tool path through the same async-aware dispatch: give _execute_single_native_tool_call an async variant that awaits tool.ainvoke(...) for coroutine tools (offloading only genuinely-sync tools), and call it on the running loop rather than wrapping the whole step in asyncio.to_thread (agent_executor.py:1232). That would make both executors share one tool-invocation path and eliminate the per-call asyncio.run().

I haven't fully mapped the step_executor batching around agent_executor.py:1232, so I'll defer to maintainers on exact placement. Could attempt a PR if this is the intended direction.

### Additional context

MacOS Tahoe

Sakshamm-Goyal:
## Summary

- dispatch awaitable native-tool results back to the active flow event loop instead of creating a per-call worker loop
- preserve synchronous native-tool execution, argument validation, usage limits, hooks, caching, and ordering
- cover async `BaseTool._run`, wrapped async functions, parallel calls, and flow-route registration

Closes #6611

## Validation

- `pytest lib/crewai/tests/agents/test_agent_executor.py -q -n 0 --timeout=60 --block-network --import-mode=importlib -k "not anthropic_provider_has_image_block_converter and not kickoff_response_format_with_planning_and_tools"` (98 passed, 2 deselected because optional Anthropic/Exa dependencies are unavailable)
- `pytest lib/crewai/tests/agents/test_agent_executor.py::TestNativeToolExecution -q -n 0 --timeout=60 --block-network --import-mode=importlib` (8 passed)
- `mypy lib/crewai/src/crewai/experimental/agent_executor.py`
- `ruff check` and `ruff format --check` on changed files

### DIFF
--- a/lib/crewai/src/crewai/experimental/agent_executor.py
+++ b/lib/crewai/src/crewai/experimental/agent_executor.py
@@ -121,6 +121,9 @@ if TYPE_CHECKING:
     from crewai.tools.tool_types import ToolResult
 
 _RouteT = TypeVar("_RouteT", bound=str)
+_native_tool_event_loop: contextvars.ContextVar[asyncio.AbstractEventLoop | None] = (
+    contextvars.ContextVar("native_tool_event_loop", default=None)
+)
 
 
 class AgentExecutorState(BaseModel):
@@ -1675,7 +1678,6 @@ class AgentExecutor(Flow[AgentExecutorState], BaseAgentExecutor):
 
         return "tool_completed"
 
-    @router("native_tool_calls")
     def execute_native_tool(
         self,
     ) -> Literal["native_tool_completed", "tool_result_is_final"]:
@@ -1849,6 +1851,17 @@ class AgentExecutor(Flow[AgentExecutorState], BaseAgentExecutor):
 
         return "native_tool_completed"
 
+    @router("native_tool_calls")
+    async def execute_native_tool_async(
+        self,
+    ) -> Literal["native_tool_completed", "tool_result_is_final"]:
+        """Run native tool calls while keeping coroutine tools on this event loop."""
+        token = _native_tool_event_loop.set(asyncio.get_running_loop())
+        try:
+            return await asyncio.to_thread(self.execute_native_tool)
+        finally:
+            _native_tool_event_loop.reset(token)
+
     def _should_parallelize_native_tool_calls(self, tool_calls: list[Any]) -> bool:
         """Determine if native tool calls are safe to run in parallel."""
         if len(tool_calls) <= 1:
@@ -1991,8 +2004,11 @@ class AgentExecutor(Flow[AgentExecutorState], BaseAgentExecutor):
         elif not from_cache and not max_usage_reached and output_tool is not None:
             if func_name in self._available_functions:
                 try:
-                    tool_func = self._available_functions[func_name]
-                    raw_result = tool_func(**args_dict)
+                    if isinstance(original_tool, BaseTool):
+                        raw_result = self._run_native_tool(original_tool, args_dict)
+                    else:
+                        tool_func = self._available_functions[func_name]
+                        raw_result = tool_func(**args_dict)
                     raw_tool_result = raw_result
 
                     # Add to cache after successful execution (before string conversion)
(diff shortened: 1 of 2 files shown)


Sakshamm-Goyal:
This PR was authored with AI assistance. The repository contribution policy requires the `llm-generated` label, but the contributor account does not have permission to apply it. Could a maintainer add that required label? Thank you.

PR #6621, opened 2026-07-23
fix: bare raise in ToolUsage._original_tool_calling causes RuntimeError instead of ToolUsageError (#6430)

Diwak4r:
## Summary

Fixes #6430: A bare `raise` statement in `ToolUsage._original_tool_calling()` (lib/crewai/src/crewai/tools/tool_usage.py:853) was outside any exception handler. When `_validate_tool_input` returns a non-dict value (which it can in edge cases or with future changes), Python raises `RuntimeError: No active exception to re-raise` instead of the intended `ToolUsageError`.

## Root Cause

At line 852-853:
```python
    if raise_error:
        raise  # Bare raise - no active exception here!
```

The `_validate_tool_input` method either returns a dict or raises an exception. The `try/except` block above catches exceptions from `_validate_tool_input`, but the non-dict check happens AFTER that try/except completes. So if somehow a non-dict is returned (future change, subclass override, etc.), the bare `raise` has no active exception to re-raise.

## Fix

Replace the bare `raise` with an explicit `ToolUsageError` that mirrors the `raise_error=False` branch:

```python
    if raise_error:
        raise ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
```

This is also flagged by ruff (PLE0704) as a bare raise outside exception handler.

## Verification

All 28 `tests/tools/test_tool_usage.py` tests pass. The fix ensures consistent error handling between `raise_error=True` and `raise_error=False` branches.

### DIFF
--- a/lib/crewai-files/src/crewai_files/uploaders/factory.py
+++ b/lib/crewai-files/src/crewai_files/uploaders/factory.py
@@ -11,6 +11,7 @@ from crewai_files.uploaders.anthropic import AnthropicFileUploader
 from crewai_files.uploaders.bedrock import BedrockFileUploader
 from crewai_files.uploaders.gemini import GeminiFileUploader
 from crewai_files.uploaders.openai import OpenAIFileUploader
+from crewai_files.processing.exceptions import PermanentUploadError
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


Diwak4r:
Superseded by #6696. The replacement (`fix/6430-bare-raise-tool-usage`) is a clean single commit rebased onto latest upstream main and fixes **both** bare-raise sites in `ToolUsage._original_tool_calling` (the bare `raise` in the failure path **and** the implicit raise on a non-dict return), with regression tests. Closing to keep a single authoritative PR per issue.

Diwak4r:
Closing in favor of #6696 (clean rebase + both raise sites fixed + tests).

PR #6625, opened 2026-07-23
fix(reasoning): use flexible regex to detect READY state to prevent agent deadlocks

Kikingodoy:
This PR fixes a bug in the `AgentReasoning` loop where the readiness check was hardcoded to expect the exact string `"READY: I am ready to execute the task."`. 

Newer local/open-source models, or models under slightly altered prompts, frequently output just `"READY"` or some other variation, causing the planning executor to fail to detect readiness and trap the agent in an infinite planning loop. 

This replaces the hardcoded exact string matches with a flexible regular expression `(?i)(?<!not )\bready\b` that safely identifies when the model is ready, gracefully ignoring "not ready".

### DIFF
--- a/lib/crewai/src/crewai/utilities/reasoning_handler.py
+++ b/lib/crewai/src/crewai/utilities/reasoning_handler.py
@@ -105,6 +105,13 @@ FUNCTION_SCHEMA: Final[dict[str, Any]] = {
 }
 
 
+def _is_ready(response: str) -> bool:
+    """Check if the response indicates the agent is ready."""
+    if re.search(r"(?i)\bnot\s+ready\b", response):
+        return False
+    return bool(re.search(r"(?i)\bready\b", response))
+
+
 class AgentReasoning:
     """
     Handles the agent planning/reasoning process, enabling an agent to reflect
@@ -410,7 +417,7 @@ class AgentReasoning:
             return (
                 response_str,
                 [],
-                bool(re.search(r"(?i)(?<!not )\bready\b", response_str)),
+                _is_ready(response_str),
             )
 
         except Exception as e:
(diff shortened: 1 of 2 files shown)


Kikingodoy:
I've pushed an update addressing the feedback! 

I've extracted the readiness check into a centralized `_is_ready` helper method. It uses the `\s+` regex token to robustly handle all whitespace variations of "not ready" (including tabs, newlines, and double spaces), ensuring we don't prematurely evaluate the readiness state. 

I've also added a comprehensive test suite in `tests/utilities/test_reasoning_handler.py` to explicitly guarantee all these edge cases (and various casing formats) are covered moving forward!

lib/crewai/src/crewai/utilities/reasoning_handler.py
```diff
@@ -104,6 +105,13 @@ class AgentReasoningOutput(BaseModel):
 }
 
 
+def _is_ready(response: str) -> bool:
+    """Check if the response indicates the agent is ready."""
+    if re.search(r"(?i)\bnot\s+ready\b", response):
+        return False
+    return bool(re.search(r"(?i)\bready\b", response))
```
gnanirahulnutakki, line 112: Could we parse an explicit readiness marker instead of treating any remaining `ready` token as permission to execute? On this exact head, both `AgentReasoning._parse_planning_response("Plan draft.\nNOT YET READY: missing token")[1]` and the same call with `"I need more context before I am ready."` return `True`; `never ready`, `hardly ready`, `will be ready`, and `NOT-READY` do too. That prematurely exits the refinement loop even though the model is explicitly unready. The prompt already defines a terminal `READY` / `NOT READY` protocol, so a safer approach is to match line-anchored markers (optionally `Status:` / `I am`) and use the last marker, rather than infer intent from unrestricted prose. I validated that approach against the 16 existing cases plus six semantic-negation cases: 22 focused tests passed, along with Ruff and mypy.

gnanirahulnutakki (commented):
Thanks for broadening support beyond the legacy exact sentence. I reproduced a remaining semantic false-positive in the centralized helper: common unready/future-readiness prose now ends planning early. I included a protocol-based parsing direction and focused validation inline.

PR #6629, opened 2026-07-24
fix: detect Anthropic preview tool-use blocks

alex-clawd:
## Summary

- Detect Anthropic tool-use content blocks by API shape (`type == "tool_use"`, `name`, `input`) instead of only concrete SDK classes.
- Apply the same handling to sync, async, and streaming Anthropic paths.
- Add regression coverage for dict-shaped tool-use blocks like preview/new model responses.

## Root cause

Fable/preview Anthropic models can return tool-use blocks whose concrete SDK type does not match the imported `ToolUseBlock` / `BetaToolUseBlock` aliases. The response can contain a valid tool-use block with no text content.

Before this change, CrewAI missed those tool-use blocks, extracted no text, returned an empty string, and the generic LLM response validator raised:

```txt
Invalid response from LLM call - None or empty.
```

## Validation

- `uv run ruff format lib/crewai/src/crewai/llms/providers/anthropic/completion.py lib/crewai/tests/llms/anthropic/test_anthropic.py`
- `uv run ruff check lib/crewai/src/crewai/llms/providers/anthropic/completion.py lib/crewai/tests/llms/anthropic/test_anthropic.py`
- `uv run pytest lib/crewai/tests/llms/anthropic/test_anthropic.py -q`
- `git diff --check`

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -124,6 +124,33 @@ def _contains_file_id_reference(messages: list[dict[str, Any]]) -> bool:
     return False
 
 
+def _is_tool_use_block(block: Any) -> bool:
+    """Return true for Anthropic tool-use blocks across SDK shapes.
+
+    New/preview Anthropic models can return content blocks whose concrete SDK
+    class is not one of the imported ``ToolUseBlock`` aliases. The stable part
+    of the API shape is the block type plus ``name`` and ``input`` fields, so use
+    duck typing instead of relying only on SDK classes.
+    """
+    if isinstance(block, (ToolUseBlock, BetaToolUseBlock)):
+        return True
+    if isinstance(block, dict):
+        return block.get("type") == "tool_use" and "name" in block and "input" in block
+    return (
+        getattr(block, "type", None) == "tool_use"
+        and hasattr(block, "name")
+        and hasattr(block, "input")
+    )
+
+
+def _tool_use_blocks(blocks: list[Any]) -> list[Any]:
+    return [block for block in blocks if _is_tool_use_block(block)]
+
+
+def _tool_use_attr(block: Any, attr: str) -> Any:
+    return block.get(attr) if isinstance(block, dict) else getattr(block, attr)
+
+
 class AnthropicThinkingConfig(BaseModel):
     type: Literal["enabled", "disabled"]
     budget_tokens: int | None = None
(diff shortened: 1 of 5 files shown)


lib/crewai/tests/llms/anthropic/test_anthropic.py
```diff
@@ -1358,6 +1358,33 @@ def test_tool_search_via_llm_class():
...
+    mock_response.stop_reason = "tool_use"
+    mock_response.id = "msg_123"
+
+    mock_client = MagicMock()
+    mock_client.messages.create.return_value = mock_response
+    llm._client = mock_client
+
+    result = llm.call("Search for CrewAI", tools=_MANY_TOOLS)
+
+    assert result == mock_response.content
```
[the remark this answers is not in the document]
alex-clawd: Addressed in f21d995. Expanded coverage for dict-shaped Anthropic tool_use blocks across sync call, async acall, sync streaming, async streaming, direct tool execution, and the follow-up conversation path. Also made thinking-block extraction tolerate dict-shaped non-thinking content so streaming final messages with dict tool_use blocks do not crash before returning tool calls.

Validation:
- uv run ruff format lib/crewai/src/crewai/llms/providers/anthropic/completion.py lib/crewai/tests/llms/anthropic/test_anthropic.py
- uv run ruff check lib/crewai/src/crewai/llms/providers/anthropic/completion.py lib/crewai/tests/llms/anthropic/test_anthropic.py
- uv run pytest lib/crewai/tests/llms/anthropic/test_anthropic.py -q (53 passed)
- git diff --check

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -124,6 +124,33 @@ def _contains_file_id_reference(messages: list[dict[str, Any]]) -> bool:
...
+    if isinstance(block, dict):
+        return block.get("type") == "tool_use" and "name" in block and "input" in block
+    return (
+        getattr(block, "type", None) == "tool_use"
+        and hasattr(block, "name")
+        and hasattr(block, "input")
+    )
+
+
+def _tool_use_blocks(blocks: list[Any]) -> list[Any]:
```
lorenzejay: Could we preserve a typed tool-use contract here instead of widening this path to `Any`? `_is_tool_use_block()` currently returns `bool`, `_tool_use_blocks()` returns `list[Any]`, and `_tool_use_attr()` accepts an arbitrary `str` and returns `Any`, so mypy can no longer verify any downstream access.

This also masks a concrete mismatch: the predicate validates `type`, `name`, and `input`, but every execution/streaming/follow-up path also requires `id`. An id-less dict is accepted and later produces `tool_use_id=None`; an id-less attribute-based object is accepted and later raises `AttributeError`.

Please narrow this to an explicit local shape—e.g. a `Protocol`/`TypedDict` union with a `TypeGuard`, or a normalized typed internal record—and validate `type`, `id`, `name`, and `input`. That would keep the preview-block compatibility without changing the handler signatures to `list[Any]`.

lorenzejay (commented):
One type-safety concern in the new duck-typed tool-use boundary.

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
joaomdmoura: I replaced the loose Any/string attr accessor with an explicit local tool-use shape: TypedDict for dict blocks, Protocol for object blocks, and a TypeGuard boundary. 
also changed the helper signatures from list[Any] to list[_AnthropicToolUseBlock] and added regression tests for missing id on both dict and object-shaped blocks.

lorenzejay (approved):

PR #6636, opened 2026-07-24
feat(tools): add newsflash corroborated news event search tool

zatmonkey:
## What

Adds `NewsflashNewsTool` (`lib/crewai-tools`), a search tool over [Newsflash](https://newsflash.sh) — real-time corroborated news events + 5-year archive, built for agents. Disclosure: I built Newsflash.

Instead of one raw article per hit, Newsflash dedupes coverage from many outlets into a single **event** with a corroboration count and a confidence score (`min(1, sources/3)`). The tool's `min_confidence` argument lets a crew require independent multi-outlet corroboration before acting on a story — agents that consume news as ground truth are only as reliable as their weakest single source, and a corroboration gate is a cheap, structural defense against fabricated or premature reporting.

## Tool

- Args: `query` (required), `semantic` (default `True`), `category`, `limit` (default 10), `min_confidence` (default 0.0)
- Output: formatted events with category, confidence, corroboration count + source list, and the first article link (details fetched only for returned events, capped at `limit`)
- **Keyless testability**: works with no API key (50 requests/day), so reviewers can run it as-is; optional `NEWSFLASH_API_KEY` (declared `required=False`) for higher limits
- No new dependencies (uses `requests` from core deps); follows the `BaseTool` + `args_schema` + `env_vars` `default_factory` pattern from `SerperDevTool`

## Local checks

- `uv run pytest tests/tools/newsflash_news_tool_test.py -vv` (in `lib/crewai-tools`) — **8 passed** (mocked `requests`, no network, matching existing tool test style)
- `tests/test_generate_tool_specs.py` + `tests/test_optional_dependencies.py` — 10 passed, 1 skipped
- `uv run ruff check` / `ruff format --check` on the changed files — clean
- `uv run mypy` on the new tool — no issues
- Live keyless smoke test against the real API verified formatting end-to-end

https://claude.ai/code/session_01KTSt4BDWJCsGjUyLCnbuj5

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/__init__.py
@@ -117,6 +117,9 @@ from crewai_tools.tools.mongodb_vector_search_tool.vector_search import (
 )
 from crewai_tools.tools.multion_tool.multion_tool import MultiOnTool
 from crewai_tools.tools.mysql_search_tool.mysql_search_tool import MySQLSearchTool
+from crewai_tools.tools.newsflash_news_tool.newsflash_news_tool import (
+    NewsflashNewsTool,
+)
 from crewai_tools.tools.nl2sql.nl2sql_tool import NL2SQLTool
 from crewai_tools.tools.ocr_tool.ocr_tool import OCRTool
 from crewai_tools.tools.oxylabs_amazon_product_scraper_tool.oxylabs_amazon_product_scraper_tool import (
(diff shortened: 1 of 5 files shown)


zatmonkey:
Addressed all review findings in the latest push:

- **Category enforcement** (CodeRabbit, Copilot): `category` is now a `Literal` of the supported set — invalid values fail Pydantic validation before any API call, with a behavior test.
- **N+1 detail requests** (CodeRabbit): article-link fetches are capped by a `max_link_fetches` attribute (default 5), so a wide result set no longer fans out into per-event requests — with a test asserting the bound.
- **Blocking `_arun`** (Copilot): now runs the synchronous path via `asyncio.to_thread`.
- **Missing-id guard** (Copilot): events without an `id` skip the detail fetch instead of hitting `/events/None`.
- **README wording** (CodeRabbit): `requests` described as the existing dependency.

All 10 tests pass locally (`uv run pytest tests/tools/newsflash_news_tool_test.py`).

PR #6649, opened 2026-07-25
Use UTF-8 for user data JSON

Ghraven:
## Summary
- read and write the crewai-core user data JSON file with explicit UTF-8 encoding
- extend the smoke test to cover a non-ASCII value round trip

## Problem
The `.crewai_user.json` helper currently relies on the platform default encoding. That can make trace-consent/user-data persistence locale-dependent on Windows or other non-UTF-8 environments.

## Before / After
Before: user-data JSON reads/writes used the process default encoding.

After: user-data JSON reads/writes use UTF-8 consistently.

## Verification
- `python -m py_compile lib/crewai-core/src/crewai_core/user_data.py lib/crewai-core/tests/test_smoke.py`
- `uv run --project lib/crewai-core pytest lib/crewai-core/tests/test_smoke.py -q`

### DIFF
--- a/lib/crewai-core/src/crewai_core/user_data.py
+++ b/lib/crewai-core/src/crewai_core/user_data.py
@@ -36,7 +36,7 @@ def _load_user_data() -> dict[str, Any]:
     p = _user_data_file()
     if p.exists():
         try:
-            return cast(dict[str, Any], json.loads(p.read_text()))
+            return cast(dict[str, Any], json.loads(p.read_text(encoding="utf-8")))
         except (json.JSONDecodeError, OSError, PermissionError) as e:
             logger.warning("Failed to load user data: %s", e)
     return {}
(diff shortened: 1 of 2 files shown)


Ghraven:
Thanks for the review. I addressed both points in the follow-up commit: `_load_user_data()` now treats `UnicodeDecodeError` like other invalid/corrupt user-data reads, and the JSON write uses `ensure_ascii=False` with a smoke-test assertion that the persisted file contains the literal non-ASCII value.

I reran:
- `python -m py_compile lib/crewai-core/src/crewai_core/user_data.py lib/crewai-core/tests/test_smoke.py`
- `uv run --project lib/crewai-core pytest lib/crewai-core/tests/test_smoke.py -q` (`16 passed`)

Ghraven:
Thanks for the stale ping. I addressed the earlier review feedback in the follow-up commit and the focused smoke test was passing (`16 passed`). I am happy to refresh the branch or make further changes if this is still useful to maintainers.

PR #6652, opened 2026-07-26
Emit skill usage events at runtime for observability

alex-clawd:
## Problem

OSS emitted skill events only for **setup**, never for **usage**:

| event | when | emitted from |
|---|---|---|
| `SkillDiscoveryStartedEvent` | scanning a dir | `skills/loader.py` |
| `SkillDiscoveryCompletedEvent` | scan done | `skills/loader.py` |
| `SkillLoadedEvent` | metadata parsed | `skills/loader.py` |
| `SkillActivatedEvent` | promoted to INSTRUCTIONS | `skills/loader.py` |
| `SkillLoadFailedEvent` | parse/load failed | `skills/loader.py` |

Nothing fired from an execution path — `agent/`, `task/`, `tools/` and
`lite_agent` contain zero skill emissions. `SkillActivatedEvent` is the closest
proxy, but activation is idempotent and happens once at setup, so an agent that
uses a skill across 20 turns produces exactly one event.

That makes it impossible to answer "which skills did this agent actually use, on
which task, and how often" — which is what traces and observability need.

## Change

Adds `SkillUsedEvent`, emitted from `Agent._finalize_task_prompt` — the point
where skill context is rendered into the prompt for a task.

```python
class SkillUsedEvent(SkillEvent):
    type: Literal["skill_used"] = "skill_used"
    disclosure_level: int = 1
```

One emit site covers both sync (`execute_task`) and async
(`execute_task_async`) execution, since both funnel through
`_finalize_task_prompt`. Inherits `SkillEvent`, so `from_agent`/`from_task`
populate `agent_id`, `agent_role` and `task_id` through the existing
`BaseEvent` plumbing — same shape as tool usage events, no new machinery.

44 lines of `src` across 3 files.

## Behaviour

- One event per skill injected, per execution.
- Re-emits on every execution (unlike activation, which is idempotent) — this is
  the point.
- No skills → no events.
- Carries agent id/role, task id, skill name/path and disclosure level.

## Deliberately not included

I checked two adjacent things and left them alone:

1. **`crews/utils.py` activates crew-level skills without an event source**, so
   `activate_skill` skips emission (it is gated on `source is not None`).
   Passing `crew` looks like the fix and the loader docstring even says
   "agent or crew" — but `BaseEvent._set_agent_params` reads `agent.role`, and
   `Crew` has neither `.role` nor `.id`, so it would raise `AttributeError`.
   That needs a wider event-source refactor; out of scope here. Crew-level
   skills are still covered at runtime, because `SkillUsedEvent` fires per agent.

2. No new listener/tracing wiring — the event flows through the existing bus and
   is picked up by anything subscribing to `SkillEvent` subclasses.

## Tests

`tests/events/test_skill_usage_events.py` (6 cases): attribution, one event per
skill, re-emission per execution, disclosure level, no-skill agents stay silent,
and package export.

Note: handlers run on the bus thread pool, so the tests call
`crewai_event_bus.flush()` before asserting. Without it the assertions race the
handler and pass or fail depending on scheduling — that also masks failures
under `-n auto`.

Verified: `tests/skills/ tests/events/` → **294 passed**. Deterministic serially,
under random order, and in parallel. Ruff check + format clean.

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -73,6 +73,7 @@ from crewai.events.types.memory_events import (
     MemoryRetrievalFailedEvent,
     MemoryRetrievalStartedEvent,
 )
+from crewai.events.types.skill_events import SkillUsedEvent
 from crewai.experimental.agent_executor import AgentExecutor
 from crewai.knowledge.knowledge import Knowledge
 from crewai.knowledge.source.base_knowledge_source import BaseKnowledgeSource
@@ -551,8 +552,36 @@ class Agent(BaseAgent):
             The fully prepared task prompt.
         """
         prepare_tools(self, tools, task)
+        self._emit_skill_usage(task)
 
         return apply_training_data(self, task_prompt)
+
+    def _emit_skill_usage(self, task: Task) -> None:
+        """Emit one SkillUsedEvent per skill injected into this task's prompt.
+
+        Skills are agent-scoped and rendered into the prompt on every execution,
+        so this is the runtime usage signal traces need — attributing each skill
+        to the agent and task that used it.
+
+        Args:
+            task: The task whose prompt the skills are being applied to.
+        """
+        if not self.skills:
+            return
+
+        for skill in self.skills:
+            if not isinstance(skill, SkillModel):
+                continue
+            crewai_event_bus.emit(
+                self,
+                event=SkillUsedEvent(
+                    from_agent=self,
+                    from_task=task,
+                    skill_name=skill.name,
+                    skill_path=skill.path,
+                    disclosure_level=skill.disclosure_level,
+                ),
+            )
 
     def _retrieve_memory_context(self, task: Task, task_prompt: str) -> str:
         """Retrieve memory context and append it to the task prompt.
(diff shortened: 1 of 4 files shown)


alex-clawd:
Good catch — the original tests only exercised the shared `_finalize_task_prompt` helper, so they never proved the event reaches the bus during a real run.

Added `TestSkillUsedEventThroughExecution` in `7adbb09e5`: behavioral coverage through the public `execute_task` and `aexecute_task` entry points, with the agent executor stubbed so no LLM call is made. Both assert the event arrives with task attribution.

Kept the helper-level tests for the per-skill details (one event per skill, re-emission per execution, disclosure level, no-skill agents staying silent), since that's where those distinctions are cheapest to pin.

`tests/skills/ tests/events/` → 296 passed, deterministic serially and in parallel.

alex-clawd:
Second point is valid — fixed in `88eb922d1`. `assert received[0].task_id` only proved non-empty, so a mapping to the *wrong* task would still have passed. Now asserts `== str(task.id)` (the event stores `str(task.id)` via `BaseEvent._set_agent_params`).

Tightened three assertions, not two — the helper-level test had the same weakness on both `agent_id` and `task_id`, so those now check identity too.

The first nitpick (exercise `execute_task` / `aexecute_task`) was already addressed in `7adbb09e5`; that review appears to have re-run against the earlier commit. `TestSkillUsedEventThroughExecution` covers both sync and async entry points with a stubbed executor.

`tests/skills/ tests/events/` → 296 passed, deterministic serially and in parallel.

joaomdmoura (approved):

PR #6655, opened 2026-07-26
fix(reasoning): use regex for plan readiness detection (closes #6204)

### LINKED ISSUE #6204: [BUG] Reasoning plan always detects "NOT READY" even though the model indicates "READY"

Diwak4r:
## Description

The reasoning plan readiness detection was matching only the exact phrase `"READY: I am ready to execute the task."` in three places within `reasoning_handler.py`. However, the **refine plan prompt** instructs the model to "Conclude with READY or NOT READY" — a much shorter format. Models responding with standalone `READY` were falsely detected as **NOT READY**, causing the system to loop through unnecessary refinement attempts.

This affected:
- The JSON parsing fallback path (when function calling result parsing fails)
- The fallback text-only LLM call path
- The pure text parsing path for LLMs that don't support function calling

## Fix

Replace all three rigid substring checks with a case-insensitive regex that detects standalone `READY` while excluding `NOT READY` via a negative lookbehind:

```python
re.search(r"(?<!\bnot\s)\bready\b", response, re.IGNORECASE)
```

This correctly handles:
- `READY` — standalone ✅
- `READY: I am ready to execute the task.` — long form ✅
- `Ready` / `ready` — case variants ✅
- `NOT READY` — correctly rejected ❌
- `I'm ready!` — in-context usage ✅
- `NOTREADY` (no space) — correctly rejected ❌

## Related Issue

Closes #6204


Diwak4r:

PR #6622, opened 2026-07-23
fix: await async tools natively in _ainvoke_loop_native_tools

rkfshakti:
## What this PR does

Adds async variants of `_handle_native_tool_calls` and `_execute_single_native_tool_call` that use `await tool.arun()` instead of `tool.run()`, and updates `_ainvoke_loop_native_tools` to call them.

## Why it's needed

When the async native tool path (`_ainvoke_loop_native_tools`) calls the sync `_handle_native_tool_calls`, which calls `tool.run()` → `asyncio.run()`, the `asyncio.run()` call crashes with `RuntimeError: asyncio.run() cannot be called from a running event loop` when the agent is invoked from an already-running event loop (e.g. via `ainvoke()`).

The ReAct executor (`_ainvoke_loop_react`) already handles this correctly via `aexecute_tool_and_check_finality()` → `tool_usage.ause()` → `await`.

## What changed

1. **`_ahandle_native_tool_calls()`** — async variant that uses `asyncio.gather` instead of `ThreadPoolExecutor` for parallel tool execution, so async tools are properly awaited rather than run through `asyncio.run()`.

2. **`_aexecute_single_native_tool_call()`** — async variant that calls `await tool.arun()` instead of `tool.run()`, avoiding the nested `asyncio.run()` crash.

3. **`_ainvoke_loop_native_tools()`** now calls `_ahandle_native_tool_calls()` instead of `_handle_native_tool_calls()`.

### How to verify

1. Create a crew with an async tool (a tool whose `_run` method is a coroutine)
2. Call `crew.kickoff_async()` with `function_calling_llm` set to a model that supports native function calling
3. Before the fix: `RuntimeError: asyncio.run() cannot be called from a running event loop`
4. After the fix: the tool executes successfully

### Tested on

| OS | Status |
| :-- | :----: |
| macOS | ✅ |
| Windows | ⚠️ |
| Linux | ⚠️ |

## Risk & Scope

- **Main risk or tradeoff**: The async path now uses `asyncio.gather` instead of `ThreadPoolExecutor` for parallel execution. This is the correct approach for async code but means parallel tool execution is cooperative rather than preemptive.
- **Not validated / out of scope**: The experimental `AgentExecutor` (Flow-based) has the same bug but is in a separate code path.
- **Breaking changes / migration notes**: None. The sync path is unchanged.

## Linked Issues

Fixes #7630 (supersedes #6611, which was stale-closed as not-planned; same bug — confirmed still present on current main)

PR #6684, opened 2026-07-27
fix(experimental): use async LLM calls in AgentExecutor

### LINKED ISSUE #6683: [BUG] AgentExecutor.ainvoke routes LLM requests through LLM.call
### Description

`AgentExecutor.ainvoke()` runs through `Flow.kickoff_async()`, while the ReAct and native-tool LLM router methods are synchronous and call `get_llm_response()`. Flow offloads these synchronous router methods to the default executor, so each in-flight provider request occupies a worker thread and routes through `LLM.call()`.

This limits async concurrency to the executor's worker capacity even when the provider implements `LLM.acall()`.

### Steps to Reproduce

1. Create a `BaseLLM` test double that records calls to `call()` and `acall()`.
2. Create an `Agent` with the test double.
3. Run `await agent.kickoff_async("probe")`.
4. Inspect the recorded method.

Current result:

```text
call
```

### Expected behavior

The async AgentExecutor path should await `aget_llm_response()` and route capable providers through `LLM.acall()`.

Custom LLMs that signal unavailable async support with `NotImplementedError` should execute `LLM.call()` in a worker thread to preserve compatibility.

### Evidence

The two shared LLM router methods in `crewai.experimental.agent_executor.AgentExecutor` currently call `get_llm_response()`:

- `call_llm_and_parse`
- `call_llm_native_tools`

`get_llm_response()` invokes `LLM.call()`, while `aget_llm_response()` invokes `LLM.acall()`.

### Possible Solution

Convert both LLM router methods to async methods and await `aget_llm_response()`. Add a narrow `NotImplementedError` fallback in `aget_llm_response()` using `asyncio.to_thread(llm.call, ...)`.

### Additional context

This issue covers experimental AgentExecutor LLM routing. Async native tool execution is tracked separately in #6616 and #6622.

suqinghen:
## Summary

- make the experimental `AgentExecutor` ReAct and native-tool LLM routers async
- route the main LLM request paths through `aget_llm_response()` and `LLM.acall()`
- preserve sync-only custom LLM support through `asyncio.to_thread(LLM.call)`
- offload synchronous request preparation, response validation, and RPM enforcement
- update provider, tool, cache, and telemetry fixtures for async execution
- avoid an event-loop deadlock in the synchronous CrewAI Files upload-cache bridge

Fixes #6683

## Root cause

`AgentExecutor.ainvoke()` uses `Flow.kickoff_async()`, while both LLM router methods called the synchronous `get_llm_response()`. Provider requests therefore occupied default-executor workers for their full duration and bypassed provider-native async clients.

Moving the routers to `LLM.acall()` also exposed two existing compatibility assumptions:

1. several tests and custom doubles only configured `LLM.call()` or synchronous provider clients
2. the CrewAI Files synchronous upload-cache bridge waited on a coroutine scheduled onto the same running event loop, which could deadlock

## Behavior

Providers with async support now run through `LLM.acall()`. Custom providers that raise `NotImplementedError` from `acall()` continue through a worker-thread `LLM.call()` fallback.

Request preparation and response finalization run in worker threads so synchronous hooks and validation do not block the event loop. ReAct parsing, native-tool response handling, structured responses, callbacks, and error routing retain their existing semantics.

The forced-final-answer path used after maximum iterations continues to use the existing synchronous `LLM.call()` behavior.

## Scope

This change covers the main LLM routing paths in `crewai.experimental.AgentExecutor`, their provider test fixtures, and the CrewAI Files upload-cache bridge reached by async multimodal execution.

Async native tool execution remains covered by #6616 and #6622.

## Test plan

- focused async executor, provider, tool, telemetry, multimodal, and upload-cache suite: `728 passed, 13 skipped`
- CrewAI Files suite: `168 passed, 4 skipped`
- full workspace integration run on Python 3.13 after applying separately scoped current-main CLI and deterministic-fixture fixes: `5970 passed, 70 skipped`
- Ruff checks on changed production and test files
- `git diff --check`


suqinghen:
@coderabbitai review

lorenzejay:
love this pr!!

lib/crewai/src/crewai/experimental/agent_executor.py
```diff
@@ -1416,7 +1418,7 @@ def call_llm_and_parse(self) -> Literal["parsed", "parser_error", "context_error
                 None if self.original_tools else self.response_model
             )
 
-            answer = get_llm_response(
+            answer = await aget_llm_response(
```
lorenzejay, line 1431: This moves more than the provider call onto the event loop: `aget_llm_response()` runs `_prepare_llm_call()` and `_validate_and_finalize_llm_response()` synchronously, and those phases execute the before/after LLM hooks. Because Flow now awaits this router directly, a hook doing blocking I/O or `time.sleep()` can stall unrelated async agents. Could we offload the hook phases (or support async hooks) and add an event-loop responsiveness test?

lorenzejay (commented):
Two async-path concerns worth addressing before merge:

lib/crewai/src/crewai/experimental/agent_executor.py
```diff
@@ -1401,7 +1401,9 @@ def ensure_force_final_answer(self) -> Literal["agent_finished"]:
         return "agent_finished"
 
     @router("continue_reasoning")
-    def call_llm_and_parse(self) -> Literal["parsed", "parser_error", "context_error"]:
+    async def call_llm_and_parse(
```
lorenzejay, line 1414: `enforce_rpm_limit()` is still synchronous, but this router is now awaited directly by Flow. `RPMController.check_or_wait()` can call `time.sleep(60)` when the limit is reached, so that wait will freeze the event loop rather than occupy a worker. Could we offload this with `await asyncio.to_thread(...)` or introduce a non-blocking async limiter? The same issue applies to `call_llm_native_tools` below.

suqinghen: I've wrapped `enforce_rpm_limit` with `asyncio.to_thread` in a follow-up commit. A non-blocking async limiter is a better idea, should I open a new PR to address this, or is it fine to keep committing to this one?

lorenzejay (approved):

suqinghen:
Refactored associated unit tests to fix mismatches between test cases and the actual code logic.

lorenzejay (approved):

PR #6687, opened 2026-07-27
feat(tools): add Mengram memory tools (search, save, learned procedures)

alibaizhanov:
## Summary

Adds three memory tools backed by [Mengram](https://mengram.io) — a memory layer with semantic (facts), episodic (events), and **procedural** memory (workflows that learn from successes and failures):

- `MengramSearchTool` — semantic search over the user's long-term memory
- `MengramSaveTool` — save durable facts/preferences/decisions
- `MengramProceduresTool` — retrieve learned workflows with a success/failure track record and *preconditions* (assumptions violated in past failures), so agents stop repeating mistakes they already made

Follows existing tool conventions (Browserbase-style): lazy vendor import, `package_dependencies`, `env_vars` with `MENGRAM_API_KEY`, optional-dependency extra `mengram`, per-tool README, registered in tool/package `__init__` and `__all__`. Multi-user isolation via `user_id` — one API key, isolated memory per end-user.

Underlying client: [`mengram-ai`](https://pypi.org/project/mengram-ai/) (Apache 2.0 core, free tier, self-hostable). Tool logic is unit-tested downstream against crewai 1.15.7 in [our repo](https://github.com/alibaizhanov/mengram/tree/main/integrations/crewai).

Happy to adjust naming, category, or add a docs page under `docs/edge/en/tools/` if you'd like.


alibaizhanov:
Thanks for the review — addressed the substantive ones in the latest commits:

- **API key at import time** → now resolved at call time (explicit `api_key` wins, else `MENGRAM_API_KEY` read when the tool runs), so it picks up the environment as configured.
- **Facts-only search** → `MengramSearchTool` now uses `search_all` and returns semantic + episodic + procedural results.
- **Preconditions extraction** → added the top-level `preconditions` fallback alongside `metadata.preconditions`.
- **`__all__`** → exported the three tools in `tools/__init__.py`.
- **Docstrings** → added to the three schemas and three tool classes.

On the `user_id` default: kept `"default"` to match every other tool's single-user ergonomics (and the other Mengram SDKs) — the README already documents passing `user_id` for per-end-user isolation in multi-tenant apps. Happy to make it required if the maintainers prefer that convention.

PR #6696, opened 2026-07-28
fix: #6430 - Replace bare raise with explicit ToolUsageError in tool_usage.py

Diwak4r:
## Summary
Fixes #6430 - Replaces bare `raise` statement with explicit `ToolUsageError` in `ToolUsage._original_tool_calling` method.

## Changes
- **lib/crewai/src/crewai/tools/tool_usage.py**: Fixed bare `raise` at line ~853 that caused `RuntimeError: no active exception to reraise` when `raise_error=True` and an exception occurred during tool calling.

## Root Cause
The bare `raise` statement (line ~853) was only valid inside an `except` block, but was executed outside one. When `raise_error=True` and an exception occurred, Python raised `RuntimeError` instead of the intended `ToolUsageError`.

## Fix
Replaced bare `raise` with explicit `raise ToolUsageError(str(e)) from e` to match the pattern already used in the `raise_error=False` branch.

## Testing
- Verified fix doesn't break existing tool calling functionality
- The fix follows the same error handling pattern established in the `raise_error=False` branch

### DIFF
--- a/lib/crewai/src/crewai/tools/tool_usage.py
+++ b/lib/crewai/src/crewai/tools/tool_usage.py
@@ -884,12 +884,12 @@ class ToolUsage:
 
         except Exception:
             if raise_error:
-                raise
+                raise ToolUsageError(f"{I18N_DEFAULT.errors('tool_arguments_error')}")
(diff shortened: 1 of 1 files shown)


Diwak4r:

PR #6697, opened 2026-07-28
fix: #6481 - Add async support for before/after_kickoff_callbacks in akickoff

### LINKED ISSUE #6481: [BUG] before/after_kickoff_callbacks do not support async callables in akickoff

Diwak4r:
## Summary
Fixes #6481 - Adds async support for `before_kickoff_callbacks` and `after_kickoff_callbacks` in `Crew.akickoff()` method.

## Changes
- **lib/crewai/src/crewai/crew.py**: Added `inspect.isawaitable()` checks for both `before_kickoff_callbacks` (via new `aprepare_kickoff`) and `after_kickoff_callbacks` to properly await async callables
- **lib/crewai/src/crewai/crews/utils.py**: Added async `aprepare_kickoff()` function with async callback support for `before_kickoff_callbacks`
- **lib/crewai/tests/crew/test_async_callbacks.py**: Added 4 comprehensive tests:
  - `test_akickoff_calls_async_before_callback` - verifies async before callback is awaited
  - `test_akickoff_calls_async_after_callback` - verifies async after callback is awaited
  - `test_akickoff_mixed_sync_and_async_callbacks` - verifies mixed sync/async callbacks work together
  - `test_akickoff_empty_callbacks` - verifies empty callbacks still function normally

## Root Cause
The `akickoff` method only awaited `task_callbacks` but not `before_kickoff_callbacks` or `after_kickoff_callbacks`, causing async callbacks to not execute properly (coroutines were never awaited).

## Fix
Applied the same `inspect.isawaitable()` pattern used for `task_callbacks` to both kickoff callback types, and created an async version of `prepare_kickoff` for the async path.

## Testing
All 4 new tests pass + all 11 existing async crew tests pass (15 total). Only Windows teardown `PermissionError` observed (environmental, not code-related).

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import inspect
 from collections.abc import Callable, Sequence
 from concurrent.futures import Future
 from copy import copy as shallow_copy
@@ -1254,7 +1255,8 @@ class Crew(FlowTrackable, BaseModel):
 
         runtime_scope = crewai_event_bus._enter_runtime_scope()
         try:
-            inputs = prepare_kickoff(self, inputs, input_files)
+            from crewai.crews.utils import aprepare_kickoff
+            inputs = await aprepare_kickoff(self, inputs, input_files)
 
             if self.process == Process.sequential:
                 result = await self._arun_sequential_process()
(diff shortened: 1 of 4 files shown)


Diwak4r:

Issue #6704, opened 2026-07-28
[BUG] kickoff_for_each_async / akickoff_for_each wipe replay data on every run


bardock-2393:
## Description

`Crew.kickoff_for_each_async` and `Crew.akickoff_for_each` both delegate to the shared helper `run_for_each_async` in `lib/crewai/src/crewai/crews/utils.py`. After gathering results from all input copies, that helper unconditionally resets the shared replay store:

https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/src/crewai/crews/utils.py#L500-L514

```python
    results = await asyncio.gather(*async_tasks)

    total_usage_metrics = UsageMetrics()
    for crew_copy in crew_copies:
        if crew_copy.usage_metrics:
            total_usage_metrics.add_usage_metrics(crew_copy.usage_metrics)
    crew.usage_metrics = total_usage_metrics

    crew._task_output_handler.reset()
    return list(results)
```

Each crew copy already resets and repopulates the shared `TaskOutputStorageHandler` (backed by a single SQLite file) as part of its own `kickoff_async`/`akickoff` run. The `crew._task_output_handler.reset()` call above then deletes whatever the copies just persisted, immediately before returning.

This is the async/native-async counterpart of #6650 (which covers the synchronous `kickoff_for_each`); the fix for #6650 only touches `Crew.kickoff_for_each` in `crew.py` and does not touch `run_for_each_async`, so this path is still broken.

## Steps to Reproduce

```python
import asyncio
from unittest.mock import patch
from crewai import Agent, Crew, Task, Process
from crewai.tasks.task_output import TaskOutput

agent = Agent(role="Researcher", goal="Research a topic.", backstory="You are a careful researcher.")
task = Task(description="Research {topic}.", expected_output="A concise research note.", agent=agent)
crew = Crew(agents=[agent], tasks=[task], process=Process.sequential)

first_output = TaskOutput(description="Research first.", raw="first result", agent="Researcher")
latest_output = TaskOutput(description="Research latest.", raw="latest result", agent="Researcher")

async def main():
    with patch.object(Task, "execute_sync", side_effect=[first_output, latest_output]):
        results = await crew.kickoff_for_each_async(inputs=[{"topic": "first"}, {"topic": "latest"}])
    print("stored:", crew._task_output_handler.load())

asyncio.run(main())
```

Observed on unmodified `main`:
```
stored: []
```

## Expected behavior

[...]

Issue #6650, opened 2026-07-25
[BUG] kickoff_for_each clears latest replay records before returning


HarperZ9:
## Description

The documented synchronous, non-streaming `Crew.kickoff_for_each` replay contract is not preserved at current `main` / CrewAI 1.15.6 (`b3aaaab023a53a08db4d36c9430b3463f1efcc7d`).

The replay guide says `kickoff_for_each` supports replay from the most recent crew run:

- [`docs/edge/.../replay-tasks-from-latest-crew-kickoff.mdx`](https://github.com/crewAIInc/crewAI/blob/b3aaaab023a53a08db4d36c9430b3463f1efcc7d/docs/edge/en/learn/replay-tasks-from-latest-crew-kickoff.mdx#L10-L14)
- [`docs/v1.15.6/.../replay-tasks-from-latest-crew-kickoff.mdx`](https://github.com/crewAIInc/crewAI/blob/b3aaaab023a53a08db4d36c9430b3463f1efcc7d/docs/v1.15.6/en/learn/replay-tasks-from-latest-crew-kickoff.mdx#L10-L14)

Each copied kickoff already resets the shared latest-output store before executing and then persists its records. Because synchronous `kickoff_for_each` runs the copies sequentially, the final copy leaves the valid latest run. The parent method then performs another [`self._task_output_handler.reset()` immediately before returning](https://github.com/crewAIInc/crewAI/blob/b3aaaab023a53a08db4d36c9430b3463f1efcc7d/lib/crewai/src/crewai/crew.py#L1106-L1110), deleting those records.

A deterministic regression on unmodified source receives both expected batch outputs but observes `[]` in the replay store after return. An ordinary-`kickoff` replay control passes.

## Steps to Reproduce

1. Pin the repository and create its Python 3.12 environment:

   ```shell
   git clone https://github.com/crewAIInc/crewAI.git
   git -C crewAI checkout --detach b3aaaab023a53a08db4d36c9430b3463f1efcc7d
   cd crewAI
   uv sync --python 3.12 --frozen --group dev
   ```

2. Copy `test_kickoff_for_each_replay.py` from the public evidence repository into `lib/crewai/tests/`.

3. Run the regression on unmodified production code:

   ```shell
   ./.venv/Scripts/python.exe -m pytest -q -n 0 --tb=short --disable-warnings \
[...]

PR #6705, opened 2026-07-28
fix(crew): preserve replay data after kickoff_for_each_async

bardock-2393:
## Summary

Contributes to #6704 by preserving replay data after `kickoff_for_each_async()` / `akickoff_for_each()` complete.

Both methods share `run_for_each_async()`. Each crew copy resets and repopulates the shared replay store as part of its own run, but `run_for_each_async()` then reset that same shared storage again right after gathering all the copies' results, discarding whatever they had just persisted — every call left the store empty. Removing that trailing reset is the async counterpart of the fix for #6650, which only covered the synchronous `kickoff_for_each()`.

Because the copies run concurrently on separate threads against one shared SQLite-backed store, which copy's output survives isn't deterministic the way it is for the sequential `kickoff_for_each()` (guaranteed last-input-wins). This PR only fixes the unconditional wipe-to-empty; it doesn't attempt to add ordering guarantees for the concurrent case, which is a separate, harder problem.

## Testing

- `uv run pytest lib/crewai/tests/test_crew.py::test_kickoff_for_each_async_persists_output_for_replay -q`
- `uv run pytest lib/crewai/tests/test_crew.py -q` (132 passed, 1 skipped)
- `uv run pytest lib/crewai/tests/crew/ lib/crewai/tests/skills/ -q` (161 passed)
- `uv run ruff check lib/crewai/src/crewai/crews/utils.py lib/crewai/tests/test_crew.py`
- `uv run ruff format --check lib/crewai/src/crewai/crews/utils.py lib/crewai/tests/test_crew.py`
- `uv run mypy lib/crewai/src/crewai/crews/utils.py`

### DIFF
--- a/lib/crewai/src/crewai/crews/utils.py
+++ b/lib/crewai/src/crewai/crews/utils.py
@@ -510,5 +510,4 @@ async def run_for_each_async(
             total_usage_metrics.add_usage_metrics(crew_copy.usage_metrics)
     crew.usage_metrics = total_usage_metrics
 
-    crew._task_output_handler.reset()
     return list(results)
--- a/lib/crewai/tests/test_crew.py
+++ b/lib/crewai/tests/test_crew.py
@@ -1566,6 +1566,65 @@ async def test_async_kickoff_for_each_async_empty_input():
     assert results == [], "Result should be an empty list when input is empty"
 
 
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


bardock-2393:
Per CONTRIBUTING.md, this should carry the `llm-generated` label — I don't have permission to add it myself. Could a maintainer please apply it?

Issue #6706, opened 2026-07-28
[BUG] Flow dict state loses fields added after a checkpoint on restore


bardock-2393:
## Description

`Flow._restore_state` (`lib/crewai/src/crewai/flow/runtime/__init__.py`) is called whenever a flow resumes from persisted state, either via `kickoff(inputs={"id": "<uuid>"})` or via `restore_from_state_id`. For unstructured (dict) flow state, it does:

https://github.com/crewAIInc/crewAI/blob/main/lib/crewai/src/crewai/flow/runtime/__init__.py#L1798-L1800

```python
if isinstance(self._state, dict):
    self._state.clear()
    self._state.update(stored_state)
```

`self._state` is populated from the flow's *current* `initial_state` at construction (`_create_initial_state`), before the restore call runs. If a developer adds a new key to `initial_state` after some runs have already been checkpointed, resuming one of those older checkpoints calls `.clear()` first, wiping out the fresh default for the new key, then only restores whatever keys happen to exist in the old, smaller `stored_state` dict — so the new field silently disappears from state entirely (not even at its default) instead of keeping the default the fresh instance was constructed with.

The `BaseModel` branch right below doesn't have this problem: `model_validate(stored_state)` fills in any field missing from `stored_state` from the model's own defaults. Only the plain-dict branch loses data.

## Steps to Reproduce

```python
from crewai.flow.flow import Flow, start
from crewai.flow.persistence import persist
from crewai.flow.persistence.sqlite import SQLiteFlowPersistence

persistence = SQLiteFlowPersistence("/tmp/flows.db")

# "Before": flow state only has `counter`.
class FlowV1(Flow):
    initial_state = {"counter": 0}

    @start()
    @persist(persistence)
    def go(self):
        self.state["counter"] = 5

flow1 = FlowV1(persistence=persistence)
flow1.kickoff()
checkpoint_id = flow1.state["id"]

# "After": a developer adds `config_value` with a default.
class FlowV2(Flow):
    initial_state = {"counter": 0, "config_value": "important_default"}

    @start()
    @persist(persistence)
    def go(self):
        pass

flow2 = FlowV2(persistence=persistence)
flow2.kickoff(inputs={"id": checkpoint_id})
print(flow2.state)
```

Observed on unmodified `main`:
```
{'counter': 5, 'id': '<checkpoint_id>'}
```
`config_value` is missing entirely.

## Expected behavior

```
[...]

PR #6707, opened 2026-07-28
fix(flow): keep fields added after a checkpoint on dict state restore

bardock-2393:
## Summary

Contributes to #6706 by keeping dict-typed flow state fields that were added to `initial_state` after an older checkpoint was saved.

`_restore_state` cleared the dict state before applying a stored checkpoint. `self._state` is populated from the flow's *current* `initial_state` at construction, so any key added since an older checkpoint was saved existed on the fresh instance with its default value — until `.clear()` wiped it out, leaving it missing entirely instead of falling back to that default. The `BaseModel` branch right below doesn't have this problem, since `model_validate` already fills unset fields from the model's own defaults; this brings the dict branch to the same behavior by merging the stored values on top of the constructed defaults instead of replacing the state outright.

## Testing

- `uv run pytest lib/crewai/tests/test_flow_persistence.py::test_dict_state_restore_keeps_fields_added_after_the_checkpoint -q`
- `uv run pytest lib/crewai/tests/test_flow_persistence.py lib/crewai/tests/test_flow_persistence_factory.py -q` (18 passed)
- `uv run pytest lib/crewai/tests/ -k flow -q` (600 passed, 22 skipped)
- `uv run ruff check lib/crewai/src/crewai/flow/runtime/__init__.py lib/crewai/tests/test_flow_persistence.py`
- `uv run ruff format --check lib/crewai/src/crewai/flow/runtime/__init__.py lib/crewai/tests/test_flow_persistence.py`
- `uv run mypy lib/crewai/src/crewai/flow/runtime/__init__.py`

### DIFF
--- a/lib/crewai/src/crewai/flow/runtime/__init__.py
+++ b/lib/crewai/src/crewai/flow/runtime/__init__.py
@@ -1796,7 +1796,10 @@ class Flow(BaseModel, Generic[T], metaclass=FlowMeta):
             raise ValueError("Stored state must have an 'id' field")
 
         if isinstance(self._state, dict):
-            self._state.clear()
+            # Merge rather than replace: a field added to `initial_state` after
+            # `stored_state` was saved should keep its fresh default instead of
+            # disappearing, matching the BaseModel branch below (model_validate
+            # fills unset fields from the model's own defaults).
             self._state.update(stored_state)
         elif isinstance(self._state, BaseModel):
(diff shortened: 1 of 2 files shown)


bardock-2393:
Per CONTRIBUTING.md, this should carry the `llm-generated` label — I don't have permission to add it myself. Could a maintainer please apply it?

PR #6734, opened 2026-07-30
fix(llm): honour response_model in the async streaming handler

### LINKED ISSUE #6733: [BUG] response_model is ignored by acall() when streaming is enabled
### Description

`await llm.acall(messages, response_model=SomeModel)` returns the model's raw text instead
of the structured output when streaming is enabled. `_ahandle_streaming_response` accepts
`response_model`, documents it, and never uses it.

`LLM.acall` forwards the argument correctly:

https://github.com/crewAIInc/crewAI/blob/ceed4a3ff71b5b4cb0ca316b4178ffcce74a53b2/lib/crewai/src/crewai/llm.py#L2027-L2034

```python
                    if self._effective_stream():
                        return await self._ahandle_streaming_response(
                            params=params,
                            callbacks=callbacks,
                            available_functions=available_functions,
                            from_task=from_task,
                            from_agent=from_agent,
                            response_model=response_model,   # <-- arrives here
                        )
```

The handler declares the parameter at
[`llm.py:1533`](https://github.com/crewAIInc/crewAI/blob/ceed4a3ff71b5b4cb0ca316b4178ffcce74a53b2/lib/crewai/src/crewai/llm.py#L1533)
and documents it as "Optional response model", but the name never appears again in the
function body. The success path returns `full_response` — the accumulated stream text —
verbatim:

https://github.com/crewAIInc/crewAI/blob/ceed4a3ff71b5b4cb0ca316b4178ffcce74a53b2/lib/crewai/src/crewai/llm.py#L1690-L1705

The other three handlers all consume it and convert via `InternalInstructor`:

| handler | line | consumes `response_model`? |
| --- | --- | --- |
| `_handle_streaming_response` | `llm.py:1040` | yes |
| `_handle_non_streaming_response` | `llm.py:1242` | yes |
| `_ahandle_non_streaming_response` | `llm.py:1397` | yes |
| **`_ahandle_streaming_response`** | — | **no** |

### Reproduction

Replace `litellm.completion`/`acompletion` with a stub that streams a prose answer (not
JSON), replace `InternalInstructor` so no second LLM call is needed, then call all four
paths with the same `response_model`:

```python
class Landmark(BaseModel):
    city: str
    population: int

# What a model streams when it answers in prose. Deliberately NOT valid JSON.
PROSE = "The Eiffel Tower is in Paris, which has about 2,100,000 residents."

llm = LLM(model="gpt-4o", is_litellm=True, api_key="test", stream=stream)
out = llm.call(msgs, response_model=Landmark)          # or: await llm.acall(...)
```

Result on `main` (`ceed4a3`):

```
sync  stream=False -> structured  | {"city":"Paris","population":2100000}
sync  stream=True  -> structured  | {"city":"Paris","population":2100000}
async stream=False -> structured  | {"city":"Paris","population":2100000}
async stream=True  -> RAW TEXT    | The Eiffel Tower is in Paris, which has about 2,100,000 residents.
```

Three of four paths honour the contract. The fourth returns prose.

### Why this matters

`response_model` is the structured-output contract, so callers act on the return value
without re-checking it:

- **Silent corruption, not a loud failure.** The return type is `str` in both cases, so
  nothing raises at the boundary. Downstream `model_validate_json`/`json.loads` either
  throws far from the cause, or — worse — a prose answer that happens to contain
  JSON-looking text parses into the wrong object.
- **Reachable through normal configuration.** No malformed input and no unusual setup:
  `stream=True` plus `response_model` plus `acall` is the ordinary combination for a
  streaming async agent that wants typed output. Both are documented, supported options.
- **Streaming is exactly where async is used.** The path that silently drops the parameter
  is the one an async streaming application takes by default.
- **The sync/async asymmetry makes it a trap.** Code developed and tested synchronously
  behaves correctly, then quietly stops honouring `response_model` when moved to `acall`
  with no signature change and no warning.

### Expected behaviour

`await llm.acall(msgs, response_model=M)` with `stream=True` returns the same structured
JSON string as the other three paths.

### Fix

Mirror the block that `_handle_streaming_response` already has, in the no-tool-call return
path of `_ahandle_streaming_response`:

```python
            if response_model and self.is_litellm:
                instructor_instance = InternalInstructor(
                    content=full_response,
                    model=response_model,
                    llm=self,
                )
                result = instructor_instance.to_pydantic()
                structured_response = result.model_dump_json()
                self._handle_emit_call_events(
                    response=structured_response,
                    ...
                )
                return structured_response
```

The `self.is_litellm` guard and the `InternalInstructor` call match the sync handler
exactly, so behaviour is consistent across all four paths and native providers are
unaffected.

I have a PR ready with this change plus five tests (one regression, four controls covering
the three already-correct paths and the no-`response_model` case).

### Environment

- crewAI `main` @ `ceed4a3ff71b5b4cb0ca316b4178ffcce74a53b2`
- Python 3.12

LHMQ878:
Closes #6733

## Problem

`await llm.acall(messages, response_model=SomeModel)` with `stream=True` returns the model's
raw text instead of the structured output. `_ahandle_streaming_response` accepts
`response_model`, documents it as "Optional response model", and then never references it
again — the success path returns the accumulated `full_response` verbatim.

`LLM.acall` forwards the argument correctly (`llm.py:2033`), so the parameter arrives and is
dropped. The other three handlers all consume it:

| handler | consumes `response_model`? |
| --- | --- |
| `_handle_streaming_response` (`llm.py:1040`) | yes |
| `_handle_non_streaming_response` (`llm.py:1242`) | yes |
| `_ahandle_non_streaming_response` (`llm.py:1397`) | yes |
| **`_ahandle_streaming_response`** | **no** |

The failure is silent. Both the correct and the incorrect return value are `str`, so nothing
raises at the boundary: a prose answer reaches a caller that has already committed to a
typed contract, and the error surfaces later at a `model_validate_json` far from the cause —
or not at all, if the prose happens to parse into the wrong object.

It is also reachable through entirely ordinary configuration. No malformed input, no unusual
setup: `stream=True` + `response_model` + `acall` is the default combination for a streaming
async agent that wants typed output, and streaming is precisely where async gets used. Code
developed and tested synchronously works, then quietly stops honouring `response_model` when
moved to `acall` — same signature, no warning.

## Fix

Mirror the block the sync streaming handler already has, in the no-tool-call return path of
the async one:

```diff
             usage_dict = self._usage_to_dict(usage_info)
             finish_reason, response_id_last = (
                 stream_finish_reason,
                 stream_response_id,
             )
+
+            if response_model and self.is_litellm:
+                instructor_instance = InternalInstructor(
+                    content=full_response,
+                    model=response_model,
+                    llm=self,
+                )
+                result = instructor_instance.to_pydantic()
+                structured_response = result.model_dump_json()
+                self._handle_emit_call_events(
+                    response=structured_response,
+                    call_type=LLMCallType.LLM_CALL,
+                    from_task=from_task,
+                    from_agent=from_agent,
+                    messages=params.get("messages"),
+                    usage=usage_dict,
+                    finish_reason=finish_reason,
+                    response_id=response_id_last,
+                )
+                return structured_response
+
             self._handle_emit_call_events(
                 response=full_response,
```

The `self.is_litellm` guard, the `InternalInstructor` construction and the
`model_dump_json()` conversion match `_handle_streaming_response` exactly, so all four paths
now behave identically and native (non-LiteLLM) providers are untouched.

Placement matters: this sits *after* the `accumulated_tool_args` block, which already
returns for tool calls. That mirrors the sync handler, where the conversion lives under
`if not tool_calls or not available_functions:` — a tool call still takes precedence over
structured-output conversion, exactly as before.

The emitted `LLMCallType.LLM_CALL` event carries the structured response rather than the raw
text, again matching the sync handler, so observers see the same payload the caller receives.

## Evidence

Stubbing `litellm.completion`/`acompletion` to stream a prose answer (deliberately not valid
JSON) and calling all four paths with the same `response_model`:

| path | before | after |
| --- | --- | --- |
| `call(..., stream=False)` | `{"city":"Paris","population":2100000}` | unchanged |
| `call(..., stream=True)` | `{"city":"Paris","population":2100000}` | unchanged |
| `await acall(..., stream=False)` | `{"city":"Paris","population":2100000}` | unchanged |
| **`await acall(..., stream=True)`** | **`The Eiffel Tower is in Paris, which has about 2,100,000 residents.`** | `{"city":"Paris","population":2100000}` |

## Tests

New file `lib/crewai/tests/llms/litellm/test_litellm_response_model_parity.py`, 5 tests. No
cassettes — `litellm.completion`/`acompletion` and `InternalInstructor` are both replaced, so
nothing touches the network.

The streamed content is prose, not JSON, which is what makes the assertion meaningful: if a
handler ignores `response_model` the returned string is the prose itself, and
`assert result != PROSE` fails with a message naming the cause.

| test | role |
| --- | --- |
| `test_async_streaming_honours_response_model` | the regression |
| `test_sync_call_honours_response_model[non_streaming]` | control: already correct |
| `test_sync_call_honours_response_model[streaming]` | control: already correct |
| `test_async_non_streaming_honours_response_model` | control: already correct |
| `test_async_streaming_without_response_model_returns_text` | control: the fix is not an unconditional conversion — with no `response_model` the raw text is still returned |

### Control experiment

Reverting `llm.py` and keeping the tests:

```
1 failed, 4 passed
FAILED test_async_streaming_honours_response_model
E   AssertionError: response_model was ignored: raw model text returned
E   assert 'The Eiffel Tower is in Paris, which has about 2,100,000 residents.' != 'The Eiffel Tower is in Paris, which has about 2,100,000 residents.'
```

Exactly the one regression test fails; all four controls pass. The last control is the one
that matters most — it rules out the fix being a blunt "always convert".

### Suite runs

`lib/crewai/tests/llms/litellm/`, `test_tool_call_streaming.py`, `openai_compatible/`:

| | result |
| --- | --- |
| clean `main` (new file ignored) | 1 failed, 42 passed, 9 skipped |
| this branch | 1 failed, **47** passed, 9 skipped |

Same single failure on both sides —
`TestAnthropicToolCallStreaming::test_anthropic_streaming_emits_tool_call_events`, which
fails identically on unmodified `main` and is unrelated to this change. +5 passing are the
new tests.

- `ruff format --check` / `ruff check` on `llm.py` — clean
- `mypy lib/crewai/src/crewai/llm.py` — `Success: no issues found`

One note on running these locally: `addopts` carries `--block-network`, and on Windows
`asyncio.run()` builds a `ProactorEventLoop` whose `_make_self_pipe()` calls
`socket.socketpair()`, which falls back to a real localhost `connect()` and trips
`pytest_recording`'s guard. I ran with `--allowed-hosts=127.0.0.1,::1` to let the event
loop's own self-pipe through. This is a Windows-only artifact — on Linux `socketpair()` is a
real syscall that never touches `connect`, so CI needs no flag.

## Duplicate check

Searched for `response_model stream`, `_ahandle_streaming_response`, `acall response_model`,
and `response_model async streaming` in titles. Two open PRs touch adjacent code but not
this defect:

- **#4819** adds a `not params.get("tools")` guard to the *existing* `response_model` blocks
  — it narrows when conversion happens on the paths that already have it, and does not add
  the missing block.
- **#6272** changes error visibility in `_ahandle_streaming_response`; a 4-line diff with no
  `response_model` involvement.

No issue or PR reports the async streaming path dropping `response_model`.

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1692,6 +1692,27 @@ class LLM(BaseLLM):
                 stream_finish_reason,
                 stream_response_id,
             )
+
+            if response_model and self.is_litellm:
+                instructor_instance = InternalInstructor(
+                    content=full_response,
+                    model=response_model,
+                    llm=self,
+                )
+                result = instructor_instance.to_pydantic()
+                structured_response = result.model_dump_json()
+                self._handle_emit_call_events(
+                    response=structured_response,
+                    call_type=LLMCallType.LLM_CALL,
+                    from_task=from_task,
+                    from_agent=from_agent,
+                    messages=params.get("messages"),
+                    usage=usage_dict,
+                    finish_reason=finish_reason,
+                    response_id=response_id_last,
+                )
+                return structured_response
+
             self._handle_emit_call_events(
                 response=full_response,
                 call_type=LLMCallType.LLM_CALL,
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -1659,6 +1659,7 @@ class LLM(BaseLLM):
             if usage_info:
                 self._track_token_usage_internal(usage_info)
 
+            tool_calls_unhandled = False
             if accumulated_tool_args:
                 tool_calls_list: list[ChatCompletionDeltaToolCall] = [
                     ChatCompletionDeltaToolCall(
@@ -1684,6 +1685,7 @@ class LLM(BaseLLM):
                         )
                         if result is not None:
                             return result
+                        tool_calls_unhandled = True
                     else:
                         return tool_calls_list
 
@@ -1693,7 +1695,12 @@ class LLM(BaseLLM):
                 stream_response_id,
             )
 
-            if response_model and self.is_litellm:
+            # A tool call the helper declined to handle — an unknown tool, or
+            # arguments it could not parse — leaves the accumulated content as
+            # whatever preceded it, which is usually nothing. Converting that would
+            # invent a structured answer the model never gave, so the tool-call path
+            # keeps precedence here exactly as it does in _handle_streaming_response.
+            if response_model and self.is_litellm and not tool_calls_unhandled:
                 instructor_instance = InternalInstructor(
                     content=full_response,
                     model=response_model,
--- a/lib/crewai/tests/llms/litellm/test_litellm_response_model_parity.py
+++ b/lib/crewai/tests/llms/litellm/test_litellm_response_model_parity.py
@@ -1,0 +1,158 @@
+"""response_model must be honoured on all four LiteLLM completion paths.
+
+The sync/async and streaming/non-streaming handlers each receive ``response_model``
+from ``LLM.call``/``LLM.acall``. These tests pin that every one of them actually
+converts the model output, rather than returning the raw text.
+
+No cassettes: ``litellm.completion``/``acompletion`` and ``InternalInstructor`` are
+both replaced, so nothing here touches the network.
+"""
+
+import json
+from typing import Any
+
+from litellm.types.utils import (
+    Delta as LiteLLMDelta,
+    ModelResponseStream,
+    StreamingChoices as LiteLLMStreamingChoices,
+)
+import pytest
+from pydantic import BaseModel
+
+from crewai import llm as llm_module
+from crewai.llm import LLM
+
+
+class Landmark(BaseModel):
+    """Target schema for the structured-output conversion."""
+
+    city: str
+    population: int
+
+
+# What a model streams when it answers in prose rather than JSON. Deliberately not
+# valid JSON: if a handler returns this verbatim, response_model was ignored.
+PROSE = "The Eiffel Tower is in Paris, which has about 2,100,000 residents."
+STRUCTURED = '{"city": "Paris", "population": 2100000}'
+
+
+class _FakeInstructor:
+    """Stand-in for InternalInstructor that converts without a second LLM call."""
+
+    def __init__(self, content: Any = None, model: Any = None, llm: Any = None, **_: Any):
+        self.model = model
+
+    def to_pydantic(self) -> BaseModel:
+        return self.model.model_validate_json(STRUCTURED)
+
+
+def _stream_chunk(content: str, finish: str | None = None) -> ModelResponseStream:
+    """A real litellm streaming chunk; the handlers type-check against these."""
+    return ModelResponseStream(
+        id="chunk-1",
+        choices=[
+            LiteLLMStreamingChoices(
+                index=0,
+                delta=LiteLLMDelta(content=content),
+                finish_reason=finish,
+            )
+        ],
+    )
+
+
+class _Message:
+    def __init__(self, content: str) -> None:
+        self.content = content
+        self.tool_calls = None
+
+
+class _Choice:
+    def __init__(self, content: str) -> None:
+        self.message = _Message(content)
+        self.finish_reason = "stop"
+
+
+class _Response:
+    """Minimal non-streaming response shape, accessed both ways by the handlers."""
+
+    def __init__(self, content: str) -> None:
+        self.choices = [_Choice(content)]
+        self.id = "resp-1"
+        self.usage = None
+
+    def __getitem__(self, key: str) -> Any:
+        return {"choices": [{"message": {"content": PROSE}}]}[key]
+
+
+@pytest.fixture
+def fake_litellm(monkeypatch: pytest.MonkeyPatch) -> None:
+    """Serve PROSE from every litellm entry point, sync and async, stream and not."""
+
+    def _chunks():
+        yield _stream_chunk(PROSE[:20])
+        yield _stream_chunk(PROSE[20:], finish="stop")
+
+    async def _achunks():
+        yield _stream_chunk(PROSE[:20])
+        yield _stream_chunk(PROSE[20:], finish="stop")
+
+    import litellm
+
+    monkeypatch.setattr(
+        litellm,
+        "completion",
+        lambda **kw: _chunks() if kw.get("stream") else _Response(PROSE),
+    )
+
+    async def _acompletion(**kw: Any) -> Any:
+        return _achunks() if kw.get("stream") else _Response(PROSE)
+
+    monkeypatch.setattr(litellm, "acompletion", _acompletion)
+    monkeypatch.setattr(llm_module, "InternalInstructor", _FakeInstructor)
+    monkeypatch.setattr(
+        "crewai.utilities.internal_instructor.InternalInstructor", _FakeInstructor
+    )
+
+
+def _assert_structured(result: Any) -> None:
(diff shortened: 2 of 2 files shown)


LHMQ878:
Thanks — I checked all three against the code and against the synchronous handler before changing anything. **One is real and is fixed in `a7cd57e`. One I'm not taking, and I'd rather explain why than skip it silently. One was already true, but the assertion was worth adding.**

### 1. Tool-call precedence — real, fixed

Confirmed by probing both handlers with the same input: a tool call naming a tool absent from `available_functions`, `response_model=Landmark`.

```
[async, before]  returned '{"city":"Paris","population":2100000}'   <-- fabricated
[sync]           returned ''
[async, after]   returned ''
```

The async branch invented a structured answer from an empty `full_response`, because a chunk carrying `tool_calls` carries no content. Since aligning the two handlers is the entire point of this PR, that divergence mattered more than the fix being small. Gated the same way `_handle_streaming_response` gates it, via a `tool_calls_unhandled` flag set when `_handle_streaming_tool_calls` returns `None`.

### 2. Re-raising conversion failures — not taking this one

The premise is right: `to_pydantic()` failures do get swallowed by the enclosing `except Exception`, which returns the raw `full_response` when `chunk_count > 0`. But **the synchronous handler behaves identically**, and I checked rather than assumed:

```
[F2-async] NO RAISE, returned 'The Eiffel Tower is in Paris, which has about 2,100,000 residents.'
[F2-sync]  NO RAISE, returned 'The Eiffel Tower is in Paris, which has about 2,100,000 residents.'
```

`llm.py:1040`'s conversion sits inside the same `try`, and `llm.py:1093`'s handler returns `full_response` whenever `full_response.strip()` is non-empty. So re-raising in the async path alone would *create* a sync/async divergence in a PR whose only purpose is to remove one — a caller would get structured output or raw text from `call()`, and an exception from `acall()`, for the same failure.

If swallowed conversion errors are worth fixing — and I think they are, silently returning prose where a `BaseModel`-shaped string was asked for is bad — it should be one change to both handlers, argued on its own merits. I'd rather open that separately than smuggle a behaviour change into a parity fix. Happy to file it.

### 3. Asserting on the emitted event — was already correct, now pinned

Spying on `_handle_emit_call_events` showed the payload was already the structured JSON, not `PROSE`, so this was not a bug. But nothing was *checking* it, and the return value alone doesn't: converting only the return value would leave traces and token accounting reporting the raw prose. Added `test_async_streaming_emits_the_structured_response`, which spies on `_handle_emit_call_events` and asserts `PROSE not in emitted` alongside the existing returned-value assertion.

### Verification

**7 passed** in `test_litellm_response_model_parity.py` (was 5), 7 passed + 9 skipped across `lib/crewai/tests/llms/litellm/`.

Both new tests are load-bearing, and they fail for different reasons — I checked by reverting each half separately:

```
revert source entirely      -> 2 failed, 5 passed
revert only the F1 gate     -> 1 failed, 6 passed   (the parity test)
```

`lib/crewai/tests/llms/test_tool_call_streaming.py` — the file most exposed to the gate change — gives 7 passed, 1 failed, and that failure is pre-existing on `origin/main` at the same commit: `ImportError: Azure AI Inference native provider not available, to install: uv add "crewai[azure-ai-inference]"`.

Issue #6529, opened 2026-07-13
after_llm_call hook prevents agents from executing tools


lucasgomide:
### Description

When an `after_llm_call` hook is registered, an agent that would otherwise call a tool stops executing that tool. The agent finishes as if the model's tool call were the final answer. Removing the hook restores normal tool execution.

### Reproduction

Requires an OpenAI API key (uses a real LLM call).

```python
from crewai import Agent
from crewai.hooks import after_llm_call
from crewai.tools import tool

executed = []


@tool("calculate_sum")
def calculate_sum(a: int, b: int) -> int:
    """Return the sum of two integers a and b."""
    executed.append((a, b))
    return a + b


@after_llm_call
def observe(context):
    return None  # observe only, no modification


agent = Agent(
    role="Math Assistant",
    goal="Perform calculations using tools",
    backstory="You always use the calculate_sum tool for addition.",
    tools=[calculate_sum],
    verbose=True,
)

result = agent.kickoff(
    messages="What is 5 + 3? Use the calculate_sum tool, then report the number."
)
print("tool executed:", bool(executed))
```

To see the contrast, comment out the `@after_llm_call def observe(...)` hook and run again.

### Expected

`tool executed: True` — the `calculate_sum` tool runs, exactly as it does when no hook is registered.

### Actual

`tool executed: False` — with the `after_llm_call` hook registered, the tool is never executed and the agent returns the model's raw tool call as its result. Removing the hook makes the tool execute normally.

PR #6737, opened 2026-07-30
fix: run after_llm_call hooks on async provider paths

### LINKED ISSUE #6736: after_llm_call hooks never run on acall(): eight native provider handlers skip _invoke_after_llm_call_hooks
### Description

`after_llm_call` hooks never run on `acall()`. The native providers invoke
`_invoke_after_llm_call_hooks` only from their **sync** handlers, so a hook registered to
redact, rewrite or audit a response is silently skipped for every async direct LLM call — the
unmodified response reaches the caller and nothing is logged.

The gap is per-handler, not per-provider. Coverage on `main` (`3266932`):

| provider | handler | sync | async |
| --- | --- | --- | --- |
| openai | `_handle_responses` / `_ahandle_responses` | yes (`:1068`) | **no** |
| openai | `_handle_streaming_responses` / `_ahandle_streaming_responses` | yes (`:1376`) | **no** |
| openai | `_handle_completion` / `_ahandle_completion` | yes (`:2008`) | **no** |
| openai | `_handle_streaming_completion` / `_ahandle_streaming_completion` | yes (`:2298`) | **no** |
| anthropic | `_handle_completion` / `_ahandle_completion` | yes (`:1104`) | **no** |
| anthropic | `_handle_streaming_completion` / `_ahandle_streaming_completion` | yes (`:1307`) | **no** |
| bedrock | `_handle_converse` / `_ahandle_converse` | yes (`:823`) | **no** |
| bedrock | `_handle_streaming_converse` / `_ahandle_streaming_converse` | **no** | yes (`:1772`) |

The bedrock row is the giveaway: within one file, `_ahandle_streaming_converse` calls the hook
and `_ahandle_converse` does not, and the sync/async polarity is inverted relative to every
other row. That is not a design decision about async — it is eight handlers each missing the
call their twin makes.

### Reproduction

No network. Stub the SDK client and compare `call()` with `acall()`:

```python
from crewai.hooks.llm_hooks import register_after_llm_call_hook

calls = []
def hook(ctx):
    calls.append(ctx.response)
    return "REDACTED BY HOOK"
register_after_llm_call_hook(hook)

llm = AnthropicCompletion(model="claude-sonnet-4-5", thinking={"type": "enabled", "budget_tokens": 1024},
                          max_tokens=2048, stream=False)
llm._client       = SimpleNamespace(messages=SyncStub())    # returns one thinking + one text block
llm._async_client = SimpleNamespace(messages=AsyncStub())   # same payload

print(llm.call("hi"))
print(asyncio.run(llm.acall("hi")))
```

On `main`:

```
anthropic sync : 'REDACTED BY HOOK'                     hook fired: 1
anthropic async: 'the model said something sensitive'   hook fired: 0
openai    sync : 'REDACTED BY HOOK'                     hook fired: 1
openai    async: 'the model said something sensitive'   hook fired: 0
```

The hook is never even invoked on the async path, so a hook that only observes (logging,
auditing, metrics) also silently records nothing.

### Second defect in the same code: anthropic loses thinking blocks on the async path

While mapping the above I found `_ahandle_completion` and `_ahandle_streaming_completion` never
call `_extract_thinking_block`, so `_previous_thinking_blocks` stays empty on the async path.
The same probe:

```
anthropic sync  thinking kept: [{'type': 'thinking', 'thinking': 'internal reasoning', 'signature': 'sig-abc'}]
anthropic async thinking kept: []
```

`_format_messages_for_anthropic:851` replays `self._previous_thinking_blocks` into the
assistant message on the next turn when `thinking` is enabled. With the async path leaving that
list empty, the signed thinking block is dropped and the follow-up request loses the reasoning
context the sync path preserves. `test_anthropic_thinking_blocks_preserved_across_turns`
(`tests/llms/anthropic/test_anthropic.py:633`) pins this behaviour precisely — for `call()` only.

### Why this matters

`after_llm_call` is the documented interception point for exactly the things that must not be
skippable: redacting PII before a response is stored or displayed, enforcing an output policy,
rewriting a response, and audit logging. A guardrail that works under `call()` and silently
stops working under `acall()` is worse than one that never worked, because the failure is
invisible: no exception, no warning, and the response looks plausible.

Async is not an edge path — `acall()` is what `kickoff_async`, async flows and any concurrent
usage reach. Moving a working crew from sync to async silently disables every registered
`after_llm_call` hook.

### Expected behaviour

`acall()` runs `after_llm_call` hooks exactly as `call()` does, and anthropic retains thinking
blocks on both paths.

### Fix

Add the missing `_invoke_after_llm_call_hooks` call to each of the eight handlers, mirroring the
placement its twin already uses (after `_emit_call_completed_event`, and behind the same
`isinstance(result, str)` guard the sync streaming handler uses so structured tool-call payloads
are not stringified), plus the two `_extract_thinking_block` loops in anthropic's async handlers.

`_invoke_after_llm_call_hooks` already returns the response unchanged when `from_agent` is not
`None`, so the agent-driven path (which dispatches `POST_MODEL_CALL` separately via
`agent_utils._setup_after_llm_call_hooks`) is unaffected — no double dispatch.

Note on scope: **azure** and **gemini** call the hook on *neither* path. That is a symmetric
gap — the provider does not support the hook at all — rather than sync/async drift, so I left it
out of this fix; it needs a separate decision about whether those providers should support
direct-call hooks. Happy to follow up if wanted.

I have a PR ready with the fix plus 11 tests (sync baselines, async regressions, and negative
controls asserting an unhooked response passes through untouched).

### Environment

- crewAI `main` @ `3266932`
- Python 3.12

LHMQ878:
## Summary

Fixes #6736.

`after_llm_call` hooks never ran on `acall()`. The native providers invoked
`_invoke_after_llm_call_hooks` only from their **sync** handlers, so a hook registered to redact,
rewrite or audit a response was silently skipped for every async direct call — the unmodified
response reached the caller and nothing was logged.

Eight handlers were missing the call their twin makes:

| provider | handler | before | after |
| --- | --- | --- | --- |
| openai | `_ahandle_responses` | no (sync had it at `:1068`) | yes |
| openai | `_ahandle_streaming_responses` | no (sync at `:1376`) | yes |
| openai | `_ahandle_completion` | no (sync at `:2008`) | yes |
| openai | `_ahandle_streaming_completion` | no (sync at `:2298`) | yes |
| anthropic | `_ahandle_completion` | no (sync at `:1104`) | yes |
| anthropic | `_ahandle_streaming_completion` | no (sync at `:1307`) | yes |
| bedrock | `_ahandle_converse` | no (sync at `:823`) | yes |
| bedrock | `_handle_streaming_converse` | no (**async** had it at `:1772`) | yes |

That last row is what shows this is drift rather than a decision about async: within one file
`_ahandle_streaming_converse` calls the hook and its sync twin does not, with the polarity
inverted relative to every other row.

Each call is placed where its twin already places it — after `_emit_call_completed_event` — and
openai's async streaming handler keeps the same `isinstance(result, str)` guard the sync one uses,
so structured tool-call payloads are not stringified (#6529).

`_invoke_after_llm_call_hooks` returns the response unchanged when `from_agent` is not `None`, so
the agent-driven path — which dispatches `POST_MODEL_CALL` separately via
`agent_utils._setup_after_llm_call_hooks` — is unaffected. No double dispatch.

### Second defect in the same code

anthropic's `_ahandle_completion` and `_ahandle_streaming_completion` never called
`_extract_thinking_block`, so `_previous_thinking_blocks` stayed empty on the async path.
`_format_messages_for_anthropic:851` replays that list into the assistant message on the next
turn when `thinking` is enabled, so the async path dropped the signed thinking block that the
sync path preserves — losing the reasoning context across turns.
`test_anthropic_thinking_blocks_preserved_across_turns` pins this for `call()` only.

### Why this is worth fixing

`after_llm_call` is the documented interception point for the things that must not be skippable:
redacting PII before a response is stored or displayed, enforcing an output policy, and audit
logging. A guardrail that works under `call()` and silently stops working under `acall()` is
worse than one that never worked, because the failure is invisible — no exception, no warning,
and the response looks plausible.

`acall()` is what `kickoff_async`, async flows and any concurrent usage reach, so moving a
working crew from sync to async silently disabled every registered `after_llm_call` hook.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Improvement

## Evidence

No network — the SDK client is stubbed and the same payload is served to both paths.

Before, on `main` (`3266932`):

```
anthropic sync : 'REDACTED BY HOOK'                     hook fired: 1  thinking kept: [{...signature...}]
anthropic async: 'the model said something sensitive'   hook fired: 0  thinking kept: []
openai    sync : 'REDACTED BY HOOK'                     hook fired: 1
openai    async: 'the model said something sensitive'   hook fired: 0
```

After:

```
anthropic sync : 'REDACTED BY HOOK'   hook: 1  thinking: [{...signature...}]
anthropic async: 'REDACTED BY HOOK'   hook: 1  thinking: [{...signature...}]
openai    sync : 'REDACTED BY HOOK'   hook: 1
openai    async: 'REDACTED BY HOOK'   hook: 1
```

The hook was not merely failing to modify the response on the async path — it was never invoked,
so an observe-only hook (logging, auditing, metrics) recorded nothing either.

## Tests

11 tests appended to `lib/crewai/tests/hooks/test_llm_hooks.py`. No network; the existing
`clear_hooks` autouse fixture handles isolation.

| test | role |
| --- | --- |
| `TestAfterLLMCallHooksOnAsyncPaths::test_sync_call_runs_after_hook` | the baseline the async path is compared against (parameterised over openai + anthropic) |
| `::test_async_call_runs_after_hook` | the regression — used to return the unredacted response |
| `::test_async_call_without_hooks_is_unchanged` | negative control: with no hook registered the response passes through untouched |
| `::test_async_call_hook_returning_none_keeps_response` | a hook returning `None` observes without replacing |
| `TestAnthropicThinkingBlocksOnAsyncPath::test_sync_call_stores_thinking_blocks` | baseline |
| `::test_async_call_stores_thinking_blocks` | the regression |
| `::test_async_thinking_blocks_are_replayed_on_the_next_turn` | asserts the stored block reaches the follow-up request with its signature intact |

The negative controls matter: without them, a "fix" that unconditionally rewrote the response, or
that stuffed something into `_previous_thinking_blocks` regardless of content, would pass.

### Control experiment

Reverting the three source files and keeping the tests:

```
6 failed, 31 passed

FAILED TestAfterLLMCallHooksOnAsyncPaths::test_async_call_runs_after_hook[openai]
FAILED TestAfterLLMCallHooksOnAsyncPaths::test_async_call_runs_after_hook[anthropic]
FAILED TestAfterLLMCallHooksOnAsyncPaths::test_async_call_hook_returning_none_keeps_response[openai]
FAILED TestAfterLLMCallHooksOnAsyncPaths::test_async_call_hook_returning_none_keeps_response[anthropic]
FAILED TestAnthropicThinkingBlocksOnAsyncPath::test_async_call_stores_thinking_blocks
FAILED TestAnthropicThinkingBlocksOnAsyncPath::test_async_thinking_blocks_are_replayed_on_the_next_turn
```

Exactly the six tests that depend on the fix fail; the sync baselines, the negative control and
all 26 pre-existing tests in the file pass unchanged.

### Suite runs

- `lib/crewai/tests/hooks/test_llm_hooks.py` — **37 passed** (26 pre-existing + 11 new)
- `lib/crewai/tests/llms/anthropic/` — 66 passed, 0 failed
- `lib/crewai/tests/llms/openai/` — 150 passed, 1 skipped, 0 failed
- `lib/crewai/tests/llms/bedrock/` — 35 passed, 8 skipped, 0 failed
- `ruff format --check` — `3 files already formatted`
- `ruff check` — `All checks passed!`

The provider runs report teardown errors on my machine (`PermissionError: [WinError 32]` while
`shutil.rmtree` removes `crewai_test_storage/latest_kickoff_task_outputs.db`) — a Windows sqlite
file-handle artifact in `conftest.py`, present on unmodified `main` and unrelated to this change.
No test failed.

## Additional notes

**Scope:** azure and gemini call the hook on *neither* path. That is a symmetric gap — those
providers do not support direct-call hooks at all — rather than sync/async drift, so I left it
out; it needs a separate decision. Happy to follow up if wanted.

Searched open PRs and issues for `after_llm_call`, `acall hooks`, `_invoke_after_llm_call_hooks`
and `thinking blocks async` — no existing report or PR. The scan that found this and the tests
were developed with AI assistance; every claim above was verified by running it.

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -1618,10 +1618,19 @@ class AnthropicCompletion(BaseLLM):
                     return result
 
         content = ""
+        thinking_blocks: list[ThinkingBlock] = []
+
         if response.content:
             for content_block in response.content:
                 if hasattr(content_block, "text"):
                     content += content_block.text
+                else:
+                    thinking_block = self._extract_thinking_block(content_block)
+                    if thinking_block:
+                        thinking_blocks.append(cast(ThinkingBlock, thinking_block))
+
+        if thinking_blocks:
+            self._previous_thinking_blocks = thinking_blocks
 
         content = self._apply_stop_words(content)
 
@@ -1639,7 +1648,9 @@ class AnthropicCompletion(BaseLLM):
         if usage.get("total_tokens", 0) > 0:
             logging.info(f"Anthropic API usage: {usage}")
 
-        return content
+        return self._invoke_after_llm_call_hooks(
+            params["messages"], content, from_agent
+        )
 
     async def _ahandle_streaming_completion(
         self,
@@ -1761,6 +1772,16 @@ class AnthropicCompletion(BaseLLM):
 
             final_message = await stream.get_final_message()
 
+        thinking_blocks: list[ThinkingBlock] = []
+        if final_message.content:
+            for content_block in final_message.content:
+                thinking_block = self._extract_thinking_block(content_block)
+                if thinking_block:
+                    thinking_blocks.append(cast(ThinkingBlock, thinking_block))
+
+        if thinking_blocks:
+            self._previous_thinking_blocks = thinking_blocks
+
         usage = self._extract_anthropic_token_usage(final_message)
         self._track_token_usage_internal(usage)
 
@@ -1828,7 +1849,9 @@ class AnthropicCompletion(BaseLLM):
             response_id=final_response_id,
         )
 
-        return full_response
+        return self._invoke_after_llm_call_hooks(
+            params["messages"], full_response, from_agent
+        )
 
     async def _ahandle_tool_use_conversation(
         self,
--- a/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
@@ -1166,7 +1166,11 @@ class BedrockCompletion(BaseLLM):
             response_id=response_id,
         )
 
-        return full_response
+        return self._invoke_after_llm_call_hooks(
+            messages,
+            full_response,
+            from_agent,
+        )
 
     async def _ensure_async_client(self) -> Any:
         """Ensure async client is initialized and return it."""
@@ -1422,7 +1426,11 @@ class BedrockCompletion(BaseLLM):
                 response_id=response_id,
             )
 
-            return text_content
+            return self._invoke_after_llm_call_hooks(
+                messages,
+                text_content,
+                from_agent,
+            )
 
         except ClientError as e:
             error_code = e.response.get("Error", {}).get("Code", "Unknown")
--- a/lib/crewai/src/crewai/llms/providers/openai/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/openai/completion.py
@@ -2557,7 +2557,12 @@ class OpenAICompletion(BaseLLM):
                     finish_reason=parsed_stream_finish_reason,
                     response_id=parsed_stream_response_id,
                 )
-                return accumulated_content
+                # Validation failed, so this returns raw model text rather than a
+                # parsed object -- the same kind of value every other path here
+                # hands to the hooks, and already reported as an LLM_CALL above.
+                return self._invoke_after_llm_call_hooks(
+                    params["messages"], accumulated_content, from_agent
+                )
 
         stream: AsyncIterator[
             ChatCompletionChunk
--- a/lib/crewai/src/crewai/llms/providers/openai/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/openai/completion.py
@@ -1212,6 +1212,10 @@ class OpenAICompletion(BaseLLM):
                 response_id=response_id,
             )
 
+            content = self._invoke_after_llm_call_hooks(
+                params.get("input", []), content, from_agent
+            )
+
         except NotFoundError as e:
             error_msg = f"Model {self.model} not found: {e}"
             logging.error(error_msg)
@@ -1510,7 +1514,9 @@ class OpenAICompletion(BaseLLM):
             response_id=response_id,
         )
 
-        return full_response
+        return self._invoke_after_llm_call_hooks(
+            params.get("input", []), full_response, from_agent
+        )
 
     def _extract_function_calls_from_response(
         self, response: Response
@@ -2436,6 +2442,10 @@ class OpenAICompletion(BaseLLM):
 
             if usage.get("total_tokens", 0) > 0:
                 logging.info(f"OpenAI API usage: {usage}")
+
+            content = self._invoke_after_llm_call_hooks(
+                params["messages"], content, from_agent
+            )
         except NotFoundError as e:
             error_msg = self._model_not_found_message(e)
             logging.error(error_msg)
@@ -2623,7 +2633,7 @@ class OpenAICompletion(BaseLLM):
                         response_id=response_id_stream,
                     )
 
-        return self._finalize_streaming_response(
+        result = self._finalize_streaming_response(
             full_response=full_response,
             tool_calls=tool_calls,
             usage_data=usage_data,
@@ -2634,6 +2644,11 @@ class OpenAICompletion(BaseLLM):
             finish_reason=stream_finish_reason,
             response_id=stream_response_id,
         )
+        if isinstance(result, str):
+            return self._invoke_after_llm_call_hooks(
+                params["messages"], result, from_agent
+            )
+        return result
 
     def supports_function_calling(self) -> bool:
         """Check if the model supports function calling."""
--- a/lib/crewai/tests/hooks/test_llm_hooks.py
+++ b/lib/crewai/tests/hooks/test_llm_hooks.py
@@ -670,3 +670,217 @@ class TestDirectLLMScopedHooks:
             )
 
         assert result == "contains [REDACTED]"
+
+
+def _openai_response(content: str) -> object:
+    """A minimal chat-completions response shaped like the OpenAI SDK's."""
+    from types import SimpleNamespace
+
+    message = SimpleNamespace(content=content, tool_calls=None, refusal=None)
+    return SimpleNamespace(
+        choices=[SimpleNamespace(message=message, finish_reason="stop", index=0)],
+        usage=SimpleNamespace(prompt_tokens=5, completion_tokens=7, total_tokens=12),
+        id="cmpl_stub",
+        model="gpt-4o",
+    )
+
+
+def _anthropic_response(content: str) -> object:
+    """A minimal messages response with one thinking block and one text block."""
+    from types import SimpleNamespace
+
+    thinking = SimpleNamespace(
+        type="thinking", thinking="internal reasoning", signature="sig-stub"
+    )
+    text = SimpleNamespace(type="text", text=content)
+    return SimpleNamespace(
+        content=[thinking, text],
+        usage=SimpleNamespace(input_tokens=5, output_tokens=7),
+        stop_reason="end_turn",
+        id="msg_stub",
+    )
+
+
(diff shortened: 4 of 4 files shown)


LHMQ878:
Went through all five findings against the code. One was a real bug and is fixed in `b3c4f1a`; three describe pre-existing behaviour that this PR deliberately mirrors rather than introduces; one I'd rather not do as suggested.

### Fixed: the structured-output streaming fallback (`accumulated_content`)

This one was right, and it was mine. In `_ahandle_streaming_completion`, when `response_model.model_validate_json` raises, the `except` returns `accumulated_content` from inside the `response_model` branch — upstream of the hook call I added at the end of the method. So a direct async call whose structured output failed to parse returned raw model text with `after_llm_call` never running.

What decided it: the method **already reports that same string as an `LLM_CALL`** via `_emit_call_completed_event` on the line before returning it. So the codebase already treats it as a model response; skipping the hooks for it was inconsistent, not intentional.

I checked whether this belonged in this PR or is pre-existing: the sync `_handle_streaming_completion`'s `response_model` branch returns `parsed_result` or `""` and has no raw-text fallback, so the gap exists only on the async path — in scope here.

Two tests. Reverting only the source change fails exactly one of them (the regression) and leaves the other (no-hook pass-through) passing.

### Not changed: hooks on string tool results

> That differs from the async non-streaming and sync paths and can corrupt tool results through redaction or rewriting.

The underlying concern is real — `_finalize_streaming_response` is documented as returning "tool execution result when `available_functions` is provided", and `_handle_tool_execution` is typed `-> str | None`, so `isinstance(result, str)` genuinely cannot tell model text from tool output.

But it does not differ from the sync path. The sync streaming path at `openai/completion.py:2303` is:

```python
if isinstance(result, str):
    return self._invoke_after_llm_call_hooks(
        params["messages"], result, from_agent
    )
return result
```

and my async block is byte-identical to it — I compared the two five-line blocks programmatically, not by eye. `git blame` puts the sync one in #6690 ("feat(tools): add WaitTool…"), not in this PR. So this PR makes the async path match the sync path; it does not create a divergence.

That leaves the question of whether the shared behaviour is wrong. I think it may well be, but fixing it means changing what the sync path does too, and the suggested remedy — "return an explicit result type indicating its origin" — changes `_finalize_streaming_response`'s return contract, which four call sites and the executor depend on. That is a different PR from "run the hooks on async paths", and doing it here would mean this PR silently changes sync behaviour as well. Happy to open it separately if a maintainer agrees the shared behaviour is a bug.

### Not changed: the Bedrock empty-response fallback

Same shape. The async early return at `bedrock/completion.py:1301` is character-for-character the sync one at `:696`, and the async hook site I added at `:1429` mirrors the pre-existing sync one at `:823` — which #6690 added, with the same apology-string return upstream of it. So the async path now behaves exactly like the sync path.

There's also a substantive reason not to route it through the hook: `"I apologize, but I received an empty response. Please try again."` is a string this client synthesises when Bedrock returns no content. It never came from the model. Passing it to `after_llm_call` would report a model response that does not exist — which for an auditing hook is worse than the current gap.

### On the test-coverage finding

Fair that the matrix is narrow. I'd push back on part of it though: a streaming structured-tool-call case would be asserting the `isinstance` guard, and per the above I don't think this PR should be pinning down behaviour it inherited from #6690 — if that guard is wrong, a test here would make it harder to change. The Bedrock and OpenAI-Responses async handlers are worth adding and I'll do that if you'd like; say the word and I'll extend the fixture matrix.

### Verification

`pytest lib/crewai/tests/hooks/test_llm_hooks.py` — 39 passed. (Locally I have to pass `-o addopts=...` to drop `--block-network`: on Windows `asyncio`'s event-loop setup uses `socket.socketpair()`, which the network guard rejects. Pre-existing async suites like `tests/a2a/test_a2a_integration.py` error the same way on this machine, so it isn't specific to these tests.) `ruff check` clean.

PR #6740, opened 2026-07-30
fix: run before_llm_call hooks on async provider paths

### LINKED ISSUE #6739: before_llm_call hooks never run on async acall(), so a blocking hook fails to block the request
### Description

`before_llm_call` hooks are the only interception point that can **abort** an LLM call: returning `False` (or raising `HookAborted` via the reducer) is supposed to stop the request from ever reaching the provider. That works on the synchronous `call()` path of every native provider, but **none of the five native providers invoke the hook on their async `acall()` path**.

The consequence is not "the hook didn't get a chance to observe the messages" — it is that **the request is actually issued**. A guardrail hook written to block a call (PII gate, spend cap, policy check, prompt-injection filter) silently permits it whenever the caller uses `acall()`/`kickoff_async`.

Coverage of `_invoke_before_llm_call_hooks` in `main` (`ebe0082`):

| provider | sync `call()` | async `acall()` |
| --- | --- | --- |
| `llms/providers/openai/completion.py` | ✅ `:465` | ❌ missing |
| `llms/providers/anthropic/completion.py` | ✅ `:377` | ❌ missing |
| `llms/providers/bedrock/completion.py` | ✅ `:380` | ❌ missing |
| `llms/providers/gemini/completion.py` | ✅ `:316` | ❌ missing |
| `llms/providers/azure/completion.py` | ✅ `:524` | ❌ missing |

`llms/base_llm.py` (the LiteLLM fallback) has the guard on both paths, so the bug is specific to the native providers.

### Steps to Reproduce

No network needed — swap the provider client for a stub that records whether a request was issued, so the observable is "did the HTTP call happen", not merely "did the response change".

```python
import asyncio
from types import SimpleNamespace

from crewai.hooks import register_before_llm_call_hook
from crewai.llms.providers.openai.completion import OpenAICompletion

ISSUED: list[str] = []


def blocking_hook(context) -> bool:
    return False  # must abort the call


register_before_llm_call_hook(blocking_hook)

llm = OpenAICompletion(model="gpt-4o", api_key="stub", stream=False)


def _response():
    msg = SimpleNamespace(content="LEAKED", tool_calls=None, refusal=None)
    return SimpleNamespace(
        choices=[SimpleNamespace(message=msg, finish_reason="stop")],
        usage=SimpleNamespace(prompt_tokens=1, completion_tokens=1, total_tokens=2),
    )


def create(**_kwargs):
    ISSUED.append("sync")
    return _response()


async def acreate(**_kwargs):
    ISSUED.append("async")
    return _response()


llm._client = SimpleNamespace(chat=SimpleNamespace(completions=SimpleNamespace(create=create)))
llm._async_client = SimpleNamespace(chat=SimpleNamespace(completions=SimpleNamespace(create=acreate)))

msgs = [{"role": "user", "content": "hi"}]

try:
    llm.call(msgs)
except ValueError as exc:
    print("sync  ->", exc)
print("sync  request issued:", ISSUED or "no")

ISSUED.clear()
print("async ->", asyncio.run(llm.acall(msgs)))
print("async request issued:", ISSUED or "no")
```

### Expected Behavior

Both paths refuse to issue the request:

```
sync  -> LLM call blocked by before_llm_call hook
sync  request issued: no
async -> LLM call blocked by before_llm_call hook
async request issued: no
```

### Actual Behavior

```
sync  -> LLM call blocked by before_llm_call hook
sync  request issued: no
async -> LEAKED
async request issued: ['async']
```

Same result on `AnthropicCompletion` (and by inspection on bedrock/gemini/azure, whose `acall()` bodies likewise have no `_invoke_before_llm_call_hooks` site).

### Notes

- This is the `before` counterpart to #6736 / #6737, which fixed the same sync/async drift for `after_llm_call`. `before` is the more severe half: an `after` hook can only fail to rewrite a response, while a `before` hook failing means a call the user explicitly blocked still went out to the provider and was billed.
- The fix is to invoke each provider's own sync-path guard from `acall()` as well. There is no double-dispatch risk on the agent path: `_invoke_before_llm_call_hooks` (`llms/base_llm.py:981`) returns `True` immediately when `from_agent is not None`, leaving the agent-executor dispatch as the single source.

I have a patch plus regression tests ready and will open a PR referencing this issue.

LHMQ878:
## Summary

`before_llm_call` hooks never ran on the async `acall()` path of any native provider. Because `before_llm_call` is the interception point that can **abort** a call, a hook returning `False` did not merely fail to observe the request — the request was actually issued and billed.

This is the `before` counterpart to #6736 / #6737 (`after_llm_call` on async paths), and the more severe half: an `after` hook that is skipped can only fail to rewrite a response, whereas a skipped `before` hook means a call the user explicitly blocked still reached the provider.

Fixes #6739

## Note on sync hooks in `acall()`

This PR runs the existing sync `before_llm_call` contract on `acall()` — same `bool` return, no new async hook type. If a hook already does I/O (logging to an external service, hitting a policy microservice), that already blocks on `call()`, and it will also block the event loop here. That's pre-existing, not something the five `acall()` sites introduce. An async hook variant is a separate API change.

## Root cause

Coverage of `_invoke_before_llm_call_hooks` on `main` (`ebe0082`):

| provider | sync `call()` | async `acall()` (before) | async `acall()` (after this PR) |
| --- | --- | --- | --- |
| `openai/completion.py` | ✅ `:465` | ❌ missing | ✅ `:603` |
| `anthropic/completion.py` | ✅ `:377` | ❌ missing | ✅ `:452` |
| `bedrock/completion.py` | ✅ `:380` | ❌ missing | ✅ `:513` |
| `gemini/completion.py` | ✅ `:316` | ❌ missing | ✅ `:402` |
| `azure/completion.py` | ✅ `:524` | ❌ missing | ✅ `:606` |

~~`llms/base_llm.py` (the LiteLLM fallback) already had the guard on both paths, so users on the fallback path were unaffected -- which is likely why this went unnoticed.~~

**Correction (2026-07-31): that sentence was wrong, in a way that understates the bug.** `llms/base_llm.py` only *defines* `_invoke_before_llm_call_hooks`; it has no callers, and its `acall()` is `raise NotImplementedError`. The LiteLLM fallback is `crewai/llm.py`, and there `call()` guards both hook families (`:1876`, `:1904`) while `acall()` at `:1959` guards **neither**. So the fallback has the same hole as the native providers rather than being the safe path, and nothing in this PR fixes it.

Probe on unmodified `origin/main` (`ebe0082`), with `litellm.completion` / `acompletion` stubbed so the observable is whether a request left the process:

```
blocking before hook (returns False) + rewriting after hook:
  call   before:1 after:0  issued:no     -> ValueError: LLM call blocked by before_llm_call hook
  acall  before:0 after:0  issued:async  -> 'the model said something sensitive'
```

`LLM(model=..., is_litellm=True)` is required to reach it -- without the flag the constructor returns a native provider, which is part of why I read the fallback as covered.

I am leaving that fix out of this PR rather than adding it: it is a different file and a different seam from the five provider `acall()` bodies under review here, and this PR already has review history against its current shape. It will follow separately. Flagging it here because "users on the fallback path were unaffected" is a claim a reviewer could reasonably act on, and it is not true.

## Proof, before any fix was written

A no-network probe that swaps the provider client for a stub recording **whether a request was issued**, so the observable is the HTTP call itself rather than the returned string:

```
anthropic call()       hook fired: 1  request issued: no    -> ValueError: LLM call blocked by before_llm_call hook
anthropic acall()      hook fired: 0  request issued: ['anthropic-async']  -> returned 'LEAKED'
openai    call()       hook fired: 1  request issued: no    -> ValueError: LLM call blocked by before_llm_call hook
openai    acall()      hook fired: 0  request issued: ['openai-async']     -> returned 'LEAKED'
```

After this PR, all four rows are identical:

```
anthropic call()       hook fired: 1  request issued: no    -> ValueError: LLM call blocked by before_llm_call hook
anthropic acall()      hook fired: 1  request issued: no    -> ValueError: LLM call blocked by before_llm_call hook
openai    call()       hook fired: 1  request issued: no    -> ValueError: LLM call blocked by before_llm_call hook
openai    acall()      hook fired: 1  request issued: no    -> ValueError: LLM call blocked by before_llm_call hook
```

## The change

Each `acall()` now runs **its own sync twin's guard, verbatim**, at the equivalent point in the body — right after the provider-specific message formatting and before the request parameters are assembled:

```python
if not self._invoke_before_llm_call_hooks(formatted_messages, from_agent):
    raise ValueError("LLM call blocked by before_llm_call hook")
```

Gemini needs the extra line its sync path uses at `:310`, because its formatted payload is `Content` objects rather than dicts:

```python
messages_for_hooks = self._convert_contents_to_dict(formatted_content)
```

Deliberately *not* changed:

- **No new hook-dispatch helper, no shared base-class wrapper.** Each provider formats messages differently and the sync paths already encode the right shape; copying the sync guard keeps the two twins textually comparable, which is what makes the next drift of this kind visible in review.
- **No double dispatch on the agent path.** `_invoke_before_llm_call_hooks` (`llms/base_llm.py:981`) returns `True` immediately when `from_agent is not None`, so the agent-executor dispatch remains the single source and adding the call to `acall()` cannot fire hooks twice.
- **Azure's `responses` delegate** returns before this block when `self.api == "responses"`, delegating to an `OpenAICompletion` — which now carries its own guard, so that path is covered too.

## Tests

`lib/crewai/tests/hooks/test_llm_hooks.py`, new class `TestBeforeLLMCallHooksOnAsyncPaths`, parametrised over `openai` and `anthropic` (10 tests total). The stubbed clients append to an `issued` list, so every assertion is about **whether the request went out**, not just what was returned.

| test | asserts |
| --- | --- |
| `test_sync_call_is_blocked_by_before_hook` | baseline: sync path blocks, `issued == []` |
| `test_async_call_is_blocked_by_before_hook` | **the regression**: `acall()` raises and `issued == []` |
| `test_async_call_runs_before_hook_that_allows` | a non-aborting hook observes the messages and the call proceeds (`issued == ["async"]`) |
| `test_async_call_without_hooks_is_unchanged` | **negative control**: with no hooks registered `acall()` behaves exactly as before |
| `test_async_call_before_hook_can_mutate_messages` | the list the hook receives via `ctx.messages` is the one handed to the provider |

### Control experiment

Reverting only the five source files and keeping the tests fails **exactly** the async-blocking tests and nothing else:

```
6 failed, 30 passed
```

The 6 are the three async-blocking assertions × 2 providers. The sync baseline, the negative control and all 26 pre-existing tests in the file pass unchanged — the tests are pinned to this bug, not to incidental behaviour.

## Verification

| suite | result |
| --- | --- |
| `tests/hooks/test_llm_hooks.py` | **36 passed** (26 pre-existing + 10 new) |
| `tests/llms/anthropic` + `tests/llms/openai` | **216 passed, 1 skipped**, 0 failed |
| `tests/llms/bedrock` + `google` + `azure` + `llms/hooks` | **244 passed, 10 skipped**, 0 failed |
| `ruff format --check` / `ruff check` on all 6 changed files | clean |

The provider suites additionally report Windows-only teardown **errors** (not failures) — `PermissionError: [WinError 32]` from `shutil.rmtree` on `crewai_test_storage/latest_kickoff_task_outputs.db` in `conftest.py:223`. These reproduce on clean `main` and are unrelated to this change; there are zero test failures.

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -449,6 +449,11 @@ class AnthropicCompletion(BaseLLM):
                     self._format_messages_for_anthropic(messages)
                 )
 
+                if not self._invoke_before_llm_call_hooks(
+                    formatted_messages, from_agent
+                ):
+                    raise ValueError("LLM call blocked by before_llm_call hook")
+
                 completion_params = self._prepare_completion_params(
                     formatted_messages, system_message, tools, available_functions
                 )
--- a/lib/crewai/src/crewai/llms/providers/azure/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/azure/completion.py
@@ -603,6 +603,11 @@ class AzureCompletion(BaseLLM):
 
                 formatted_messages = self._format_messages_for_azure(messages)
 
+                if not self._invoke_before_llm_call_hooks(
+                    formatted_messages, from_agent
+                ):
+                    raise ValueError("LLM call blocked by before_llm_call hook")
+
                 completion_params = self._prepare_completion_params(
                     formatted_messages, tools, effective_response_model
                 )
--- a/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/bedrock/completion.py
@@ -510,6 +510,11 @@ class BedrockCompletion(BaseLLM):
                     messages
                 )
 
+                if not self._invoke_before_llm_call_hooks(
+                    formatted_messages, from_agent
+                ):
+                    raise ValueError("LLM call blocked by before_llm_call hook")
+
                 body: BedrockConverseRequestBody = {
                     "inferenceConfig": self._get_inference_config(),
                 }
--- a/lib/crewai/src/crewai/llms/providers/gemini/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/gemini/completion.py
@@ -397,6 +397,13 @@ class GeminiCompletion(BaseLLM):
                     self._format_messages_for_gemini(messages)
                 )
 
+                messages_for_hooks = self._convert_contents_to_dict(formatted_content)
+
+                if not self._invoke_before_llm_call_hooks(
+                    messages_for_hooks, from_agent
+                ):
+                    raise ValueError("LLM call blocked by before_llm_call hook")
+
                 config = self._prepare_generation_config(
                     system_instruction, tools, effective_response_model
                 )
--- a/lib/crewai/src/crewai/llms/providers/openai/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/openai/completion.py
@@ -600,6 +600,11 @@ class OpenAICompletion(BaseLLM):
 
                 formatted_messages = self._format_messages(messages)
 
+                if not self._invoke_before_llm_call_hooks(
+                    formatted_messages, from_agent
+                ):
+                    raise ValueError("LLM call blocked by before_llm_call hook")
+
                 if self._effective_api() == "responses":
                     return await self._acall_responses(
                         messages=formatted_messages,
--- a/lib/crewai/tests/hooks/test_llm_hooks.py
+++ b/lib/crewai/tests/hooks/test_llm_hooks.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from types import SimpleNamespace
+from typing import Any
 from unittest.mock import Mock
 
 from crewai.hooks import (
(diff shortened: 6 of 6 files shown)


LHMQ878:
Went through both findings against the code. The test-coverage one was right and is fixed in `ae92404`; the Bedrock recursion one describes a real gap that is pre-existing and identical on the sync path, so I'd rather not widen this PR into it.

### Fixed: coverage for all five providers

> `TestBeforeLLMCallHooksOnAsyncPaths` is parametrized only over `["openai", "anthropic"]`

Correct, and it matters more here than it usually would: the regression class is *per-provider* -- each `acall()` had its own independently missing guard -- so two providers of five left three of the source changes asserting nothing. `ae92404` parametrises over all five, all built the same way (a real completion object whose SDK client is a stub appending to an `issued` list, so the assertion stays "was the request issued").

Two provider-specific details worth noting for review:

- **Bedrock** raises `NotImplementedError` at `completion.py:492` before reaching the hook when `aiobotocore` is absent, and that is an optional extra. The gate under test is upstream of any AWS I/O and the client is a stub, so the builder patches `AIOBOTOCORE_AVAILABLE` and presets `_async_client_initialized` rather than depending on the extra being installed.
- **Gemini** serves both paths from one client with the async surface under `.aio`, matching `_get_async_client` delegating to `_get_sync_client` at `:141-143`.

`test_async_call_runs_before_hook_that_allows` now locates the prompt inside the last message instead of comparing against one provider's content layout, since the formatted shape genuinely differs (plain string vs Converse content blocks vs Gemini part dicts).

**Control experiment, rerun at five providers:** reverting only the five source files fails exactly the three async-blocking tests on every provider -- **15 failed / 36 passed** -- with the sync baseline, the negative control and all 26 pre-existing tests in the file passing unchanged.

### Not changed: the recursive Bedrock follow-up

> After tool execution, `_ahandle_converse()` recursively issues another async `converse` request without invoking `_invoke_before_llm_call_hooks`

The behaviour is real. `_ahandle_converse` appends the `toolResult` message and recurses at `:1404` and `:1713`, and the guard I added sits in `acall()`, which the recursion does not re-enter -- so a hook that would block based on tool results does not see the follow-up request.

Two reasons it does not belong in this PR:

1. **It is not async-specific.** `_handle_converse` recurses identically at `:802` and `:1112`, and the sync path has never gated those either. So this is not the sync/async drift this PR exists to close -- it is a symmetric gap present on both paths since the recursion was written. Fixing only the async half here would *create* drift in the opposite direction.
2. **Bedrock is the only provider that does this.** I checked the other four: openai, anthropic, gemini and azure all execute tools and return through `_process_*_response` without re-issuing a request from inside the handler. So the fix is not "apply the guard at the shared async request boundary" -- there is no shared boundary to apply it at, and the change would be Bedrock-local design work about whether a tool-result turn counts as a new call for hook purposes.

That second question is a real one and I don't think it should be answered silently inside a drift fix. Happy to file it separately with the line references above if maintainers agree it's worth closing; my read is that gating the follow-up turn is probably right, since a `before_llm_call` hook exists precisely to inspect what is about to be sent and tool results are attacker-influenced content. But it is a behaviour change for both paths, not a parity repair.

LHMQ878:
The payload-assertion finding is correct, and thanks for pushing on it — the test was weaker than I thought. Fixed in 9a9a9f9.

**The gap was real.** `test_async_call_before_hook_can_mutate_messages` asserted on `captured[0][-1]`, and `captured[0]` *is* the list the hook appended to. That assertion holds no matter what `acall()` does with the list afterwards, so it would have passed if the guard were handed a copy and the mutation silently dropped — precisely the failure the test was supposed to catch. Its docstring ("the hook's messages list is the one handed to the provider") claimed something it never checked.

**What I did.** The provider stubs now record their request kwargs via a small `_Recorder` instead of discarding them, and the mutation tests assert against what the stubbed client actually received. That turned up something I want to flag rather than quietly encode:

Whether a `before_llm_call` mutation reaches the payload is a **pre-existing, per-provider** property, not something this PR decides. Measured on this branch:

| provider | mutation reaches payload (sync `call()`) | (async `acall()`) |
|---|---|---|
| openai | yes | yes |
| anthropic | yes | yes |
| azure | yes | yes |
| bedrock | yes | yes |
| gemini | **no** | **no** |

Gemini differs because its guard is fed `_convert_contents_to_dict(formatted_content)` — its payload is `types.Content` objects, not dicts, so the hook necessarily gets a converted copy. That is what `main`'s sync path at `gemini/completion.py:314` already does; I mirrored it in `acall()` deliberately, and it means a hook can *inspect* and *block* on Gemini but cannot *rewrite* its payload. Worth fixing, but it is upstream of this change and identical on both paths, so I have left it alone rather than widening the PR.

So the tests are split to say exactly that, and no more:

- `test_async_before_hook_mutation_reaches_the_provider_as_on_sync` — pins async propagation to whatever the *same provider's sync path* does. This is the invariant this PR owns: `acall()` gates and forwards exactly as `call()` does. It fails if the async guard is ever wired to a different list than its sync twin, without asserting Gemini's copy is correct.
- `test_async_before_hook_mutation_reaches_the_provider` — the symmetry check above cannot distinguish "both propagate" from "neither does", so this pins the four propagating providers by name, with Gemini skipped and the reason stated in the skip message rather than papered over.

`test_async_call_runs_before_hook_that_allows` also now asserts the prompt is present in the recorded payload, so the hook is known to observe the request that is really about to be issued.

**Discrimination check.** Reverting only the five provider source files (tests untouched) fails **18** tests and no sync test: both mutation tests on the four propagating providers (8), plus the blocking and the allowing test on all five (10). With the fix in place: `55 passed, 1 skipped` in `tests/hooks/test_llm_hooks.py`.

On the Bedrock recursion finding, my position from [the earlier reply](https://github.com/crewAIInc/crewAI/pull/6740#issuecomment-5135620393) stands: `_ahandle_converse()`'s follow-up request genuinely is unguarded, but `_handle_converse()` has the identical gap on the sync path, so it is not a regression this PR introduces and fixing it here would mean changing sync behaviour too. Happy to do it as a follow-up if you'd like it — it is a one-line guard per path plus a tool-loop test fixture.

PR #6743, opened 2026-07-30
fix: stop UploadCache sync wrappers deadlocking inside a running event loop

### LINKED ISSUE #6742: UploadCache sync wrappers deadlock when called from inside a running event loop
## Description

`UploadCache._run_sync` has a branch for being called while an event loop is already running. Its docstring says it runs the coroutine *"without blocking event loop"*. It does the opposite: it schedules the coroutine on the caller's own running loop and then blocks the caller's thread waiting for it. That thread **is** the thread running the loop, so the loop can never advance the coroutine it was just handed. The call stalls for the full 30-second timeout and raises `TimeoutError`.

`lib/crewai-files/src/crewai_files/cache/upload_cache.py:402-413` on `main` (`ebe0082`):

```python
@staticmethod
def _run_sync(coro: Any) -> Any:
    """Run an async coroutine from sync context without blocking event loop."""
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        loop = None

    if loop is not None and loop.is_running():
        future = asyncio.run_coroutine_threadsafe(coro, loop)
        return future.result(timeout=30)
    return asyncio.run(coro)
```

`asyncio.run_coroutine_threadsafe` is documented as being for submitting a coroutine to a loop running **in a different thread**. Here `loop` was obtained from `asyncio.get_running_loop()`, i.e. it is this thread's loop, so the precondition is inverted.

## Scope

Every one of the ten synchronous wrappers funnels through `_run_sync`, so all of them hang when called from async code:

| wrapper | line |
| --- | --- |
| `get` | `:417` |
| `get_by_hash` | `:424` |
| `set` | `:438` |
| `set_by_hash` | `:453` |
| `remove` | `:462` |
| `remove_by_file_id` | `:467` |
| `clear_expired` | `:472` |
| `clear` | `:477` |
| `get_all_for_provider` | `:482` |

`UploadCache` and `cleanup_expired_files` are public exports of `crewai_files`, so this is reachable from user code: a FastAPI handler, a notebook cell, or an async crew callback that touches the upload cache synchronously will stall for 30s and then raise.

## Steps to reproduce

```python
import asyncio

from crewai_files import FileBytes, ImageFile
from crewai_files.cache.upload_cache import UploadCache

MINIMAL_PNG = (
    b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01\x00\x00\x00\x01"
    b"\x08\x06\x00\x00\x00\x1f\x15\xc4\x89\x00\x00\x00\nIDATx\x9cc\x00"
    b"\x01\x00\x00\x05\x00\x01\r\n-\xb4\x00\x00\x00\x00IEND\xaeB`\x82"
)


def use_the_cache() -> None:
    cache = UploadCache()
    file = ImageFile(source=FileBytes(data=MINIMAL_PNG, filename="test.png"))
    cache.set(file=file, provider="gemini", file_id="file-123")
    print("cached:", cache.get(file=file, provider="gemini"))


use_the_cache()  # no running loop -> works


async def main() -> None:
    use_the_cache()  # a loop is running -> hangs 30s, then TimeoutError


asyncio.run(main())
```

Output:

```
cached: CachedUpload(file_id='file-123', ...)
  ...
  File ".../crewai_files/cache/upload_cache.py", line 412, in _run_sync
    return future.result(timeout=30)
  File ".../concurrent/futures/_base.py", line 458, in result
    raise TimeoutError()
TimeoutError
```

The bridging logic alone reproduces it, with no cache involved:

```python
import asyncio


async def work():
    return "ok"


def run_sync(coro):
    loop = asyncio.get_running_loop()
    return asyncio.run_coroutine_threadsafe(coro, loop).result(timeout=3)


async def main():
    try:
        return run_sync(work())
    except BaseException as e:
        return f"{type(e).__name__} (deadlock)"


print("inside a running loop:", asyncio.run(main()))
# -> inside a running loop: TimeoutError (deadlock)
```

## Expected behavior

The synchronous wrappers return their result whether or not the calling thread has a running event loop. Providing a sync surface over the async cache is the entire purpose of these wrappers, and the `loop.is_running()` branch exists specifically to handle this case.

## Suggested fix

Drive the coroutine on a worker thread with its own loop, so the thread that blocks is not the thread that must make progress:

```python
if loop is not None and loop.is_running():
    executor = ThreadPoolExecutor(max_workers=1)
    try:
        return executor.submit(asyncio.run, coro).result(timeout=30)
    finally:
        executor.shutdown(wait=False)
```

`wait=False` matters: the context-manager form's `shutdown(wait=True)` blocks until the worker finishes, which would make the 30s bound meaningless on exactly the timeout it is supposed to guard.

Loop affinity is not a concern here — the pre-existing `else` branch already ran every coroutine under a fresh `asyncio.run`, so the `aiocache` backend was already required to be loop-agnostic, and the default memory backend is a plain dict.

`lib/crewai-files/tests/test_upload_cache.py` currently has no async coverage at all, which is why a branch that always timed out went unnoticed.

## Additional context

This is the deadlock half of the same sync/async bridging class as camel-ai/camel#4239: a `run_coroutine_threadsafe` call whose target loop is the caller's own. I have a patch with regression tests ready and will open a PR referencing this issue.

LHMQ878:
## Description

`UploadCache._run_sync`'s branch for being called while an event loop is already running could not succeed. It scheduled the coroutine on the caller's **own** running loop with `asyncio.run_coroutine_threadsafe`, then blocked that same thread on `.result(timeout=30)`. The blocked thread is the thread running the target loop, so the loop can never advance the coroutine it was just handed. Every such call stalled for the full 30 seconds and raised `TimeoutError`.

The docstring claimed the opposite of the behaviour: *"Run an async coroutine from sync context without blocking event loop."*

Fixes #6742

## Root cause

`lib/crewai-files/src/crewai_files/cache/upload_cache.py:402-413` on `main` (`ebe0082`):

```python
@staticmethod
def _run_sync(coro: Any) -> Any:
    """Run an async coroutine from sync context without blocking event loop."""
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        loop = None

    if loop is not None and loop.is_running():
        future = asyncio.run_coroutine_threadsafe(coro, loop)
        return future.result(timeout=30)
    return asyncio.run(coro)
```

`asyncio.run_coroutine_threadsafe` is for submitting work to a loop running in a **different** thread. `loop` here came from `asyncio.get_running_loop()`, so it is this thread's loop — the precondition is inverted.

All ten synchronous wrappers funnel through `_run_sync` (`get`, `get_by_hash`, `set`, `set_by_hash`, `remove`, `remove_by_file_id`, `clear_expired`, `clear`, `get_all_for_provider`), and `UploadCache` / `cleanup_expired_files` are public exports of `crewai_files`. So any of them called from a FastAPI handler, a notebook cell, or an async crew callback hung for 30s and then raised.

## The change

The coroutine runs on a worker thread with its own loop, so the thread that blocks is not the thread that has to make progress:

```python
if loop is not None and loop.is_running():
    executor = ThreadPoolExecutor(max_workers=1)
    try:
        return executor.submit(asyncio.run, coro).result(timeout=30)
    finally:
        # ``shutdown(wait=True)`` -- what the context-manager form
        # does -- would block until the worker finishes, so a
        # coroutine that overran the timeout would still hang the
        # caller and make the bound meaningless.
        executor.shutdown(wait=False)
```

The `wait=False` shutdown is deliberate rather than stylistic. Writing this as `with ThreadPoolExecutor(...) as executor:` reads better but reintroduces the bug it is fixing on the timeout path: `__exit__` calls `shutdown(wait=True)`, which blocks until the worker finishes, so a coroutine that overran the 30s bound would still hang the caller indefinitely — the bound would only appear to be enforced.

Deliberately *not* changed:

- **The `else` branch is untouched.** With no running loop, `asyncio.run(coro)` was already correct.
- **The 30s timeout value is kept**, so this PR changes only whether the call can complete, not how long a caller may wait.
- **No persistent background loop.** A per-call worker keeps the thread's lifetime tied to the call and adds no global state; these wrappers are coarse-grained and already paid for a fresh loop per call on the `else` path.

Loop affinity was checked before choosing this shape: the pre-existing `else` branch already ran every coroutine under a fresh `asyncio.run`, so the `aiocache` backend was already required to be loop-agnostic, and the default memory backend is `Cache(serializer=PickleSerializer(), namespace=namespace)` — a plain dict.

## Tests

`lib/crewai-files/tests/test_upload_cache.py` had no async coverage at all, which is why a branch that always timed out went unnoticed. Five tests added in `TestRunSyncFromInsideAnEventLoop`:

| test | asserts |
| --- | --- |
| `test_set_and_get_outside_event_loop` | baseline: with no running loop the wrappers already worked |
| `test_set_and_get_inside_running_event_loop` | **the regression**: `set`/`get` from async code complete, bounded by a wall clock assertion so a stall fails with a diagnosis rather than just a timeout |
| `test_sync_wrappers_do_not_use_the_callers_loop` | the coroutine runs on a loop that is *not* the caller's — the property whose absence causes the stall, checked directly rather than inferred from elapsed time |
| `test_remaining_sync_wrappers_inside_running_event_loop` | the other wrappers (`get_all_for_provider`, `get_by_hash`, `remove`, `remove_by_file_id`, `clear_expired`, `clear`) share `_run_sync` and are covered too |
| `test_exceptions_propagate_from_inside_running_event_loop` | a failure inside the coroutine surfaces to the sync caller instead of being swallowed by the bridge |

### Control experiment

Reverting only `upload_cache.py` and keeping the tests:

```
4 failed, 15 passed   (122.68s)
```

The 4 failures are exactly the four in-loop tests, and each fails at `concurrent/futures/_base.py:458: TimeoutError` — the reported defect, not an incidental assertion. The sync baseline and all 14 pre-existing tests pass unchanged.

Worth flagging for review: that control run takes ~2 minutes because the unfixed code *stalls* for its 30s timeout on each affected test rather than failing fast. The wall clock bound in `test_set_and_get_inside_running_event_loop` (`elapsed < 10`) exists so the headline regression reports the cause rather than only timing out.

## Verification

| suite | result |
| --- | --- |
| `lib/crewai-files/tests/test_upload_cache.py` | **19 passed** (14 pre-existing + 5 new) |
| `ruff check` / `ruff format --check` on both changed files | clean |
| `mypy lib/crewai-files/src/crewai_files/cache/upload_cache.py` | clean |

On Windows these tests need `--allowed-hosts=127.0.0.1,::1` (the `ProactorEventLoop` self-pipe uses `socket.socketpair()`, which `pytest-recording` blocks) and a working `libmagic`; both are pre-existing environment concerns unrelated to this change.

### DIFF
--- a/lib/crewai-files/src/crewai_files/cache/upload_cache.py
+++ b/lib/crewai-files/src/crewai_files/cache/upload_cache.py
@@ -6,11 +6,12 @@ import asyncio
 import atexit
 import builtins
 from collections.abc import Iterator
-from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures import TimeoutError as FuturesTimeoutError
 from dataclasses import dataclass
 from datetime import datetime, timezone
 import hashlib
 import logging
+import threading
 from typing import TYPE_CHECKING, Any
 
 from aiocache import Cache  # type: ignore[import-untyped]
@@ -24,6 +25,31 @@ if TYPE_CHECKING:
     from crewai_files.core.types import FileInput
 
 logger = logging.getLogger(__name__)
+
+# How long a sync wrapper waits for its coroutine when it had to hand the work
+# to a worker loop. These are cache reads and writes, so the bound exists to
+# turn a hang into an error rather than to police a plausible duration.
+_RUN_SYNC_TIMEOUT_SECONDS = 30
+# How long to wait for the worker loop to stop after the task is done or
+# cancelled. Reached only if the coroutine ignores cancellation.
+_RUN_SYNC_JOIN_TIMEOUT_SECONDS = 5
+
+
+async def _settling(coro: Any, done: threading.Event) -> Any:
+    """Run ``coro``, signalling ``done`` once it has fully unwound.
+
+    The concurrent future that :func:`asyncio.run_coroutine_threadsafe` returns
+    is not a reliable "the coroutine has stopped" signal after a cancellation:
+    it is completed when the cancellation is *chained* to it, while the task may
+    still be running its own ``finally`` blocks. Stopping the loop on that signal
+    destroys the task mid-unwind ("Task was destroyed but it is pending").
+    Setting the flag from inside the coroutine's own ``finally`` cannot be early
+    by construction.
+    """
+    try:
+        return await coro
+    finally:
+        done.set()
 
 
 @dataclass
@@ -409,23 +435,70 @@ class UploadCache:
         loop runs on, so the loop could never advance the coroutine and the
         wait would time out. The coroutine is driven on a worker thread with
         its own loop instead.
+
+        The loop is owned here rather than handed to ``asyncio.run`` on a
+        pooled thread, because the timeout has to be able to *stop* the work.
+        ``asyncio.run`` is opaque from the outside: the only handle on it is
+        the thread, and threads cannot be interrupted, so a coroutine that
+        overran the bound would keep running -- and, being a non-daemon pool
+        thread, hold up interpreter exit until it finished. Scheduling through
+        :func:`asyncio.run_coroutine_threadsafe` keeps a handle on the task
+        itself, so a timeout cancels it and the loop shuts down for real.
         """
         try:
             loop = asyncio.get_running_loop()
         except RuntimeError:
             loop = None
 
-        if loop is not None and loop.is_running():
-            executor = ThreadPoolExecutor(max_workers=1)
+        if loop is None or not loop.is_running():
+            return asyncio.run(coro)
+
+        worker_loop = asyncio.new_event_loop()
+        # Daemon so that a worker wedged in uncancellable work (a blocking
+        # call inside the coroutine) cannot keep the process alive; the
+        # cancellation below is what handles the ordinary case.
+        thread = threading.Thread(
+            target=worker_loop.run_forever,
+            name="crewai-files-upload-cache",
+            daemon=True,
+        )
+        thread.start()
+        unwound = threading.Event()
+        try:
+            future = asyncio.run_coroutine_threadsafe(
+                _settling(coro, unwound), worker_loop
+            )
             try:
-                return executor.submit(asyncio.run, coro).result(timeout=30)
-            finally:
-                # ``shutdown(wait=True)`` -- what the context-manager form
-                # does -- would block until the worker finishes, so a
-                # coroutine that overran the timeout would still hang the
-                # caller and make the bound meaningless.
-                executor.shutdown(wait=False)
-        return asyncio.run(coro)
+                return future.result(timeout=_RUN_SYNC_TIMEOUT_SECONDS)
+            except FuturesTimeoutError:
+                # Cancelling the task, not just abandoning the thread: the
+                # coroutine touches the shared cache, so letting it run on
+                # after the caller gave up would mutate state nobody is
+                # waiting for.
+                future.cancel()
+                # ``cancel()`` only *requests* it -- the exception is thrown in
+                # on the worker loop's next pass. Stopping the loop right away
+                # would race that, leaving the coroutine suspended with its
+                # cleanup never run. The bound keeps a coroutine that swallows
+                # cancellation from turning this into a second hang.
+                unwound.wait(timeout=_RUN_SYNC_JOIN_TIMEOUT_SECONDS)
+                raise
+        finally:
+            worker_loop.call_soon_threadsafe(worker_loop.stop)
+            thread.join(timeout=_RUN_SYNC_JOIN_TIMEOUT_SECONDS)
+            if thread.is_alive():
+                # Closing a loop whose thread is still in ``run_forever``
+                # would raise, and would leave the loop's internals being
+                # torn down underneath it. The join timeout is only reached
+                # when the coroutine ignores cancellation, so the loop is
+                # leaked deliberately in preference to that.
+                logger.warning(
+                    "Upload cache worker did not stop within "
+                    f"{_RUN_SYNC_JOIN_TIMEOUT_SECONDS}s; its event loop is "
+                    "left open because closing a running loop is unsafe."
+                )
+            else:
+                worker_loop.close()
 
     def get(self, file: FileInput, provider: ProviderType) -> CachedUpload | None:
         """Sync wrapper for aget."""
--- a/lib/crewai-files/src/crewai_files/cache/upload_cache.py
+++ b/lib/crewai-files/src/crewai_files/cache/upload_cache.py
@@ -6,6 +6,7 @@ import asyncio
 import atexit
 import builtins
 from collections.abc import Iterator
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass
 from datetime import datetime, timezone
 import hashlib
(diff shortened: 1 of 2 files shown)


LHMQ878:
Both findings were right. Fixed in `cf0182f`.

## The major one: `shutdown(wait=False)` cannot stop the work

Confirmed, and the reason matters for the shape of the fix. `asyncio.run` on a pooled thread is opaque from the outside — the only handle on it is the thread, and threads cannot be interrupted — so there was nothing to cancel *with*. Measured on `b262f4e`:

```
caller unblocked after 0.50s: TimeoutError
1.0s after the timeout, marks = []                        <- still running
about to exit;              marks = []
interpreter exit took 1.50s, marks = ['COMPLETED AFTER TIMEOUT']
```

That last line is worse than the finding states and is what decided me. The coroutine did not merely outlive its caller: pool threads are non-daemon, so it **held up interpreter exit** and then ran to completion *during teardown*, writing to the shared cache there. A 30s timeout would have added 30s to the exit of a process that had already given up on the call.

After the fix, same probe:

```
timed out after 0.22s
settled? True  outcome = ['cancelled']
threads: ['MainThread']
interpreter exit took 0.00s
```

## Two things the fix turns on

I went with `run_coroutine_threadsafe` on an owned loop rather than a lifecycle-managed pool, since keeping a handle on the *task* is the whole point. Two details cost me a debugging pass each and are worth stating:

**1. The concurrent future is not a "the coroutine has stopped" signal.** It completes when the cancellation is *chained* to it, while the task may still be running its own `finally`. Waiting on `future.exception()` and then stopping the loop still produced:

```
Task was destroyed but it is pending!
task: <Task cancelling ... wait_for=<Future cancelled>>
```

So the flag is now set from inside the coroutine's own `finally` (`_settling`), which cannot be early by construction.

**2. The loop is left open, not closed, if the join times out.** Closing a loop whose thread is still in `run_forever` raises, and would tear down its internals underneath the running task. That path is only reachable when the coroutine ignores cancellation; the thread is a daemon so it still cannot hold the process open.

## Tests

- `test_timeout_cancels_the_coroutine_instead_of_abandoning_it`
- `test_timed_out_worker_is_cleaned_up_before_returning`
- `test_successful_call_leaves_no_worker_thread_behind`

Two things I had to correct in my own tests, since both would have passed for the wrong reason:

- With a 0.2s bound the future gets cancelled *before the worker loop picks it up*. That is a correct outcome, but it means no running task was ever interrupted — which is the property under test. The bound is now 2.0s and a `started` event pins the premise.
- Filtering leaked threads by the name prefix this implementation happens to use would pass for any implementation that leaks a differently-named thread. Comparison is now by identity across `threading.enumerate()`.

Discrimination: swapping the old `ThreadPoolExecutor` + `shutdown(wait=False)` mechanism back in while keeping the new constants fails all three (19 passed / 3 failed); with the fix, 22 passed. Whole package: 175 passed / 4 skipped. `ruff check`, `ruff format --check` and `mypy` clean.

## The minor one

Taken as suggested — `test_remaining_sync_wrappers_inside_running_event_loop` now calls `set_by_hash` instead of `set` twice, so that wrapper is exercised directly rather than being listed in scope and left uncovered.

PR #6744, opened 2026-07-30
Fix MCP connection timeouts and disable tool call retries by default

lorenzejay:
> [!NOTE]
> **Medium Risk**
> Touches core MCP connection and tool execution paths used by agents; default no-retry on tools changes behavior for callers that relied on implicit retries, though timeouts and cleanup reduce flaky double-execution risk.
> 
> **Overview**
> Hardens **MCP connect/teardown** so `connect_timeout` applies to transport SDK startup (HTTP/SSE/stdio) via shared `async_timeout`, avoiding `asyncio.wait_for` crossing tasks and anyio “cancel scope in a different task” failures. **HTTP** teardown now forwards the unwinding exception into the SDK context exit; connection failures get clearer classification (timeout, TLS, auth) and cleanup that does not mask the root error.
> 
> **`MCPClient`** propagates timeout to transports, bounds session `initialize()` with remaining budget, and **stops retrying tool calls by default** (`retry_tool_calls=False`) to avoid replaying mutating tools after a lost response; retries are opt-in.
> 
> Adds **`test_client.py`** lifecycle coverage and switches MCP adapter integration tests to **`sys.executable`** instead of `uv run python`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d46f6117192e2c7e2b88ad36d8b32126f51b2ae7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/crewai/src/crewai/mcp/client.py
```diff
@@ -76,6 +113,7 @@ def __init__(
         max_retries: int = MCP_MAX_RETRIES,
         cache_tools_list: bool = False,
         logger: logging.Logger | None = None,
+        retry_tool_calls: bool = False,
```
joaomdmoura, line 116: is this intended? jsut asking because disabling tool-call retries by default is a behavior change, so want to point it out

joaomdmoura (approved):

lib/crewai/src/crewai/mcp/client.py
lorenzejay: it stops MCPClient from its own internal retires different from blocking Agent.max_retry_limit:

example of why we do False,

1. CrewAI calls an MCP tool.
2. The server successfully performs the action.
3. The response is lost because the connection drops.
4. CrewAI assumes the operation failed and calls it again.
5. The action happens twice.

PR #6748, opened 2026-07-31
fix(memory): keep the best duplicate score in recall dedup, not the first

### LINKED ISSUE #6747: Memory recall dedup keeps whichever duplicate a thread finished first, not the best score
## Description

`RecallFlow.synthesize_results` deduplicates by keeping the **first** occurrence of each `record.id`, but the findings it iterates arrive in **thread-completion order**. When the same record is reachable from more than one `(sub-query, scope)` search task, which of its scores survives is decided by thread timing, so the same query can return the same record at a different rank on consecutive runs.

`lib/crewai/src/crewai/memory/recall_flow.py`, `synthesize_results`:

```python
seen_ids: set[str] = set()
matches: list[MemoryMatch] = []
for finding in self.state.chunk_findings:
    ...
    for item in results:
        ...
        if isinstance(record, MemoryRecord) and record.id not in seen_ids:
            seen_ids.add(record.id)          # <-- first wins
            composite, reasons = compute_composite_score(record, float(score), self._config)
            matches.append(MemoryMatch(record=record, score=composite, match_reasons=reasons))
matches.sort(key=lambda m: m.score, reverse=True)
final_results = matches[: self.state.limit]
```

And `_do_search`, which produces `chunk_findings`:

```python
with ThreadPoolExecutor(max_workers=min(len(tasks), 4)) as pool:
    futures = {pool.submit(...): (emb, sc) for emb, sc in tasks}
    for future in as_completed(futures):     # <-- completion order, not task order
        ...
        findings.append({"scope": scope, "results": results, "top_score": top_composite})
```

## Why duplicates are guaranteed, not hypothetical

Two independent reasons, both structural:

1. **Scope prefix matching overlaps by design.** `lib/crewai/src/crewai/memory/storage/lancedb_storage.py`:

   ```python
   if scope_prefix is not None and scope_prefix.strip("/"):
       like_val = prefix + "%"
       query = query.where(f"scope LIKE '{like_val}'")
   ```

   A record stored at `/a/b` is returned for scope `/a` **and** for scope `/a/b`. `filter_and_chunk` routinely selects both (`selected_scopes = candidates[:20]`).

2. **Each sub-query embedding scores the same record differently.** `analyze_query_step` produces up to 3 sub-query embeddings, and `_do_search` runs the full `embeddings × scopes` cross product — up to 60 tasks against overlapping scopes.

So the same `record.id` legitimately arrives from several tasks carrying *different* semantic scores, and `compute_composite_score` is monotonic in `semantic_score`:

```python
composite = (
    config.semantic_weight * semantic_score
    + config.recency_weight * decay
    + config.importance_weight * record.importance
)
```

Keeping the first-arriving one therefore does not merely reorder — it can keep a strictly worse score and **under-rank** the record, or push it out of `matches[: self.state.limit]` entirely. The duplicate also consumed nothing visible, but the lower score it left behind is what the caller ranks on.

## Reproduction

Standalone, no storage backend needed — it models `_do_search`'s dispatch plus the first-wins dedup, and only the sleep order differs between the two runs:

```python
"""Models _do_search's dispatch plus synthesize_results' first-wins dedup.

One record 'R' lives at /a/b, so lancedb's `scope LIKE 'prefix%'` returns it for
scope /a AND for scope /a/b, and each sub-query embedding scores it differently.
Only the per-task delays differ between the two runs below.
"""

from concurrent.futures import ThreadPoolExecutor, as_completed
import time


def run(tasks):
    """tasks: list of (scope, semantic_score, delay_seconds)."""

    def one(scope, score, delay):
        time.sleep(delay)
        return {"scope": scope, "results": [("R", score)]}

    findings = []
    with ThreadPoolExecutor(max_workers=2) as pool:
        futures = [pool.submit(one, *t) for t in tasks]
        for future in as_completed(futures):
            findings.append(future.result())

    seen, out = set(), []
    for finding in findings:
        for rid, score in finding["results"]:
            if rid not in seen:          # first wins -- the bug
                seen.add(rid)
                out.append((rid, score, finding["scope"]))
    return out


print("record 'R': scope /a scores 0.9, scope /a/b scores 0.4")
print("  /a finishes first: ", run([("/a", 0.9, 0.00), ("/a/b", 0.4, 0.05)]))
print("  /a/b finishes first:", run([("/a", 0.9, 0.05), ("/a/b", 0.4, 0.00)]))
```

Measured output:

```
record 'R': scope /a scores 0.9, scope /a/b scores 0.4
  /a finishes first:  [('R', 0.9, '/a')]
  /a/b finishes first: [('R', 0.4, '/a/b')]
```

Same inputs, different surviving score, purely from which thread finished first.

## Impact on callers

`UnifiedMemory` consumes `flow.state.final_results` directly (`lib/crewai/src/crewai/memory/unified_memory.py`), then calls `touch_records` on it and emits `MemoryQueryCompletedEvent`, so the nondeterminism reaches user-visible recall results, `last_accessed` bookkeeping, and telemetry.

There is a second, smaller ordering dependence: `matches.sort(key=lambda m: m.score, reverse=True)` is stable, so records with *equal* composite scores keep their arrival order — also `as_completed` order.

## Suggested fix

Keep the **best** score per `record.id` rather than the first, and give the sort a deterministic tiebreak on `record.id`. Best-wins is both order-independent and the score ranking should be using anyway.

## Out of scope

`self.state.confidence = max((f["top_score"] for f in findings), default=0.0)` in `_do_search` is already order-independent (`max` over the whole list) and needs no change.

## Environment

- crewAI `main`
- Python 3.12, Windows 11

I have a fix plus regression tests ready and will open a PR referencing this issue.

LHMQ878:
Closes #6747

## Problem

`RecallFlow.synthesize_results` deduplicated by `record.id` keeping the **first** occurrence, but `_do_search` appends findings inside `for future in as_completed(futures)` — i.e. in thread-completion order. So when the same record arrived from more than one `(sub-query, scope)` task, **thread timing decided which score survived**.

Duplicates are structural rather than incidental:

- `lancedb_storage.search` filters with `query.where(f"scope LIKE '{prefix}%'")`, so a record stored at `/a/b` is returned for scope `/a` **and** `/a/b` — and `filter_and_chunk` routinely selects both.
- `analyze_query_step` yields up to 3 sub-query embeddings and `_do_search` runs the full `embeddings × scopes` cross product, so the same record arrives carrying *different* semantic scores.

`compute_composite_score` is monotonic in `semantic_score`, so keeping the first-arriving duplicate can keep a strictly worse score — under-ranking the record, or dropping it out of `matches[: self.state.limit]`. This reaches users through `UnifiedMemory` (`flow.state.final_results` → `touch_records` → `MemoryQueryCompletedEvent`).

A second, smaller dependence: `sort(key=..., reverse=True)` is stable, so equal-scoring records kept their arrival order too.

## Fix

Keep the **best** score per `record.id` instead of the first, and break sort ties on `record.id`:

```python
best: dict[str, MemoryMatch] = {}
...
    previous = best.get(record.id)
    if previous is None or composite > previous.score:
        best[record.id] = MemoryMatch(record=record, score=composite, match_reasons=reasons)
matches: list[MemoryMatch] = list(best.values())
matches.sort(key=lambda m: (-m.score, m.record.id))
```

Best-wins is both order-independent and the score ranking should have been using. Behaviour that already worked is untouched: dedup still collapses duplicates to one entry, ranking is still by descending composite, `limit` is still applied last, and the defensive shape checks (non-dict finding, non-list `results`, short tuple, non-`MemoryRecord`) are all preserved.

`state.confidence` is deliberately unchanged — `max((f["top_score"] for f in findings), default=0.0)` is already order-independent.

## Tests

New `lib/crewai/tests/memory/test_recall_flow_dedup.py`, 6 tests. Each drives `synthesize_results` on a hand-built `chunk_findings` list rather than through a thread pool, so the assertions pin the dedup rule itself instead of racing the scheduler.

| Test | Pins |
|---|---|
| `test_duplicate_record_keeps_the_best_score_not_the_first_seen` | the regression: `[high, low]` and `[low, high]` both resolve to the higher score |
| `test_equal_scores_rank_by_record_id_rather_than_arrival` | ties order by `record.id` in both a forward and a reversed findings list |
| `test_record_reachable_from_several_scopes_appears_once` | best-wins is not keep-both — 4 scopes still collapse to one match |
| `test_distinct_records_are_all_kept_and_ranked_by_score` | descending-composite ranking did not regress |
| `test_limit_is_applied_after_deduplication` | a duplicate does not consume one of the caller's `limit` slots |
| `test_malformed_findings_are_skipped_without_raising` | the defensive shapes the original loop tolerated are still tolerated |

The composite carries a recency term read from the wall clock at scoring time, so the score assertions compare with `pytest.approx(..., abs=1e-6)` against separately-scored baselines — a tolerance three orders of magnitude below the 0.9/0.1 gap being asserted — rather than against hard-coded floats.

**Control run** — with `git checkout origin/main -- lib/crewai/src/crewai/memory/recall_flow.py` and the new test file in place:

```
2 failed, 4 passed
FAILED test_recall_flow_dedup.py::test_duplicate_record_keeps_the_best_score_not_the_first_seen
FAILED test_recall_flow_dedup.py::test_equal_scores_rank_by_record_id_rather_than_arrival
```

So both new assertions fail on unfixed source, and the four guarding existing behaviour pass on it — they are not tautologies.

**Suite comparison** — `lib/crewai/tests/memory/` before and after, comparing failing test *names*:

```
before: 5 errors (all test_memory_root_scope.py, pre-existing on main)
after : 134 passed, 19 skipped, 5 errors  <- same 5, no new failures
```

## Checks

- `ruff check` on both files: `All checks passed!`
- `ruff format --check` on the source file: already formatted (`lib/crewai/tests/` is ruff-excluded)

### DIFF
--- a/lib/crewai/src/crewai/memory/recall_flow.py
+++ b/lib/crewai/src/crewai/memory/recall_flow.py
@@ -342,9 +342,15 @@ class RecallFlow(Flow[RecallState]):
 
     @listen("synthesize")
     def synthesize_results(self) -> list[MemoryMatch]:
-        """Deduplicate, composite-score, rank, and attach evidence gaps."""
-        seen_ids: set[str] = set()
-        matches: list[MemoryMatch] = []
+        """Deduplicate, composite-score, rank, and attach evidence gaps.
+
+        A record reachable from more than one (sub-query, scope) task is scored
+        once per task, and the scores differ. ``_do_search`` collects findings in
+        ``as_completed`` order, so keeping the first occurrence would let thread
+        timing pick the surviving score. The best score is kept instead, which is
+        both deterministic and the one ranking should use.
+        """
+        best: dict[str, MemoryMatch] = {}
         for finding in self.state.chunk_findings:
             if not isinstance(finding, dict):
                 continue
@@ -356,19 +362,22 @@ class RecallFlow(Flow[RecallState]):
                     record, score = item[0], item[1]
                 else:
                     continue
-                if isinstance(record, MemoryRecord) and record.id not in seen_ids:
-                    seen_ids.add(record.id)
-                    composite, reasons = compute_composite_score(
-                        record, float(score), self._config
+                if not isinstance(record, MemoryRecord):
+                    continue
+                composite, reasons = compute_composite_score(
+                    record, float(score), self._config
+                )
+                previous = best.get(record.id)
+                if previous is None or composite > previous.score:
+                    best[record.id] = MemoryMatch(
+                        record=record,
+                        score=composite,
+                        match_reasons=reasons,
                     )
-                    matches.append(
-                        MemoryMatch(
-                            record=record,
-                            score=composite,
-                            match_reasons=reasons,
-                        )
-                    )
-        matches.sort(key=lambda m: m.score, reverse=True)
+        matches: list[MemoryMatch] = list(best.values())
+        # ``record.id`` breaks ties, so equal-scoring records do not order by
+        # which task happened to finish first either.
+        matches.sort(key=lambda m: (-m.score, m.record.id))
         final_results = matches[: self.state.limit]
         self.state.final_results = final_results
 
--- a/lib/crewai/tests/memory/test_recall_flow_dedup.py
+++ b/lib/crewai/tests/memory/test_recall_flow_dedup.py
@@ -1,0 +1,149 @@
+"""Tests for ``RecallFlow.synthesize_results`` deduplication.
+
+A record reachable from more than one ``(sub-query, scope)`` task is scored once
+per task, and the scores differ. ``_do_search`` appends findings in
+``as_completed`` order, so a dedup that kept the first occurrence let thread
+completion timing decide which score survived — the same query could return the
+same record at a different rank on consecutive runs.
+
+Each test drives ``synthesize_results`` on a hand-built ``chunk_findings`` list
+rather than through a thread pool, so the assertions pin the dedup rule itself
+instead of racing the scheduler.
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+
+import pytest
+
+from crewai.memory.recall_flow import RecallFlow, RecallState
+from crewai.memory.types import MemoryConfig, MemoryRecord
+
+
+_FIXED_CREATED_AT = datetime(2026, 1, 1, 12, 0, 0)
+
+
+def _record(record_id: str) -> MemoryRecord:
+    """A record with everything but ``id`` held constant.
+
+    ``compute_composite_score`` also reads ``created_at`` and ``importance``, so
+    pinning both keeps the composite a function of the semantic score alone.
+    """
+    return MemoryRecord(
+        id=record_id,
+        content=f"content for {record_id}",
+        created_at=_FIXED_CREATED_AT,
+        importance=0.5,
+    )
+
+
+def _flow_with_findings(findings: list[dict]) -> RecallFlow:
+    flow = RecallFlow(
+        storage=object(),
+        llm=object(),
+        embedder=object(),
+        config=MemoryConfig(),
+    )
+    flow._state = RecallState(limit=10, chunk_findings=findings)
+    return flow
+
+
+def test_duplicate_record_keeps_the_best_score_not_the_first_seen() -> None:
+    """The regression: the surviving score must not depend on finding order."""
+    record = _record("R")
+    high = {"scope": "/a", "results": [(record, 0.9)], "top_score": 0.9}
+    low = {"scope": "/a/b", "results": [(record, 0.1)], "top_score": 0.1}
+
+    high_first = _flow_with_findings([high, low]).synthesize_results()
+    low_first = _flow_with_findings([low, high]).synthesize_results()
+    # What the 0.9 and the 0.1 semantic scores are worth on their own. The
+    # composite also carries a recency term read from the wall clock at scoring
+    # time, so these are compared with a tolerance well below the 0.9/0.1 gap
+    # rather than for exact equality.
+    only_high = _flow_with_findings([high]).synthesize_results()[0].score
+    only_low = _flow_with_findings([low]).synthesize_results()[0].score
+
+    assert len(high_first) == 1
+    assert len(low_first) == 1
+    # Both orderings resolve to the higher score, not to whichever came first.
+    assert high_first[0].score == pytest.approx(only_high, abs=1e-6)
+    assert low_first[0].score == pytest.approx(only_high, abs=1e-6)
+    assert only_high - only_low > 0.1
+
+
+def test_record_reachable_from_several_scopes_appears_once() -> None:
+    """Dedup still collapses duplicates; keeping the best score is not keeping both."""
+    record = _record("R")
+    findings = [
+        {"scope": f"/s{i}", "results": [(record, 0.1 * i)], "top_score": 0.1 * i}
+        for i in range(1, 5)
+    ]
+
+    matches = _flow_with_findings(findings).synthesize_results()
+
+    assert [m.record.id for m in matches] == ["R"]
+
+
+def test_equal_scores_rank_by_record_id_rather_than_arrival() -> None:
+    """Ties must not fall back to whichever task finished first."""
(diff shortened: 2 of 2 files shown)


LHMQ878:
Two findings, and they land differently. The flakiness one is right and was mine — fixed in `dac5e8c0`. The invalid-score one describes a real hazard but the patch is at the wrong site and nothing in tree can reach it; declining with the measurements below.

---

### 1. Recency-drift flakiness in the tie test — confirmed, fixed

Right, and worse than "can drift apart": the drift has a **direction**, which is what makes it a flake rather than simply noise. `compute_composite_score` reads `datetime.utcnow()` per call, so of two records created at the same instant the one scored *first* is read a few hundred nanoseconds younger, gets a larger decay, and wins.

Measured on the exact two-record shape the test builds:

```
composite diff over 2000 pairs:  6 nonzero, max 8.1e-11
                                 6 positive (first-scored wins), 0 negative
```

So the test passes for the wrong reason in the forward case — `aaa` is scored first and wins on drift, not on `record.id` — and the *reversed* assertion is the one that breaks. Running the synthesize loop 20000 times:

```
forward order wrong:   0 / 20000  (0.000%)
reverse order wrong: 200 / 20000  (1.000%)   <-- the flake
```

Fixed at the test layer rather than in the product: sub-nanosecond drift is meaningless at a 30-day half-life, and the sort already carries a deterministic tiebreaker — the test was just failing to hold the clock still long enough to reach it. A `_frozen_clock` fixture substitutes a `datetime` subclass whose `utcnow` returns a fixed instant **one day after** `_FIXED_CREATED_AT`, so the decay term is still exercised instead of zeroed.

Two checks that the fixture is actually in effect, not silently a no-op:

```
unfrozen composite: 0.352308425586
frozen   composite: 0.643147990530
expected (1-day age, half_life 30, computed by hand): 0.643147990530   match: True
distinct diffs over 50000 tied pairs under frozen clock: {0.0}
```

Exactly `0.0`, not "small" — the flake is gone by construction. 40 consecutive runs of the file: `0/40 failed`.

I also corrected `_record`'s docstring, which claimed pinning `created_at` and `importance` made the composite a function of the semantic score alone. It does not — the wall clock is the third input — and that wrong claim is why the flake got written in the first place.

### 2. Invalid numeric scores — declining, with the reason

The hazard is real but the suggested guard does not prevent the abort, because the abort happens **earlier than the patched line**. `_do_search` computes `top_score` at `recall_flow.py:134` (and `:163` on the threaded path) by passing the score straight through, with no `float()`:

```python
top_composite, _ = compute_composite_score(results[0][0], results[0][1], self._config)
```

Measured against that call:

```
_do_search top_score with score=None      -> TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
_do_search top_score with score='invalid' -> TypeError: can't multiply sequence by non-int of type 'float'
```

Neither is caught — the surrounding `try/except` wraps `_search_one` / `future.result()`, and `compute_composite_score` sits after it. So a malformed score aborts recall before `synthesize_results` is ever entered, and a guard at :367 would run only on entries that already got past :134.

And nothing in tree can produce one. The `MemoryStorage.search` ABC types the return as `list[tuple[MemoryRecord, float]]`, and both implementations coerce at the source:

- `lancedb_storage.py:404` — `score = 1.0 / (1.0 + float(distance)) if distance is not None else 1.0`
- `qdrant_edge_storage.py:352` — `(sp.payload or {}, float(sp.score), is_local)`

So `None` / `"invalid"` reaching this loop means a third-party backend already violated the typed contract. Hardening for that is worth doing, but it belongs at the boundary that trusts the contract (`_do_search`, or a coercion in `search`'s caller) and it is out of scope for a dedup fix — this PR does not touch `float(score)`, which is pre-existing at `HEAD~1:361-362`.

Happy to open it separately as a `_do_search`-level guard if you'd like it — that version would actually stop the abort, whereas the one suggested here would not.

`isfinite` is a fair separate point: `nan` and `inf` do propagate (`composite=nan` / `composite=inf` respectively). A `nan` score also breaks the sort silently rather than loudly, since every comparison with it is `False`. Same conclusion though — the place to reject it is where the score enters, not where dedup reads it.

Issue #6664, opened 2026-07-26
[BUG] Incorrect number of workspaces in CONTRIBUTING.md


jaceklaskowski:
### Description

In [Repository Structure](./blob/main/.github/CONTRIBUTING.md#repository-structure):

> This is a uv workspace with four packages under lib/:

There are 6 per [pyproject.toml](https://github.com/crewAIInc/crewAI/blob/main/pyproject.toml#L237-L245):

```yaml
[tool.uv.workspace]
members = [
    "lib/crewai",
    "lib/crewai-tools",
    "lib/devtools",
    "lib/crewai-files",
    "lib/cli",
    "lib/crewai-core",
]
```

### Steps to Reproduce

Open the files

### Expected behavior

n/a

### Screenshots/Code snippets

n/a

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

main

### crewAI Tools Version

main

### Virtual Environment

Venv

### Evidence

n/a

### Possible Solution

fix the number

### Additional context

n/a

KXHXK:
I confirmed this is still present on current main: the repository-structure section says four packages and lists four rows, while [tool.uv.workspace] contains six members. I plan to update the count and add the missing crewai-cli and crewai-core rows, keeping the change limited to .github/CONTRIBUTING.md. I also searched open PRs by the issue number and repository-structure wording and found no competing implementation. This contribution is AI-assisted and will carry the required llm-generated label.

KXHXK:
Correction: my earlier open-PR search missed existing implementations because the query was too narrow. A final search by the issue number shows several open PRs, including #6678, that already update the workspace package list. I will not open a competing PR and have removed my unpublished branch. Sorry for the noise.

Issue #6662, opened 2026-07-26
[BUG] Python version discrepancy


jaceklaskowski:
### Description

No idea where to post this, so reporting here as a possible bug 🤷‍♂️ 

There are two different Python versions used:
* [.python-version](blob/main/.python-version): `3.13`
* [.github/workflows/publish.yml](./blob/main/.github/workflows/publish.yml#L34): `python-version: "3.12"`
* [.github/CONTRIBUTING.md#prerequisites](./blob/main/.github/CONTRIBUTING.md#prerequisites): `development targets 3.12`

[pyproject.toml](./blob/main/pyproject.toml#L4) is even more relaxed and claims the following:

```yaml
requires-python = ">=3.10,<3.14"
```

It should be consistent, shouldn't it?

### Steps to Reproduce

Open the files mentioned and compare the values.

### Expected behavior

Consistent version range across the main build files.

### Screenshots/Code snippets

n/a

### Operating System

Ubuntu 20.04

### Python Version

3.10

### crewAI Version

main branch

### crewAI Tools Version

n/a

### Virtual Environment

Venv

### Evidence

already given

### Possible Solution

already stated

### Additional context

n/a

medlouaynjima:
I noticed this issue already has several linked PRs addressing the discrepancy.

The inconsistency is clear: `.python-version` targets 3.13, while `publish.yml` and `CONTRIBUTING.md` reference 3.12, and `pyproject.toml` allows 3.10–3.13.

For consistency, the project should ideally have one clearly defined supported/development Python policy, with `.python-version`, CI/CD, contributor documentation, and `requires-python` aligned accordingly.

I'll avoid claiming the issue since there are already active PRs addressing it.

parthiban-sivakumar:
Hi @jaceklaskowski, @Vidit-Ostwal,

I ran into another instance of this while filing a couple of bug reports today — the issue template's dropdowns are out of step with the rest of the repo.

**Python.** The template offers only `3.10`, `3.11`, `3.12`, but:

- `pyproject.toml`: `requires-python = ">=3.10,<3.14"`
- `.github/workflows/tests.yml`: `python-version: ['3.10', '3.11', '3.12', '3.13']`

So someone on 3.13 — a version CI actually tests — can't select it and ends up picking a version they aren't running. I'm on 3.13.13 and had to put the real one in Additional context both times.

[...]

Issue #6643, opened 2026-07-25
[BUG] 20 tests fail in crewai-cli-1.15.6


yurivict:
### Description

```
_________________________________________________________________________ test_json_create_provider_preselects_default_model _________________________________________________________________________

tmp_path = PosixPath('/tmp/pytest-of-yuri/pytest-784/test_json_create_provider_pres0'), monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x21f7f71c0680>

    def test_json_create_provider_preselects_default_model(tmp_path, monkeypatch):
        monkeypatch.chdir(tmp_path)
        with mock.patch(
            "crewai_cli.create_json_crew._wizard_agents_and_tasks"
        ) as mock_wizard:
            mock_wizard.return_value = (
                [
                    {
                        "name": "researcher",
                        "role": "Researcher",
                        "goal": "Research",
                        "backstory": "Researcher",
                        "llm": "openai/gpt-5.5",
                        "tools": [],
                        "planning": False,
                        "allow_delegation": False,
                    }
                ],
                [
                    {
                        "name": "research_task",
                        "description": "Research",
                        "expected_output": "Findings",
                        "agent": "researcher",
                        "context": [],
                    }
                ],
                {"process": "sequential", "memory": False, "inputs": {}},
            )
    
            json_crew.create_json_crew("JSON Crew", provider="openai", skip_provider=True)
    
        mock_wizard.assert_called_once_with(
            skip_provider=True,
            default_llm="openai/gpt-5.5",
        )
        assert (tmp_path / "json_crew" / "crew.jsonc").exists()
        assert not (tmp_path / "json_crew" / "src").exists()
        assert not (tmp_path / "json_crew" / "tests").exists()
        assert not (tmp_path / "json_crew" / "config.jsonc").exists()
        generated_paths = {
            path.relative_to(tmp_path / "json_crew").as_posix()
            for path in (tmp_path / "json_crew").rglob("*")
            if path.is_file()
        }
        assert not any(
            path.endswith("/crew.py") or path == "crew.py" for path in generated_paths
        )
        assert not any(
            path.endswith("/agents.yaml") or path == "agents.yaml"
            for path in generated_paths
        )
        assert not any(
            path.endswith("/tasks.yaml") or path == "tasks.yaml"
            for path in generated_paths
        )
        assert not any(path.startswith("src/") for path in generated_paths)
    
[...]

PR #6751, opened 2026-07-31
fix(tools/cli/docs): JSON serialization, dep range, workspace docs

Diwak4r:
## Summary

Four independent fixes across tools, CLI, and documentation.

**docs: workspace count and Python dev target (#6664 + #6662)**
- `CONTRIBUTING.md` table listed four packages; the uv workspace now has six (`cli` and `crewai-core` were added). Updated the table.
- Python development target read 3.12 while `.python-version` pins 3.13. Updated docs to match.

**fix(tools): serialize dict/list outputs as JSON (#6267)**
- Plain `dict`/`list` results without a `result_schema` were formatted with `str()`, producing Python repr (single quotes, `True`/`False`/`None`) instead of valid JSON.
- Changed `_format_tool_output_for_agent` to use `json.dumps(raw_result, default=str)` for dict/list, falling back to `str()` only on `ValueError` (circular references). Applied to both the no-schema path and the schema-validation exception fallback.

**fix(experimental): ToolFailure context on error path (#6611 follow-up)**
- Error dict in the parallel tool execution path was missing `tool_failure`, so downstream failure-policy handling was skipped on exceptions.

**fix(cli): minor-baseline for generated dependency range (#6643)**
- `get_crewai_dependency_range("1.15.6")` returned `>=1.15.6,<2.0.0` instead of `>=1.15.0,<2.0.0`, breaking 20 CLI tests and causing generated projects to fail resolution on every patch bump.
- Normalize lower bound to `major.minor.0`. Add `encoding="utf-8"` when reading generated files in tests (Windows fix).

## Test plan

- `TestToolOutputSchema` — 14 tests, all green
- `test_generated_project_dependency_uses_minor_baseline_for_patch_versions` — green against source
- `test_generated_project_dependency_uses_next_major_upper_bound` — green

> This PR carries the `llm-generated` label per CONTRIBUTING.md requirements.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -12,7 +12,7 @@ Thank you for your interest in contributing to CrewAI. This guide covers everyth
 
 ## Prerequisites
 
-- Python 3.10–3.14 (development targets 3.12)
+- Python 3.10–3.14 (development targets 3.13)
 - [uv](https://docs.astral.sh/uv/) for package management
 - [pre-commit](https://pre-commit.com/) for Git hooks
 
@@ -29,7 +29,7 @@ uv run pre-commit install
 
 ## Repository Structure
 
-This is a uv workspace with four packages under `lib/`:
+This is a uv workspace with six packages under `lib/`:
 
 | Package | Path | Description |
 |---------|------|-------------|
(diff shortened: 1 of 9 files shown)


Diwak4r:
Addressed the CodeRabbit and Copilot review feedback in the latest push:

**Cancellation timing** — replaced \syncio.gather(return_exceptions=True)\ with \syncio.wait(return_when=FIRST_EXCEPTION)\ so a \ToolExecutionFailedError\ surfaces immediately. Pending siblings are now cancelled before they execute rather than after all tasks have already completed.

**Test quality** — replaced the \inspect.getsource\ assertion (which only checked that the words existed in source code) with a behavioral test: spawns a failing tool and a slow sibling concurrently, then asserts the sibling never records a start when the first task raises \ToolExecutionFailedError\.

The sync duplicate of the dispatch body (CodeRabbit nitpick) is a separate concern I can address in a follow-up once the async path is confirmed correct — happy to add it here if preferred.

Vidit-Ostwal (commented):
Closing as a duplicate of the focused PRs: #6678 (workspace docs), #6663 (Python version), and #6647 (CLI dependency range). This branch also bundles unrelated tools/CLI changes. Follow along on those three.

PR #6757, opened 2026-07-31
docs: add approval-bound tool execution example

jw-ond:
## Summary
- Adds a small approval-bound side-effect example to the Tool Call Hooks guide
- Shows how to build a stable action object and SHA-256 digest before tool execution
- Maps external checkpoint decisions to allow, reviewer approval, or HookAborted block behavior

## Checks
- Parsed all Python code blocks in docs/edge/en/learn/tool-hooks.mdx with ast.parse
- Verified heading numbering in the Common Use Cases section
- Verified the fork branch file matches the local edited file

### DIFF
--- a/docs/edge/en/learn/tool-hooks.mdx
+++ b/docs/edge/en/learn/tool-hooks.mdx
@@ -185,7 +185,7 @@ ApprovalDecision = Literal["allow", "require_approval", "deny"]
 
 
 def stable_digest(action: dict[str, Any]) -> str:
-    payload = json.dumps(action, sort_keys=True, separators=(",", ":"))
+    payload = json.dumps(action, sort_keys=True, separators=(",", ":"), default=str)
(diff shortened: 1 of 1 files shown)


jw-ond:
Thanks for the review. The current PR head already applies the suggested `default=str` fallback in both places: `stable_digest()` preserves the existing sorted compact serialization, and the reviewer prompt serialization preserves the existing indentation/sorting. That should keep the example from raising `TypeError` on non-JSON-native tool input values while keeping the digest behavior deterministic for common primitive inputs.

jw-ond:
Following up on the CodeRabbit note: the current PR head already uses `default=str` for both digest serialization and the human-readable prompt serialization, so non-JSON-native values such as datetimes, UUIDs, Decimals, Enums, or Paths should no longer raise during the hook example.

No actionable review comments remain from the latest automated pass. If maintainers prefer a different placement or shorter example, I can trim it.

PR #6758, opened 2026-07-31
fix(agents): recover real tool calls from fabricated observation continuations

HUAN2022A:
## Summary

`process_llm_response` had a dead recovery block, so when a model that does not support stop words fabricated an `Observation:` + `Final Answer:` continuation after a real `Action`, the fabricated answer won and the real tool call was never executed. This PR replaces the dead exception path with an explicit shape check: when stop words are unsupported and a `Final Answer:` follows a parsed `Action`, everything from that `Final Answer:` onward is dropped so the real action executes.

Related: #3154, #6449.

Note: the issue author's PR #6450 is stale and unmerged; this is an independent implementation with a slightly different truncation strategy.

## Validation

- Reproduction confirmed on unmodified `origin/main`
- `test_agent_utils.py`: 78/78 passed, including 5 new regression tests
- `ruff` clean on changed files

### DIFF
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -17,7 +17,10 @@ from crewai_core.settings import Settings
 from pydantic import BaseModel
 from rich.console import Console
 
-from crewai.agents.constants import FINAL_ANSWER_AND_PARSABLE_ACTION_ERROR_MESSAGE
+from crewai.agents.constants import (
+    ACTION_INPUT_REGEX,
+    FINAL_ANSWER_ACTION,
+)
 from crewai.agents.parser import (
     AgentAction,
     AgentFinish,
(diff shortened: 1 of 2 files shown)


HUAN2022A:
Addressed in 58c866214: the recovery now requires a line-delimited `Observation:` after the action followed by a `Final Answer:`, and truncates at the `Observation:` marker so the fabricated Observation/Thought suffix can no longer leak into the parsed tool input via the DOTALL action-input capture. A literal `Final Answer:` inside a JSON action input no longer truncates valid arguments. Added regression assertions for both (`test_recovered_response_excludes_fabricated_continuation`, `test_final_answer_inside_action_input_is_not_truncated`).

HUAN2022A:
Addressed in bc20ac05c: the recovery now only fires when the candidate action input between the capture start and the line-delimited `Observation:` marker parses as a complete JSON value — proof that the real input ended before the marker. Free-form multiline input (including one carrying a line-start `Observation:` and a later `Final Answer:`) is preserved untouched, with a regression test covering exactly that case. The fabricated-continuation tests (JSON tool inputs) still recover as before.

HUAN2022A:
Hi maintainers 👋

The CI workflows on this PR (Lint, Run Type Checks, Run Tests, etc.) are currently stuck in `action_required` and haven't executed since the latest push. Could someone please approve the workflow runs so the checks can run?

Happy to address anything that comes up once they execute. Thanks!

PR #6770, opened 2026-08-02
fix(anthropic): count billed cache tokens in token usage (#6768)

### LINKED ISSUE #6768: Conversational Flow golden use case improvements - Anthropic total_tokens omits cache read/write tokens, undercounting cost
## Description

For the Anthropic provider, `total_tokens` is computed as `input_tokens + output_tokens` only. Anthropic reports cache reads (`cache_read_input_tokens`) and cache writes (`cache_creation_input_tokens`) as **separate** counters that are *not* included in `input_tokens` — and both are billed (cache writes at 1.25×, cache reads at 0.1×).

The result: any prompt-cached request reports a `total_tokens` that is wildly lower than what was actually billed. Anyone computing cost from CrewAI's usage metrics undercounts, and the undercount grows with how well caching works.

This bit us while instrumenting per-session cost for a conversational Flow app — we ended up wrapping the Anthropic SDK ourselves to get real numbers. (Filing under the same banner as our other conversational-flow findings, though the bug itself is provider-level, not conversational-specific.)

### Root cause

`crewai/llms/providers/anthropic/completion.py` ~L1971:

```python
cache_creation_tokens = (
    getattr(usage, "cache_creation_input_tokens", 0) or 0
)
result: dict[str, Any] = {
    "input_tokens": input_tokens,
    "total_tokens": input_tokens + output_tokens,   # <-- omits both cache counters
}
```

Both cache values are captured correctly and surfaced individually — they're just excluded from the total.

## Steps to Reproduce

1. Make an Anthropic call large enough to cache, with a `cache_breakpoint`.
2. Repeat it so the second call is a cache read.
3. Compare `total_tokens` against the sum of all four counters.

## Expected behavior

`total_tokens` should account for every billed token, i.e.

```python
total_tokens = input_tokens + output_tokens + cache_read_tokens + cache_creation_tokens
```

(or, if `total_tokens` is deliberately meant to exclude cache traffic, that should be documented loudly, since it makes the field unusable for cost estimation on any cached workload.)

## Screenshots/Code snippets

```python
"""Minimal repro: total_tokens omits Anthropic cache read/write tokens."""
from crewai import LLM

llm = LLM(model="anthropic/claude-haiku-4-5", max_tokens=64)
BIG = "You are a helpful assistant.\n" + ("Filler context line for the cache. " * 500)

for i in (1, 2):
    llm.call([
        {"role": "system", "content": BIG, "cache_breakpoint": True},
        {"role": "user", "content": "Reply with the single word OK."},
    ])
    u = llm._token_usage
    billed = (u["prompt_tokens"] + u["completion_tokens"]
              + u["cached_prompt_tokens"] + u["cache_creation_tokens"])
    print(f"call {i}: reported total_tokens={u['total_tokens']:5} | "
          f"prompt={u['prompt_tokens']:5} completion={u['completion_tokens']:4} "
          f"cache_write={u['cache_creation_tokens']:5} cache_read={u['cached_prompt_tokens']:5} "
          f"-> actually billed={billed}")
```

Actual output (counters are cumulative across calls):

```
call 1: reported total_tokens=   17 | prompt=   13 completion=   4 cache_write= 4509 cache_read=    0 -> actually billed=4526
call 2: reported total_tokens=   34 | prompt=   26 completion=   8 cache_write= 4509 cache_read= 4509 -> actually billed=9052
```

`total_tokens` reports **17** where **4526** tokens were billed — a 266× undercount on the first call.

The same pattern is visible in production traces. An `llm_call_completed` event from one of our agent turns:

```json
"usage": {
  "input_tokens": 2,
  "total_tokens": 42,
  "output_tokens": 40,
  "cached_prompt_tokens": 0,
  "cache_creation_tokens": 2285
}
```

2285 billed cache-write tokens, reported total of 42.

## Environment

- crewai `1.15.10`
- Python 3.12
- macOS (Darwin 25.1)
- Provider: `anthropic` (native provider path, not LiteLLM)

Anai-Guo:
Fixes #6768.

## Problem

Anthropic reports cache reads (`cache_read_input_tokens`) and cache writes (`cache_creation_input_tokens`) as counters that are **separate** from `input_tokens`, and bills both (writes at 1.25×, reads at 0.1×). `_extract_anthropic_token_usage` captured both values and surfaced them individually, but left them out of the totals — so every prompt-cached request reported a `total_tokens` well below what was billed, and the undercount grew with cache hit rate.

## Why the one-line fix in the issue isn't enough

The issue points at `total_tokens: input_tokens + output_tokens`. Fixing only that line does **not** change what users actually read (`llm._token_usage` / `get_token_usage_summary()`), because `UsageMetrics.from_provider_dict` discards the provider's total and recomputes it:

```python
return cls(
    total_tokens=prompt_tokens + completion_tokens,   # provider total_tokens ignored
    ...
)
```

Verified against `main`:

```
provider dict total_tokens : 200      # after the issue's suggested one-line fix
UsageMetrics.total_tokens  : 150      # unchanged
```

So the cache tokens have to be part of the **prompt count** to survive normalization.

## Fix

Report the *billed* input from the Anthropic extractor:

```python
billed_input_tokens = input_tokens + cache_read_tokens + cache_creation_tokens
```

`cached_prompt_tokens` and `cache_creation_tokens` stay as they are, now as a breakdown of that prompt count. This is the convention the other providers already follow — OpenAI's `prompt_tokens` and Gemini's `prompt_token_count` both include cached tokens and expose the cached figure separately — and it is exactly what `from_provider_dict` assumes. It also keeps the change inside the Anthropic provider: no shared normalizer change, so OpenAI/Gemini/Azure numbers are untouched.

## Verification

Ran an old-vs-new harness against the installed `crewai==1.15.9`, whose `completion.py`, `usage_metrics.py` and `base_llm.py` are byte-identical to `main`:

| case (input / output / cache read / cache write) | `total_tokens` before | after |
|---|---|---|
| 100 / 50 / 30 / 20 | 150 | **200** |
| 40 / 20 / 0 / 0 | 60 | 60 |
| missing cache fields (`None`) | 60 | 60 |
| 10 / 5 / 0 / 200 (cache write only) | 15 | **215** |

Non-cached requests are unaffected; the two cached rows were undercounting 50 and 200 billed tokens respectively.

Tests (run against the patched module, then re-run against unpatched `main` to confirm they actually guard the regression):

- `test_anthropic_total_tokens_includes_cache_tokens` — new; asserts end-to-end through `_track_token_usage_internal` → `get_token_usage_summary()`, which is the path the issue's repro reads.
- `test_anthropic_total_tokens_unchanged_without_cache` — new; pins the no-cache path.
- `test_anthropic_cache_creation_tokens_extraction` — existing; its `input_tokens`/`total_tokens` numbers moved to 150/200. That test's subject is cache-field *extraction* (`cached_prompt_tokens == 30`, `cache_creation_tokens == 20`, both unchanged); the totals were incidental and encoded the undercount.
- `test_anthropic_missing_cache_fields_default_to_zero` — existing, unchanged, still passes.

```
4 passed          # patched
2 failed, 2 passed  # unpatched main — the two cache tests fail
```

`ruff==0.15.1 format --check` clean; `ruff check` output on the touched file is identical to `main`.

## Note

If you'd rather `input_tokens` keep mirroring the raw Anthropic field, the alternative is to have `from_provider_dict` honor an explicit provider `total_tokens`. I did not take that route: it is shared code, and Gemini's `total_token_count` includes thinking tokens, so honoring it there would move Gemini's numbers too — a bigger change than this issue calls for. Happy to switch if you prefer that shape.

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -1971,10 +1971,19 @@ class AnthropicCompletion(BaseLLM):
             cache_creation_tokens = (
                 getattr(usage, "cache_creation_input_tokens", 0) or 0
             )
+            # Anthropic reports cache reads/writes as counters *separate* from
+            # ``input_tokens``, and bills both (writes at 1.25x, reads at 0.1x).
+            # Other providers (OpenAI, Gemini) already fold cached tokens into
+            # their prompt count and expose the cached figure as a breakdown, so
+            # report the billed input here to match. Leaving them out made
+            # ``total_tokens`` undercount every prompt-cached request.
+            billed_input_tokens = (
+                input_tokens + cache_read_tokens + cache_creation_tokens
+            )
             result: dict[str, Any] = {
-                "input_tokens": input_tokens,
+                "input_tokens": billed_input_tokens,
                 "output_tokens": output_tokens,
-                "total_tokens": input_tokens + output_tokens,
+                "total_tokens": billed_input_tokens + output_tokens,
                 "cached_prompt_tokens": cache_read_tokens,
                 "cache_creation_tokens": cache_creation_tokens,
             }
--- a/lib/crewai/tests/llms/anthropic/test_anthropic.py
+++ b/lib/crewai/tests/llms/anthropic/test_anthropic.py
@@ -1653,11 +1653,68 @@ def test_anthropic_cache_creation_tokens_extraction():
     mock_response.model = None
 
     usage = llm._extract_anthropic_token_usage(mock_response)
-    assert usage["input_tokens"] == 100
+    # Anthropic keeps cache reads/writes out of ``input_tokens``; both are
+    # billed, so the reported prompt count is the billed input (100 + 30 + 20)
+    # and the total covers every billed token.
+    assert usage["input_tokens"] == 150
     assert usage["output_tokens"] == 50
-    assert usage["total_tokens"] == 150
+    assert usage["total_tokens"] == 200
     assert usage["cached_prompt_tokens"] == 30
     assert usage["cache_creation_tokens"] == 20
+
+
+def test_anthropic_total_tokens_includes_cache_tokens():
+    """Cache reads/writes are billed, so they must reach total_tokens.
+
+    Regression test for the undercount reported in #6768: ``total_tokens`` was
+    ``input_tokens + output_tokens``, which omitted both Anthropic cache
+    counters and made the field unusable for cost estimation on any cached
+    workload.
+    """
+    llm = LLM(model="anthropic/claude-3-5-sonnet-20241022")
+
+    mock_response = MagicMock()
+    mock_response.content = [MagicMock(text="test response")]
+    mock_response.usage = MagicMock(
+        input_tokens=100,
+        output_tokens=50,
+        cache_read_input_tokens=30,
+        cache_creation_input_tokens=20,
+    )
+    mock_response.stop_reason = None
+    mock_response.model = None
+
+    usage = llm._extract_anthropic_token_usage(mock_response)
+    llm._track_token_usage_internal(usage)
+    summary = llm.get_token_usage_summary()
+
+    # The shared normalizer recomputes the total as prompt + completion, so the
+    # billed input has to be in ``input_tokens`` for the total to survive it.
+    assert summary.total_tokens == 200
+    assert summary.prompt_tokens == 150
+    assert summary.completion_tokens == 50
+    assert summary.cached_prompt_tokens == 30
+    assert summary.cache_creation_tokens == 20
+
+
+def test_anthropic_total_tokens_unchanged_without_cache():
+    """A request that touches no cache keeps input + output as the total."""
+    llm = LLM(model="anthropic/claude-3-5-sonnet-20241022")
+
+    mock_response = MagicMock()
+    mock_response.content = [MagicMock(text="test response")]
+    mock_response.usage = MagicMock(
+        input_tokens=40,
+        output_tokens=20,
+        cache_read_input_tokens=0,
+        cache_creation_input_tokens=0,
+    )
+    mock_response.stop_reason = None
+    mock_response.model = None
+
+    usage = llm._extract_anthropic_token_usage(mock_response)
+    assert usage["input_tokens"] == 40
+    assert usage["total_tokens"] == 60
 
 
 def test_anthropic_missing_cache_fields_default_to_zero():


ErenAta16 (commented):
The direction is right and, unusually for this kind of fix, it is complete in one edit. `_extract_anthropic_token_usage` is the only place `cache_read_input_tokens` and `cache_creation_input_tokens` are read in `completion.py`, and it is called from all six paths:

```
1006  sync completion
1240  streaming final_message
1445  tool-call follow-up
1554  async completion
1764  async streaming final_message
1870  async tool-call follow-up
```

so streaming and the tool follow-ups are covered without separate edits. That is worth stating in the PR body, since the usual failure mode for a usage fix is landing it on the non-streaming path only.

**The detail that makes the fix necessary rather than cosmetic** is the one your second test pins: the shared normalizer recomputes the total as prompt plus completion, so putting the cache counts anywhere other than `input_tokens` would have them silently dropped again one layer up. Asserting through `get_token_usage_summary()` rather than only on the dict is what actually proves that, and it is the assertion I would protect if this ever gets refactored.

**One correction to the stated motivation.** The comment cites the billing multipliers:

> writes at 1.25x, reads at 0.1x

and the test docstring says the field was "unusable for cost estimation". Both are true, but folding at face value does not make `total_tokens` usable for cost estimation either. 100 cache-read tokens and 100 fresh input tokens now contribute equally to `total_tokens` while differing tenfold in price. What the fix actually delivers is a correct *token* count, which is the right thing for a field named `total_tokens` to be, and it makes `cached_prompt_tokens` and `cache_creation_tokens` load-bearing rather than informational, since a cost estimate has to reconstruct the split from them. I would reword the docstring to claim the token count rather than the cost, otherwise the next person reading it will assume `total_tokens * price` is meaningful.

**The double-count trap, and why it deserves a changelog line.** `input_tokens` now includes both cache counters while `cached_prompt_tokens` and `cache_creation_tokens` still sit beside it. Anyone who wrote

```python
usage["input_tokens"] + usage["cached_prompt_tokens"]
```

against the current shape, which was the correct way to get billed input before this PR, now double counts. The changed assertion in `test_anthropic_cache_creation_tokens_extraction` from `100` to `150` is the honest signal that this is a public-number change rather than a bug fix that nobody can observe. Existing dashboards will show a step change on cached workloads. That is the right outcome, but it should be announced rather than discovered.

`test_anthropic_total_tokens_unchanged_without_cache` is a good inclusion for exactly that reason: it draws the line showing uncached workloads see no movement at all, which is what most users will want to confirm first.

Vidit-Ostwal:
Thanks @Anai-Guo, for this PR, I have added some additional changes on the PR.
Closing this one at this moment.

PR #6771, opened 2026-08-02
fix(anthropic): include cache tokens in total usage

### LINKED ISSUE #6768: Conversational Flow golden use case improvements - Anthropic total_tokens omits cache read/write tokens, undercounting cost

gambletan:
## Summary

- include Anthropic cache-read and cache-creation tokens in provider-level total usage
- preserve the cache-inclusive total through shared UsageMetrics normalization and Flow aggregation
- keep OpenAI-style cached prompt accounting unchanged by applying the extra counters only to the Anthropic input_tokens shape

## Problem

Anthropic reports cache reads and cache writes separately from input_tokens. CrewAI exposed those counters but calculated total_tokens from input_tokens + output_tokens only, substantially undercounting cached workloads. The shared UsageMetrics normalizer then recomputed the same incomplete total, so public LLM summaries and Flow metrics were affected as well.

## Testing

- 76 Anthropic provider and Flow usage tests passed
- 29 usage-shape and LLM usage-event tests passed
- ruff check passed for all changed files
- ruff format --check passed for changed source files
- git diff --check passed
- two independent Codex review passes; the first identified the shared normalizer gap, which is covered by the final tests

Fixes #6768

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -1974,7 +1974,12 @@ class AnthropicCompletion(BaseLLM):
             result: dict[str, Any] = {
                 "input_tokens": input_tokens,
                 "output_tokens": output_tokens,
-                "total_tokens": input_tokens + output_tokens,
+                "total_tokens": (
+                    input_tokens
+                    + output_tokens
+                    + cache_read_tokens
+                    + cache_creation_tokens
+                ),
                 "cached_prompt_tokens": cache_read_tokens,
                 "cache_creation_tokens": cache_creation_tokens,
             }
--- a/lib/crewai/src/crewai/types/usage_metrics.py
+++ b/lib/crewai/src/crewai/types/usage_metrics.py
@@ -145,12 +145,18 @@ class UsageMetrics(BaseModel):
             if isinstance(details, dict):
                 cached_prompt_tokens = _coerce_int(details.get("cached_tokens"))
 
+        cache_creation_tokens = _coerce_int(usage_data.get("cache_creation_tokens"))
+        total_tokens = prompt_tokens + completion_tokens
+        if "input_tokens" in usage_data:
+            # Anthropic reports cache reads and writes separately from input_tokens.
+            total_tokens += cached_prompt_tokens + cache_creation_tokens
+
         return cls(
-            total_tokens=prompt_tokens + completion_tokens,
+            total_tokens=total_tokens,
             prompt_tokens=prompt_tokens,
             completion_tokens=completion_tokens,
             cached_prompt_tokens=cached_prompt_tokens,
             reasoning_tokens=_coerce_int(usage_data.get("reasoning_tokens")),
-            cache_creation_tokens=_coerce_int(usage_data.get("cache_creation_tokens")),
+            cache_creation_tokens=cache_creation_tokens,
             successful_requests=1,
         )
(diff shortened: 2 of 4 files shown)


gambletan:
Closing as a duplicate of #6770, which was opened first and covers the same issue and end-to-end UsageMetrics path. This PR was created after a final duplicate-check/create command did not abort when the issue timeline returned #6770. Thanks @Anai-Guo for the thorough fix.

PR #6775, opened 2026-08-02
fix(agents): keep null in the task output schema embedded in the prompt

### LINKED ISSUE #6774: [BUG] Task output schema in the prompt marks Optional fields as required and strips null, so the model cannot express "not applicable"
### Description

`build_task_prompt_with_schema` embeds a JSON schema of `task.output_pydantic` into the
prompt. That schema is produced by `generate_model_description`, whose default
`strip_null_types=True` removes `null` from every `anyOf`, while
`ensure_all_properties_required` adds every property to `required`.

The combined result tells the model that an `Optional[str] = None` field is a
**required, non-nullable string**. The model is then given no legal way to say "this
field does not apply to this item" — while the provider-side response schema, generated
straight from the same Pydantic model, still says the field is optional and nullable.

Two contradictory contracts reach the model in one request. Note also that the emitted
property is internally inconsistent on its own: `"type": "string"` sits next to
`"default": null`.

That sanitizer exists for **OpenAI strict function-calling schemas**. This call site is
not a function-calling schema — it is prose inside a prompt, where strict-mode
constraints do not apply.

### Reproducer

```python
# pip install crewai==1.15.10
import json
from typing import Optional

from crewai import Task
from crewai.agent.utils import build_task_prompt_with_schema
from crewai.utilities.pydantic_schema_utils import generate_model_description


class Step(BaseModel):
    name: str = Field(description="What the step does")
    note: Optional[str] = Field(default=None, description="Optional note")


class Plan(BaseModel):
    steps: list[Step]


native = Plan.model_json_schema()
print("pydantic  required:", native["$defs"]["Step"]["required"])
print("pydantic  note    :", json.dumps(native["$defs"]["Step"]["properties"]["note"]))

prompt_schema = generate_model_description(Plan)["json_schema"]["schema"]
step = prompt_schema["properties"]["steps"]["items"]
print("in-prompt required:", step["required"])
print("in-prompt note    :", json.dumps(step["properties"]["note"]))

task = Task(description="Produce a plan.", expected_output="JSON.", output_pydantic=Plan)
print(build_task_prompt_with_schema(task, ""))
```

### Actual

```
pydantic  required: ['name']
pydantic  note    : {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, ...}

in-prompt required: ['name', 'note']
in-prompt note    : {"default": null, "type": "string", ...}
```

### Expected

The schema shown to the model should permit `null` for fields the model itself declares
optional — as it did before. On `crewai==1.8.1` the same model produced
`anyOf: [{"type": "string"}, {"type": "null"}]` in the prompt. We did not bisect the exact
release in which this changed.

### Impact

On a task whose output model has several optional fields, the model has to put *something*
non-null in every one of them. In our case a planner started concatenating unrelated
parameters into a single required string field, dropped a step it would otherwise have
emitted, and in a few runs extended one JSON string value until it hit the output-token
ceiling. One run wrote its confusion into the payload itself:

> `"value": "... Wait, the schema requires step_description, element_description, value, k`

The failure is silent — the task still returns and still validates, because the
provider-side schema does allow the nulls that never get emitted.

### Proposed fix

Pass the flag the codebase already ships. There is precedent in the same tree:
`lib/crewai/src/crewai/utilities/agent_utils.py` already calls
`generate_model_description(tool.args_schema, strip_null_types=False)` for tool schemas.

```python
schema_dict = generate_model_description(task.output_pydantic, strip_null_types=False)
```

With that flag, `note` becomes `anyOf: [{"type": "string"}, {"type": "null"}]`. `required`
still lists every property, which is now harmless and is the standard strict-mode idiom:
the model must emit the key, and may emit `null`.

### Secondary observation

1.8.1 appended a short output-format priming block (ending *"I MUST use these formats, my
job depends on it!"*) that the current executor no longer applies, though the string is
still shipped in `crewai/translations/en.json`. Possibly intentional; flagging in case a
step was lost.

### Environment

`crewai==1.15.10`, Python 3.13, LiteLLM-backed `LLM`.

monkscode:
`build_task_prompt_with_schema` embeds the task output schema into the prompt using
`generate_model_description`, whose `strip_null_types` defaults to `True`. Combined with
`ensure_all_properties_required`, an `Optional[str] = None` field is presented to the
model as a required, non-nullable string — leaving it no way to express "not
applicable", and contradicting the provider-side response schema generated from the
same model.

That sanitizer targets OpenAI strict function-calling schemas. This call site produces
prompt prose, where those constraints do not apply.

This passes `strip_null_types=False`, matching the existing call for tool schemas at
`utilities/agent_utils.py:268`. `required` still lists every property, which is the
standard strict-mode idiom once the type is nullable.

### Before / after

For `class Step(BaseModel): name: str; note: Optional[str] = None`:

```
before   note: {"default": null, "type": "string"}                              <- cannot be null
after    note: {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null}
```

`required` is `['name', 'note']` in both cases — unchanged by this PR.

### Tests

Adds `lib/crewai/tests/agents/test_agent_utils.py`. No existing test referenced
`build_task_prompt_with_schema`, so the file is new; maintainers may prefer it
elsewhere. It fails on `main` with `KeyError: 'anyOf'` and passes with this change.

Existing cassette-based tests are unaffected: `vcr_config` matches on
`["method", "scheme", "host", "port", "path"]`, not on the request body.

Fixes #6774


monkscode:
@lorenzejay @greysonlalonde — flagging this for triage since it touches `agent/utils.py`, which you two have most recently worked in.

One concrete blocker before the change itself: **CI has never run here.** All seven workflows on the head commit sit at `action_required` (first-time-contributor fork gate), so CodeRabbit is the only check and this reads as unverified. One "Approve and run workflows" click would give it a real signal.

The change is two lines, and there is precedent in-tree: `utilities/agent_utils.py:268` already passes `strip_null_types=False` for tool schemas. This does the same at both branches of `build_task_prompt_with_schema`. `required` is unchanged — every property is still listed, which is the standard strict-mode idiom once the type is nullable.

On why it is worth the look: the bug is silent, not cosmetic. The provider-side response schema still permits nulls, so every task validates and returns normally — the damage is in what the model writes to avoid emitting a null it has been told is illegal. On our planner task we measured unrelated parameters concatenated into one required string field, a step silently dropped, and some runs extending a single JSON string value until it hit the output-token ceiling. Our end-to-end pass rate went 96.7% → 86.7% on 1.15.10, which is why we are still pinned to 1.8.1.

Happy to move or split the test file, or rebase — just say where.

percymcn:
Ran #6774's repro on `crewai==1.15.14` and simulated this diff verbatim before writing. It does what it says — post-fix the model can express "not applicable" again. Four things I measured that aren't visible from the diff.

**1. This fixes the nullability half of #6774; the `required` half survives.**

```
pydantic    required: ['name']
pre-fix     required: ['name', 'note']
POST-FIX    required: ['name', 'note']

POST-FIX    note: {"anyOf": [{"type": "string"}, {"type": "null"}], "default": null, ...}
```

`ensure_all_properties_required` runs regardless of `strip_null_types`. To be fair to the diff: required-but-nullable is the ordinary strict-mode encoding and the model *can* now emit `null`, so the user-facing harm in the issue is genuinely gone. But #6774's other complaint is that the in-prompt and provider-side contracts disagree, and on `required` they still do — `['name']` vs `['name','note']`. Worth deciding deliberately rather than by omission; the new test doesn't assert on `required`.

**2. Eight call sites feed this schema into the same prompt slice. This fixes two.**

`I18N.slice("formatted_task_instructions")` is also fed by `utilities/converter.py:549` and `:556`, `lite_agent.py:698`, `agent/core.py:1841`, and `utilities/evaluators/task_evaluator.py:100` / `:174` — none of which pass `strip_null_types=False`. `agents/agent_adapters/base_converter_adapter.py:152`/`:154` return the same description to their caller. So after this merges, the same prompt built through `Converter`, `LiteAgent` or the task evaluator still strips null.

Useful precedent already in the tree: `utilities/agent_utils.py:267` passes `strip_null_types=False` for tool parameters. The pattern exists; it's applied at one of the sites that needs it.

**3. `strip_null_types` isn't the only pass that shouldn't run on a prompt, and the other one destroys data.**

`generate_model_description` also runs `force_additional_properties_false`, which is unconditional on any `type == "object"` node. A `Dict[str, str]` field:

```
raw      : {"type": "object", "additionalProperties": {"type": "string"}, "title": "Tags"}
pre-fix  : {"type": "object", "additionalProperties": false, "properties": {}, "required": [], "title": "Tags"}
POST-FIX : (unchanged)
```

`additionalProperties: false` on a node with no `properties` doesn't *close* the object, it *empties* it — the only legal instance is `{}`. The prompt therefore tells the model that a dictionary field must be empty, which is the same class of contradiction #6774 reports, one step worse: the model can comply perfectly and still produce nothing.

Two details that make this hard to catch later:

- The pass also **adds** `properties: {}` and `required: []`, so the emptied map comes out byte-identical to a genuinely empty `BaseModel`. Nothing downstream can tell the two apart.
- On the strict paths, OpenAI's own transformer (`openai@7.4.0`, `openai/lib/transform` → `toStrictJsonSchema()`) **throws** on the raw open map (`must set additionalProperties: false`) and **accepts the emptied form verbatim**. So the sanitizer turns a precise, fixable 400 into a silent 200 with a field that can never be populated. That function is an offline oracle — no API key — if you want to pin any of this in CI.

**4. The new test passes either way on that shape.** `_Output` is `name: str` + `note: str | None`, so it stays green while the Dict damage remains. Adding `tags: dict[str, str]` to `_Output`, or a second case, would pin it.

None of this argues against merging — the diff is correct for what it does. It's about whether the remaining call sites and `force_additional_properties_false` belong here or in a follow-up.

monkscode:
@percymcn — I loaded `pydantic_schema_utils.py` from `main` @ `f7ba8e3` (v1.15.14) and ran the real `generate_model_description` over both shapes rather than reading the diff. All four points reproduce. The one thing I did not check is the `openai-node` `toStrictJsonSchema()` behaviour you cite in (3); everything below I ran.

**1. `required` — unchanged by design, and I'll pin it if that's wanted.**

```
pydantic required : ['name']
pre-fix  required : ['name', 'note']
POST-FIX required : ['name', 'note']
```

`ensure_all_properties_required` runs at `pydantic_schema_utils.py:679`, above the `strip_null_types` branch at `:681`, so this flag structurally cannot reach it. Required-plus-nullable is the encoding I was aiming for, but you're right that the test leaves it implicit: `assert schema["required"] == ["name", "note"]` passes today and would state it. I'll add it on request.

Making the in-prompt `required` mirror Pydantic's `['name']` instead means changing `ensure_all_properties_required`, which is shared with the provider strict paths. Different blast radius — not something I'd fold into this.

**2. Other call sites — scoped deliberately, happy to widen.**

Confirmed on `main`, with one addition: `lite_agent.py:866` also embeds an unstripped schema, through the `lite_agent_response_format` slice rather than `formatted_task_instructions`. The full prompt-facing set that still strips null after this merges:

`utilities/converter.py:549`, `:556` · `lite_agent.py:698`, `:866` · `agent/core.py:1841` · `utilities/evaluators/task_evaluator.py:100`, `:174` · `agents/agent_adapters/base_converter_adapter.py:152`, `:154` (reaches the model via `enhance_system_prompt`)

The `llms/providers/*` and `tools/structured_tool.py` sites are correctly excluded — those build provider response schemas, where stripping is the right behaviour.

I scoped this to the site #6774 reproduces on to keep a first contribution reviewable. Extending to the nine above is mechanical; I'll push it here if a maintainer prefers one PR, otherwise a follow-up.

**3. `force_additional_properties_false` — reproduced, but the fix isn't prompt-only.**

```
raw      tags: {"additionalProperties": {"type": "string"}, "type": "object", ...}
pre-fix  tags: {"additionalProperties": false, "properties": {}, "required": [], "type": "object", ...}
POST-FIX tags: identical
```

An object with `properties: {}` and `additionalProperties: false` admits exactly one instance, `{}` — so the prompt states that a dict field must be empty. Same class of contradiction as #6774, and worse, since the model can comply perfectly and still produce nothing.

Why I'd keep it out of this PR isn't scope hygiene. `force_additional_properties_false` is called at `:670` from `generate_model_description` and also at `:610` from `_common_strict_pipeline`, which backs `sanitize_tool_params_for_openai_strict` / `_anthropic_strict` / `_bedrock_strict`. On those paths the closure is arguably load-bearing — OpenAI strict mode cannot express an open map at all, so the real choice is fail-loudly versus degrade-silently. That is a maintainer decision, not a flag flip, and it needs its own issue. I'll file it with the repro unless you'd rather own it, since the measurements are yours.

**4. Test coverage for that shape — lands with (3).** Adding `tags: dict[str, str]` to `_Output` pins nothing on its own; the assertion that would pin it fails on this branch.

Blocker unchanged: the seven workflows on the head commit are still `action_required` behind the first-time-contributor fork gate, so nothing here has been executed by CI. @lorenzejay @greysonlalonde — one "Approve and run workflows" click would give this a real signal, and #6774 is still untriaged.

monkscode:
@lucasgomide — flagging this to you specifically rather than broadly: #4579 is where `strip_null_types` was introduced, together with the `utilities/agent_utils.py` call site this PR cites as precedent. This is that same fix applied to the schema embedded in the task prompt — the one site of that pair which never received it.

@theCyberTech — a smaller, separate ask for whoever reaches it first: the seven workflows on the head commit are all still `action_required` behind the first-time-contributor fork gate. One "Approve and run workflows" click gives this a real signal and commits no one to a review.

**What has changed since 4 Aug**

The defect is still on `main` at HEAD: both branches of `build_task_prompt_with_schema` call `generate_model_description(...)` with the default. It also ships in the released `1.15.16`, verified against the wheel.

The surface is wider than the PR title suggests:

- When the agent has tools, the provider-side model is dropped entirely — `effective_response_model = None if self.original_tools else self.response_model` (`agents/crew_agent_executor.py:356`, and `:1170` async). The stripped prompt schema is then the *only* contract the model sees; there is no second schema to fall back on.
- On the native OpenAI paths (`llms/providers/openai/completion.py:875` and `:1824`) the provider schema is built from the same `generate_model_description(...)` default and shipped with `strict: True`. There the required non-nullable string is enforced, not merely suggested.

Measured on 1.15.16, offline, for `class Step(BaseModel): name: str; note: str | None = None`:

```
                    required           note
pydantic            ['name']           {"anyOf":[{"type":"string"},{"type":"null"}],"default":null}
prompt, pre-fix     ['name','note']    {"default":null,"type":"string"}
prompt, post-#6775  ['name','note']    {"anyOf":[{"type":"string"},{"type":"null"}],"default":null}
```

**Scope is yours to set.** As it stands this is two lines plus a regression test in `tests/agents/test_agent_utils.py`, matching that directory's existing layout. Nine other prompt-facing call sites still strip null — enumerated earlier in this thread — and I will extend the change here or leave it as a follow-up, whichever you prefer. The `force_additional_properties_false` case raised above I would file separately, since that pass is shared with the OpenAI/Anthropic/Bedrock strict sanitizers and the decision there is not a flag flip.

monkscode:
@Vidit-Ostwal — thanks for picking this up. One thing blocks it that only a maintainer can clear: all seven workflows on head 55d0319 are still action_required behind the first-time-contributor fork gate, so CI has never run on this PR and CodeRabbit is the only check. One "Approve and run workflows" click gives it a real signal.

Still current as of 1.15.20 (released yesterday): both call sites unchanged at agent/utils.py:77 and :83, the precedent this copies still at utilities/agent_utils.py:268, and the branch merged up to main and MERGEABLE.

Happy to fold in the two open CodeRabbit nitpicks (an output_json case, and an order-independent anyOf assertion) if you want them before review — just say so.

Vidit-Ostwal (commented):
The regression test currently covers only the `output_pydantic` branch of `build_task_prompt_with_schema`, while this PR changes both `output_pydantic` and `output_json` branches. Please add a matching `output_json` test or parameterize the existing test so we verify nullable fields are preserved in both prompt-schema paths.

monkscode:
@Vidit-Ostwal — done, pushed as `fa911d5`. Parameterized rather than duplicated, so both prompt-schema branches run the same assertions.

```
test_optional_fields_stay_nullable_in_the_prompt_schema[output_pydantic] PASSED
test_optional_fields_stay_nullable_in_the_prompt_schema[output_json]     PASSED
```

To check the new case actually exercises the branch rather than passing either way, I reverted both `strip_null_types=False` on this branch and re-ran:

```
[output_pydantic] FAILED  KeyError: 'anyOf'
[output_json]     FAILED  KeyError: 'anyOf'
```

Red without the fix, green with it, on both branches. Ran on 3.12 against the repo's pytest config minus the xdist split (`--tb=short --timeout=60 --block-network --import-mode=importlib`).

Same commit also folds in the other open CodeRabbit nitpick, since it touched the same assertion: the `anyOf` members are now compared as a set, so member order is no longer part of the test contract.

CI still has not executed any of this — all seven workflows on `fa911d5` are `action_required` behind the first-time-contributor fork gate, so CodeRabbit remains the only check. One "Approve and run workflows" click would give the branch a real signal.

Vidit-Ostwal (approved):

monkscode:
Thanks @Vidit-Ostwal for merging this quickly! Unblocks our dependency upgrade path once the release lands.

PR #6781, opened 2026-08-02
fix: record task and crew failures instead of reporting them as OK

joaomdmoura:
## Why

`error_count` is **zero for every month** in `crew_task_executions_daily_target` — going back to at least Sept 2025, across ~230M task executions. That isn't a low error rate, it's broken instrumentation.

Root cause: failures were being recorded as successes.

## Three separate defects

**1. Task failures were recorded as `OK`.**

`TaskFailedEvent` routed to `Telemetry.task_ended`, which calls `close_span()`:

```python
def close_span(span: Span) -> None:
    span.set_status(Status(StatusCode.OK))   # unconditional
    span.end()
```

Every failed task was exported with `otel.status_code: OK`. No downstream query could ever count one — which is exactly what the zeros show.

**2. Crew failures weren't recorded at all, and leaked their span.**

`on_crew_failed` never touched telemetry. A crew that raised left `_execution_span` open — never ended, never exported. The failure was invisible *and* the span was lost.

**3. Some task failures leaked their span too.**

`on_task_failed` only ended the span when `source.agent.crew` was present. A task failing without one got popped from the span map and never closed.

## Changes

- `close_span_with_error()` — sets `StatusCode.ERROR`, optionally records `error_type`.
- `Telemetry.task_failed()` and `Telemetry.crew_failed()`. `crew_failed` clears `_execution_span` so it can't be double-closed, and no-ops safely when no span exists (`share_crew=False`).
- Wire both failure events to them, closing spans **unconditionally** so neither path can leak.
- Optional `error_type` on `TaskFailedEvent` and `CrewKickoffFailedEvent`, populated with `type(e).__name__` at the four emit sites. Defaults to `None` — existing callers unaffected.

## Privacy

Only the exception **class name** is recorded, never the message. `TaskFailedEvent.error` already carries `str(e)`, which routinely contains prompts, model output, and credentials — that field is untouched and never reaches telemetry.

`close_span_with_error` drops any value failing `str.isidentifier()`, so a message can't be recorded even if passed by mistake. Every builtin exception name is a valid identifier, so nothing legitimate is lost. Tested against six message-shaped inputs including one containing `sk-live-1234`.

## Verification

End-to-end through the real telemetry path — two successful tasks and one `TimeoutError` whose message contained a fake API key:

```
total=3 errors=1  <- error_count is now countable
recorded error_type: TimeoutError
secret leaked anywhere? False
```

## Testing

New `tests/telemetry/test_failure_instrumentation.py` — 16 tests covering error status, the success/failure distinction, the PII guard, span-leak regressions for both task and crew, and event backwards compatibility.

- `tests/telemetry/`: **47 passed, 1 skipped**, stable across repeated randomized runs.
- `tests/events/`: 184 passed, 2 failed — both pre-existing on `main` (verified by stashing).
- `tests/utilities/test_events.py`: 46 passed.
- `tests/a2a/`: collection error from a missing optional `a2a` module, pre-existing.
- Ruff and mypy clean.

One thing worth knowing for future telemetry tests: the suite runs with `OTEL_SDK_DISABLED=true`, and the root conftest **pops** it on teardown. So only the *first* test in a session sees a disabled SDK and silently gets non-recording spans. This module sets it explicitly rather than depending on that leak.

## Not included: `total_duration_ms`

Also zero, but it's a **pipeline-side** bug, not an SDK one. The data is there — the raw `duration` column is a Go-style string:

```
duration: "2.026641s"   (99.99% of rows use the `s` suffix; 21 in 200k are empty)
```

`toInt64OrZero("2.026641s")` returns `0`, so the MV sums to nothing. The fix belongs in the ClickHouse materialized views:

```sql
toFloat64OrZero(replaceRegexpOne(duration, 's$', '')) * 1000  AS duration_ms
```

I can't deploy that from here — flagging it so it gets picked up alongside this.

https://claude.ai/code/session_01UNumDnNbiyw3pv1WakAe6t

---

> [!NOTE]
> **Medium Risk**
> Changes how execution spans are closed on failure paths in telemetry and the event listener; behavior is well-tested but affects observability pipelines that depend on span status and export.
> 
> **Overview**
> Fixes broken failure metrics by closing OpenTelemetry spans with **ERROR** instead of **OK**, and by ending spans that were previously leaked or never touched on crew kickoff failure.
> 
> Adds **`close_span_with_error()`**, which sets `StatusCode.ERROR` and optionally records **`error_type`** only when the value passes **`isidentifier()`** (so error messages with prompts or secrets never reach telemetry). **`Telemetry.task_failed()`** and **`Telemetry.crew_failed()`** use this helper; crew failure also clears **`_execution_span`** after close.
> 
> **`TaskFailedEvent`** and **`CrewKickoffFailedEvent`** gain optional **`error_type`** (defaults **`None`**); kickoff and task emit sites set **`type(e).__name__`**. The event listener routes task failures to **`task_failed`** unconditionally (no longer gated on **`source.agent.crew`**) and calls **`crew_failed`** on **`CrewKickoffFailedEvent`**.
> 
> New **`tests/telemetry/test_failure_instrumentation.py`** covers error vs success spans, PII guard, span-leak regressions, and backwards compatibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 328e0e36e9751eb9c9300bd98769cd4043cda098. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -1065,6 +1065,7 @@ class Crew(FlowTrackable, BaseModel):
                 self,
                 CrewKickoffFailedEvent(
                     error=str(e),
+                    error_type=type(e).__name__,
                     crew_name=self.name,
                     started_event_id=self._kickoff_event_id,
                 ),
@@ -1279,6 +1280,7 @@ class Crew(FlowTrackable, BaseModel):
                 self,
                 CrewKickoffFailedEvent(
                     error=str(e),
+                    error_type=type(e).__name__,
                     crew_name=self.name,
                     started_event_id=self._kickoff_event_id,
                 ),
--- a/lib/crewai/src/crewai/events/event_listener.py
+++ b/lib/crewai/src/crewai/events/event_listener.py
@@ -198,6 +198,11 @@ class EventListener(BaseEventListener):
 
         @crewai_event_bus.on(CrewKickoffFailedEvent)
         def on_crew_failed(source: Any, event: CrewKickoffFailedEvent) -> None:
+            # Previously this handler never touched telemetry, so a crew that
+            # raised left its execution span open: never ended, never exported,
+            # and the failure invisible downstream.
+            self._telemetry.crew_failed(source, event.error_type)
+
             self.formatter.handle_crew_status(
                 event.crew_name or "Crew",
                 source.id,
@@ -261,8 +266,10 @@ class EventListener(BaseEventListener):
         def on_task_failed(source: Any, event: TaskFailedEvent) -> None:
             span = self.execution_spans.pop(source, None)
             if span:
-                if source.agent and source.agent.crew:
-                    self._telemetry.task_ended(span, source, source.agent.crew)
+                # Closed unconditionally: previously the span was only ended
+                # when source.agent.crew was present, so any task that failed
+                # without one leaked its span and was never exported.
+                self._telemetry.task_failed(span, source, event.error_type)
 
             task_name = get_task_name(source)
             self.formatter.handle_task_status(
--- a/lib/crewai/src/crewai/events/types/crew_events.py
+++ b/lib/crewai/src/crewai/events/types/crew_events.py
@@ -52,6 +52,13 @@ class CrewKickoffFailedEvent(CrewBaseEvent):
     """Event emitted when a crew fails to complete execution"""
 
     error: str
+    error_type: str | None = None
+    """Exception class name (e.g. "ValidationError").
+
+    Kept separate from ``error`` so telemetry can record what kind of failure
+    occurred without ever touching the message, which routinely contains
+    prompts, model output, or credentials.
+    """
     type: Literal["crew_kickoff_failed"] = "crew_kickoff_failed"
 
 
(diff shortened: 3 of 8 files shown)


joaomdmoura:
Closing this for now — not because the finding is wrong, but because it needs to be re-cut.

Re-verified against `main` today; all three defects are still live:

- `close_span()` sets `StatusCode.OK` unconditionally (`lib/crewai/src/crewai/telemetry/utils.py:222`)
- `on_crew_failed` never touches the execution span (`lib/crewai/src/crewai/events/event_listener.py:200`)
- `on_task_failed` drops the span when `source.agent.crew` is absent (`event_listener.py:261`)

And the consequence is confirmed in the warehouse: **240.0M task executions across 13 months in
`crew_task_executions_daily_target`, `error_count` = 0 in every month.**

Two reasons this branch isn't the right vehicle:

1. **The crew half is structurally dead for the default population.** `crew_execution_span()` returns
   `None` unless `share_crew=True` (`telemetry.py:1019-1021`), so `crew._execution_span` is `None`
   for nearly everyone and the new `crew_failed()` exits at its `if span is None: return`. No
   materialized view reads `operationName = 'Crew Execution'` either. Landing this would not have
   produced crew failure data.
2. The task half — which *is* correct and covers the 240M-execution surface — is worth landing on
   its own, rebased, rather than carried behind the crew work.

Needs a Linear ticket to re-cut as: task-failure instrumentation only, plus a separate decision on
whether crew-level outcome should stop being `share_crew`-gated.

Branch is preserved; reopen or cherry-pick from `fix/task-crew-failure-telemetry`.

PR #6795, opened 2026-08-03
fix(security): close SSRF in ArxivPaperTool.download_pdf, plus a DNS-rebinding gap in safe_get

Eman-Yousaf:
Fixes #6694.

## What's actually vulnerable

The issue names both `urllib` calls in `ArxivPaperTool`, but only one is a real SSRF vector:

- **`fetch_arxiv_data()`**'s `urlopen` call targets a hardcoded `BASE_API_URL` (`export.arxiv.org`); `search_query` only ever lands in the query string, percent-encoded. **Not an SSRF vector** for destination redirection. It was plain `http://`, though — a real but separate weakness (a network MITM can tamper with the API response). Switched to `https://` and migrated to `safe_get()` for consistency and redirect-safety while touching this.
- **`download_pdf()`**'s `urlretrieve` call is the real vector: `pdf_url` comes from parsing the arxiv API's **XML response** (an `href` attribute), not directly from `search_query`. Whatever URL shows up there was fetched and written to disk with zero validation — reachable via the same plain-HTTP MITM angle, or a malicious link ever indexed upstream. Migrated to `safe_download()` (new helper, see below).

## Additional finding: DNS-rebinding TOCTOU in the existing `safe_requests` infra

`crewai_tools.security` already has real SSRF protection (`validate_url`, `safe_get`) that both call sites above should have been using from the start. While wiring the arxiv tool into it, I found a real gap in the shared infrastructure itself: `validate_url()` resolves DNS, checks the IP, and returns the *original URL string* — it doesn't pin the actual connection to the address it just validated. `requests.get()` then re-resolves DNS itself at connection time. An attacker controlling DNS (or using a short-TTL record) can present a safe IP for validation and a private one for the real connection moments later.

Fixed by:
- `safe_path.validate_and_resolve(url) -> (url, ip)`: shares `validate_url`'s exact logic but also returns the checked IP. `validate_url`/new `resolve_validated_ip` are now thin wrappers around it (`validate_url`'s public signature/behavior is unchanged).
- `safe_requests._pin_dns(hostname, ip)`: a context manager that pins `socket.getaddrinfo(hostname, ...)` to the validated IP for the duration of one request. `safe_get()` now validates+pins on the initial URL *and* on every redirect hop, closing the window for both direct requests and redirect-based rebinding.
- `safe_requests.safe_download(url, dest_path)`: new helper for streaming a validated, pinned download to disk — `safe_get` didn't have an equivalent for `arxiv_paper_tool`'s use case.

## Testing

- New regression tests:
  - `test_safe_get_pins_dns_against_rebinding`: proves the fix directly — mocks `getaddrinfo` to return a safe IP on the first lookup and a private one on any subsequent lookup (simulating rebinding), and asserts the connection never sees the private IP. **Fails against the pre-fix code.**
  - `test_safe_get_restores_real_resolver_after_pinning`: pinning doesn't leak past the request it was applied to.
  - `safe_download`: writes-to-disk, blocks private IP, blocks redirect to private IP.
  - `ArxivPaperTool.download_pdf`: blocks `127.0.0.1` and the AWS/GCP/Azure metadata endpoint **end-to-end** (not mocked at the `safe_download` level), proving the tool itself is protected, not just the helper in isolation.
- Existing `arxiv_paper_tool_test.py` tests updated to mock `safe_get`/`safe_download` instead of `urllib` (the code path changed; test intent didn't).
- Full suite: `uv run --package crewai-tools pytest lib/crewai-tools/tests/utilities/test_safe_path.py lib/crewai-tools/tests/utilities/test_safe_requests.py lib/crewai-tools/tests/tools/arxiv_paper_tool_test.py` — **55 passed**.
- `ruff check` clean on all changed files.

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/security/safe_path.py
+++ b/lib/crewai-tools/src/crewai_tools/security/safe_path.py
@@ -180,6 +180,13 @@ def _is_private_or_reserved(ip_str: str) -> bool:
         return True  # If we can't parse, block it
 
 
+def _is_ipv4(ip_str: str) -> bool:
+    try:
+        return ipaddress.ip_address(ip_str).version == 4
+    except ValueError:
+        return False
+
+
 def validate_and_resolve(url: str) -> tuple[str, str | None]:
     """Validate that a URL is safe to fetch, and return the specific IP address
     checked so a caller can pin its actual connection to it.
@@ -202,8 +209,16 @@ def validate_and_resolve(url: str) -> tuple[str, str | None]:
         url: The URL to validate.
 
     Returns:
-        A ``(validated_url, ip)`` tuple. ``ip`` is the first address checked,
-        or ``None`` when validation is bypassed via the escape hatch (nothing
+        A ``(validated_url, ip)`` tuple. Every resolved address is checked,
+        but ``ip`` prefers an IPv4 address if one was returned (falling back
+        to the first address otherwise) -- ``getaddrinfo`` on a dual-stack
+        host often returns IPv6 first, and pinning the connection to a single
+        address (see ``safe_requests.safe_get``) forgoes the normal
+        multi-address fallback an unpinned connection attempt would get, so
+        picking the address most likely to actually be reachable (many CI
+        runners, containers, and networks have working IPv4 but broken or
+        absent IPv6 routes) matters more here than it would otherwise.
+        ``None`` when validation is bypassed via the escape hatch (nothing
         was resolved, so there is nothing to pin).
 
     Raises:
@@ -243,7 +258,7 @@ def validate_and_resolve(url: str) -> tuple[str, str | None]:
     except socket.gaierror as exc:
         raise ValueError(f"Could not resolve hostname: '{parsed.hostname}'") from exc
 
-    first_ip: str | None = None
+    checked_ips: list[str] = []
     for _family, _, _, _, sockaddr in addrinfos:
         ip_str = str(sockaddr[0])
         if _is_private_or_reserved(ip_str):
@@ -252,10 +267,12 @@ def validate_and_resolve(url: str) -> tuple[str, str | None]:
                 f"Access to internal networks is not allowed. "
                 f"Set {_UNSAFE_PATHS_ENV}=true to bypass."
             )
-        if first_ip is None:
-            first_ip = ip_str
+        checked_ips.append(ip_str)
 
-    return url, first_ip
+    pinned_ip = next((ip for ip in checked_ips if _is_ipv4(ip)), None) or (
+        checked_ips[0] if checked_ips else None
+    )
+    return url, pinned_ip
 
 
 def validate_url(url: str) -> str:
--- a/lib/crewai-tools/src/crewai_tools/security/safe_path.py
+++ b/lib/crewai-tools/src/crewai_tools/security/safe_path.py
@@ -180,18 +180,31 @@ def _is_private_or_reserved(ip_str: str) -> bool:
         return True  # If we can't parse, block it
 
 
-def validate_url(url: str) -> str:
-    """Validate that a URL is safe to fetch.
+def validate_and_resolve(url: str) -> tuple[str, str | None]:
+    """Validate that a URL is safe to fetch, and return the specific IP address
+    checked so a caller can pin its actual connection to it.
 
-    Blocks ``file://`` scheme entirely. For ``http``/``https``, resolves
-    DNS and checks that the target IP is not private or reserved (prevents
-    SSRF to internal services and cloud metadata endpoints).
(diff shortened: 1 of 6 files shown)


Eman-Yousaf:
Thanks both — addressed everything in 83d1f53:

**The real one** (flagged independently by both @coderabbitai and @Copilot): `_pin_dns` patched `socket.getaddrinfo` process-wide with no synchronization. Two concurrent `safe_get`/`safe_download` calls could race — one thread's `finally` restoring over another's still-active pin, silently dropping DNS-rebinding protection for whichever request lost the race. My original docstring's claim that concurrent requests were "unaffected" was wrong. Fixed with a `threading.Lock` serializing the whole patch/use/restore window, plus `test_safe_get_dns_pin_is_thread_safe` (two real threads, widened race window) to guard it.

Also fixed:
- `safe_download` now writes to a `.part` temp file and renames into place only on success — no truncated file left at `dest_path` on a failed/interrupted download.
- `safe_download` hard-coded `stream=True` but also forwarded `**kwargs`, so a caller-supplied `stream=` would raise a `TypeError`. Now explicitly overrides `kwargs["stream"]`.
- `download_pdf`'s `save_path` was typed `str` but always called with a `Path` (needing a `type: ignore`); widened to `str | Path`.
- `test_download_pdf_success` now asserts the exact args passed to `safe_download`, not just that it was called.
- The two end-to-end SSRF tests now clear `CREWAI_TOOLS_ALLOW_UNSAFE_PATHS` via `monkeypatch` so they stay hermetic regardless of environment.

58 passed (was 55), `ruff check` clean.

Eman-Yousaf:
Thanks for the follow-up round — addressed in c7dc446:

**Real bug, caught by CodeRabbit**: my own previous fix for the truncated-download issue introduced a new one — `tmp_path` was derived only from `dest.name`, so two concurrent `safe_download` calls targeting the *same* destination would race on writing to the *same* `.part` file, interleaving/corrupting both. Fixed by including a `uuid4` in the temp filename, so every call gets its own temp file regardless of thread/process/how many callers target the same `dest_path` at once. New test `test_safe_download_concurrent_calls_to_same_dest_dont_corrupt` (two real threads) asserts the final file is always one complete, uncorrupted payload — never a byte-level mix of both (last-writer-wins on the final rename is expected/fine; corruption is not).

**Copilot's resource-hygiene suggestion**: `safe_get` now closes each redirect response right after appending it to `history`, instead of leaving it open until GC. Redirect bodies are never read in this code path, so `status_code`/`headers`/`url` stay accessible on `history` entries — only body access would be affected, and nothing reads it.

**Docs nitpick**: added docstrings to `fetch_arxiv_data`/`download_pdf` covering the params/returns/exceptions that changed when they moved off `urllib` onto `safe_get`/`safe_download`.

59 passed (was 58), `ruff check` clean.

Eman-Yousaf:
Addressed both in 895773d — thanks for surfacing these even though they were marked suppressed:

**IPv4 preference**: `validate_and_resolve` was pinning to whichever address `getaddrinfo` returned first. On a dual-stack host that's often IPv6, and pinning forgoes the normal multi-address connection fallback an unpinned request gets — plenty of real environments (CI runners, containers, some networks) have working IPv4 but broken/absent IPv6 routes, so this could make an otherwise-fine request fail. Now prefers an IPv4 address among the validated results, falling back to IPv6 only if that's genuinely all that's offered. Every returned address is still validated regardless of which one ends up pinned — a private IPv6 address anywhere in the results still blocks the request.

**Port typing in the pinned resolver**: `socket.getaddrinfo`'s real signature allows a service name or `None` for `port`, not just an int — the pinned resolver's `port: int` annotation was inaccurate, and a non-int port for the pinned hostname during the pin window would've built a malformed sockaddr. Widened to `int | str | None` and added an `isinstance` check: matching-host lookups with a non-int port now fall through to the real resolver instead.

New tests: `TestValidateAndResolve` (dual-stack prefers IPv4, IPv6-only still works, a private address anywhere in the results still blocks regardless of preference) and `test_pinned_resolver_falls_through_for_non_int_port`.

63 passed (was 59), `ruff check` clean.

Eman-Yousaf:
Good catch, fixed in a7fdf3c: the test asserted on the *returned IP* from the non-int-port lookup, but the \`public_dns\` fixture resolves \`public.example\` to the same address for both the pinned request and the fallback-delegation path — so a broken \`pinned_getaddrinfo\` that incorrectly intercepted a non-int port would have returned that same IP and passed anyway. It never actually proved delegation happened.

Fixed by wrapping the fixture-installed resolver with a call-tracking function *before* invoking \`safe_get\`, and asserting the non-int-port lookup actually reached it directly — exactly your suggested fix.

63 passed, \`ruff check\` clean.

PR #6796, opened 2026-08-04
docs: add Secure Agent Design guide

theCyberTech:
Adds a Secure Agent Design guide under Guides → Agents. The guide covers trusted vs untrusted input, prompt injection, tool abuse, output checks, approval gates, delegation limits, and agent isolation.

This page is the design and threat-model guide. Execution limits (`max_rpm`, `max_iter`, `max_execution_time`), verbosity, and agent settings stay on [Agents](/en/concepts/agents) and [Customize Agents](/en/learn/customizing-agents), so an Advanced config page can own those knobs without overlapping this guide.

HITL is approval, not a control. That warning lives in the overview and in section 6 only. Task `human_input=True` reviews the final answer after tools ran; it does not gate tool execution. Pre-execution blocking stays on tool hooks (`HookAborted`). This matches the `human_input` contract called out in #7096.

The isolation Flow example now stores validated `OutreachNotes` (`response_format` + `result.pydantic`) and passes only `claims` and `sources` to the sender. It does not dump `result.raw`.

The prose is simplified technical English: short sentences, defined terms, and no jargon such as “fail open” or “ACL”. The same APIs and execution-path limits remain.

Accuracy notes (matched to code):

- `agent.kickoff()` uses `AgentExecutor` / `LiteAgentOutput`. It does not create a Task or a Crew.
- `Agent.guardrail` runs on kickoff only. Task guardrails and `human_input` run on the Task path.
- Execution boundary hooks run on Crew and Flow only.
- `@CrewBase` `@on` hooks are added to the global hook list when the crew is created.
- `HookAborted` blocks that tool call. The run continues.
- If a tool hook raises any exception other than `HookAborted`, CrewAI ignores the error and the tool still runs.
- On the Task path, `memory=False` on an agent does not turn off crew memory.
- `allow_delegation` defaults to `False`. A2A `trust_remote_completion_status` defaults to `False`.

Also in this update:

- The first `backstory` example is labeled as a soft control.
- pt-BR `concepts/agents.mdx` now has the direct `kickoff()` section and the `AgentExecutor` / `LiteAgentOutput` note.

Locales: edge `en`, `pt-BR`, `ko`, and `ar` pages plus nav entries. Frozen `docs/v*/` is unchanged.

Locale links to EN-only A2A and E2B pages keep the `/en/...` prefix so `mint broken-links` passes. The Arabic kickoff note now matches the English `AgentExecutor` behavior.

<div><a href="https://cursor.com/agents/bc-019fcb80-13e4-7b83-b99b-5ed37041612d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-019fcb80-13e4-7b83-b99b-5ed37041612d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


Nakagawa-master (commented):
The approval section is already careful about two important boundaries: `human_input=True` happens after tools, and `request_human_input` does not authenticate the approver. One additional boundary may be worth stating because approval can be a long pause:

```text
the user approved this exact tool call
!=
the call is still authorized under current mutable state when execution resumes
```

For the synchronous console example the `ToolCallHookContext` keeps the same tool input, which is good. But an application can also have mutable authorization state outside that input: the user's role can be revoked, an allowlist can change, a ticket/file/payment target can advance to a new revision, or a spend/policy limit can be consumed by another session while the human is deciding.

A short production note after the example could make the safe contract explicit:

```text
candidate tool + exact args + target/current revision
→ show/record approval
→ after the human returns, re-read current permissions/policy/target revision
→ execute only if both the approved proposal and current authority still match
→ otherwise abort and ask again
```

A useful regression is: render approval while authorized → revoke the relevant permission or mutate the target revision during the approval wait → type `yes` → `HookAborted`, no side effect.

This is a TOCTOU/current-authority distinction, not a request to turn CrewAI HITL into an identity system. The guide can keep identity/authentication application-owned while still warning that an approval should not freeze mutable execution authority.

Reusable public framing for the same boundary: https://github.com/Nakagawa-master/nakagawa-theory-archive/blob/main/CURRENT_AUTHORITY_REUSE_KIT.md

Issue #6767, opened 2026-08-01
Conversational Flow golden use case improvements - route labels collide with method names in @listen validation


lorenzejay:
## Description

In a conversational `Flow`, the string passed to `@listen("...")` is a **router intent label**, not a method name. But `FlowDefinition._validate_trigger_namespace` treats the two as one namespace, so naming a handler after the route it serves — the most natural naming there is — fails at instantiation:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyFlow
  Value error, methods.create_video.listen must not reference itself
```

The self-reference guard makes sense for method-to-method triggers (it prevents a method re-triggering itself). For conversational routes the two namespaces are unrelated: `@listen("create_video")` means "run when the router picks the `create_video` intent", which has nothing to do with the method being called `create_video`.

The failure is also late and indirect — it surfaces as a pydantic `ValidationError` at `MyFlow()` construction, not at class definition, and the message reads as if the developer wrote a self-referential trigger.

## Steps to Reproduce

1. Define a conversational `Flow` with a custom route.
2. Name the handler method the same as the route label.
3. Instantiate the flow.

## Expected behavior

Either:
- route labels and method names are treated as separate namespaces (so `@listen("create_video")` on `def create_video` is allowed), or
- the error explicitly explains the collision, e.g. *"'create_video' is both a route label and a method name; conversational route labels share a namespace with method names — rename the handler."*

Right now the message sends you looking for a self-referential trigger you never wrote.

## Screenshots/Code snippets

```python
from crewai import LLM
from crewai.experimental.conversational import ConversationConfig, ConversationState, RouterConfig
from crewai.flow.flow import Flow, listen

llm = LLM(model="anthropic/claude-haiku-4-5")

@ConversationConfig(llm=llm, router=RouterConfig(llm=llm,
    route_descriptions={"create_video": "User wants a new video."}))
class MyFlow(Flow[ConversationState]):
    conversational = True

    @listen("create_video")
    def create_video(self) -> str:      # named after the route it listens to
        return "made a video"

print("class defined OK")   # <-- this prints
f = MyFlow()                # <-- ValidationError raised here
```

Output:

```
class defined OK
Traceback (most recent call last):
  ...
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyFlow
  Value error, methods.create_video.listen must not reference itself
```

Relevant validator (`crewai/flow/flow_definition.py` ~L773):

```python
[...]

PR #6799, opened 2026-08-04
separate conversational route labels from method trigger namespace

Vidit-Ostwal:
<h1 data-pm-slice="1 1 []">What Is the Issue?</h1><h2>One sentence</h2><p>In conversational Flows, <code>@listen("create_video")</code> means <strong>“run when the router picks intent <code>create_video</code>”</strong>, but validation treats that string as <strong>“run when method <code>create_video</code> completes”</strong> — and rejects the handler because the names match.</p><hr><h2>Two namespaces, one check</h2><p>Flow uses the same string type for two different concepts:</p>
Concept | Example | Meaning
-- | -- | --
Router intent label | @listen("create_video") on a conversational handler | Fire when route_conversation returns "create_video"
Method trigger | @listen("begin") on process | Fire when method begin finishes

<p>The self-listen guard exists for the second case: <code>@listen("process") def process</code> would re-fire when <code>process</code> completes → infinite loop.</p><p>For conversational routes, the first case is what developers intend:</p><pre><code class="language-python">@listen("create_video")
def create_video(self) -&gt; str:   # natural: handler named after the route it serves
    return "made a video"</code></pre><p>Here <code>"create_video"</code> is <strong>not</strong> a self-referential method trigger. It is the output label of the conversational router (<code>route_conversation</code> in <code>[conversational_mixin.py](lib/crewai/src/crewai/experimental/conversational_mixin.py)</code>).</p><pre data-language="mermaid" data-type="mermaid" class="language-mermaid"><code class="language-mermaid">sequenceDiagram
    participant User
    participant RouteConversation as route_conversation
    participant Handler as create_video

    User-&gt;&gt;RouteConversation: handle_turn message
    RouteConversation-&gt;&gt;RouteConversation: route_turn / LLM picks intent
    RouteConversation--&gt;&gt;Handler: emit trigger "create_video"
    Handler-&gt;&gt;Handler: run handler once
    Note over Handler: NOT meant to re-trigger itself</code></pre><hr><h2>Where it breaks</h2><p>Validation in <code>[flow_definition.py](lib/crewai/src/crewai/flow/flow_definition.py)</code> (~L773):</p><pre><code class="language-python">def _validate_trigger_namespace(self) -&gt; FlowDefinition:
    for method_name, method in self.methods.items():
        if _condition_references(method.listen, method_name):
            raise ValueError(
                f"methods.{method_name}.listen must not reference itself"
            )</code></pre><p><code>_condition_references</code> is pure string equality — it does not know about <code>conversational = True</code> or router intent labels:</p><pre><code class="language-python">def _condition_references(condition, name) -&gt; bool:
    if isinstance(condition, str):
        return condition == name   # "create_video" == "create_video" → reject</code></pre><p>So every conversational handler whose method name matches its route label fails — even though no method-to-method self-trigger was written.</p><hr><h2>Why the failure feels wrong</h2><p>Three UX problems stack on top of the logic bug:</p><ol><li><p><strong>Late</strong> — Class definition succeeds; failure happens on first <code>MyFlow()</code> because <code>FlowDefinition</code> is built lazily in <code>_flow_post_init</code> → <code>flow_definition()</code> (<code>[runtime/__init__.py](lib/crewai/src/crewai/flow/runtime/__init__.py)</code> ~L763), not when the class body is parsed.</p></li><li><p><strong>Misleading message</strong> — <code>methods.create_video.listen must not reference itself</code> sounds like you wired a method to listen to its own completion. You did not; you wired it to a router intent.</p></li><li><p><strong>No doc hint</strong> — The docs show handlers with <em>different</em> names (<code>handle_order</code> listens to <code>"order"</code>), but nothing warns that matching names are forbidden. The framework itself avoids the collision internally (<code>converse_turn</code> listens to <code>"converse"</code>, not <code>def converse</code>).</p></li></ol><hr><h2>Why the guard exists (and what must stay protected)</h2><p>The guard is correct for ordinary Flows. These should keep failing:</p><pre><code class="language-python">@listen("process")
def process(self): ...          # method listens to its own completion

@router("route")
def route(self): ...            # router listens to its own trigger string</code></pre><p>Existing tests in <code>[test_flow.py](lib/crewai/tests/test_flow.py)</code> (<code>test_self_listening_method_is_rejected</code>, etc.) cover this and must not regress.</p><hr><h2>Secondary concern (runtime, not just validation)</h2><p>Even if validation were loosened, runtime has the same string collision:</p><ul><li><p>Router emits <code>"create_video"</code> → handler runs (correct)</p></li><li><p>Handler completes → <code>_execute_listeners("create_video", ...)</code> is called (<code>[runtime/__init__.py](lib/crewai/src/crewai/flow/runtime/__init__.py)</code> ~L3248)</p></li><li><p><code>_find_triggered_methods</code> would match <code>@listen("create_video")</code> again → potential loop (currently caught by <code>max_method_calls</code> / <code>RecursionError</code>)</p></li></ul><p>So a complete fix needs <strong>both</strong>:</p><ul><li><p>Validation: exempt conversational route handlers from the self-listen check</p></li><li><p>Runtime: skip re-triggering a listener on its own completion when <code>listen</code> references its own name</p></li></ul><hr><h2>Workaround today</h2><p>Rename the handler to something other than the route label:</p><pre><code class="language-python">@listen("create_video")
def direct_new_video(self) -&gt; str: ...</code></pre><p>Works, but fights the most natural naming and cost time in the golden-use-case exercise.</p><hr><h2>Scope boundary</h2><p>This issue is <strong>conversational-specific</strong> in the bug report. Non-conversational flows where a <code>@router</code> emits <code>"create_video"</code> and a handler is also named <code>create_video</code> hit the same validation rule today — that may stay rejected unless explicitly expanded later.</p>


Vidit-Ostwal:
@coderabbit review

Vidit-Ostwal:
Resolves https://github.com/crewAIInc/crewAI/issues/6767

Vidit-Ostwal:
## Implementation update

This PR now includes the full fix beyond the initial design proposal.

### What changed

**Validation** (`flow_definition.py`)
- Exempt conversational, non-router handlers when the `@listen` label matches the method name (route intent binding, not a method self-trigger)
- Routers with self-referencing listen conditions are still rejected

**Runtime** (`runtime/__init__.py` + `types.py`)
- Added `FlowDispatchTrigger` with `method_emit` vs `router_emit` provenance
- Router intents fire handlers once; method completion no longer re-triggers self-referencing handlers

**Tests**
- 4 validation tests in `test_flow_definition.py`
- 4 dispatch/runtime tests in `test_flow.py`

**Docs**
- Added `<Note>` guidance in `concepts/flows.mdx` and `guides/flows/conversational-flows.mdx` recommending different handler names from `@listen` labels (matching names are supported in conversational flows, but split names are preferred for clarity)

Also removed the internal design doc from `lib/crewai/docs/` now that the implementation is in the PR.

Ready for review!

PR #6801, opened 2026-08-04
fix(anthropic): include cache tokens in totals

### LINKED ISSUE #6788: [BUG] total_tokens omits cache reads and writes on the native Anthropic provider
On the native Anthropic provider, `UsageMetrics.total_tokens` counts neither cache reads nor cache writes. Those are tokens you are billed for, and on a well-cached workload they are almost all of them.

The LiteLLM route for the same model does count them, so the two routes report different totals for identical traffic.

Read from `main` on 2026-08-03.

## Where

`lib/crewai/src/crewai/llms/providers/anthropic/completion.py:1977`:

```python
result: dict[str, Any] = {
    "input_tokens": input_tokens,
    "total_tokens": input_tokens + output_tokens,     # <--
}
```

Anthropic's `usage.input_tokens` is the **uncached remainder**, not the input total. One prompt, three consecutive calls against the API directly, `claude-haiku-4-5`, anthropic-sdk 0.120.2:

| call | `input_tokens` | `cache_creation` | `cache_read` | `output` | actually moved |
|---|---|---|---|---|---|
| 1, unmarked | 17,111 | 0 | 0 | 4 | 17,115 |
| 2, marked | 9 | 17,102 | 0 | 4 | 17,115 |
| 3, marked | 9 | 0 | 17,102 | 8 | 17,119 |

So on the third call `total_tokens` is `9 + 8 = 17` for a request that moved 17,119 tokens. Raw usage for every call is in [`prompt-tokens-semantics.json`](https://github.com/Tanisha-Katara/cacheeconomics/blob/main/tier-b/evidence/prompt-tokens-semantics.json).

The dict is consumed by `base_llm.py:954-964` via `UsageMetrics.from_provider_dict`, which recomputes `total_tokens = prompt_tokens + completion_tokens` with `prompt_tokens` resolved from the `input_tokens` alias (`usage_metrics.py:127-129, 149`). So the undercount survives normalisation and lands in `self._token_usage`.

## Why the LiteLLM route disagrees

`utilities/token_counter_callback.py:54` uses `usage.prompt_tokens`, and LiteLLM reports that as the **total** including both cache classes — 17,111 on all three calls above, never moving. It then adds cached tokens to a separate counter without touching the total (`base_token_process.py:25-26`), which is right and does not double-count.

Same crew, same model, `is_litellm=True` versus the native provider (`llm.py:478`), and `total_tokens` differs by three orders of magnitude on a fully-cached call.

## How big in practice

Those ratios come from a near-100% cache hit on a deliberately large static prefix, which is the ceiling rather than a typical run — the exact multiple is a property of the fixture, not of CrewAI. What generalises is the sign and the mechanism.

For a realistic figure: across 14,375 requests of my own agent traffic, cache reads and writes were 97% of all input tokens. A `total_tokens` built this way would report about 3% of what was sent.

CrewAI reports tokens rather than dollars, so nothing here is a wrong invoice. It is a wrong denominator for anything computed from it — cost estimates downstream, the evaluation metrics in `experimental/evaluation/metrics/reasoning_metrics.py:68-69` that divide `total_tokens` by call count, and any OTel span consuming the same field.

## Reproduction

```bash
git clone https://github.com/Tanisha-Katara/cacheeconomics
python3 cacheeconomics/disclosure/verify_crewai_tokens.py
```

No key, no network, no CrewAI install — it transcribes the extraction and replays it against the committed usage artifact.

## A test currently pins the behaviour

`lib/crewai/tests/llms/anthropic/test_anthropic.py:1655-1660`:

```python
mock_response.usage = MagicMock(
    input_tokens=100, output_tokens=50,
    cache_read_input_tokens=30, cache_creation_input_tokens=20,
)
usage = llm._extract_anthropic_token_usage(mock_response)
assert usage["total_tokens"] == 150
```

That request sent 150 input tokens (100 + 30 + 20) and produced 50 output, so 200 passed through it. The assertion encodes 150, which is the undercount, and would need to change with the fix. Flagging it rather than leaving it for review to discover — I have shipped this exact shape myself, where the test that should have caught a bug had recorded it as the contract instead.

## Suggested fix

```python
input_tokens = getattr(usage, "input_tokens", 0)
output_tokens = getattr(usage, "output_tokens", 0)
cache_creation_tokens = getattr(usage, "cache_creation_input_tokens", 0) or 0

# Anthropic reports input_tokens as the uncached remainder; the cache classes
# are billed separately and are additional to it.
prompt_tokens = input_tokens + cache_read_tokens + cache_creation_tokens

result: dict[str, Any] = {
    "input_tokens": prompt_tokens,
    "total_tokens": prompt_tokens + output_tokens,
}
```

This makes the native route agree with the LiteLLM route. `cached_prompt_tokens` and `cache_creation_tokens` keep reporting the split, so nothing reading them today changes.

One thing to decide rather than something I would assert: whether `input_tokens` in this dict should become the total, or stay the remainder with only `total_tokens` corrected. The alias table at `usage_metrics.py:127-129` maps `input_tokens` onto `prompt_tokens`, so leaving it as the remainder fixes `total_tokens` and leaves `prompt_tokens` still undercounting. That is the reason for the version above, but it is your call.

## Scope and caveats

- Measured against Anthropic directly (sdk 0.120.2) and through LiteLLM 1.83.9, recorded in the artifact linked above.
- The measurement is of the Anthropic API's reporting, not of CrewAI running. I have not executed a crew end to end and watched `UsageMetrics`; the link from those usage fields to `total_tokens` is read from source and traced above. If you run one and see something different, I would want to know.
- The Bedrock provider (`llms/providers/bedrock/completion.py:2071`) reads `usage.get("totalTokens", ...)`, which Bedrock populates itself, so it is probably unaffected. I have not verified that against a live Bedrock call and am not claiming it.
- I have not checked the Gemini or OpenAI native providers for the same shape.

Happy to open a PR if that is easier than a patch in an issue.

Linxiushen:
## Summary

- include Anthropic cache reads and cache writes in normalized input and total token counts
- preserve cache read and cache creation counters as separate usage dimensions
- cover both provider extraction and the final UsageMetrics summary

Closes #6788

## Testing

- targeted Anthropic token usage tests: 3 passed
- Ruff check: passed
- Ruff format check: passed
- Mypy on the changed provider: passed

## AI assistance

Developed with AI assistance and reviewed and tested by the contributor. External contributors cannot apply labels in this repository, so I have requested the required llm-generated label in a comment below.

### DIFF
--- a/docs/edge/en/concepts/crews.mdx
+++ b/docs/edge/en/concepts/crews.mdx
@@ -322,7 +322,7 @@ Caches can be employed to store the results of tools' execution, making the proc
 
 After the crew execution, you can access the `usage_metrics` attribute to view the language model (LLM) usage metrics for all tasks executed by the crew. This provides insights into operational efficiency and areas for improvement.
 
-`total_tokens` is the sum of normalized input and output tokens; for Anthropic, normalized input includes uncached, cache-read, and cache-creation input tokens, while `cached_prompt_tokens` and `cache_creation_tokens` expose those cache categories separately and should not be added again.
+Across providers, `total_tokens` is normalized as `prompt_tokens + completion_tokens`. Breakdown fields such as `cached_prompt_tokens`, `cache_creation_tokens`, and `reasoning_tokens` are already represented in those normalized totals and should not be added again. Provider adapters account for API-specific usage shapes; for example, Anthropic prompt usage combines uncached, cache-read, and cache-creation input tokens.
 
 ```python Code
 # Access the crew's usage metrics
--- a/docs/edge/en/concepts/crews.mdx
+++ b/docs/edge/en/concepts/crews.mdx
@@ -322,6 +322,8 @@ Caches can be employed to store the results of tools' execution, making the proc
 
 After the crew execution, you can access the `usage_metrics` attribute to view the language model (LLM) usage metrics for all tasks executed by the crew. This provides insights into operational efficiency and areas for improvement.
 
+`total_tokens` is the sum of normalized input and output tokens; for Anthropic, normalized input includes uncached, cache-read, and cache-creation input tokens, while `cached_prompt_tokens` and `cache_creation_tokens` expose those cache categories separately and should not be added again.
+
 ```python Code
 # Access the crew's usage metrics
 crew = Crew(agents=[agent1, agent2], tasks=[task1, task2])
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -1971,10 +1971,13 @@ class AnthropicCompletion(BaseLLM):
             cache_creation_tokens = (
                 getattr(usage, "cache_creation_input_tokens", 0) or 0
             )
+            total_input_tokens = (
+                input_tokens + cache_read_tokens + cache_creation_tokens
+            )
             result: dict[str, Any] = {
-                "input_tokens": input_tokens,
+                "input_tokens": total_input_tokens,
                 "output_tokens": output_tokens,
-                "total_tokens": input_tokens + output_tokens,
+                "total_tokens": total_input_tokens + output_tokens,
                 "cached_prompt_tokens": cache_read_tokens,
                 "cache_creation_tokens": cache_creation_tokens,
             }
--- a/lib/crewai/tests/llms/anthropic/test_anthropic.py
+++ b/lib/crewai/tests/llms/anthropic/test_anthropic.py
@@ -1657,14 +1657,13 @@ def test_anthropic_cache_creation_tokens_extraction():
     mock_response.stop_reason = None
     mock_response.model = None
 
-    usage = llm._extract_anthropic_token_usage(mock_response)
-    assert usage["input_tokens"] == 150
-    assert usage["output_tokens"] == 50
-    assert usage["total_tokens"] == 200
-    assert usage["cached_prompt_tokens"] == 30
-    assert usage["cache_creation_tokens"] == 20
+    with (
+        patch.object(llm._client.messages, "create", return_value=mock_response),
+        patch("crewai.llms.base_llm.crewai_event_bus.emit"),
+    ):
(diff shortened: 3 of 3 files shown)


Linxiushen:
AI assistance disclosure: this contribution was developed with AI assistance and personally reviewed and tested. GitHub denied my attempt to add the required llm-generated label because external contributors do not have AddLabelsToLabelable permission. Maintainers, please apply the llm-generated label.

Linxiushen:
Addressed the review suggestion in `eaddeb5`: the cache-token regression now mocks the Anthropic client, calls the public `llm.call()` path, and asserts the resulting `UsageMetrics`. Event emission is isolated so the test remains deterministic under the suite's `--block-network` setting. Focused result: 2 passed.

Vidit-Ostwal:
Hey @Linxiushen, I think it would be better to add a doc sharing how exactly we are calculating the total tokens, Just a one liner would be fine at this moment.
Lmk if you wanna work on this, can assign to you.

Linxiushen:
Added the requested one-line calculation note in 5f25218. It documents that Anthropic normalized input includes uncached, cache-read, and cache-creation input tokens, and clarifies that the two cache metrics are separate breakdowns rather than values to add to total_tokens again.

Vidit-Ostwal:
I don't think so that it's just a case for anthropic, I think it general way we have decided for each provider, can you check and update ?

Linxiushen:
Checked the shared normalization path and updated the note in a9a70d0. It now states the provider-independent rule: `total_tokens = prompt_tokens + completion_tokens`, while cache/reasoning fields are breakdowns already represented in those totals and must not be added again. Anthropic remains as an example of provider-specific normalization (uncached + cache-read + cache-creation input), rather than being presented as the general rule itself.

coolhandle01:
I am keen to see this, I have a branch too but you beat me to the PR. Let me know if you need a hand :)

Vidit-Ostwal:
Closing as superseded by #6844, which contains the fix.

PR #6823, opened 2026-08-05
fix(llm): infer native provider from name patterns for unprefixed models

### LINKED ISSUE #6813: ANTHROPIC_MODELS has drifted from the CLI model list, so models offered by crewai create route to the OpenAI client
> Per CONTRIBUTING: this issue was written with AI assistance. The `llm-generated` label is required but outside contributors cannot apply labels (403). Could a maintainer add it. Findings below were verified by hand against the files and line numbers cited.

---

## What happens

`LLM(model="claude-sonnet-4-6")` constructs an `OpenAICompletion`, not `AnthropicCompletion`. No error is raised.

`claude-sonnet-4-6` is the second entry in the Anthropic list that `crewai create` offers, so this is reachable straight from the CLI picker.

## Two lists, out of sync

`lib/cli/src/crewai_cli/constants.py:163` offers `claude-opus-4-6` and `claude-sonnet-4-6` at the top of the Anthropic menu.

Neither is in `ANTHROPIC_MODELS` in `lib/crewai/src/crewai/llms/constants.py:212`, which stops at `claude-opus-4-5`. The `AnthropicModels` Literal at `:184` has the same gap.

## Why the mismatch routes wrong

`LLM._infer_provider_from_model` in `lib/crewai/src/crewai/llm.py:634` checks the five constant lists and falls back to `"openai"`:

```python
if model in ANTHROPIC_MODELS:
    return "anthropic"
...
return "openai"
```

Its docstring describes different behavior:

> This method first checks the hardcoded constants list for known models. If not found, it uses pattern matching to infer the provider from model name patterns. This allows supporting new models and "latest" versions without hardcoding.

There is no pattern matching in the function.

The sibling `_validate_model_in_constants` does exactly what that docstring describes: it checks the lists, then falls through to `_matches_provider_pattern` at `:617`. That helper already handles the `claude-` and `anthropic.` prefixes at `:538`. `_infer_provider_from_model` just never calls it.

The bare-model path at `:473` sets `use_native = True`, so the inferred provider goes straight to `_get_native_provider` and the wrong client is constructed.

## Also in `ANTHROPIC_MODELS`

`claude-4-sonnet-20250514` and `claude-4-opus-20250514` are not valid Anthropic ids. Anthropic moved to `claude-<family>-<version>` at Claude 3, and the correct `claude-sonnet-4-20250514` and `claude-opus-4-20250514` are already present in the same list.

Missing beyond the CLI gap: `claude-opus-4-7`, `claude-opus-4-8`, and the Claude 5 family (`claude-opus-5`, `claude-sonnet-5`, `claude-fable-5`).

## Suggested fix

Have `_infer_provider_from_model` fall through to `_matches_provider_pattern` before defaulting to openai, which is what its docstring already promises and what makes the constants list a fast path rather than a gate. Then refresh `ANTHROPIC_MODELS` and drop the two invalid ids.

Happy to send a PR for this. Assign it to me if you want me to take it.

Anai-Guo:
## What

`LLM._infer_provider_from_model` checked only the hardcoded constants lists (`OPENAI_MODELS`, `ANTHROPIC_MODELS`, …) and then defaulted to `"openai"`. Its docstring promises a pattern-matching fallback:

> This method first checks the hardcoded constants list for known models. If not found, it uses pattern matching to infer the provider from model name patterns.

…but no pattern matching ever ran. The sibling `_validate_model_in_constants` already falls through to `_matches_provider_pattern`; this method never did.

## Impact

A bare Anthropic id that is not yet in `ANTHROPIC_MODELS` — e.g. `claude-sonnet-4-6`, offered near the top of the `crewai create` Anthropic menu — was inferred as `"openai"`, so `LLM(model="claude-sonnet-4-6")` silently constructed an `OpenAICompletion` instead of `AnthropicCompletion`, with no error.

## Fix

Fall through to `_matches_provider_pattern` for the providers whose prefixes are unambiguous (`anthropic`, `gemini`) before defaulting to `openai`. `bedrock` (`"." in name`) and `azure` (gpt-/o-series) are intentionally excluded here because their patterns overlap other providers and would reroute models that currently resolve to openai; those remain on the explicit-provider path. Models already present in the constants lists are unaffected (they return before the new fallback).

## Test

Adds `test_infer_provider_from_model_falls_back_to_patterns`, mirroring the existing `test_validate_model_in_constants`: in-constants ids keep resolving as before, `claude-sonnet-4-6` / `claude-future-5` / `gemma-3-latest` now resolve to their real provider, and a non-matching id still defaults to `openai`.

Fixes #6813

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -659,6 +659,17 @@ class LLM(BaseLLM):
         if model in AZURE_MODELS:
             return "azure"
 
+        # Models absent from the constants lists (new releases, "latest"
+        # aliases) still fall back to provider naming patterns, as this
+        # method's docstring describes and the sibling
+        # _validate_model_in_constants already does. Only providers whose
+        # prefixes are unambiguous are checked here; bedrock ("." in name)
+        # and azure (gpt-/o-series) patterns overlap other providers and are
+        # left to the explicit-provider path.
+        for provider in ("anthropic", "gemini"):
+            if cls._matches_provider_pattern(model, provider):
+                return provider
+
         return "openai"
 
     @classmethod
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -662,13 +662,17 @@ class LLM(BaseLLM):
         # Models absent from the constants lists (new releases, "latest"
         # aliases) still fall back to provider naming patterns, as this
         # method's docstring describes and the sibling
-        # _validate_model_in_constants already does. Only providers whose
-        # prefixes are unambiguous are checked here; bedrock ("." in name)
-        # and azure (gpt-/o-series) patterns overlap other providers and are
-        # left to the explicit-provider path.
-        for provider in ("anthropic", "gemini"):
-            if cls._matches_provider_pattern(model, provider):
-                return provider
+        # _validate_model_in_constants already does. bedrock ("." in name)
+        # and azure (gpt-/o-series) patterns overlap other providers, so
+        # they are left to the explicit-provider path.
+        # Only unambiguous prefixes fall through here: the "anthropic."
+        # prefix that _matches_provider_pattern also accepts is Bedrock's
+        # namespaced form and stays on the explicit "bedrock/" path.
+        if model.lower().startswith("claude-"):
+            return "anthropic"
+
+        if cls._matches_provider_pattern(model, "gemini"):
+            return "gemini"
 
         return "openai"
 
--- a/lib/crewai/tests/test_llm.py
+++ b/lib/crewai/tests/test_llm.py
@@ -974,6 +974,30 @@ def test_validate_model_in_constants():
         is True
     )
 
+
+def test_infer_provider_from_model_falls_back_to_patterns():
+    """Unprefixed models not in the constants lists still infer their provider
+    from naming patterns, matching this method's docstring and the behavior of
+    _validate_model_in_constants.
+
+    Regression test for #6813: a bare ``claude-*`` id offered by
+    ``crewai create`` (e.g. ``claude-sonnet-4-6``) is not in ANTHROPIC_MODELS,
+    so it fell through to the ``"openai"`` default and built the wrong client.
+    """
+    # In-constants ids keep resolving exactly as before.
+    assert LLM._infer_provider_from_model("gpt-4o") == "openai"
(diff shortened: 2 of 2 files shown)


Vidit-Ostwal (commented):
Keeping this PR for #6813 — it is the narrower fix (anthropic + gemini pattern fallback only), which avoids the azure/openai overlap.

Please rebase onto current `main` (`mergeStateStatus` is BEHIND). Also address the CodeRabbit note: `anthropic.*` ids should stay on the explicit Bedrock path rather than matching the Anthropic prefix fallback.

Anai-Guo:
Thanks @Vidit-Ostwal — CodeRabbit's note is addressed in 778aa4e.

`_matches_provider_pattern(model, "anthropic")` accepts two prefixes, `claude-` **and** `anthropic.`, and the second one is Bedrock's namespaced form. Delegating the whole predicate therefore pulled Bedrock ids into the new fallback. The fallback now only uses the bare `claude-` prefix, so `anthropic.*` ids keep resolving exactly as they do on `main` and stay on the explicit `bedrock/` path:

```python
if model.lower().startswith("claude-"):
    return "anthropic"

if cls._matches_provider_pattern(model, "gemini"):
    return "gemini"
```

The test pins both Bedrock forms so this cannot regress again:

```python
assert LLM._infer_provider_from_model("anthropic.claude-opus-4-1-20250805-v1:0") == "bedrock"   # in BEDROCK_MODELS
assert LLM._infer_provider_from_model("anthropic.claude-future-v1:0") == "openai"               # not in constants
```

I ran `test_infer_provider_from_model_falls_back_to_patterns` against three revisions of `_infer_provider_from_model` (real file, real `llms/constants.py`):

| revision | result |
|---|---|
| `main` (no fallback) | ❌ fails on `claude-sonnet-4-6` → the reported bug |
| previous commit (delegated to `_matches_provider_pattern`) | ❌ fails on `anthropic.claude-future-v1:0` → the regression you flagged |
| this commit | ✅ passes |

On the rebase: the branch still shows `BEHIND`, which I can't clear from my side — my token can't push a tree containing the updated `.github/workflows/*` that syncing my fork's `main` would bring along. It should be a no-op for review purposes: `llms/constants.py` is byte-identical between my base and current `main`, and the drift in `llm.py` / `test_llm.py` is entirely outside the hunks here (the PR still reports `MERGEABLE`, and I re-applied this same patch on top of current `main`'s `llm.py` and the test passes there too). If a fast-forward is required before merge, the **Update branch** button on this PR does it.

PR #6825, opened 2026-08-05
fix(flow): clarify conversational route/handler name collision errors

### LINKED ISSUE #6767: Conversational Flow golden use case improvements - route labels collide with method names in @listen validation

Vidit-Ostwal:
## Summary

- Improve validation errors when a Flow handler's `@listen(...)` trigger includes its own method name — the common conversational mistake of naming a handler after its route label (e.g. `@listen("create_video")` on `def create_video`).
- Two message variants: **conversational** (route label vs handler name) and **default** (method self-listen loop). Both include the listen condition in the text.
- Wrap Python DSL `build_flow_definition` failures with `Invalid flow definition for {ClassName}:` so `MyFlow()` errors reference the class the author wrote.
- Document the `handle_*` naming pattern in conversational flows guide.

## When each error surfaces

| Message | When | Example | How it appears |
|---------|------|---------|----------------|
| **Conversational** | `conversational.enabled` and listen references handler name | `@listen("create_video")` on `def create_video` | See paths below |
| **Default** | Non-conversational flow, same trigger check | `@listen("publish")` on `def publish` | See paths below |
| **Class-name wrapper** | Python Flow class built via DSL only | `MyFlow()` or `MyFlow.flow_definition()` | Prepends `Invalid flow definition for MyFlow:` |

**Paths:**
- `FlowDefinition.from_declaration(yaml/dict)` → pydantic `ValidationError` for `FlowDefinition`
- `MyFlow.flow_definition()` → `ValueError` with class-name wrapper
- `MyFlow()` → same underlying message, wrapped by pydantic as `ValidationError for MyFlow` on instantiation

## Test plan

- [x] `pytest lib/crewai/tests/test_flow_definition.py -k "self_listen or conversational_route or build_flow_definition" -n0`
- [x] `pytest lib/crewai/tests/test_flow.py::test_self_listening_method_is_rejected -n0`
- [x] Manual: conversational `@listen("create_video")` / `def create_video` shows conversational wording with listen condition

### DIFF
--- a/docs/edge/en/guides/flows/conversational-flows.mdx
+++ b/docs/edge/en/guides/flows/conversational-flows.mdx
@@ -376,6 +376,27 @@ def handle_internet_search(self) -> str:
     ...
 ```
 
+### Naming handlers
+
+The string in `@listen("…")` is a **router route label** (an event name), not the Python method name. Route labels and method completion events share one trigger namespace, so naming a handler the same as its route causes the handler to re-trigger itself in a loop.
+
+Use a different method name — the docs examples use a `handle_*` prefix:
+
+```python
+@listen("create_video")
+def handle_create_video(self) -> str:
+    """User wants a new video."""
+    ...
+```
+
+Do **not** mirror the route label on the method:
+
+```python
+@listen("create_video")
+def create_video(self) -> str:  # rejected at flow instantiation
+    ...
+```
+
 …and the router LLM sees:
 
 ```
--- a/lib/cli/src/crewai_cli/templates/AGENTS.md
+++ b/lib/cli/src/crewai_cli/templates/AGENTS.md
@@ -627,6 +627,26 @@ class MyFlow(Flow):
 | `@listen(method)` | Triggers when specified method completes. Receives output as argument |
 | `@router(method)` | Conditional branching. Returns string labels that trigger `@listen("label")` |
 
+### `@listen` labels vs handler names
+
+The string in `@listen("...")` is an **event or route label**, not the Python method name. Router return values, route labels, and method completion events share one trigger namespace.
+
+**Never** use the same name for the `@listen` label and the handler method:
+
+```python
+# ❌ Wrong — validation error; handler would re-trigger itself in a loop
+@listen("create_video")
+def create_video(self):
+    ...
+
+# ✅ Correct — distinct handler name (handle_* prefix is a common pattern)
+@listen("create_video")
+def handle_create_video(self):
+    ...
+```
+
+This applies to all flows. It is especially common in **conversational flows** (`conversational = True`), where `@listen("...")` is a router intent name — do not name the handler after the route it serves.
+
 ### Structured State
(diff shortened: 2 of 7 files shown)


lorenzejay (approved):
lgtm

PR #6838, opened 2026-08-06
fix(anthropic): count cached input tokens in the reported totals

arthi-arumugam-git:
## The problem

Anthropic reports `input_tokens` **net of the prompt cache** and bills `cache_read_input_tokens` and `cache_creation_input_tokens` on top of it. It never sends a total.

`_extract_anthropic_token_usage` read both cache counters, surfaced them as their own keys, and then built `total_tokens` from `input_tokens + output_tokens` only. So the total is short by the cached amount, and nothing raises. The number is simply smaller than what was charged for.

On a call with 200 cached, 50 fresh input and 100 output tokens:

| | reported | billed |
|---|---|---|
| `total_tokens` | 150 | 350 |

A second symptom makes it self-evident: on that same call `cached_prompt_tokens` comes back as **200** while `prompt_tokens` is **50**, so the cached subset is larger than the set it belongs to.

## Why this is a cross-provider inconsistency, not just a missing addition

Every other provider here reports a prompt count that already contains its cached portion. OpenAI nests `cached_tokens` inside `prompt_tokens`; Gemini uses `total_token_count` and Bedrock uses the API's own total. Anthropic is the only one where the cached tokens sit outside the number that gets mapped onto `prompt_tokens`.

`UsageMetrics.from_provider_dict` states the invariant this breaks in its own docstring:

> Mirrors `BaseLLM._track_token_usage_internal` so per-LLM totals, flow-level aggregation, and OTel spans agree on every provider.

They do not agree. The same billed work reported through the Anthropic path and the OpenAI path produces 150 and 350.

## The fix

Normalise Anthropic to the OpenAI-compatible shape at the point of extraction: `prompt_tokens` becomes the full billed input, and `cached_prompt_tokens` stays a **subset** of it rather than a disjoint number a consumer might add on top. That keeps `cache_creation_tokens` available separately for anyone pricing cache writes at their own rate.

It flows through `UsageMetrics`, `BaseLLM._track_token_usage_internal`, `CrewOutput.token_usage` and the OTel span without further change.

## Tests

Eight, in `tests/llms/anthropic/test_anthropic_cached_token_totals.py`. **Four fail before this change**, including a direct parity assertion that the Anthropic and OpenAI shapes of identical work produce identical `UsageMetrics`.

The other four pin the edges so the fix cannot regress them: an uncached call is unchanged, absent cache fields are treated as zero, `None` cache fields are treated as zero, and a response with no usage still returns `{"total_tokens": 0}`.

Verified against `develop` at `18c52c4`. The three failures in `tests/test_flow_usage_metrics.py` are present on a clean checkout without this change and are unrelated.

## Context

This is the sixth framework I have found with the same defect. The equivalent fixes are merged in [Pipecat](https://github.com/pipecat-ai/pipecat/pull/5163), [LiveKit](https://github.com/livekit/agents/pull/6663) and [Haystack](https://github.com/deepset-ai/haystack-core-integrations/pull/3717), the last of which took exactly this OpenAI-compatible approach.

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
+++ b/lib/crewai/src/crewai/llms/providers/anthropic/completion.py
@@ -1971,10 +1971,19 @@ class AnthropicCompletion(BaseLLM):
             cache_creation_tokens = (
                 getattr(usage, "cache_creation_input_tokens", 0) or 0
             )
+            # Anthropic reports input_tokens NET of the prompt cache and bills
+            # cache_read_input_tokens and cache_creation_input_tokens on top of
+            # it, so input_tokens alone is not what the request was charged for.
+            # OpenAI and Gemini report a prompt count that already contains its
+            # cached portion, so normalise to that shape here: prompt_tokens is
+            # the full billed input and cached_prompt_tokens is the subset of it
+            # that was served from cache. Without this, total_tokens is short by
+            # the cached amount on Anthropic alone and nothing raises.
+            prompt_tokens = input_tokens + cache_read_tokens + cache_creation_tokens
             result: dict[str, Any] = {
-                "input_tokens": input_tokens,
+                "input_tokens": prompt_tokens,
                 "output_tokens": output_tokens,
-                "total_tokens": input_tokens + output_tokens,
+                "total_tokens": prompt_tokens + output_tokens,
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:
Fixed here https://github.com/crewAIInc/crewAI/pull/6844, thanks

PR #6841, opened 2026-08-06
fix(crewai-files): replace bare raises with PermanentUploadError in get_uploader

jpcj223:
## Problem

Bare `raise` statements in `get_uploader()` are used outside of `except` blocks, which causes a confusing `RuntimeError: No active exception to re-raise` instead of a meaningful error message.

Two affected locations:
1. When an unsupported provider is passed (line 216)
2. When bedrock provider is used without S3 bucket config (line 199)

## Solution

Replace both bare `raise` statements with `PermanentUploadError` so users get a clear, actionable error message.

## Changes

- `lib/crewai-files/src/crewai_files/uploaders/factory.py`:
  - Import `PermanentUploadError` from `crewai_files.processing.exceptions`
  - Replace bare raise for unsupported provider with `PermanentUploadError` listing supported providers
  - Replace bare raise for unconfigured bedrock with `PermanentUploadError` explaining how to configure
- `lib/crewai-files/tests/test_uploader_factory.py`:
  - Add unit tests for `get_uploader()` covering all supported providers and error paths

## Testing

All 7 new tests pass:

### DIFF
--- a/lib/crewai-files/src/crewai_files/uploaders/factory.py
+++ b/lib/crewai-files/src/crewai_files/uploaders/factory.py
@@ -189,9 +189,8 @@ def get_uploader(
     if "bedrock" in provider_lower or "aws" in provider_lower:
         import os
 
-        if (
-            not os.environ.get("CREWAI_BEDROCK_S3_BUCKET")
-            and "bucket_name" not in kwargs
+        if not os.environ.get("CREWAI_BEDROCK_S3_BUCKET") and not kwargs.get(
+            "bucket_name"
         ):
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:
Closing in favor of #7283.

Same bare-`raise` bug, but we are taking the discussion there. Direction: raise a `ValueError` with a concrete reason (unknown provider / unconfigured Bedrock), not `PermanentUploadError` (the upload path swallows that) and not a silent `None`. Follow along on #7283.

Issue #6826, opened 2026-08-05
[FEATURE] Add OutageDeck status tool


koko3tallah:
### Feature Area

Integration with external tools

### Is your feature request related to an existing bug? Please link it here.

NA

### Describe the solution you'd like

Add a built-in `OutageDeckStatusTool` to `crewai-tools` so agents can check current infrastructure-provider status without relying on stale model knowledge or configuring credentials.

The tool should use OutageDeck's public, read-only API and support three operations through one compact schema:

- current provider status with services and active incidents
- paginated incidents with optional provider, lifecycle, and severity filters
- individual service status with bounded recent-incident context

It should fix requests to the production API origin, strictly validate slugs and filters before any network call, return structured JSON, and include sync and async execution. No OutageDeck account or API key should be required.

### Describe alternatives you've considered

- Directly querying each vendor status API preserves direct-vendor provenance but requires separate parsers and lifecycle mappings for every provider.
- CrewAI's MCP adapter can connect to the remote OutageDeck MCP server, but it requires the MCP extra and explicit adapter setup. A native tool gives ordinary Python crews a zero-credential, in-process path.
- A generic HTTP tool can call the API, but every user must reproduce the validation, response shaping, error handling, and LLM-facing operation descriptions.

### Additional context

The public API contract is documented at https://outagedeck.com/docs/api?utm_source=crewai&utm_medium=integration&utm_campaign=crewai_tool. The proposed methods are GET-only, use a fixed HTTPS origin, and do not accept a configurable base URL.

This feature request was authored by an AI agent and is labeled `llm-generated` accordingly.

### Willingness to Contribute

Yes, I'd be happy to submit a pull request.

koko3tallah:
AI-authorship disclosure: this issue was authored by an AI agent. I attempted to apply the required `llm-generated` label, but GitHub does not grant outside contributors permission to label upstream issues. Maintainers or automation: please apply `llm-generated`. The disclosure is also present in the issue body.

jpcj223:
I've submitted a PR implementing this feature: #6842

It adds `OutageDeckStatusTool` with three operations:
- `provider_status` — overall provider status with services and active incidents
- `list_incidents` — paginated incidents with provider/lifecycle/severity filters
- `service_status` — individual service status

Both sync and async execution are supported. No API key required (uses the public read-only API).

Vidit-Ostwal:
This is a third-party product pitch (OutageDeck), not a CrewAI bug or feature gap.

[...]

PR #6842, opened 2026-08-06
feat(tools): add OutageDeckStatusTool

jpcj223:
## Description

Adds a built-in `OutageDeckStatusTool` that lets agents check the current status of infrastructure providers and services using the OutageDeck public read-only API. No API key or account required.

The tool uses a compact schema with three operations:
- **`provider_status`** — overall provider status, including services list and active incidents
- **`list_incidents`** — paginated incident list with optional `provider`, `lifecycle`, and `severity` filters
- **`service_status`** — individual service status with limited recent incident context

## Design decisions

- **Single tool + action parameter** — compact schema, all three operations are conceptually the same "status check" category
- **Strict input validation before network calls** — slugs are validated against a regex, lifecycle/severity are validated against known enums, using Pydantic field validators
- **Fixed base URL** — always points at the production API, not configurable
- **Sync + async** — both `_run` and `_arun` are implemented, using httpx
- **Structured JSON output** — returns the full API response as formatted JSON so the agent has all the data it needs

## Changes

- `lib/crewai-tools/src/crewai_tools/tools/outage_deck_status_tool/outage_deck_status_tool.py` — tool implementation
- `lib/crewai-tools/src/crewai_tools/tools/outage_deck_status_tool/__init__.py` — package init
- `lib/crewai-tools/tests/tools/test_outage_deck_status_tool.py` — 14 unit tests
- `lib/crewai-tools/src/crewai_tools/tools/__init__.py` — register tool
- `lib/crewai-tools/src/crewai_tools/__init__.py` — register tool

## Testing

13 passed, 1 skipped (async test skipped on Windows due to pytest-recording + asyncio event loop compatibility — CI on Linux covers async paths).

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/__init__.py
@@ -123,6 +123,9 @@ from crewai_tools.tools.multion_tool.multion_tool import MultiOnTool
 from crewai_tools.tools.mysql_search_tool.mysql_search_tool import MySQLSearchTool
 from crewai_tools.tools.nl2sql.nl2sql_tool import NL2SQLTool
 from crewai_tools.tools.ocr_tool.ocr_tool import OCRTool
+from crewai_tools.tools.outage_deck_status_tool.outage_deck_status_tool import (
+    OutageDeckStatusTool,
+)
 from crewai_tools.tools.oxylabs_amazon_product_scraper_tool.oxylabs_amazon_product_scraper_tool import (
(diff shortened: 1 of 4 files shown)


kingsoul0018-jpg (commented):
[](url) @

Vidit-Ostwal:
Closing with #6826. Same OutageDeck listing as the issue: a public GET API does not need a first-party CrewAI tool. Use `MCPServerAdapter` or `UrlReadTool` instead.

PR #6862, opened 2026-08-07
Fix conversational @persist dropping return-only @listen replies

### LINKED ISSUE #6766: Conversational Flow golden use case improvements - @persist silently drops assistant replies from custom @listen routes
## Description

When a conversational `Flow` is combined with `@persist`, replies produced by **custom `@listen` routes are never written to the database**, while replies from the **built-in `converse_turn` route are**. Across turns the assistant side of the conversation silently disappears, leaving a user-only monologue — so the agent can no longer recall anything it previously said.

This only shows up when each turn runs on a **fresh `Flow` instance** — the normal pattern for a web server, where a request handler constructs the flow and calls `handle_turn(...)`. In-memory state looks correct, so the bug is invisible in a single long-lived process.

### Root cause

`handle_turn` appends the assistant message *after* `self.kickoff()` returns:

```python
# crewai/experimental/conversational_mixin.py ~L325
object.__setattr__(self, "_assistant_reply_appended", False)
result = self.kickoff(inputs={"id": sid}, **kickoff_kwargs)
if (
    result is not None
    and not self._assistant_reply_appended
    and self._is_public_turn_result(result)
):
    self.append_assistant_message(self._stringify_result(result))
```

`@persist` snapshots state per-method *inside* `kickoff()`. So this append lands **after the final snapshot** and is never persisted.

The built-in `converse_turn` is unaffected because it calls `append_assistant_message()` *inside* the method body (`conversational_mixin.py` L229/L242/L251), so its reply is part of that method's snapshot. A custom `@listen` handler that simply returns a string relies on the post-`kickoff` fallback — and loses the message.

Net effect: **built-in routes persist their replies; custom routes silently don't.**

## Steps to Reproduce

1. Define a conversational `Flow` with `@persist` and one custom `@listen` route.
2. Run two turns, each on a **new** flow instance with the same `session_id`.
3. Dump `state.messages` from the persistence DB.

## Expected behavior

The assistant reply should be persisted regardless of which route produced it. After two turns the stored history should be `['user', 'assistant', 'user', 'assistant']`.

## Screenshots/Code snippets

```python
"""Minimal repro: @persist + conversational=True loses assistant messages."""
import json, sqlite3, tempfile, os
from crewai import LLM
from crewai.experimental.conversational import ConversationConfig, ConversationState, RouterConfig
from crewai.flow.flow import Flow, listen
from crewai.flow.persistence import persist
from crewai.flow.persistence.sqlite import SQLiteFlowPersistence

DB = os.path.join(tempfile.mkdtemp(), "repro.db")
llm = LLM(model="anthropic/claude-haiku-4-5")

@ConversationConfig(llm=llm, router=RouterConfig(llm=llm,
    route_descriptions={"greet": "The user greets or says hello."},
    default_intent="greet", fallback_intent="greet"))
@persist(SQLiteFlowPersistence(DB))
class GreetFlow(Flow[ConversationState]):
    conversational = True

    @listen("greet")
    def do_greet(self) -> str:
        return "Hello! I am the assistant."

SID = "repro-session"
for turn, msg in enumerate(["hi there", "hello again"], 1):
    flow = GreetFlow()                      # fresh instance per turn (web-server pattern)
    flow.handle_turn(msg, session_id=SID)
    print(f"turn {turn} in-memory : {[m.role for m in flow.state.messages]}")

print("\n--- what actually persisted ---")
c = sqlite3.connect(DB)
for m, s in c.execute("select method_name, state_json from flow_states order by timestamp"):
    print(f"  {m:20}", [x.get("role") for x in json.loads(s).get("messages", [])])
```

Actual output:

```
turn 1 in-memory : ['user', 'assistant']
turn 2 in-memory : ['user', 'user', 'assistant']     <-- turn 1's assistant reply is gone

--- what actually persisted ---
  route_conversation   ['user']
  do_greet             ['user']                       <-- custom route: reply NOT persisted
  route_conversation   ['user', 'user']
  converse_turn        ['user', 'user', 'assistant']  <-- built-in route: reply IS persisted
```

Note how turn 2 restored `['user', 'user']` — the assistant turn never made it to disk — and how `converse_turn` behaves differently from `do_greet`.

## Impact

The concrete failure is **the agent cannot recall its own prior statements**. Once the assistant side is missing from restored history, `converse_turn` (and anything else answering from history) has no record of what was already said:

```python
# turn 1 -> custom route replies:
#   "Directed a first cut of 'TaskPilot' — 4 scenes, 28 seconds, coral accent."
VF().handle_turn("build a launch video for my todo app TaskPilot", session_id=SID)

# turn 2, fresh instance, same session:
f2 = VF()
f2.handle_turn("how long did you say it was?", session_id=SID)
print([(m.role, m.content[:40]) for m in f2.state.messages])
```

```
[('user', 'build a launch video for my todo app Tas'),
 ('user', 'how long did you say it was?'),          <-- turn 1's reply is gone
 ('assistant', "I haven't actually given a length yet")]
```

The model answers: *"I haven't actually given a length yet — this is the first I'm hearing about TaskPilot, so there's no prior estimate for me to point back to."* — one turn after stating 28 seconds.

It goes unnoticed for a long time because in-memory behaviour is correct; it only appears once state round-trips through persistence.

For what it's worth, we did **not** manage to demonstrate degraded *router* accuracy from the missing assistant turns — across a set of clear and ambiguous follow-ups the router picked the same intent with and without assistant messages in history. The reproducible damage is to recall / history-answering, not intent classification.

## Workaround

Call the framework's own API *inside* the handler, so the append happens before the snapshot:

```python
@listen("greet")
def do_greet(self) -> str:
    text = "Hello! I am the assistant."
    self.append_assistant_message(text)   # lands inside the @persist snapshot
    return text
```

This also sets `_assistant_reply_appended`, which suppresses `handle_turn`'s duplicate append, and emits the `conversation_message_added` trace event. It's the right fix for user code, but it isn't discoverable — the failure is silent and only appears across process/instance boundaries.

## Suggested fix

Either:
- have `handle_turn` append the assistant reply **before** the final persistence snapshot, or
- persist once at end-of-turn (after `handle_turn` completes) in addition to per-method snapshots, or
- at minimum, document that custom `@listen` routes must call `append_assistant_message()` themselves when combined with `@persist`.

## Environment

- crewai `1.15.10`
- Python 3.12
- macOS (Darwin 25.1)
- `SQLiteFlowPersistence`

Vidit-Ostwal:
## Summary
- Re-persist flow state after `handle_turn` / `stream_turn` fallback append so return-only `@listen` handlers survive `@persist` restore on the next turn
- Reuse `_persist_method_completion` with the last executed listener method from `_method_outputs`
- Add SQLite regression test covering fresh Flow instances sharing the same `session_id`

## Test plan
- [x] `pytest tests/test_flow_conversation.py::TestHandleTurnReplyFallback -o addopts=`
- [x] `pytest tests/test_flow_conversation.py::TestHandleTurnReplyFallback::test_return_only_handler_persists_assistant_for_fresh_instance -o addopts=`


anujbolewar (commented):
Re-persisting after the fallback append so a return-only listen handler is visible to restore on the next turn is a reasonable correctness fix. The guard that only persists when the last method output maps to a method is sensible. One concern: calling into the runtime flow persist from within the mixin reintroduces a runtime import dependency, so confirm the circular import path is fully safe when the mixin is used without a flow, and add a test that simulates the snapshot-before-fallback ordering you are compensating for.

PR #6872, opened 2026-08-08
fix(tools): block CGNAT and non-global IPs in validate_url

SashaMIT:
## Summary

Sibling of #6870. Shared `validate_url` (used by host scrapers via `safe_get`, and by Selenium once #6870 lands) only checked a hand-maintained IPv4 list. That list missed:

- CGNAT / shared address space `100.64.0.0/10` (`is_private` is False there)
- Documentation / benchmarking / Class E style non-global ranges
- The rest of `0.0.0.0/8` beyond `0.0.0.0/32`

So a prompt-injected scrape URL to e.g. `http://100.64.0.1/` passed validation at tip `92012aec`.

## Fix

- Treat any address with `not ipaddress…is_global` as blocked (covers CGNAT)
- Keep explicit networks as belt-and-suspenders; expand `0.0.0.0/8` and add `100.64.0.0/10`

## Testing

`uv run pytest tests/utilities/test_safe_path.py` → 36 passed (new CGNAT / 198.18 / 192.0.2 / 0.0.0.1 cases).

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/security/safe_path.py
+++ b/lib/crewai-tools/src/crewai_tools/security/safe_path.py
@@ -162,21 +162,36 @@ _BLOCKED_IPV6_NETWORKS = [
 ]
 
 
+_NAT64_WELL_KNOWN = ipaddress.ip_network("64:ff9b::/96")
+
+
+def _unwrap_embedded_ipv4(addr: ipaddress.IPv4Address | ipaddress.IPv6Address):
+    """Return embedded IPv4 for IPv4-mapped or well-known NAT64 addresses."""
+    if isinstance(addr, ipaddress.IPv6Address):
+        if addr.ipv4_mapped:
+            return addr.ipv4_mapped
+        if addr in _NAT64_WELL_KNOWN:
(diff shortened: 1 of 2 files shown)


manjunathbhaskar:
the `is_global` check you just added (safe_path.py:182ish) does not catch multicast — `224.0.0.1`, `239.255.255.250` still get through since `ipaddress` reports `is_global=True` for multicast.

also `64:ff9b::/96` (NAT64 prefix) reads as global even when it's wrapping `169.254.169.254` or `127.0.0.1` underneath:

`ipaddress.ip_address("64:ff9b::a9fe:a9fe").is_global` -> `True`

same unwrap trick you're doing for `::ffff:.../96` a few lines up would work here too, plus an `is_multicast` check. worth folding in before merge.

SashaMIT:
Good catch @manjunathbhaskar — folded in on `05c67d4`:

- reject `is_multicast` (covers `224.0.0.1` / `239.255.255.250`)
- unwrap well-known NAT64 `64:ff9b::/96` the same way as `::ffff:` so embeds like `64:ff9b::a9fe:a9fe` (169.254.169.254) fail closed

Added regression tests for both. Ready for another look.

PR #6875, opened 2026-08-08
Draft pull: feature request_human_approval [work in progress]

### LINKED ISSUE #6859: [FEATURE] Tool-Level Human-in-the-Loop (HITL) for Sensitive Action approval
### Feature Area

Core functionality

Problem: Currently, the Crew AI system handles the HITL interctions on the task level, once the model runs a query, the execution of different tasks within the query are operated. In many real world scenarios the tasks can have different level of sensitivity as follows:

Case 1: Scrap data from kaggle and create a summary of statistical features of the data.

In this case, there is no sensitive execution as of now, the model performs as per the query, requires no additional approval. 

Case 2: Analyze the transaction pattern and freeze the crypto wallet. 

In such cases, where there is a high stake or financial executions, a subtle inaccuracy in the prompt or the model understanding can cause the model to freeze a wallet or make a bigger wire transfer [for example].

The solution is to propose a parameter called "requires_human_approval" directly inside the tool decorator, which defaults to "false" for base cases such as case 1, and can be set as true for sensitive cases like case 2.

When an agent attempts to use a tool with this flag set to True, the execution pauses, shows the user the exact arguments the LLM wants to use, and waits for a Y/n or text feedback. If rejected, the feedback is sent back to the agent as an observation so it can course-correct.

Timeout Fallback: Introduce a human_approval_timeout (e.g., default 60 seconds). If no human responds, the tool safely aborts and returns an "Approval Timeout" observation to the agent.

Callback Support: Allow the approval prompt to trigger a callback function instead of a terminal input(), so developers can route the approval request to a Slack webhook, a Streamlit button, or a custom UI.

None for now, open to alternative ideas as per the feedbacks.

### Additional context

Design examples:

**Case 1: Default Behavior (Fully Autonomous)**
```python
from crewai.tools import tool

# Defaults to False. Runs automatically without pause.
@tool("Kaggle Scraper")
def scrape_kaggle(dataset_url: str):
    """Scrapes statistical features from Kaggle data."""
    return fetch_data(dataset_url)
```

Case 2:

**Case 2: Sensitive Execution (Requires Approval)**
```python
# User explicitly opts-in to the safety check
@tool("Freeze Wallet", requires_human_approval=True)
def freeze_wallet(wallet_address: str):
    """Freezes a suspected illicit crypto wallet."""
    return execute_freeze(wallet_address)
```

### Willingness to Contribute

AryanSharma21:
closes #6859


sattyamjjain:
Thanks for turning this around quickly, and for taking the predicate suggestion seriously. The `Any`-typed field accepting a callable, fail-closed on predicate error, and the rejected/timeout tests all go the right way. Notes below are mostly about paths that don't line up yet rather than the direction.

**1. The predicate receives different arguments depending on which path runs.**

`base_tool.py` calls `_should_require_approval(*args, **kwargs)` with raw arguments. `structured_tool.py` calls it with `**parsed_args` after schema validation. So a predicate written as `lambda **kwargs: kwargs.get("amount", 0) > 1000` works on one path, and on the other a positional call raises `TypeError`, gets caught, and returns `True`.

Fail-closed, so nothing unsafe happens. But the predicate never actually evaluates, and every call ends up requiring approval, which is the approval-fatigue outcome the predicate was meant to avoid. Worth binding to the schema first and evaluating against named arguments on both paths.

**2. There isn't a timeout.**

`input()` blocks indefinitely, and nothing in that `try` can time out, so "Fail closed on timeout" isn't what the code does. What `except Exception` actually catches is `EOFError`, which `input()` raises immediately when stdin isn't a TTY. So in CI or any non-interactive `kickoff()`, every approval-gated tool raises `TimeoutError("Approval timed out")` on the first call. Failing closed there is right; the message just isn't true, and `human_approval_timeout` from the issue isn't implemented yet.

**3. Rejection burns the usage budget.**

Both paths do `_increment_usage_count()` before the approval check, so denying three times on a `max_usage_count=3` tool exhausts it without the tool ever running. Moving the check above the increment covers it.

Two structural things I'd raise now because they get harder later.

The two methods are copy-pasted into both classes and have already drifted. The missing `time` import is in the copy, and the indentation differs too. Every fix in this review has to be applied twice. One shared mixin removes that class of bug.

And on the audit trail: `handle_tool_failure` in `tool_failure.py` already builds a `ToolFailureRecord` with `tool_name`/`tool_args`/`agent_role`/`task_id` and emits it on `crewai_event_bus`. Emitting a structured approval event the same way gets you a real record instead of an f-string, keeps raw arguments out of stdout and the application log, and gives the Slack/Streamlit path somewhere to hook without subclassing.

One correction to something I said on the issue, now that I've read `tool_failure.py`: your `raise PermissionError` does give you the "rejection comes back as an observation" behaviour, because `ToolFailureReason.EXCEPTION` is documented as "the framework caught it and fed the text to the agent" and the default `WARN` policy keeps going. So that part works. What it doesn't do is make the denial stick. The agent is told it was rejected and can call the same tool again on the next step. Fine to leave for a follow-up, but worth writing down as a known limit rather than leaving it implied.

Smaller ones: `requires_human_approval: Any` will fight the required type-checker job (`bool | Callable[..., bool]`), `import logging, time` trips ruff E401 and `lint` is required, `audit_data` is built then unused, `ainvoke` calls blocking `input()` on the event loop, and the human's typed text is discarded when it isn't in the approve-list even though the issue asked for text feedback.

Good direction overall. 1 and 3 are the ones I'd fix before asking a maintainer to look.

AryanSharma21:
Hey @sattyamjjain ... thanks for the in depth review.. I'll adjust accordingly

Vidit-Ostwal:
Closing with #6859. Tool-level HITL already exists via `PRE_TOOL_CALL` hooks and `request_human_input()` (https://docs.crewai.com/en/learn/tool-hooks). No need for a parallel `@tool(requires_human_approval=True)` path in core.

PR #6881, opened 2026-08-09
fix: make pre-commit hooks portable on Windows

### LINKED ISSUE #6863: [BUG] pre-commit hooks fail on Windows due to hardcoded Unix virtual environment path
### Description

The hooks currently use:

```bash
source .venv/bin/activate
```

### Steps to Reproduce

4. Run:

```bash
```

### Expected behavior

### Screenshots/Code snippets

Current hook configuration:

```yaml
```

Error:

```text
```

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

1.15.12

### crewAI Tools Version

N/A

### Virtual Environment

Venv

### Evidence

Running:

```bash
```

produces:

```text

```

```bash
uv run ruff check lib/
```

### Possible Solution

```bash
source .venv/bin/activate
```

### Additional context

Vaishnavi220506:
## Summary

Fixes #6863.

The local pre-commit hooks always sourced `.venv/bin/activate`, which is not present in virtual environments created on Windows. They now select `.venv/Scripts/activate` for MSYS, Cygwin, and Win32 shells, while keeping `.venv/bin/activate` on Unix-like systems before running the existing `uv run` commands.

## Validation

- `pre-commit validate-config` — passed
- Ruff hook through Git Bash on Windows — passed
- Ruff Format hook through Git Bash on Windows — passed
- The previous full GitHub Actions run passed; fresh checks will run for the latest scope-only commit


Vaishnavi220506:
@coderabbitai review

Vidit-Ostwal (commented):
Thanks — this is the right fix. Dropping `source .venv/bin/activate` *and* the `bash -c` wrapper is the portable version (Git Bash is no longer required). `pass_filenames: true` still appends paths onto `uv run ruff/mypy ...`, so that behavior is preserved.

Please rebase onto `main` before we can merge (`mergeStateStatus` is CONFLICTING). The `pip-audit` ignore list on `main` has grown (e.g. `PYSEC-2026-139`, `GHSA-2wm9-hf6c-p5cr`, `GHSA-36p7-vc44-83pf`, `GHSA-xph7-9rjv-w5fr`). Keep those entries; only change the `entry` wrapper from `bash -c 'source .venv/bin/activate && uv run ...'` to `uv run ...`.

CodeRabbit's nit on the regression test is optional: asserting each hook id maps to the matching tool (`ruff`, `ruff format`, `mypy`, `pip-audit`) would make the test fail if an `entry` were copy-pasted onto the wrong hook. Not a merge blocker.

The `parents[3]` walk from `lib/crewai/tests/` to the repo root is a bit brittle if this file ever moves, but it's fine for this check.

Nikhi00718:
Rebased onto the latest `main` and kept the newer `pip-audit` ignore entry. I also tightened the regression test to check the expected command for each local hook. The focused test and `pre-commit validate-config` both pass.

Vaishnavi220506:
Hi, just checking in on PR #6881. I rebased it onto the latest `main` and addressed the test suggestion. All checks are passing. Please let me know if anything else is needed. Thanks!

Vaishnavi220506:
Hi @Vidit-Ostwal, the required GitHub Actions runs for this fork PR are expiring while waiting for maintainer approval. CodeRabbit has passed and there are no new code comments; the branch is already rebased onto the latest `main`. Could you please approve the workflow runs and review the PR when convenient? No further code changes are needed from my side. Thanks!

Vaishnavi220506:
Hi @Vidit-Ostwal, I pushed commit a6043630 with a small test-isolation fix. The failed test matrix was caused by test_bedrock_completion_module_is_imported removing the Bedrock completion module from sys.modules without restoring it; it now uses pytest's monkeypatch cleanup. The original Windows pre-commit changes are unchanged. Could you please approve the new fork workflow runs so CI can execute? Thanks!

.pre-commit-config.yaml
```diff
@@ -3,19 +3,19 @@ repos:
     hooks:
       - id: ruff
         name: ruff
-        entry: bash -c 'source .venv/bin/activate && uv run ruff check --config pyproject.toml "$@"' --
+        entry: uv run ruff check --config pyproject.toml
```
Vidit-Ostwal: The bug is the Unix-only path, not `source` itself. Please keep `source .venv/bin/activate` on macOS/Linux and branch on `$OSTYPE` for Windows (`source .venv/Scripts/activate`), then the same `uv run …`.

Retarget `test_pre_commit_config.py` so it requires both activate paths instead of asserting the Unix one is gone. No extra tests needed.

Vidit-Ostwal (changes requested):
Keep Unix `source .venv/bin/activate` and add a Windows `$OSTYPE` branch instead of dropping activation.

.pre-commit-config.yaml
Vaishnavi220506: Updated in 10a0385e. The hooks now keep activation and select `.venv/Scripts/activate` for msys/cygwin/win32, falling back to `.venv/bin/activate` on Unix-like systems. I retargeted the existing regression test to require both paths and verified the Ruff hook through Git Bash on Windows.

Vidit-Ostwal (commented):
The `.pre-commit-config.yaml` change looks good for the reported failure. Local hooks already run under `bash -c`, and on Git Bash (`OSTYPE=msys` / `cygwin`) they now `source .venv/Scripts/activate` instead of the Unix path that does not exist. That should unblock `ruff`, `ruff-format`, `mypy`, and `pip-audit` on Windows (#6863). macOS/Linux still use `.venv/bin/activate`.

`lib/crewai/tests/test_pre_commit_config.py` only asserts that the YAML contains `$OSTYPE` and both activate strings. It does not show the hooks actually run on Windows, and it would fail if someone later dropped `source` and relied on `uv run` (which the issue already suggested). Fine as a string lock, not as behavior coverage.

The Bedrock/OpenAI test edits are not required for this fix. They only tighten `sys.modules` isolation when those tests force a re-import of the native completion module. That is unrelated to venv activation and should not land in this PR.

**Ask:** drop all the test changes and keep only `.pre-commit-config.yaml`. After that this is a focused, acceptable fix for #6863.

Vaishnavi220506:
Hi @Vidit-Ostwal, I synced the branch with the latest `main` and verified the requested `$OSTYPE` activation branch is in place. The review thread is resolved, the PR description now matches the current implementation, and all checks passed on the previous head. The fresh workflows for `57a2c0374` are waiting for maintainer approval. Could you please approve the runs and re-review when convenient? Thanks!

Vidit-Ostwal:
https://github.com/crewAIInc/crewAI/pull/6881#pullrequestreview-5151300172

can you check this

Vaishnavi220506:
Thanks for pointing this out. Updated in `4430b637e`: I removed the pre-commit config test and the unrelated Bedrock/OpenAI test changes. The PR diff now contains only `.pre-commit-config.yaml`. I also reran `pre-commit validate-config`, Ruff, and Ruff Format through Git Bash on Windows; they pass.

Vidit-Ostwal (approved):
Looks good. Local hooks now pick `.venv/Scripts/activate` on Windows shells and keep `.venv/bin/activate` elsewhere, so ruff/mypy/pip-audit can run without a Unix-only venv path. Approving.

PR #6884, opened 2026-08-09
fix: remove hardcoded Unix venv activation from pre-commit hooks

### LINKED ISSUE #6863: [BUG] pre-commit hooks fail on Windows due to hardcoded Unix virtual environment path

shubandaniil-beep:
## Summary
- The local `pre-commit` hooks (`ruff`, `ruff-format`, `mypy`, `pip-audit`) sourced `.venv/bin/activate` directly, which doesn't exist on Windows (activation script lives under `.venv\Scripts\` there), so all hooks failed before running.
- `uv run` already resolves the project's virtual environment on its own, so the explicit `source .venv/bin/activate &&` step is unnecessary and platform-specific. Removed it from all four hook entries.

Fixes #6863

## Test plan
- [x] Ran `uv run --with pre-commit pre-commit run ruff --all-files` locally — hook executes and reports lint results instead of failing with `.venv/bin/activate: No such file or directory`
- [x] Validated `.pre-commit-config.yaml` still parses as valid YAML

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,19 +3,19 @@ repos:
     hooks:
       - id: ruff
         name: ruff
-        entry: bash -c 'source .venv/bin/activate && uv run ruff check --config pyproject.toml "$@"' --
+        entry: bash -c 'uv run ruff check --config pyproject.toml "$@"' --
         language: system
         pass_filenames: true
         types: [python]
       - id: ruff-format
         name: ruff-format
-        entry: bash -c 'source .venv/bin/activate && uv run ruff format --config pyproject.toml "$@"' --
+        entry: bash -c 'uv run ruff format --config pyproject.toml "$@"' --
         language: system
         pass_filenames: true
         types: [python]
       - id: mypy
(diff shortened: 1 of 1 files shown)


Vidit-Ostwal (commented):
Closing as a duplicate of #6881, which already removes the Unix venv activation, drops the `bash -c` wrapper so hooks work without Git Bash, and adds a regression test. This PR still requires `bash -c`, so it would not fully fix Windows. Follow along on #6881.

PR #6953, opened 2026-08-10
fix: emit FlowStartedEvent when a boundary hook aborts the flow

lucasgomide:
A policy deny at the `execution_start` boundary aborted `Flow.kickoff_async` before `FlowStartedEvent` was emitted, so a denied execution left logs but no record — unlike crews, which report the start first. Instead of reordering the lifecycle, the abort handler now stamps the state id, emits the started event, and re-raises: a deny surfaces as a started → failed execution, while normal runs keep the existing ordering — the started event carries hook-resolved inputs and `id` rewrites keep redirecting persistence restoration. The docs spell out the flow ordering and steer rewriting to `INPUT`.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 5f38ead6f596720dce9398a13312a8bd96d8a1f9. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/crewai/src/crewai/flow/runtime/__init__.py
```diff
@@ -2309,6 +2279,41 @@ async def kickoff_async(
...
+            # started -> failed rather than an execution that never existed.
+            # ``inputs`` aliases the same object as ``payload`` (not a fresh
+            # ``{}`` from ``or``) so in-place edits survive read-back.
+            start_ctx = ExecutionStartContext(
+                flow=self,
+                inputs=inputs if inputs is not None else {},
+                payload=inputs,
+            )
+            dispatch(InterceptionPoint.EXECUTION_START, start_ctx)
+            execution_start_dispatched = True
```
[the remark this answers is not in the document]
lucasgomide: This is pre-existing deferred-session design, not something this PR changes — the diff only reorders the hook dispatches and never touches `_emit_flow_failed`. Per-turn terminal events are suppressed symmetrically for deferred sessions: the success path skips `FlowFinishedEvent` under the same `_should_defer_trace_finalization()` guard, and any method failure in a deferred turn already takes this suppression path on main. Terminal emission happens once, at `finalize_session_traces()`.

Keeping `_deferred_flow_started_event_id` set is intentional and load-bearing. Clearing it as proposed would introduce two bugs: `finalize_session_traces()` only emits the session-end `FlowFinishedEvent` when the stash is set, so the already-emitted `FlowStartedEvent` would never be closed; and the next turn after a deny would re-emit a second `FlowStartedEvent` for the same session, since `should_emit_flow_started` keys off the stash.

I pinned this lifecycle as a conformance test in b97f0fd (`test_deferred_session_deny_at_execution_start_keeps_session_open`): after a first-turn deny the session recovers on the next turn without a duplicate started event, and `finalize_session_traces()` emits exactly one `FlowFinishedEvent` paired with the original opener. Emitting a per-turn `FlowFailedEvent` for deferred sessions would change failure semantics for all deferred failures, not just hook aborts, so it's out of scope here.

docs/edge/ar/learn/execution-boundary-hooks.mdx
```diff
@@ -0,0 +1,203 @@
...
+
+clear_all_hooks()  # Clears every point, including boundaries
+```
+
+## وثائق ذات صلة
+
+- [نظرة عامة على خطافات التنفيذ →](/edge/ar/learn/execution-hooks)
+- [خطافات الخطوات →](/edge/en/learn/step-hooks)
+- [خطافات استدعاء LLM →](/edge/ar/learn/llm-hooks)
+- [خطافات استدعاء الأدوات →](/edge/ar/learn/tool-hooks)
```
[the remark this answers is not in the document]
lucasgomide: Fixed in 1f0264c3b by pointing the `HookAborted` links to the English section (`/edge/en/learn/execution-hooks#aborting-an-operation`), the same fallback these pages already use for the untranslated step-hooks page.

The localized `execution-hooks.mdx` pages predate the English rewrite that introduced the "Aborting an operation" section — they don't mention `HookAborted` at all — so adding the section there means re-translating that page, which is a broader localization task out of scope for this PR.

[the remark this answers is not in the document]
lucasgomide: Revisiting this: linking to the English section kicked locale readers out of their language, so 746421969 takes a different approach. `HookAborted` is now plain inline code in the localized pages (the abort contract is fully explained in place), and the step-hooks entry is dropped from their related links — matching the locale-navigation convention of omitting untranslated targets rather than pointing at English. No cross-language or dead-anchor links remain. Restoring the deep links belongs with a refresh of the stale `execution-hooks` translations, tracked separately.

Vidit-Ostwal:
@lucasgomide, I was trying to understand more about the diff using cursor.
got these 2 concerns while digging deeper.

If you were using the hook to hide a secret, it stops hiding it.
Say you pass an API key into kickoff() and your hook swaps it for [REDACTED] so it won't be stored anywhere. Before, the dashboard showed api_key: [REDACTED] for that run. After, it shows api_key: sk-live-abc123. The real key, sitting in your trace history on the server. Your flow still runs with the redacted value — nothing breaks, you get no warning — the secret just quietly appears in a place you thought you'd cleaned.

If your hook was pointing runs at saved state by ID, it stops working silently.
A hook that set or changed inputs["id"] to resume a previous run's state used to work. Now that ID is read before your hook runs, so your change is ignored and the flow starts fresh instead of resuming. No error message — you'd only notice because the run behaves like a brand new one.

lucasgomide:
Thanks for digging into this.. your first concern is a genuine regression, and I verified it empirically. On `main`, a boundary hook that rewrites `api_key` to `[REDACTED]` means `FlowStartedEvent.inputs` carries the redacted dict; on this branch the event now snapshots the pre-hook inputs. And since `flow_started` is serialized wholesale into the trace batch, the raw value would persist in trace history exactly as you described. That's not acceptable.

The second one is the trade-off we documented (flow identity pinned before the hooks), but your framing is fair — it is a silent behavior change, and the rework below removes it as well.

instead of hoisting `FlowStartedEvent` above the hooks, keep the original ordering... hooks run first, the started event carries post-hook inputs, and `id` rewrites keep redirecting restoration and emit `FlowStartedEvent` just-in-time when a boundary hook aborts

lib/crewai/src/crewai/flow/runtime/__init__.py
```diff
@@ -2144,120 +2144,131 @@ async def kickoff_async(
...
+                inputs = start_ctx.payload
+
+                input_ctx = InputContext(
+                    flow=self,
+                    inputs=inputs if inputs is not None else {},
+                    payload=inputs,
+                )
+                dispatch(InterceptionPoint.INPUT, input_ctx)
+                inputs = input_ctx.payload
+            except HookAborted as aborted:
```
lorenzejay: Because these assignments only run after `dispatch()` returns, a later `HookAborted` drops a replacement returned by an earlier hook, while an in-place mutation survives. I reproduced this with a rewrite hook followed by an abort: `FlowStartedEvent` received the original ID. Please preserve the active context's payload on the abort path and add a chained-hooks regression. Handling the abort locally after extracting the flow-start emission would also avoid carrying `boundary_abort` through the rest of this function.

docs/edge/ar/learn/execution-boundary-hooks.mdx
```diff
@@ -0,0 +1,201 @@
...
+لتشغيل Crew يكون ترتيب الحدود:
+
+```
+EXECUTION_START → before_kickoff callbacks → INPUT → tasks execute → OUTPUT → EXECUTION_END
+```
+
+لتشغيل Flow، تُؤطر أحداث دورة الحياة خطافات الحدود:
+
+```
+FlowStartedEvent → EXECUTION_START → INPUT → flow methods execute → OUTPUT → EXECUTION_END → FlowFinishedEvent
```
lorenzejay: This ordering, also present in the Korean and Portuguese translations, contradicts the English page and implementation: hooks run before `FlowStartedEvent`, and an ID rewrite redirects persistence restoration. Please resync all three localized versions.

lorenzejay (commented):
Left two inline comments on the abort control flow and localized docs.

lib/crewai/src/crewai/flow/runtime/__init__.py
lucasgomide: Tks! fixed in eda7fdd33.

docs/edge/ar/learn/execution-boundary-hooks.mdx
lucasgomide: fixed in a30407299

lorenzejay (approved):
looks good to me

PR #6954, opened 2026-08-10
fix(core): scope span export to our own tracer provider

joaomdmoura:
- Stops CrewAI installing its `TracerProvider` as the global one.
- Assistant/runtime detection and the shared span-attribute processor move into `crewai_core` so both packages use one implementation. 
- CLI-only processes now report the same process context they previously could not.
- Tests cover provider isolation, the re-export shims, and the CLI attributes.
- Keeps this intentionally small: no new attributes, no span-parenting change, no pipeline or dashboard work.

---

> [!NOTE]
> **Medium Risk**
> Changes how OpenTelemetry is wired process-wide; wrong behavior could drop CrewAI telemetry or still cross-export spans, but behavior is heavily regression-tested.
> 
> **Overview**
> **Stops CrewAI from registering a global OpenTelemetry `TracerProvider`.** Spans are created from a private provider via `self.provider.get_tracer(TRACER_NAME)`, so third-party instrumented libraries no longer export to CrewAI’s collector and CrewAI spans no longer leak into the user’s observability backend (Langfuse, Phoenix, etc.). `set_tracer()` is kept as an idempotent “telemetry live” flag only.
> 
> **Centralizes process context in `crewai_core`.** Assistant/runtime detection moves to `crewai_core.runtime_env`; `CommonAttributesSpanProcessor` and cached `common_span_attributes()` (`coding_agent`, `runtime_context`, optional `project_id`) live in `crewai_core.telemetry` so **CLI-only** processes get the same attributes without importing `crewai`. Existing import paths in `crewai.telemetry.utils` and `crewai.utilities.constants` re-export the moved symbols.
> 
> **Docs** add a short “isolation from your own OpenTelemetry setup” section (en/ar/ko/pt-BR). **Tests** cover global provider untouched, third-party spans not exported, CrewAI spans still exported when the app has its own provider, re-export shims, and CLI span attributes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c9952096431abad31c4c357a24b453d2ce9b8627. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joaomdmoura:
Both findings fixed in 5122effa4.

**Marker precedence docstring** — valid, and it predates this PR: the same sentence is on `main` at `lib/crewai/src/crewai/telemetry/utils.py:63`, left behind when the table went from 3 assistants to 9. Verified the real order is `claude_code, codex, cline, gemini_cli, augment, opencode, antigravity, junie, cursor`. Corrected it to state the ordering rule rather than dropping the enumeration, since *why* Cursor is last is the part worth keeping: `CURSOR_*` is set for every integrated terminal, so checking it earlier would mask an assistant running inside one.

**OTLP export and provider shutdown** — valid, and there was already evidence I had missed. `Telemetry.__init__` wires a `BatchSpanProcessor` around the real `SafeOTLPSpanExporter`, so every test in that module attempted a live export; a full-suite run showed `ERROR crewai.telemetry.telemetry:telemetry.py:90 Network is disabled` appearing only on this branch. Now patched before construction with a null exporter, and `provider.shutdown()` runs in a `finally` so the batch worker thread does not outlive the test. Applied the same fix to `lib/crewai-core/tests/test_runtime_env.py`, which had the identical problem.

Confirmed the isolation tests still discriminate after the fixture change: restoring the global-provider behaviour makes 3 of the 4 fail again.

joaomdmoura:
Both code-quality findings fixed in 451a24bbf — same root cause in each file: I had imported the telemetry module twice, once aliased as the `monkeypatch` target and once via `from ... import` for the names.

Took a different fix than suggested. Rather than assigning module-level aliases (`Telemetry = telemetry_module.Telemetry`), I dropped the module import entirely and used `monkeypatch`'s dotted-string target:

```python
monkeypatch.setattr(
    "crewai.telemetry.telemetry.SafeOTLPSpanExporter",
    lambda **_kwargs: _NullExporter(),
)
```

That leaves a single import form per module and keeps every reference unqualified, so no call sites change. Applied to `lib/crewai-core/tests/test_runtime_env.py` and `lib/crewai/tests/telemetry/test_tracer_isolation.py`.

Verified after the change: `ruff check`/`format` clean, mypy clean on every touched file, 205 passed across three random-order runs, and zero blocked-network errors logged.

lorenzejay (approved):

lorenzejay (approved):

PR #6959, opened 2026-08-11
fix(llm): infer provider from naming patterns and sync anthropic models

### LINKED ISSUE #6813: ANTHROPIC_MODELS has drifted from the CLI model list, so models offered by crewai create route to the OpenAI client

poorvith-mp:
### Summary

Fixes #6813.

### Problem
1. In `LLM._infer_provider_from_model`, model provider resolution only checked hardcoded constant lists and defaulted directly to `"openai"` without calling `_matches_provider_pattern()`. This caused models like `claude-sonnet-4-6` or custom Anthropic/Gemini models to incorrectly route to the OpenAI native client.
2. `ANTHROPIC_MODELS` and `AnthropicModels` literal in `lib/crewai/src/crewai/llms/constants.py` had drifted from `lib/cli/src/crewai_cli/constants.py`, missing `claude-opus-4-6`, `claude-sonnet-4-6`, `claude-opus-4-7`, `claude-opus-4-8`, `claude-opus-5`, `claude-sonnet-5`, `claude-fable-5`, and contained two invalid IDs (`claude-4-sonnet-20250514`, `claude-4-opus-20250514`).

### Solution
- Updated `LLM._infer_provider_from_model` to fall through to `_matches_provider_pattern` for known providers before defaulting to `"openai"`, as documented in its docstring.
- Synchronized `ANTHROPIC_MODELS` with the CLI model list and updated the `AnthropicModels` type alias.
- Added comprehensive unit tests in `test_llm.py` covering provider inference for Anthropic, Gemini, OpenAI, and custom prefix models.

Closes #6813


Vidit-Ostwal (commented):
Closing this as a duplicate of #6823.

The pattern fallback there already fixes the silent OpenAI routing for bare `claude-*` ids from `crewai create`. This PR also iterates azure/bedrock, so unlisted `gpt-*` names can match Azure first. The extra Anthropic ids are real, but they are not required once pattern matching works. Please follow along on #6823.

PR #6961, opened 2026-08-11
feat(flow): report flow outcome, duration and human-in-the-loop signals

joaomdmoura:
- Reports what happens to a flow, not just that one started. A flow previously emitted only `Flow Execution` at kickoff; `FlowFinishedEvent`, `FlowFailedEvent`, `MethodExecutionFailedEvent`, `MethodExecutionPausedEvent` and `FlowPausedEvent` reached the console formatter and stopped there, and the input and conversation-failure events had no listener at all. Adds `flow:completed`, `flow:failed`, `flow:method_failed`, `flow:paused`, `flow:hitl_paused`, `flow:resumed`, `flow:input_requested`, `flow:input_received` and `flow:conversation_turn_failed` through the existing feature-usage aggregation, so no new pipeline is needed to read them.
- Adds a `Flow Completed` span carrying `flow_name`, `duration_ms`, `outcome` and `origin`, emitted on finish and on failure. Elapsed time comes from a monotonic stamp taken at flow start and cleared on use. It is a separate span rather than `Flow Execution` held open to completion: that span is emitted and closed at start and the daily aggregate counts it, so holding it would drop every killed or crashed run from the execution count. A killed run now simply has no `Flow Completed` row. Duration is an explicit attribute rather than the span's own duration, which the ingestion pipeline stores as a suffixed string that aggregation cannot parse.
- Adds `origin` (`internal` / `user`) to `Flow Execution` and `Flow Completed`. CrewAI's own `AgentExecutor` is itself a Flow and runs once per agent execution, so it dominates flow counts and was previously indistinguishable from flows a user wrote. Both span methods take `origin` with a default, so their signatures stay backward compatible.
- Preserves every existing count and attribute. `Flow Execution` is untouched, so `flow_executions_daily_target` counts exactly what it counted before. A span-level capture of everything reaching the exporter shows the change is purely additive: nothing lost, `Flow Execution` identical on both sides.
- Covered by 14 tests driving real `Flow` executions through the event bus — completion, failure, method failure, input request/receive, a real HITL pause, a real pause-and-resume cycle against SQLite persistence, `origin` for both a user flow and a live `AgentExecutor`, and that no method name, flow-state or error text reaches telemetry. Docs updated in `docs/edge/en` and synced to `ar`, `ko` and `pt-BR`.
- Keeps this intentionally small: no crew or task duration, no change to `Flow Execution` or its start-time semantics, no span parenting, no warehouse or dashboard work.
- Next: a materialized view over `Flow Completed`, and `origin` added to the sort key of `flow_executions_daily_target` so the existing count can be split internal vs user.

---

> [!NOTE]
> **Medium Risk**
> Expands default-on telemetry for flow lifecycle and HITL signals, which affects collected data and metric semantics. Privacy-sensitive fields are intentionally excluded, and existing Flow Execution counts are preserved.
> 
> **Overview**
> Previously flows only reported that they **started**. This wires finish, fail, pause, method-failure, resume, input, and conversation-turn signals into telemetry so outcomes and HITL usage are measurable.
> 
> Adds a separate **`Flow Completed`** span with `duration_ms`, `outcome`, and `origin`, plus **`Flow Paused`** / **`Flow Method Failed`**. `Flow Execution` stays start-only (so daily counts stay intact) and gains `origin` (`internal`/`user`) and a string `resumed` flag. Internal flows (`AgentExecutor`, memory encode/recall) are marked via `is_crewai_internal` so they no longer swamp user metrics.
> 
> Feature-usage spans cover input request/receive, HITL pause, and conversation-turn failure. Method names, error text, and flow state are never recorded. Docs updated across locales; covered by real Flow execution tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5849f48ecef6364fceb615255cf2eb0fe7572bed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lorenzejay (approved):

joaomdmoura:
Both Bugbot findings confirmed against the code and fixed in `e0d0bde9a`. Branch also rebased onto `main` (now includes #6953, which changes when `FlowStartedEvent` fires — my hook point — and #6850).

**Internal flows pollute outcome signals — correct, and worse than described.** It isn't only `AgentExecutor`: `EncodingFlow` (`memory/encoding_flow.py:75`) and `RecallFlow` (`memory/recall_flow.py:58`) are also Flows, and all three set `suppress_flow_events=True`. Memory encode/recall run on most executions, so `flow:completed` would have been mostly bookkeeping.

`flow:completed`, `flow:failed` and `flow:method_failed` are now emitted only for flows the caller wrote, gated on `suppress_flow_events` — the codebase's own marker for infrastructure flows, rather than my module-based `origin`, since it is the authored intent. Internal outcomes are **not** lost: they are still on the `Flow Completed` span, which carries `origin`. Verified on a real crew run — `flow:completed` went from 2 to 1 (the user flow), while `Flow Completed` still shows `origin=internal` for `AgentExecutor`.

The HITL and input signals were left ungated deliberately: infrastructure flows do not pause for humans, so there is nothing to filter.

**Resume signal broader than HITL — correct.** `_is_execution_resuming` is set at two sites: `from_pending` (`runtime/__init__.py:1234`, a human pause) and a checkpoint restore (`:2209`, which never paused for anyone). Counting the latter would push resumes above pauses and make `flow:paused - flow:resumed` unusable as an abandonment rate.

Now keyed off `_pending_feedback_context`, which only `from_pending` sets. Confirmed by probing both flags at `FlowStartedEvent` during a real pause/resume cycle: fresh start `resuming=False pending_ctx=False`, HITL resume `resuming=True pending_ctx=True`.

Two tests added for these specifically, bringing the file to 16. `ruff`/`format`/`mypy` clean, 221 passed across three random-order runs.

joaomdmoura:
All three fixed in `1ca4762a8`. Two of them invalidated a fix I made earlier in this PR, so thanks — both were right.

**Gating on `suppress_flow_events` (CodeRabbit) — correct.** That flag asks for console quiet and is a public field (`runtime/__init__.py:557`), so a caller who set it on their own flow would have silently lost `flow:completed`, `flow:failed` and `flow:method_failed`. I had even written this risk down and shipped it anyway.

**Declarative flows tagged internal (Bugbot) — correct, and it made the above fix worse.** Verified: `Flow.from_declaration()` returns a `Flow` typed in `crewai.flow.flow`, so the module check reported a caller's declarative flow as one of CrewAI's own. Gating on `origin` as suggested would have inherited that bug.

Both had the same root cause — the discriminator was **inferred**. `Flow` now declares `is_crewai_internal`, set on `AgentExecutor`, `EncodingFlow` and `RecallFlow`, and one helper serves both the `origin` attribute and the outcome gate. Verified: declarative → `user`; all three internal → `internal`.

**Failed chat reported completed (Bugbot) — correct.** `finalize_session_traces()` emits `FlowFinishedEvent` regardless of outcome, so a failed session produced `flow:conversation_turn_failed` and `flow:completed` together. The turn failure is now recorded on the flow and read back when the session closes, so the outcome is `failed`.

Three tests added for exactly these: a user flow with `suppress_flow_events=True` still reports; a declarative flow is not internal; a failed conversational session is never reported completed. 19 tests in the file.

`ruff`/`format`/`mypy` clean. Full suite: 385 failed / 4801 passed against main's 385 / 4732 — identical failure set, +69 passed.

joaomdmoura:
Both addressed.

**Stale `flow_execution_span` assertion** — correct, and already fixed in `1d3a41a6f` before this comment landed. Adding the `resumed` argument changed a signature that `test_events.py` asserts on exactly, and I had not re-run that caller before pushing. CI is green on it now.

**Resume guard test no longer checks** — correct, and the more serious of the two. `assert "flow:resumed" not in features` could no longer fail once that signal moved onto the span, so the test passed unconditionally and guarded nothing.

Fixed in `9f097e17e`, and it turned up a second problem in the same test: the manual `crewai_event_bus.emit(...)` dispatches asynchronously, so the assertion was also reading its result before the listener had run. It now waits via `wait_for_event_handlers()` and asserts the `resumed` attribute.

Verified it actually discriminates rather than assuming: keying `resumed` back off `_is_execution_resuming` fails it with

```
AssertionError: assert [('RestoredFlow', True)] == [('RestoredFlow', False)]
```

`ruff`/`format`/`mypy` clean, 270 passed across three random-order runs over the telemetry suites plus every caller of the changed signatures.

lorenzejay (approved):

lorenzejay (approved):

PR #6962, opened 2026-08-12
feat: bump versions to 1.15.15

joaomdmoura:
> [!NOTE]
> **Medium Risk**
> Expands default telemetry collection for flows, a privacy-sensitive area, though method names, errors, and flow state are explicitly excluded. Changes event-listener wiring that runs on every flow execution.
> 
> **Overview**
> Previously, flows only reported that they *started*. Finish, fail, pause, resume, method-failure, and input events reached the console and stopped there. This wires those lifecycle signals into telemetry and tags each flow as **internal** (agent executor, memory encode/recall) or **user**-authored so internal bookkeeping no longer swamps daily counts.
> 
> **New spans and features**
> - `Flow Completed` with `duration_ms` and `outcome` (completed/failed), emitted separately from start so aborted runs still count
> - `Flow Paused` and `Flow Method Failed` (flow name only — no method names or error text)
> - Feature usage for HITL pause, input requested/received, and conversation-turn failure
> - Start spans now carry `origin` and a string `resumed` flag (keyed off pending feedback, not checkpoint restore)
> 
> Marks `AgentExecutor`, `EncodingFlow`, and `RecallFlow` with `is_crewai_internal`. Documents the new defaulted **Flow Lifecycle Signals** row in telemetry docs (EN/AR/KO/PT-BR). Bumps all packages to **1.15.15**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77c68bd8b2e2fd3b5d2b894908cb76c3fd3fee12. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #6965, opened 2026-08-12
fix(events): stop a failed turn from marking the next one failed

joaomdmoura:
- Follows up the three resolved review threads on #6962, all of which were still live on `main`.
- Stops a failed conversational turn from marking the *next* turn failed. A session that opts out of deferred finalization (`ConversationConfig(defer_trace_finalization=False)`) ends each turn with its own `FlowFailedEvent`, emitted inside `kickoff()` **before** `handle_turn()` emits `ConversationTurnFailedEvent` — so the turn-failure flag was set after the run that owned it had already cleared it, survived on the instance, and made the next healthy turn report `outcome="failed"`. Reproduced on `77c68bd` as `['failed', 'failed']` for a fail-then-succeed pair; now `['failed', 'completed']`.
- Covers both emitters of `ConversationTurnFailedEvent` — `handle_turn()` and `stream_turn()` — since the listener they share is the changed code. Both regression tests fail on `77c68bd` with `['failed', 'failed']` and pass here; the deferred-session test passes either way, which is the point of it.
- Fixes `test_requesting_input_reports_both_sides`, which built `InputResponse(value=...)` — the dataclass takes `text`/`metadata`, so it raised `TypeError`, `ask()` swallowed it and returned `None`, and the test asserted both its signals while actually exercising the provider-error path. Now asserts the answer comes back too, so a raising provider can't pass it.
- Brings the flow lifecycle row in `docs/edge/{en,ar,ko,pt-BR}/telemetry.mdx` in line with the wording convention every other row in that table already follows; all four languages stay at parity.
- Unchanged: deferred sessions are the default for conversational flows and still report a failed turn at session end — pinned by a new test, since the flag is the only thing carrying that outcome when there is no per-turn terminal event. No span names, attributes, or public imports change.
- Tests: `lib/crewai/tests/telemetry/` + `tests/utilities/test_events.py` (258 passed, 1 skipped, stable across 3 random-order seeds), `test_flow_conversation.py` + `hooks/test_interception_conformance.py` (74 passed, 21 skipped), `lib/cli/tests/test_crew_run_tui.py` (53 passed). `ruff check lib/` and `ruff format --check lib/` clean. `mypy lib/` reports 560 errors in 88 files; a clean `origin/main` worktree reports the same 560/88 summary, so this adds none — I compared the summary counts, not the full error lists.
- Verified on Python 3.13.5 only; CI covers 3.10–3.13. No network, no new fixtures or cassettes.
- Confirmed end to end against the live collector and ClickHouse, not just at the exporter boundary: spans emitted from this branch were read back out of `crewai-telemetry` by `traceId`. A session whose turn failed records `Flow Completed outcome="failed"` alongside `Flow Method Failed`, and a subsequent healthy session on a fresh instance records `outcome="completed"` — the failure does not carry over. Ordinary flows are unaffected (`completed` / `failed` with real `duration_ms`), and every row carries `crewai_version=1.15.15`, `origin="user"` and `resumed="false"`. 17 spans, all present.
- Marks conversational turns so a completion rate stays meaningful. A conversational flow kicks off once per turn (`conversational_mixin.py:326`) but reports a single completion for the session at `finalize_session_traces()`, so its `Flow Execution` and `Flow Completed` spans run many-to-one. Mixed in with ordinary flows they drag any rate computed across all of them, and nothing distinguished the two shapes. Both spans now carry `conversational`, read from the flow's own `_is_conversational_enabled()` — which covers the `conversational = True` class attribute and a conversational definition alike — recorded as a string for the same reason as `resumed`. Confirmed in ClickHouse: a conversational turn reads `conversational="true"`, an ordinary flow `"false"`.
- **Rejects** the CodeRabbit `_emit_flow_failed()` proposal from #6961 rather than deferring it. Emitting `FlowFailedEvent` while finalization is deferred would end a session on a failed turn, and in a multi-turn session a failed turn is not a failed session — which is precisely what deferral exists to prevent. The underlying observation is real: a session that is never finalized emits a start with no terminal span. That is a caller-protocol issue (`chat()` finalizes in a `finally`; direct `handle_turn()` users must call `finalize_session_traces()`, as the conversational-flows guide documents), not something telemetry should paper over by changing when lifecycle events fire.
- Keeps this intentionally small: no change to span names or attributes; no edit to the frozen `docs/v1.15.15/` snapshot; no version bump; and the CodeRabbit `StubLLM`/`_run_crew` extraction nitpick is **rejected** — the two tests read better standalone than behind a helper that hides the crew setup they are asserting on.

---

> [!NOTE]
> **Medium Risk**
> Touches flow telemetry outcome reporting for conversational sessions, which can affect aggregated success/failure metrics. Does not change auth, data collection of user content, or public APIs.
> 
> **Overview**
> Fixes a telemetry bug where a failed conversational turn could mark the **next** healthy turn as `failed`. When `defer_trace_finalization=False`, `FlowFailedEvent` clears the run before `ConversationTurnFailedEvent` fires, so the failure flag was set too late and leaked onto the next turn. The listener now only sets `_telemetry_turn_failed` while a start stamp is still present.
> 
> Also adds a **`conversational`** attribute on Flow Execution / Flow Completed spans so multi-turn kickoffs can be filtered out of completion-rate metrics. Aligns the flow-lifecycle telemetry docs (en/ar/ko/pt-BR) with the existing "should not include personal info" wording, and fixes a broken `InputResponse` constructor in a related test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db24cac35ceefad51b165f7d4b90eb0149b4a01c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joaomdmoura:
Verified this branch end to end — emitted from a local run on `bdf2908e3` to the live collector, then read back out of `crewai-telemetry` by `traceId`. Previous rounds only checked what *would* be exported; this checks what actually landed.

**The fix, in the warehouse**

| flow | operationName | outcome | duration_ms |
|------|---------------|---------|-------------|
| `…ConvFixS1` — turn failed | `Flow Completed` | **failed** | 15.13 |
| `…ConvFixS1` | `Flow Method Failed` | | |
| `…ConvFixS2` — healthy | `Flow Completed` | **completed** | 8.50 |

The failure is recorded, and it does not carry into the next session.

**Ordinary emission is unaffected**

| flow | spans | outcome |
|------|-------|---------|
| `…FollowupOk` | Creation, Execution, Completed | `completed`, 6.59 ms |
| `…FollowupFail` | Creation, Execution, Method Failed, Completed | `failed`, 6.64 ms |

17 spans across both probes, all present, all `crewai_version=1.15.15`, `origin="user"`, `resumed="false"`, `duration_ms` readable from `vFloat64`.

**Scope of what this proves.** Cross-*session* isolation, not cross-*turn*: without an LLM configured, `handle_turn()` short-circuits with a canned reply and never re-runs the flow, so a second turn on the same instance was not exercised here. The `['failed', 'failed']` → `['failed', 'completed']` regression tests in this PR cover that path; this is warehouse-level confirmation on top of them, not a substitute.

**One thing the probe surfaced.** My first run drove a conversational turn without calling `finalize_session_traces()` and produced `Flow Creation`, `Flow Execution` and `Flow Method Failed` with **no terminal span** — a start with no end. Not a regression from this branch, and it disappears once the session is finalized, but it means an unfinalized session is counted as executed and never as completed or failed. Added to the PR body as a deferred item; it wants its own ticket.

Cleanup note: these probes wrote ~17 rows into production telemetry under `ClaudeFollowup*` / `ClaudeConvFix*` flow names.

joaomdmoura:
Added `db24cac35` for the conversational gap — but **not** the change that thread asked for, and the reason matters.

**Why not `_emit_flow_failed()`.** Emitting `FlowFailedEvent` while finalization is deferred would end a session on a failed turn. In a multi-turn session a failed turn is not a failed session: turn 1 can fail and turns 2–3 succeed. Firing a terminal event mid-session and then continuing is exactly what deferral exists to prevent, so that proposal is rejected on semantics rather than on scope.

**What the investigation actually turned up.** `defer_trace_finalization` defaults to `True` (`conversational.py:96`, `conversation.py:48`, `conversational_definition.py:40`), so deferral is the norm. Each turn calls `kickoff()` (`conversational_mixin.py:326`), while the session emits one `FlowFinishedEvent` at `finalize_session_traces()`. So conversational `Flow Execution` and `Flow Completed` spans run **many-to-one by design**, and until now nothing in the data distinguished them from ordinary flows. Any naive `Flow Completed / Flow Execution` completion rate is therefore wrong for conversational flows — a measurement bug in what #6961 just shipped, and a different thing from what the thread was pointing at.

Both spans now carry `conversational`, read from the flow's own `_is_conversational_enabled()`, which covers the `conversational = True` class attribute and a conversational definition alike. String-valued, for the same reason as `resumed`.

Confirmed in ClickHouse:

| span | flow | conversational | outcome |
|------|------|----------------|---------|
| `Flow Execution` | `…Chat` | **true** | |
| `Flow Completed` | `…Chat` | **true** | completed |
| `Flow Execution` | `…Plain` | **false** | |
| `Flow Completed` | `…Plain` | **false** | completed |

**On the unfinalized-session observation.** Real, and I over-called it earlier as "the gap". `chat()` finalizes in a `finally`; direct `handle_turn()` callers must call `finalize_session_traces()` themselves, which the conversational-flows guide documents (`conversational-flows.mdx:86, 301`). A caller skipping it is a protocol issue, not something telemetry should paper over by moving lifecycle events.

**What I could not verify.** The N:1 ratio itself. Driving two turns produced only one `Flow Execution`, because without an LLM configured the second `handle_turn()` short-circuits with a canned reply and never kicks off. The ratio is read from the code path, not measured. The marker is verified.

Tests: `flow_execution_span` and `flow_completed_span` gained a defaulted parameter, so every caller assertion and spy was updated and re-run — including `tests/utilities/test_events.py`, which CI caught the last time this signature moved. 281 passed across three random-order seeds; full suite 385 failed / 4812 passed against main's 385 / 4732 — identical failure set, +80. `ruff`, `format --check` and `mypy` clean on every changed file.

Vidit-Ostwal (approved):

PR #6968, opened 2026-08-12
fix(llm): isolate caller file messages

ump45nose:
## Summary

Copy caller-supplied message dictionaries before LiteLLM file formatting can replace content or remove the `files` field.

## Root cause

The LiteLLM sync and async paths passed the caller's message dictionaries directly to their file-processing helpers, which mutate those dictionaries. Reusing the same message list therefore dropped attachments after the first call.

## Validation

- `uv run --with litellm --with anthropic pytest lib/crewai/tests/test_llm.py -k "does_not_mutate_file_messages" -q`
- `uv run ruff format --check lib/crewai/src/crewai/llm.py lib/crewai/tests/test_llm.py`
- `uv run ruff check lib/crewai/src/crewai/llm.py lib/crewai/tests/test_llm.py`

### DIFF
--- a/lib/crewai/src/crewai/llm.py
+++ b/lib/crewai/src/crewai/llm.py
@@ -763,6 +763,8 @@ class LLM(BaseLLM):
         """
         if isinstance(messages, str):
             messages = [{"role": "user", "content": messages}]
+        else:
+            messages = [message.copy() for message in messages]
         if not skip_file_processing:
             messages = self._process_message_files(messages)
         formatted_messages = self._format_messages_for_provider(messages)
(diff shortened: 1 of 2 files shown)


ump45nose:
Rebased onto current `main` (84a1a8cca) with no conflicts. The focused sync/async file-message isolation tests pass locally (2/2) using the repository-declared provider extras.

ump45nose:
@lorenzejay This focused sync/async file-message isolation fix is ready for human review; the regression tests passed 2/2 before the latest conflict-free rebase, and CodeRabbit reports no actionable comments on the current head. The contribution guide requires the `llm-generated` label, but GitHub returns HTTP 403 when I try to add it as an external contributor. Could you add that label and review the change when convenient?

PR #6973, opened 2026-08-12
fix: append trailing user turn in native Gemini provider

### LINKED ISSUE #6984: [BUG] Gemini native provider never appends trailing user turn -> 400 'Requests ending with a model turn are not supported'
### Description

GeminiCompletion._format_messages_for_gemini (the native Google Gen AI provider used for gemini/... and google/... model strings) converts assistant messages to Gemini's model role but never guards against the resulting contents list ending on a model turn. CrewAI's own agent loop can produce exactly that history -- e.g. handle_max_iterations_exceeded appending an assistant message and re-calling the LLM, or a task guardrail/guardrail_max_retries re-invocation after the last LLM turn -- and the malformed history is sent to Gemini's generateContent API verbatim, which rejects it.

Other providers already guard this exact case in LLM._format_messages_for_provider (Mistral and Ollama append a synthetic trailing user message when the last message is assistant), but that method is only used on the LiteLLM fallback path. The native Gemini/Google provider (crewai/llms/providers/gemini/completion.py), which is what actually handles gemini/* and google/* model strings once google-genai is installed, has no equivalent guard.

### Steps to Reproduce

- Build a CrewAI Agent/Task/Crew using LLM(model="gemini/gemini-flash-latest") (or any google/.../gemini/... model, with the google-genai extra installed so the native provider is used).

- Give the task a guardrail with guardrail_max_retries set (or otherwise get the agent loop to hit handle_max_iterations_exceeded), so the executor re-invokes the LLM after an assistant/tool-call turn was already appended to history without an intervening user turn.

- Run the crew against a real Gemini endpoint (or reproduce directly, see snippet below).

Minimal reproduction without needing a live agent run -- call the formatter directly with a history ending on an assistant turn:

```python

llm = GeminiCompletion(model="gemini-flash-latest", api_key="...")
    {"role": "user", "content": "hi"},
    {"role": "assistant", "content": "partial"},
])
print([c.role for c in contents])  # ['user', 'model'] -- ends on 'model'
# Sending these `contents` via self.client.models.generate_content(...) 400s.
```

### Expected behavior

_format_messages_for_gemini should guarantee the returned contents list ends on a user turn (mirroring the Mistral/Ollama guard already present in LLM._format_messages_for_provider), so the agent loop can safely re-invoke the LLM after an assistant/tool-call turn without a 400.

### Screenshots/Code snippets

Actual error surfaced when this happens in a running agent (via the Gemini API, both AI Studio and Vertex enforce the same contents contract):

```
400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'Requests ending with a model turn are not supported.', 'status': 'INVALID_ARGUMENT'}}
```

We hit this specifically with a single-agent Crew whose Task sets guardrail=<completeness_check> and guardrail_max_retries=2 -- the guardrail retry path is what most reliably leaves contents ending on a model turn. It does not reproduce on short, guardrail-free crews (e.g. Google's own CrewAI+Gemini example at https://ai.google.dev/gemini-api/docs/crewai-example uses the identical LLM(model='gemini/...') setup but never hits the re-invoke-after-assistant paths, so the same latent gap doesn't surface there).

### Operating System

### Python Version

3.11

### crewAI Version

1.6.0

### crewAI Tools Version

1.6.0

### Virtual Environment

Venv

### Evidence

Traced the missing guard to crewai/llms/providers/gemini/completion.py::_format_messages_for_gemini (both the crewai==1.6.0 release and current main): it maps role == "assistant" to Gemini's "model" role in two branches (the plain-text branch and the tool-call branch) but never checks whether the final entry in contents ends up being "model". Compare with crewai/llm.py::LLM._format_messages_for_provider, which already has:

```python
# Handle Mistral models - they require the last message to have a role of 'user' or 'tool'
    if messages and messages[-1]["role"] == "assistant":
        return [*messages, {"role": "user", "content": "Please continue."}]
    return messages
```

That guard is never reached for Gemini because gemini/google model strings resolve to the native GeminiCompletion class, not the LiteLLM fallback path.

### Possible Solution

Add an equivalent guard at the end of `GeminiCompletion._format_messages_for_gemini`, appending a synthetic trailing user `types.Content when contents[-1].role == "model"`.

### Additional context

This surfaced running a CrewAI "bring your own" agent under kagent, but the bug is entirely in crewai's own message formatting and reproduces with plain CrewAI + google-genai installed, independent of kagent.

hvlcrs:
## Summary
Fixes #6984.

Gemini's `generateContent` API rejects a request whose history ends on a `model` turn (`400 'Requests ending with a model turn are not supported.'`). CrewAI's own agent loop (`handle_max_iterations_exceeded`, task `guardrail`/`guardrail_max_retries` retries) can produce exactly that history. This covers both places a Gemini call can be dispatched from:

- **Native provider**: `GeminiCompletion._format_messages_for_gemini` maps `assistant` messages to Gemini's `model` role (in both the plain-text and tool-call branches) but never checked whether the resulting `contents` list ends on `model`. Now appends a trailing `user` `types.Content` when it does.
- **LiteLLM fallback**: reached when `google-genai` isn't installed, or the model name isn't recognized as native (custom deployments, new previews, fine-tunes). `LLM._format_messages_for_provider` already guards this exact case for Mistral and Ollama, but had no Gemini branch — and litellm's own Vertex/Gemini transformation doesn't guard it either (checked `litellm/llms/vertex_ai/gemini/transformation.py`). Added a Gemini branch mirroring the existing Mistral one.

## Test plan
- [x] `uv run --package crewai pytest lib/crewai/tests/llms/google/test_google.py -v` — 46 passed, 1 skipped, 1 pre-existing unrelated failure (`test_gemini_raises_error_when_model_not_supported`, fails identically on `main` without this change; requires the `litellm` extra which isn't installed for that test module).
- [x] `uv run --package crewai pytest lib/crewai/tests/test_llm.py` — 66 passed, 3 skipped, no failures.
- [x] Added `test_gemini_message_formatting_appends_user_turn_after_trailing_model_turn` and `test_gemini_message_formatting_leaves_user_terminated_history_unchanged` in `tests/llms/google/test_google.py` (native provider).
- [x] Added `test_gemini_litellm_appends_user_message_when_last_is_assistant` and `test_gemini_litellm_does_not_modify_when_last_is_user` in `tests/test_llm.py` (LiteLLM fallback), mirroring the existing Ollama tests.


Vidit-Ostwal (commented):
Keeping this PR for #6984 — it is the most complete of the four: native formatter, LiteLLM fallback, and the unresolved-function-call case.

Closed #6985, #7039, and #7087 as duplicates.

Please rebase onto current `main` (`mergeStateStatus` is BEHIND). Happy to review once that is in.

hvlcrs:
Hi @Vidit-Ostwal, I've rebased the PR. Let me know if there's any other changes needed for this PR.

Vidit-Ostwal (changes requested):
Type-check is red on 3.10–3.13 and this change introduced it — that's the merge blocker. The rest of the approach is the right fix for #6984.

**Must fix**

`contents[-1].parts` is typed `list[Part] | None`. This new loop is the mypy failure:

```
lib/crewai/src/crewai/llms/providers/gemini/completion.py:679:
error: Item "None" of "list[Part] | None" has no attribute "__iter__"  [union-attr]
```

The same file already narrows with `and contents[-1].parts` before touching `parts`. Guard `None` (or treat missing parts as no `function_call`) so `uv run mypy lib/` is clean.

**Looks good**

- Native trailing-user append covers the reporter's `[user, assistant]` repro and `handle_max_iterations_exceeded`.
- Raising on an unresolved `function_call` is the right extra — a synthetic `"Please continue."` after a pending functionCall is also invalid for Gemini. Tests for that are solid.
- LiteLLM Gemini branch is extra vs the issue text but correct; that path has the same 400 when `google-genai` is not installed.
- Rebase is in. Tests, lint, and CodeRabbit are green. The earlier CodeRabbit notes (unresolved function call + stronger assertions) are addressed.

**Nits (non-blocking)**

- LiteLLM still appends `"Please continue."` when the last assistant message has `tool_calls` (no function-response check). Native raises. Fine as a follow-up — the max-iter path appends a text assistant message, not a bare tool call.
- `_format_messages_for_gemini` docstring does not mention the new `ValueError`.

Once mypy is green this is mergeable.

Vidit-Ostwal (approved):
Looks good. Fixes the Gemini 400 when history ends on a model turn (ReAct retry / max-iter), with the unresolved function_call path correctly failing closed instead of appending text.

LGTM.

PR #6977, opened 2026-08-12
chore(telemetry): remove unused _safe_telemetry_operation from crewai_core

Vidit-Ostwal:
## Summary
- Remove unused `_safe_telemetry_operation` from `lib/crewai-core/src/crewai_core/telemetry.py`.
- In `crewai_core`, this method was only defined and never called. The class uses `_safe_telemetry_procedure` instead.
- There is a same-named method on the older `Telemetry` in `lib/crewai/src/crewai/telemetry/telemetry.py` that is still in use; this PR does not touch that path or any tests.

## Test plan
- [ ] Confirm `_safe_telemetry_operation` has no call sites under `lib/crewai-core`
- [ ] Confirm existing telemetry tests still pass without changes


lorenzejay (approved):

lorenzejay (changes requested):
wait i see this being used in many places

Vidit-Ostwal:
Good catch — that usage is on a different class.

There are two `_safe_telemetry_operation` definitions:

1. **`lib/crewai/src/crewai/telemetry/telemetry.py`** — the older `crewai.telemetry.Telemetry`. This one is still called in many places (and covered by tests). **This PR does not remove it.**
2. **`lib/crewai-core/src/crewai_core/telemetry.py`** — the `crewai_core.Telemetry` helper. This one was only defined and never called; `crewai_core` uses `_safe_telemetry_procedure` instead. **This PR only removes that unused method.**

So the call sites you are seeing are the legacy `crewai` telemetry path, not the `crewai_core` method being deleted here.

lorenzejay (approved):
lgtm

PR #6979, opened 2026-08-12
fix(tools): run async tool results safely under a running event loop

### LINKED ISSUE #6978: BaseTool.run / StructuredTool raise 'asyncio.run() cannot be called from a running event loop' when a tool returns a coroutine
### Description

### Location

```python
```

### Reproduction

```python
import asyncio
from crewai.tools import tool

@tool("echo")
    """Echo the value."""
    await asyncio.sleep(0.01)
    return f"echo: {value}"

    return echo.run(value="hi")

asyncio.run(handler())
```

### Expected

### Environment

Shailendra005:
### What

Tool execution paths that resolve a coroutine result now use the project's established running-loop guard instead of a bare `asyncio.run()`, which raised `RuntimeError: asyncio.run() cannot be called from a running event loop` when a tool was invoked inside a running loop (FastAPI, Jupyter, any `async def`).

### Why

`BaseTool.run`, the `@tool`-decorator `Tool.run`, and `CrewStructuredTool.invoke` all did `asyncio.run(result)` unconditionally. The codebase already has the correct guard (`tasks/llm_guardrail.py::_run_coroutine_sync`, also in `mcp_native_tool.py`, `a2a/utils/*`, `mcp/tool_resolver.py`, `project/*`); the tool paths just didn't use it.

### How

- Added `crewai/utilities/async_utils.py::run_coroutine_sync` — a generic (`TypeVar`) version of the existing guard. It lives in `utilities` because `base_tool` already imports `structured_tool`, so a helper in either module would create a circular import; both now import from the shared util.
- When no loop is running it falls back to `asyncio.run`; when a loop is running it runs the coroutine to completion in a one-worker `ThreadPoolExecutor` with a copied `contextvars` context, so the caller's loop is neither blocked nor re-entered.
- No public API change.

### Tests

New `TestToolRunUnderRunningLoop` in `tests/tools/test_async_tools.py` exercises all three call sites from inside `asyncio.run(...)`. Verified failing before the change (`RuntimeError`) and passing after.

```
BEFORE: 3 failed (RuntimeError: asyncio.run() cannot be called from a running event loop)
AFTER : 15 passed  (test_async_tools.py)
        73 passed  (test_base_tool.py + test_structured_tool.py)
```

### Notes

The other duplicate guard call sites are intentionally left untouched to keep this to one logical change; consolidating them onto `run_coroutine_sync` could be a follow-up.

Fixes #6978


Shailendra005:
Thanks for the review — both points addressed in 81cab34:

1. **Docstring accuracy (`async_utils.py`):** you're right that `Future.result()` blocks the calling thread, so the running loop makes no progress while it waits. Reworded the docstring to say exactly that — it now states the worker-thread execution avoids *re-entering* the running loop, and that the calling thread blocks on `Future.result()` until the coroutine completes (so the loop does not progress meanwhile). The blocking behaviour matches the existing `llm_guardrail._run_coroutine_sync` pattern this generalizes.

2. **Untested branch in `CrewStructuredTool.invoke`:** added `test_structured_tool_invoke_returned_coroutine_inside_running_loop`, which uses a plain (non-`async def`) function that returns a coroutine, exercising the `result = self.func(...)` → `asyncio.iscoroutine(result)` → `run_coroutine_sync(result)` branch at `structured_tool.py:444-447` (the existing test only hit the `iscoroutinefunction` path). Full file now 16 passed; ruff clean.

Vidit-Ostwal (commented):
Please rebase onto `main` — this branch is behind.

CodeRabbit's docstring and extra `CrewStructuredTool.invoke` returned-coroutine test notes look addressed in 81cab34. No further CodeRabbit work from here.

Vidit-Ostwal (commented):
Closing in favor of #7236. That PR already has the shared running-loop helper for `BaseTool.run` / `Tool.run` / `CrewStructuredTool.invoke`, and it also covers `MCPToolWrapper` plus tests. Follow along there — please have #7236 link #6978 as the surviving tracker.

Vidit-Ostwal:
Duplicate of #7236. That PR already does this shared-helper fix and also covers `MCPToolWrapper`. Follow along there.

Vidit-Ostwal (commented):
Reopening — this is the surviving PR for #6978.

Please rebase onto `main` (branch is behind). Also pull in the `MCPToolWrapper` running-loop fix and tests from #7236; that site still does a bare `asyncio.run()` and swallows the error as the tool result.

CodeRabbit's docstring and returned-coroutine `invoke` test notes look already addressed.

Vidit-Ostwal:
I will keep this in radar, but if you get a response in 3 days, feel free to drop me an email vidit.ostwal@crewai.com

parthiban-sivakumar:
Hi @Shailendra005,

Vidit closed my #7236 in favour of this PR and asked me to bring one piece over.

The only site not covered here is `MCPToolWrapper._run` at `mcp_tool_wrapper.py:79`. Same bare `asyncio.run()`, but it fails silently — line 82 catches the error and returns it as the tool's result string, so the agent gets an error message as though it were the tool's output:

```
MCP sync context  -> MCP tool execution failed after 3 attempts: Connection timed out
MCP async context -> Error executing MCP tool echo: asyncio.run() cannot be called from a running event loop
```

It's a one-liner with your `run_coroutine_sync`, plus a test that patches `_run_async` and calls `_run()` inside a running loop. I can't push to your branch — happy to paste the exact diff here, or open a follow-up PR once this merges, whichever you prefer.

Everything else I had, you already covered — including the plain-`def`-returning-a-coroutine test. Sorry for the duplicate work; I should have searched for an existing issue before filing.

Thanks!

Vidit-Ostwal:
@Shailendra005 any update?, otherwise will assign this to @parthiban-sivakumar

Shailendra005:
Hey @Vidit-Ostwal sorry for the inconvenience. I will get it done by today. Thank you.

PR #6983, opened 2026-08-13
fix: make pre-commit hooks cross-platform (Windows)

### LINKED ISSUE #6863: [BUG] pre-commit hooks fail on Windows due to hardcoded Unix virtual environment path

RajeshShrirao:
## Summary

Fixes #6863 — pre-commit hooks (ruff, ruff-format, mypy, pip-audit) failed on Windows before executing because they wrapped every command in:

```bash
bash -c 'source .venv/bin/activate && uv run ...' --
```

`.venv/bin/activate` is Unix-only; Windows virtual environments use `.venv\\Scripts\\activate`.

## Change

Dropped the `bash -c 'source .venv/bin/activate && ...'` wrapper entirely and invoke `uv run` directly, which already activates the project virtual environment on every platform:

- `uv run ruff check --config pyproject.toml`
- `uv run ruff format --config pyproject.toml`
- `uv run mypy --config-file pyproject.toml`
- `uv run pip-audit --skip-editable ...`

## Why this is safe

- `uv run` resolves and activates the project env itself — the explicit `source` was redundant on POSIX too.
- Behavior on Linux/macOS is unchanged (same tools, same configs).
- Hooks now work under cmd/PowerShell on Windows without Git Bash.

## Verification

- YAML parses cleanly; entries render as expected.
- `ruff check` / `ruff format --check` run with the new entry shape.
- CI on the PR will exercise the hooks on all platforms.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,19 +3,19 @@ repos:
     hooks:
       - id: ruff
         name: ruff
-        entry: bash -c 'source .venv/bin/activate && uv run ruff check --config pyproject.toml "$@"' --
+        entry: uv run ruff check --config pyproject.toml
         language: system
         pass_filenames: true
         types: [python]
       - id: ruff-format
         name: ruff-format
-        entry: bash -c 'source .venv/bin/activate && uv run ruff format --config pyproject.toml "$@"' --
+        entry: uv run ruff format --config pyproject.toml
         language: system
         pass_filenames: true
         types: [python]
       - id: mypy
         name: mypy
-        entry: bash -c 'source .venv/bin/activate && uv run mypy --config-file pyproject.toml "$@"' --
+        entry: uv run mypy --config-file pyproject.toml
         language: system
         pass_filenames: true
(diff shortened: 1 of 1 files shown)


Vidit-Ostwal (commented):
Closing as a duplicate of #6881, which already makes the same `.pre-commit-config.yaml` change (`uv run` without `bash -c` / `.venv/bin/activate`) and includes a regression test. Follow along there.

PR #7005, opened 2026-08-16
fix(llms): support BaseTool in direct LLM calls

### LINKED ISSUE #7004: [BUG] LLM().call() doesn't work with Tools
### Description

LLM().call() doesn't work with Tools

### Steps to Reproduce

code

### Expected behavior

I want to use the LLM() class without Agent() for more control over the "messages" (lower level).

### Screenshots/Code snippets

```python
import crewai as c
import crewai.flow as cf
import crewai_tools as ct
import logging
from pathlib import Path

logging.basicConfig(level=logging.DEBUG)

llm = c.LLM(model="ollama/my", base_url="http://127.0.0.1:8080")
print(c.__version__)

ROOT = Path(Path(__file__).parent, "_temp").resolve()
ROOT.mkdir(parents=True, exist_ok=True)
fwriter = ct.FileWriterTool(base_dir=str(ROOT))
freader = ct.FileReadTool(base_dir=str(ROOT))

print(llm.call(
    messages=[{"role":"user","content":"Who are you?"}],
    tools=[fwriter, freader],
))
```

### Operating System

Windows 10

### Python Version

3.12

### crewAI Version

1.15.16

### crewAI Tools Version

1.15.16

### Virtual Environment

Venv

### Evidence

DEBUG:asyncio:Using proactor: IocpProactor
ERROR:root:OpenAI: Error extracting tool info: Tool must be a dictionary
ERROR:root:OpenAI: Tool structure: name='File Writer Tool' description="A tool to write content to a specified file. Accepts filename, content, and optionally a directory path and overwrite flag as input. Writes are confined to the tool's allowed directory; a filename or directory that resolves outside it is rejected." env_vars=[] args_schema=<class 'crewai_tools.tools.file_writer_tool.file_writer_tool.FileWriterToolInput'> result_schema=None description_updated=False cache_function=<function _default_cache_function at 0x000001EB2B7C5F80> result_as_answer=False max_usage_count=None tool_failure_policy=None current_usage_count=0 base_dir='/cw/_temp' encoding='utf-8' tool_type='crewai_tools.tools.file_writer_tool.file_writer_tool.FileWriterTool'
ERROR:root:OpenAI: Error converting tool: Tool must be a dictionary
ERROR:root:OpenAI API call failed: Tool must be a dictionary

### Possible Solution

None

### Additional context

None

varshinimp2402:
## Summary

Fixes direct `LLM.call()` usage with CrewAI `BaseTool` instances.

Previously, passing tools such as `FileWriterTool` or `FileReadTool`
directly through `LLM.call(tools=[...])` failed with:

`ValueError: Tool must be a dictionary`

This change allows `BaseTool` instances to be processed through the
existing provider tool-conversion path using their `name`,
`description`, and `args_schema`.

## Testing

- Added regression coverage for `BaseTool` conversion
- Verified existing dictionary-based tool conversion still works
- Verified `FileWriterTool` and `FileReadTool` with Ollama
- Ollama `/v1/chat/completions` returned HTTP 200 and a valid tool call
- Ruff passed
- mypy passed
- Relevant pytest tests passed

Fixes #7004

### DIFF
--- a/lib/crewai/src/crewai/llms/providers/utils/common.py
+++ b/lib/crewai/src/crewai/llms/providers/utils/common.py
@@ -2,6 +2,7 @@ import logging
 import re
 from typing import Any
 
+from crewai.tools.base_tool import BaseTool
 from crewai.utilities.pydantic_schema_utils import generate_model_description
 from crewai.utilities.string_utils import sanitize_tool_name
 
@@ -41,7 +42,9 @@ def validate_function_name(name: str, provider: str = "LLM") -> str:
     return name
 
 
-def extract_tool_info(tool: dict[str, Any]) -> tuple[str, str, dict[str, Any]]:
+def extract_tool_info(
+    tool: dict[str, Any] | BaseTool,
+) -> tuple[str, str, dict[str, Any]]:
     """Extract tool information from various schema formats.
 
     Handles both OpenAI/standard format and direct format:
@@ -57,11 +60,22 @@ def extract_tool_info(tool: dict[str, Any]) -> tuple[str, str, dict[str, Any]]:
     Raises:
         ValueError: If tool format is invalid
     """
+
+    if isinstance(tool, BaseTool):
+        parameters: dict[str, Any] = {}
+
+        if tool.args_schema is not None:
+            schema_output = generate_model_description(tool.args_schema)
+            parameters = schema_output.get("json_schema", {}).get("schema", {})
+
+        return tool.name, tool.description, parameters
+
     if not isinstance(tool, dict):
         raise ValueError("Tool must be a dictionary")
 
     if "function" in tool:
         function_info = tool["function"]
+
         if not isinstance(function_info, dict):
             raise ValueError("Tool function must be a dictionary")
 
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal (commented):
Keeping this PR for #7004 since it was opened first.

Please rebase onto current `main` (`mergeStateStatus` is BEHIND). The conversion in `extract_tool_info` is the right crash fix, but Anthropic `_convert_tools_for_interference` still does `tool.get(...)` before conversion, so a `BaseTool` will fail there. Gate those dict checks with `isinstance(tool, dict)` (and `tool.get("function", {})` the same way), then add a regression for that path. CodeRabbit's notes on schema assertions are worth picking up too.

`LLM.call()` should not grow into a full Agent-style tool loop; `available_functions` as a list and multi-tool execution in one call are out of scope for this crash.

Vidit-Ostwal (commented):
Follow-up: still BEHIND `main`, and the Anthropic path from the last review is still open.

Please rebase onto current `main`, then:

- In Anthropic `_convert_tools_for_interference`, gate `tool.get(...)` / `"input_schema" in tool` with `isinstance(tool, dict)` so a `BaseTool` reaches `safe_tool_conversion` instead of `AttributeError`. Same for `tool.get("function", {})`.
- Add a regression for that Anthropic path (the coverage in closed #7024 is a good template).
- Pick up CodeRabbit's schema assertions in `test_llm_tool_conversion.py` (`type`, property types, `required`).

Closed #7112 as a duplicate — it only wrapped the LiteLLM facade, which does not fix the native OpenAI/Ollama crash on #7004. `available_functions` as a list and multi-tool execution in one `LLM.call()` remain out of scope here.

Vidit-Ostwal (commented):
The CodeRabbit comments on this PR are still unresolved. Please address those, then ping us here when the PR is ready for another look.

Vidit-Ostwal:
@varshinimp24 any update?
Anything I can help you with?

PR #7012, opened 2026-08-17
[OSS-129] Map GPT-5.6 family to the official 1.05M context window

Vidit-Ostwal:
## Summary
- Map GPT-5.6 Sol, Terra, Luna, and the `gpt-5.6` alias to the official 1,050,000-token window on the LiteLLM, native OpenAI, and Azure paths.
- Use longest-prefix matching on native OpenAI so `gpt-5.6-*` does not inherit the shorter `gpt-5` window.
- Add the family to CLI curated catalogs; offline default stays `gpt-5.5`.

## Test plan
- [ ] Confirm `gpt-5.6`, `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna` resolve to `int(1_050_000 * 0.85)` on LiteLLM, native OpenAI, and Azure.
- [ ] Confirm `openai/gpt-5.6-luna` strips the prefix and uses the same window.
- [ ] Confirm `gpt-5` and `gpt-5.4-mini` keep their existing windows.
- [ ] Confirm CLI catalogs list Sol, Terra, Luna, and the alias, and `--provider openai` still defaults to `openai/gpt-5.5`.


lib/crewai/src/crewai/llm.py
```diff
@@ -170,6 +170,7 @@ def _ensure_litellm() -> bool:
     "gpt-4o": 128000,
     "gpt-4o-mini": 200000,
     "gpt-5.4-mini": 200000,
+    "gpt-5.6": 1050000,  # sol, terra, luna, and the gpt-5.6 alias
```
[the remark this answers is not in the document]
Vidit-Ostwal: This should be fixed now.

lib/crewai/tests/test_llm.py
```diff
@@ -343,6 +343,49 @@ def test_context_window_validation():
     assert "must be between 1024 and 2097152" in str(excinfo.value)
 
 
+@pytest.mark.parametrize(
+    "model",
+    ["gpt-5.6", "gpt-5.6-sol", "gpt-5.6-terra", "gpt-5.6-luna"],
+)
+def test_gpt56_family_uses_official_context_window(model: str) -> None:
+    """GPT-5.6 Sol, Terra, Luna, and the alias share a 1.05M window."""
+    llm = LLM(model=model, is_litellm=True)
```
lorenzejay, line 352: we should no longer be testing with `is_litellm=True` test if we respect context window from the native providers

Vidit-Ostwal: I have added test cases for both LiteLLm and the for native provider in the test_openai.py file

lorenzejay (approved):

PR #7017, opened 2026-08-17
fix(knowledge): record collection_name and embedder on the Knowledge model

tonycoder-hub:
## Summary
`Knowledge.__init__` accepts `collection_name` and `embedder`, uses them to build storage, then drops them. Both model fields stay `None` on every normally constructed instance, including `Crew.create_crew_knowledge` and `Agent.set_knowledge`.

A serialized `Knowledge` therefore carries `collection_name: null`, and rebuilding it points storage at the shared `"knowledge"` collection instead of `"knowledge_crew"` / `"knowledge_<role>"`.

Two lines assign the constructor arguments onto the declared fields, matching how `storage` and `sources` are already assigned.

## Test plan
- [x] New tests fail on current main (`None == 'docs'`) and pass after
- [x] `uv run pytest lib/crewai/tests/knowledge/test_knowledge_init.py -q` → 2 passed

## AI disclosure

### DIFF
--- a/lib/crewai/src/crewai/knowledge/knowledge.py
+++ b/lib/crewai/src/crewai/knowledge/knowledge.py
@@ -117,6 +117,8 @@ class Knowledge(BaseModel):
         **data: object,
     ) -> None:
         super().__init__(**data)
+        self.collection_name = collection_name
+        self.embedder = embedder
         if storage is not None:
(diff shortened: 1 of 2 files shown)


tonycoder-hub:

PR #7018, opened 2026-08-17
fix: reset agent retry budget per task

tonycoder-hub:
## Summary
`Agent._times_executed` is incremented on every transient failure and never cleared. A crew reuses the same Agent across tasks, so retries spent on task 1 reduce the budget for task 2. With the default `max_retry_limit=2`, two recovered failures anywhere in a run leave the agent with zero retries for everything that follows.

The field description says this is the maximum retries for an agent to execute **a task**. `_check_execution_error` now resets the counter when a different task starts failing, and `_finalize_task_execution` clears it when a task completes. The reset cannot go at the top of `execute_task` because retries recurse through that method.

Distinct from #6997 (event scope / hook aborts on the same function) and from #5822/#6048 (tool idempotency on retry).

## Test plan
- [x] New test fails on current main (`Exception: transient failure` on the second task) and passes after
- [x] `uv run pytest lib/crewai/tests/agents/test_agent.py -k max_retry_limit -p no:randomly -n 0` → 2 passed

## AI disclosure

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -19,6 +19,7 @@ from typing import (
     NoReturn,
     cast,
 )
+from uuid import UUID
 import warnings
 
 from pydantic import (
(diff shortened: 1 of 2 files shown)


tonycoder-hub:

PR #7019, opened 2026-08-17
fix: allow unregistering hooks that were registered with a filter

tonycoder-hub:
## Summary
`@before_tool_call(tools=["delete_file"])` (and the other three legacy filtered decorators) register an internal wrapper but return the original function. `unregister_*_hook` then searched for the original function in a queue that only holds wrappers, so it returned `False` and the hook kept firing.

Stash the wrapper on `_registered_hook` the same way `@on(...)` already does, and route the four legacy unregister helpers through `dispatch.unregister`. Distinct from #7017/#7018.

## Test plan
- [x] New `TestUnregisterFilteredHooks` cases fail on current main and pass after
- [x] `uv run pytest lib/crewai/tests/hooks/ lib/crewai/tests/project/ -q` → 237 passed

## AI disclosure

### DIFF
--- a/lib/crewai/src/crewai/hooks/decorators.py
+++ b/lib/crewai/src/crewai/hooks/decorators.py
@@ -70,6 +70,9 @@ def _create_hook_decorator(
 
                 if not is_method:
                     register_function(filtered_hook)
+                    # Remember the actually-registered callable so the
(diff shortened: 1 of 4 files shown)


tonycoder-hub:

PR #7020, opened 2026-08-17
fix: clear Agent.tools_results between tasks

tonycoder-hub:
Agent.tools_results is never cleared between tasks. process_tool_results scans the whole list for result_as_answer=True, so a tool used in one task silently overrides the answer of every later task on that agent. LiteAgent.kickoff already clears it per run.

Reset the list in _prepare_task_execution so both execute_task and aexecute_task are covered. Tests: uv run pytest lib/crewai/tests/agents/test_agent.py -k result_as_answer_does_not_leak -q -> passed.

Distinct from #7018 (retry budget) and #7019.

This change was generated with Claude Opus 5 via a Cursor cloud agent.

### DIFF
--- a/lib/crewai/src/crewai/agent/core.py
+++ b/lib/crewai/src/crewai/agent/core.py
@@ -556,6 +556,9 @@ class Agent(BaseAgent):
         get_env_context()
 
         self.reset_tool_failures()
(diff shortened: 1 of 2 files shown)


tonycoder-hub:

PR #7021, opened 2026-08-17
fix: reset upload-cache LRU order when the cache is cleared

tonycoder-hub:
UploadCache.aclear() reset _provider_keys but left _key_access_order populated. Eviction then popped stale keys that no longer exist, so max_entries stopped being enforced after a clear (cleanup_uploaded_files calls clear()).

Now also clears _key_access_order.

Tests: uv run pytest lib/crewai-files/tests/ -q — 168 passed, 4 skipped.

## AI Usage
This PR was fully AI-generated (Claude Opus 5, Cursor cloud agent).

PR #7022, opened 2026-08-17
fix: resolve file constraints from routing prefix, not model id

tonycoder-hub:
get_constraints_for_provider substring-scanned the whole string, so a provider name inside the model id beat the routing prefix. bedrock/anthropic.claude-* resolved to Anthropic (supports_file_upload=True, 32MB PDF) instead of Bedrock (False, 3.84MB). azure/openai/gpt-4o resolved to OpenAI.

Prefix is matched first (exact or startswith); substring scan stays as fallback for unprefixed ids.

Tests: uv run pytest lib/crewai-files/tests -q — 169 passed, 4 skipped. Distinct from #7015-#7021.

## AI Usage
This change is AI-generated.

### DIFF
--- a/lib/crewai-files/src/crewai_files/processing/constraints.py
+++ b/lib/crewai-files/src/crewai_files/processing/constraints.py
@@ -355,13 +355,18 @@ def get_constraints_for_provider(
 
     provider_lower = provider.lower()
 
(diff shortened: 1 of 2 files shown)


tonycoder-hub:

PR #7023, opened 2026-08-17
feat(security): deterministic ActionBoundary & hash-chained action ledger for agent tools

AAH20:
### Summary
Adds `ActionBoundary`, `ActionGate`, and `ActionLedger` to `crewai.security` to provide deterministic execution boundaries, non-destructive simulation fallbacks, and append-only SHA-256 hash-chained compliance action ledgers for CrewAI agent tool executions.

### Problem Solved
As autonomous agent crews in CrewAI execute mutating tools (databases, APIs, bash commands), treating model confidence as execution authorization introduces acute operational risks and compliance blockers (failing SOC 2 Type II, ISO 42001, and NIST AI RMF).

This contribution enforces:
1. **Never Equate Intent to Approval**: Model confidence does not authorize high-blast operations.
2. **Deterministic Tool Classification**: Automatically categorizes tools into `READ`, `WRITE_IDEMPOTENT`, `WRITE_MUTATING`, and `DESTRUCTIVE`.
3. **Simulation Fallback**: Unapproved writes execute in safe simulation mode (`simulation_mode: True`) rather than failing or performing unauthorized mutations.
4. **Cryptographic Action Ledger**: Emits append-only SHA-256 hash-chained JSONL records (`receipt_hash`) linking previous hashes for verifiable auditability.
5. **Atomic Kill-Switch**: Real-time agent execution freeze (`AAG_KILL_SWITCH=1` or `artifacts/KILL`).

### Testing & Validation
- Added `lib/crewai/tests/security/test_action_gate.py` verifying:
  - Read tools allowed by default.
  - Unapproved mutating writes downgraded to simulation.
  - Prove-token authorization for mutating and destructive tools.
  - Destructive tool denial without prove tokens.
  - Kill-switch activation blocking all execution.
  - SHA-256 hash-chain integrity verification across chained actions.
  - ActionBoundary wrapper decorator behavior.
- Verified all 10 unit tests pass cleanly.

### Upstream & Commercial Context
Maintained by [A2Z SOC](https://a2zsoc.com) for AI runtime governance and compliance audit readiness.

For teams deploying autonomous agent crews requiring AI safety audits or ISO 42001 / SOC 2 readiness sprints:
- **Instant Audit ($499):** https://a2zsoc.com/productized-services#instant-audit-tripwire
- **Consultation & AI Governance Sprints:** https://a2zsoc.com/consultation

### DIFF
--- a/lib/crewai/src/crewai/security/__init__.py
+++ b/lib/crewai/src/crewai/security/__init__.py
@@ -4,11 +4,28 @@ CrewAI security module.
 This module provides security-related functionality for CrewAI, including:
 - Fingerprinting for component identity and tracking
 - Security configuration for controlling access and permissions
-- Future: authentication, scoping, and delegation mechanisms
+- ActionGate & ActionBoundary for deterministic execution boundaries
+- ActionLedger for cryptographic SHA-256 hash-chained compliance audit trails
 """
 
+from crewai.security.action_gate import (
+    ActionBoundary,
+    ActionGate,
+    ActionLedger,
+    Disposition,
+    GateDecision,
+    ToolTier,
+)
 from crewai.security.fingerprint import Fingerprint
 from crewai.security.security_config import SecurityConfig
 
-
-__all__ = ["Fingerprint", "SecurityConfig"]
+__all__ = [
+    "Fingerprint",
+    "SecurityConfig",
+    "ActionGate",
+    "ActionBoundary",
+    "ActionLedger",
+    "GateDecision",
+    "ToolTier",
+    "Disposition",
+]
--- a/lib/crewai/src/crewai/security/action_gate.py
+++ b/lib/crewai/src/crewai/security/action_gate.py
@@ -1,0 +1,316 @@
+"""Action Gate & Cryptographic Action Ledger Module for CrewAI.
+
+Provides deterministic execution boundaries, simulation fallbacks, and
+append-only SHA-256 hash-chained action ledgers for CrewAI agent tool executions.
+"""
+
+from __future__ import annotations
+
+import functools
+import hashlib
+import json
+import os
+import time
+from dataclasses import asdict, dataclass
+from enum import Enum
+from pathlib import Path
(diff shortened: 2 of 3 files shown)


AAH20:
Addressed the security/audit-integrity findings from the automated review. I verified every claim against the actual code before fixing anything by reproducing two live exploits:

```python
gate = ActionGate(prove_token="real-secret")
gate.evaluate("delete_production_database", {}, prove_token="anything")
# -> allowed=True. Any non-empty string authorized a destructive op; the
#    token was checked for truthiness, never compared against the real secret.

gate.evaluate("search_and_terminate_instance", {})
# -> tier=READ, allowed=True. Dict-iteration-order substring matching
#    classified an obviously destructive tool as read-only, which bypasses
#    ALL gating -- READ tools execute unconditionally with no token check.
```

Both are now closed, along with the other three findings:

- **Token check** — a per-call `prove_token` is now verified with `hmac.compare_digest` against the gate's configured secret, not merely checked for truthiness. The gate's own configured secret still authorizes calls with no per-call override (the original intended deployment-secret behavior) — all 10 pre-existing tests pass unchanged.
- **Tool classification** — now checks tiers most-severe-first (DESTRUCTIVE → ... → READ) against whole tokens, not raw substrings, so a name matching more than one tier's keywords is always classified by its most dangerous match.
- **Ledger tamper-evidence** — added `verify_chain()`, which recomputes every entry's hash and checks `prev_hash` linkage. Recovery now calls this and raises rather than silently trusting the last line.
- **Concurrency** — `record()` now takes an advisory file lock and re-reads the true tail under it, instead of trusting an instance-cached `last_hash`. Verified with 8 threads × 20 concurrent calls → 160 entries, chain still verifies.
- **Sensitive arguments** — values for password/secret/token/api_key/credential/auth-shaped keys are now redacted before being logged or hashed. The real values still reach the guarded function; only the audit trail is redacted.

11 new regression tests added (21 total, all passing), each reproducing the original gap and asserting it's closed. Lint clean (ruff).

PR #7025, opened 2026-08-17
fix: handle asyncio.run() RuntimeError when async tools called from running event loop

### LINKED ISSUE #6978: BaseTool.run / StructuredTool raise 'asyncio.run() cannot be called from a running event loop' when a tool returns a coroutine

silentiris:
## Problem

Fixes #6978

When an async tool (a `BaseTool` subclass with an async `_run`, or a `CrewStructuredTool` wrapping a coroutine-returning function) is invoked from within an already-running event loop — e.g. a FastAPI request handler, a Jupyter cell, or any `async def` — `asyncio.run()` raises:

```
RuntimeError: asyncio.run() cannot be called from a running event loop
```

Synchronous scripts are unaffected, which is why this slips through basic testing but breaks every async-server deployment.

## Root cause

`BaseTool.run()` (two call sites) and `CrewStructuredTool.invoke()` (two call sites) all use the bare pattern:

```python
    result = asyncio.run(result)   # raises under a running loop
```

## Fix

The codebase already solves this exact problem in `llm_guardrail.py` with `_run_coroutine_sync`, which detects a running event loop and offloads the coroutine to a dedicated thread. This PR:

1. Extracts that pattern into a shared `crewai.utilities.async_utils.run_coroutine_sync` module
2. Replaces all four `asyncio.run()` call sites in `base_tool.py` and `structured_tool.py` with `run_coroutine_sync()`

The helper preserves `contextvars` and falls back to plain `asyncio.run()` when no loop is running, so sync behavior is unchanged.

## Tests

Three new tests cover the three affected entry points:

- `test_run_from_running_event_loop_does_not_raise` — `BaseTool` subclass with async `_run`, called from inside `async def`
- `test_tool_decorator_run_from_running_event_loop_does_not_raise` — `@tool` decorator wrapping an async function, called from inside `async def`
- `test_invoke_from_running_event_loop_does_not_raise` — `CrewStructuredTool.from_function` with an async function, `.invoke()` called from inside `async def`

All 76 existing tool tests continue to pass.

### DIFF
--- a/lib/crewai/src/crewai/tools/base_tool.py
+++ b/lib/crewai/src/crewai/tools/base_tool.py
@@ -40,6 +40,7 @@ from crewai.tools.structured_tool import (
 )
 from crewai.tools.tool_failure import ToolFailure, ToolFailurePolicy, ToolFailureReason
 from crewai.types.callback import SerializableCallable, _resolve_dotted_path
+from crewai.utilities.async_utils import run_coroutine_sync
 from crewai.utilities.string_utils import sanitize_tool_name
 
 
@@ -338,7 +339,7 @@ class BaseTool(BaseModel, ABC):
         result = self._run(*args, **kwargs)
 
         if asyncio.iscoroutine(result):
-            result = asyncio.run(result)
+            result = run_coroutine_sync(result)
 
         return result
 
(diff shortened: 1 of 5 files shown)


Vidit-Ostwal (commented):
Closing as a duplicate of #6979, which already applies the same `run_coroutine_sync` helper to all four `asyncio.run` call sites in `base_tool.py` and `structured_tool.py`, and has broader regression coverage (including the sync-func-returns-coroutine `invoke` path). Follow along there.

PR #7031, opened 2026-08-18
feat(flow): make conversational opt-in unmistakable

joaomdmoura:
- Makes `@ConversationConfig(...)` opt a Flow into conversational mode on its own, so chat no longer needs two statements that must agree. Every field on the config is consumed only by the conversational graph, so a decorated non-conversational Flow could only ever discard it.
- Public API consequence, called out deliberately: `ConversationConfig` is re-exported from `crewai.experimental`, and decorating a Flow now sets `conversational = True` on it. `handle_turn()` also raises `ValueError` on a non-conversational Flow instead of returning `None` — matching the guards `chat()` and `stream_turn()` already have. Every in-repo `handle_turn` caller is on a conversational Flow; an out-of-repo caller relying on the `None` return will now see the error, which is the point. And `FlowConversationalDefinition.enabled` now defaults to `true`, so declaring the block in JSON/YAML is the opt-in; `enabled: false` keeps the config while turning chat off.
- Preserved: `conversational = True` by hand still opts in with no config; a Flow with neither stays non-conversational with `conversational is None` in its definition; both existing `FlowConversationalDefinition` constructors already passed `enabled=True` explicitly, and nothing asserted the old `False` default.
- Tests: 194 lines of new tests against 40 of production. Covers the decorator alone enabling the mode, registering the built-in graph, running a full turn and deferring trace finalization; the bare flag still working; a Flow with neither staying plain; the decorator not leaking the flag onto later Flow subclasses or onto `Flow` itself; the `handle_turn` guard and its parity with `chat`/`stream_turn`; and the three JSON cases (block without `enabled`, explicit `enabled: false`, no block at all). Ran green: `test_flow_conversation` + `test_flow_definition` + `test_flow_from_definition` = 290 passed; all flow + telemetry suites = 726 passed; mirrored CLI tests = 269 passed. `ruff check`/`format` clean; `mypy` on the three changed files adds zero errors over the clean-tree baseline.
- No docs touched — the declarative conversational surface is documented in a later PR in this sequence, once it executes.
- Keeps this intentionally small: no inference from `Flow[ConversationState]` (verified unsafe — `Flow[ChatState]` + `conversational = True` raises today), no deprecation of the bare `conversational = True` flag, no removal of the `enabled` field, no conversational resolver, no built-in graph synthesis, no state-precedence fix.
- Next: route every conversational config read through a resolver that prefers the instance `_definition` over the class projection, so a JSON/YAML declaration can actually drive chat — and fix state precedence in the same change, because flipping the gates without it destroys declared state fields.

Stacked on #7030; retarget to `main` once that merges.

---

> [!NOTE]
> **Medium Risk**
> Breaking change for callers that invoked handle_turn on non-conversational flows and relied on None; conversational JSON semantics change because enabled now defaults to true when the block is present.
> 
> **Overview**
> **`@ConversationConfig` alone opts a Flow into conversational mode** by setting `conversational = True` on the class, so decorators no longer need a separate flag that could disagree and silently drop config.
> 
> **`handle_turn()` now raises `ValueError` on non-conversational flows** instead of returning `None`, aligning with `chat()` and `stream_turn()`.
> 
> **Declarative flows:** `FlowConversationalDefinition.enabled` defaults to `true`, so a JSON/YAML `conversational` block is the opt-in; `enabled: false` keeps settings while turning chat off. Omitting the block still leaves `conversational` as `None`.
> 
> Tests cover decorator-only opt-in, guards, JSON declaration cases, and that the flag does not leak to other Flow subclasses.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eafa3c8a6f1e8bc34780cd1d091c7f460163514f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/experimental/conversational.py
+++ b/lib/crewai/src/crewai/experimental/conversational.py
@@ -96,8 +96,17 @@ class ConversationConfig:
     defer_trace_finalization: bool = True
 
     def __call__(self, flow_cls: type[Any]) -> type[Any]:
-        """Use this config as a class decorator."""
+        """Use this config as a class decorator.
+
+        Decorating a Flow also opts it into conversational mode. Every field
+        here is consumed only by the conversational runtime, so a decorated Flow
+        that was not conversational would silently discard its whole config:
+        ``handle_turn()`` would return ``None`` without appending or raising.
+        Setting ``conversational = True`` explicitly is still supported and is
+        the way to opt in without a config.
+        """
         flow_cls.conversational_config = self
+        flow_cls.conversational = True
         return flow_cls
 
 
--- a/lib/crewai/src/crewai/experimental/conversational_mixin.py
+++ b/lib/crewai/src/crewai/experimental/conversational_mixin.py
@@ -293,11 +293,22 @@ class _ConversationalMixin:
            conversational ``Flow``. Signature and semantics may change before
            the feature graduates from ``crewai.experimental``.
 
-        Available only when ``conversational = True`` is set on the subclass.
-        Stashes the message + session_id as pending turn state, runs kickoff
-        (which restores from persist and then applies the pending turn), and
-        promotes the result to an assistant message when the handler didn't.
+        Available only when ``conversational = True`` is set on the subclass
+        (``@ConversationConfig`` sets it for you). Stashes the message +
+        session_id as pending turn state, runs kickoff (which restores from
+        persist and then applies the pending turn), and promotes the result to
+        an assistant message when the handler didn't.
+
+        Raises:
+            ValueError: If the flow is not conversational. Without this the
+                turn is a silent no-op: no route handlers are registered, the
+                message is never appended, and ``None`` comes back.
         """
+        if not self._is_conversational_enabled():
+            raise ValueError(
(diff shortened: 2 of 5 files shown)


joaomdmoura:
@coderabbitai review

gabemilani (dismissed):

vinibrsl (dismissed):

Vidit-Ostwal:
**TLDR**

Conversational opt-in is now a single signal instead of two that could disagree.

- `@ConversationConfig(...)` sets `conversational = True` on the Flow, so the decorator alone enables chat (bare `conversational = True` still works).
- `handle_turn()` raises `ValueError` on a non-conversational Flow, matching `chat()` / `stream_turn()` — no more silent `None`.
- JSON/YAML: declaring a `conversational:` block opts in (`enabled` defaults to `true`); `enabled: false` keeps the config and turns chat off. Omitting the block stays non-conversational.

Python chat path is complete here. A declaration driving runtime chat is #7032.

Vidit-Ostwal (approved):

PR #7042, opened 2026-08-18
fix(tools): support coroutine results inside active event loops

### LINKED ISSUE #6978: BaseTool.run / StructuredTool raise 'asyncio.run() cannot be called from a running event loop' when a tool returns a coroutine

Yash-Chindam:
## Summary
- run coroutine-returning tools safely when a synchronous API is called inside an active event loop
- share the bridge across BaseTool, @tool, and CrewStructuredTool
- add regression coverage for async-context calls

## Root cause
The synchronous tool paths always called `asyncio.run`, which raises when the caller already owns a running event loop (for example FastAPI or Jupyter).

## Validation
- Python bytecode compilation passed
- Targeted pytest was attempted but the repository test harness requires the missing `vcr` dependency in this checkout

Fixes #6978

### DIFF
--- a/lib/crewai/src/crewai/tools/base_tool.py
+++ b/lib/crewai/src/crewai/tools/base_tool.py
@@ -41,6 +41,7 @@ from crewai.tools.structured_tool import (
 from crewai.tools.tool_failure import ToolFailure, ToolFailurePolicy, ToolFailureReason
 from crewai.types.callback import SerializableCallable, _resolve_dotted_path
 from crewai.utilities.string_utils import sanitize_tool_name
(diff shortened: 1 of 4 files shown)


Yash-Chindam:
Closing this duplicate implementation in favor of the more complete existing PR #6979, which covers all affected coroutine paths and regression tests.

PR #7048, opened 2026-08-19
fix: use hostname as MCP HTTP/SSE server_name

lucasgomide:
HTTP and SSE MCP connections used the raw URL as server_name, so traces titled the row with the full endpoint including query params. `_get_server_info` now emits the hostname and keeps the full URL on `server_url`.


Vidit-Ostwal (approved):

renatonitta (commented):
👍

PR #7056, opened 2026-08-20
fix(events): always emit project_id so absent and empty stay distinct

joaomdmoura:
- Makes `project_id` always present on every span, as the empty string when the project declares none. It was previously omitted, which made two different situations indistinguishable downstream: **a client too old to report a project id at all, versus a current client whose project simply declares none.**
- **Why this is not cosmetic.** The share of clients that *could* have reported an id is the denominator of every attribution rate. With both cases collapsed into "key absent" that denominator cannot be computed — it can only be inferred from a version floor, which is fragile and silently wrong for anyone who pins or backports. Measured on the warehouse before making the change: on spans that carry `coding_agent`, `coding_agent` is present on **100%**, `runtime_context` on **81.6%** and `project_id` on **3.465%** — same processor, same dict, and no way to tell which part of that 96.5% is "too old" and which is "no project".
- **It still never invents an identity.** `get_project_id()` stays read-only and minting remains with the CLI commands a user explicitly invoked. An empty string means *"asked, and there is none"* — it is not a placeholder id.
- Consistent with the two attributes beside it: `coding_agent` and `runtime_context` already always emit a value, so this makes the third member of `common_span_attributes()` behave like the other two.
- **Two existing tests asserted the old contract and are updated, not deleted** — `test_project_id_is_omitted_when_absent` is renamed to `test_project_id_is_empty_rather_than_absent_when_undeclared` because its name described the behaviour that changed, and `test_project_id_lookup_never_breaks_telemetry` now asserts a failed lookup degrades to an empty id rather than an omitted key. A third test is added pinning the distinction itself.
- **One existing test deliberately left alone:** the one asserting a foreign application's spans are never annotated. This changes what our processor stamps, not where it is attached, so that guarantee is untouched and still passes.
- Verified locally: `96 passed` across `lib/crewai/tests/telemetry/test_coding_agent_detection.py` and `lib/crewai-core/tests/test_runtime_env.py`; `ruff check`, `ruff format --check` and `mypy` all clean.
- Keeps this intentionally small: no change to where the processor is attached, no change to minting, no new attribute, and no change to `get_project_id()` itself.

---

> [!NOTE]
> **Low Risk**
> Small telemetry attribute contract change with no auth, minting, or execution-path impact; downstream must treat empty string as "no id" rather than absent key.
> 
> **Overview**
> **Telemetry spans now always include `project_id`**, using `""` when the project has no id or lookup fails, instead of omitting the attribute.
> 
> That separates **legacy clients** (key missing) from **current clients with no declared id** (key present, empty), so warehouse metrics can compute attribution denominators without inferring from version floors. `get_project_id()` stays read-only; empty string does not mint an id. Behaviour aligns with `coding_agent` and `runtime_context`, which already always emit values.
> 
> Tests in `test_coding_agent_detection.py` were updated for the new contract and a case was added asserting declared vs undeclared ids differ.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 58a16d1666ec96b53f7121ec454382a092211814. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai-core/src/crewai_core/telemetry.py
+++ b/lib/crewai-core/src/crewai_core/telemetry.py
@@ -136,8 +136,10 @@ def common_span_attributes() -> dict[str, str]:
 
     Returns:
         Attributes to stamp on every span. ``project_id`` is always present and
-        is the empty string for projects that do not declare one -- see the
-        comment at the assignment for why absent and empty must stay distinct.
+        is the empty string whenever no id is available -- both for projects that
+        declare none and when the lookup itself failed, which are deliberately
+        indistinguishable here because neither yields an id. See the comment at
+        the assignment for why empty and *absent* must stay distinct.
     """
     attributes = {
         "coding_agent": detect_coding_agent(),
@@ -155,9 +157,11 @@ def common_span_attributes() -> dict[str, str]:
     # Always set the key, even when empty. Absent and empty mean different things
     # and only this distinction can tell them apart: absent means the client is too
     # old to report a project id at all, empty means the client asked and the project
-    # declares none. Collapsing both into "absent" makes the share of clients that
-    # COULD have reported one unknowable, and that share is the denominator every
-    # attribution rate needs.
+    # declares none -- or the lookup failed, which lands here too and is treated the
+    # same, since an unreadable pyproject.toml also means no id is available.
+    # Collapsing empty into "absent" makes the share of clients that COULD have
+    # reported one unknowable, and that share is the denominator every attribution
+    # rate needs.
     attributes["project_id"] = project_id or ""
 
     return attributes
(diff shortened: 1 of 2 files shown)


joaomdmoura:
**CodeRabbit — document lookup-failure semantics: FIXED** (`ed1ff7a`-ish, pushed).

Verified against the code first and the finding is correct: the `except` branch sets `project_id = None`, which then lands on the same `or ""` and produces an empty string. My docstring described empty only as "a project that declares none", so the failure case was undocumented.

Both now say so explicitly, and I kept the *reason* they are deliberately indistinguishable — neither yields an id, so there is nothing for a consumer to do differently. What must stay distinct is **empty versus absent**, not the two ways of arriving at empty. That distinction is why this PR exists, so I made sure the wording does not blur it.

One thing worth noting for anyone debugging: an unreadable `pyproject.toml` now looks identical to a project with no id, so `project_id = ""` should not be read as "the user declared nothing" on its own.

`93 passed` on the telemetry suite; `ruff` clean.

alex-clawd (approved):
Telemetry contract change is scoped and well-tested: current clients always emit project_id, using an empty string when unavailable, while legacy clients remain distinguishable by an absent key. All checks are green.

PR #7057, opened 2026-08-20
feat(cli): backfill project_id from every user-invoked project command

joaomdmoura:
- Extends the `project_id` backfill from `crewai run` alone to **every user-invoked command that touches a project**: `train`, `replay`, `test`, `login`, `deploy create`, `deploy push`, `flow add-crew`, `enterprise configure`, and `traces enable`.
- **The gap this closes.** `crewai run` has always backfilled — a project declaring `[tool.crewai]` without a `project_id` gets one minted on first run. No other command did, so a project driven entirely through `crewai test`, or only ever deployed, never acquired an id and **every one of its runs stayed unattributable**. That is a denominator problem rather than a cosmetic one: coverage is currently ~1.9% of runs and is *falling* as version adoption spreads, because the share is governed by how many projects have an id rather than by which client version they run.
- **No new policy.** Each site is an action the user explicitly invoked, which is exactly the condition `run_crew` already relies on. It is still **never** called from the SDK during kickoff, and `get_or_create_project_id` still **refuses to create the `[tool.crewai]` table** — so a directory that merely happens to contain a `pyproject.toml` is never rewritten. The comment at each call site records both constraints.
- **Two things deliberately not done.** `crewai flow kickoff` is untouched because it delegates to `run_crew` and already inherits the backfill — a test pins that, so the delegation is not accidentally duplicated into a second mint under a lock `run_crew` is about to take. And there is **no `crewai evaluate` command**; `crewai test` is that path, so the approved list maps to nine sites rather than eleven.
- **Placement matters and is intentional:** the call is the first statement in each command body, so a command that later fails still leaves the project with an id.
- **Test design.** The patched backfill *raises*, which proves the call happened and simultaneously guarantees nothing after it runs — so no test touches real user settings, spawns a subprocess, or reaches the network. Table-driven across all nine commands, plus two guard tests (that `crewai run` still backfills, and that `flow kickoff` still delegates rather than duplicating). **Verified against the unpatched module: the 9 command tests fail and the 2 guard tests still pass**, which is the intended split.
- **Tests live under `lib/crewai/tests/cli/` on purpose.** The required `tests` job runs only `lib/crewai` and `lib/crewai-tools`; **no workflow runs `lib/cli/tests/`**, so a test placed there would never execute in CI. Flagging that as a pre-existing gap rather than fixing it here. Relatedly, `lib/cli`'s own venv cannot run its suite standalone (missing `aiohttp` among others) — `lib/cli/tests/test_run_declarative_flow.py` gives `31 passed` from the root workspace env both with and without this change, so those failures are environmental and not caused by this PR.
- Verified locally: `280 passed` for `lib/crewai/tests/cli/`, `11 passed` for the new file, `ruff check`, `ruff format --check` and `mypy` all clean.
- Keeps this intentionally small: no change to `get_or_create_project_id` itself, no new minting policy, no attempt to create `[tool.crewai]` where it is absent, and no TUI change — the TUI traces-confirmation site is tracked separately because its code path still needs locating.

---

> [!NOTE]
> **Low Risk**
> Behavior is limited to explicit CLI entry points using the existing backfill helper; it may write `project_id` into `pyproject.toml` when `[tool.crewai]` already exists.
> 
> **Overview**
> Extends **`project_id` backfill** so projects with `[tool.crewai]` but no id get one minted on the **first statement** of nine additional user-invoked commands—not only `crewai run`. Affected entry points: `train`, `replay`, `test`, `login`, `deploy create`, `deploy push`, `flow add-crew`, `enterprise configure`, and `traces enable`. Each call site documents that backfill stays a no-op without `pyproject.toml`, never creates `[tool.crewai]`, and is not invoked from the SDK during kickoff.
> 
> **`crewai flow kickoff` is unchanged** (still delegates to `run_crew` for backfill). New coverage lives in `lib/crewai/tests/cli/test_project_id_backfill.py`: parametrized checks that all nine commands invoke backfill, ordering on `login`, and guards that `run` still backfills and `flow kickoff` does not duplicate it.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cade8fbac8440c1f334a854850c06bd2b6bee383. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/cli/src/crewai_cli/cli.py
+++ b/lib/cli/src/crewai_cli/cli.py
@@ -17,6 +17,7 @@ from crewai_cli.user_data import (
 from crewai_cli.utils import (
     build_env_with_all_tool_credentials,
     enable_prompt_line_editing,
+    get_or_create_project_id,
     is_dmn_mode_enabled,
     read_toml,
     warn_deprecated,
@@ -331,6 +332,11 @@ def train(
     filename: str,
 ) -> None:
     """Train the crew."""
+    # Backfills a project_id for projects that have [tool.crewai] but no id yet.
+    # Safe in every command the user explicitly invoked: get_or_create_project_id
+    # is a no-op without a pyproject.toml and refuses to create [tool.crewai], so it
+    # never rewrites an unrelated directory. Never called from the SDK during kickoff.
+    get_or_create_project_id()
     if deprecated_n_iterations is not None:
         warn_deprecated(kind="flag", old="--n_iterations", new="--n-iterations")
         n_iterations = deprecated_n_iterations
@@ -378,6 +384,11 @@ def replay(
         task_id: The ID of the task to replay from.
         trained_agents_file: Optional trained-agents pickle path.
     """
+    # Backfills a project_id for projects that have [tool.crewai] but no id yet.
+    # Safe in every command the user explicitly invoked: get_or_create_project_id
+    # is a no-op without a pyproject.toml and refuses to create [tool.crewai], so it
+    # never rewrites an unrelated directory. Never called from the SDK during kickoff.
+    get_or_create_project_id()
     if deprecated_task_id is not None:
         warn_deprecated(kind="flag", old="--task_id", new="--task-id")
         task_id = deprecated_task_id
@@ -593,6 +604,11 @@ def test(
     trained_agents_file: str | None,
 ) -> None:
     """Test the crew and evaluate the results."""
+    # Backfills a project_id for projects that have [tool.crewai] but no id yet.
+    # Safe in every command the user explicitly invoked: get_or_create_project_id
+    # is a no-op without a pyproject.toml and refuses to create [tool.crewai], so it
+    # never rewrites an unrelated directory. Never called from the SDK during kickoff.
+    get_or_create_project_id()
     if deprecated_n_iterations is not None:
         warn_deprecated(kind="flag", old="--n_iterations", new="--n-iterations")
         n_iterations = deprecated_n_iterations
@@ -669,6 +685,11 @@ def update() -> None:
 @crewai.command()
 def login() -> None:
     """Sign Up/Login to CrewAI AMP."""
+    # Backfills a project_id for projects that have [tool.crewai] but no id yet.
+    # Safe in every command the user explicitly invoked: get_or_create_project_id
+    # is a no-op without a pyproject.toml and refuses to create [tool.crewai], so it
+    # never rewrites an unrelated directory. Never called from the SDK during kickoff.
+    get_or_create_project_id()
     Settings().clear_user_settings()
     AuthenticationCommand().login()
 
@@ -703,6 +724,11 @@ def deploy() -> None:
 )
 def deploy_create(yes: bool, skip_validate: bool) -> None:
     """Create a Crew deployment."""
+    # Backfills a project_id for projects that have [tool.crewai] but no id yet.
+    # Safe in every command the user explicitly invoked: get_or_create_project_id
+    # is a no-op without a pyproject.toml and refuses to create [tool.crewai], so it
+    # never rewrites an unrelated directory. Never called from the SDK during kickoff.
+    get_or_create_project_id()
     deploy_cmd = DeployCommand()
     deploy_cmd.create_crew(yes, skip_validate=skip_validate)
 
@@ -723,6 +749,11 @@ def deploy_list() -> None:
 )
 def deploy_push(uuid: str | None, skip_validate: bool) -> None:
     """Deploy the Crew."""
+    # Backfills a project_id for projects that have [tool.crewai] but no id yet.
+    # Safe in every command the user explicitly invoked: get_or_create_project_id
+    # is a no-op without a pyproject.toml and refuses to create [tool.crewai], so it
+    # never rewrites an unrelated directory. Never called from the SDK during kickoff.
+    get_or_create_project_id()
     deploy_cmd = DeployCommand()
     deploy_cmd.deploy(uuid=uuid, skip_validate=skip_validate)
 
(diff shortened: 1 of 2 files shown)


joaomdmoura:
Both findings **FIXED**, and the second one was a fair criticism of my own test design.

**CodeRabbit — use runtime assertions instead of source-text checks: FIXED.** The two guard tests grepped module source for a call string. That asserts on *formatting*: a reformat would break them, and a real regression could slip past. They now invoke the commands in an isolated project and assert on observed calls.

The flow-kickoff test is the interesting one. It patches the **two distinct import sites separately** — `crewai_cli.cli.get_or_create_project_id` and `crewai_cli.run_crew.get_or_create_project_id` — and asserts run_crew's is called **exactly once** while cli's is **not called at all**. That is what makes "delegates" and "duplicates" distinguishable at runtime; a single mock could not tell them apart.

**Verified both catch what they claim**, rather than assuming:
- injecting a duplicate `get_or_create_project_id()` into `flow_run` → the delegation test fails
- removing `run_crew`'s own call → the run test fails
- restored → `11 passed`

**github-code-quality — mixed import styles for `crewai_cli.cli`: FIXED**, and by the same change rather than separately. The rewrite removed the need for `import crewai_cli.cli as cli_module` altogether, so there is no longer a second import style to reconcile. I preferred that over the suggested `from crewai_cli.cli import __file__ as cli_file`, which would have kept the source-text approach the other finding was right to object to.

`280 passed` for `lib/crewai/tests/cli/`; `ruff` and `ruff format` clean.

joaomdmoura:
**CodeRabbit — make the exact-once assertion observable: FIXED, and the finding was right in a way that mattered.**

With `side_effect=_BackfillReached` the mock raised on first use, so `call_count == 1` was guaranteed by the mock rather than by the code under test. A second backfill call **inside the same `run_crew` execution could never have been observed at all**. The duplicate protection I had actually came from `assert not in_cli.called`, so the count assertion read stronger than it was.

Both backfill mocks now return normally, and execution stops at the first call *after* the backfill (`configured_project_json_crew`, which run_crew calls immediately next). The recorded count is therefore real.

**Verified the difference this makes rather than assuming it:**

| injected regression | before | after |
|---|---|---|
| duplicate `get_or_create_project_id()` **inside `run_crew`** | not detected | **both tests fail** |
| duplicate in `flow_run` | detected | still detected |
| no regression | 11 passed | 11 passed |

So the previous version was blind to exactly the case the assertion claimed to cover.

**Also rebased onto `main`**, which now includes #7056 (`7c72d57b7`). `280 passed` for `lib/crewai/tests/cli/` after the rebase.

---

On the earlier CI red: both this PR and #7056 failed the identical test, `tests/tracing/test_trace_enable_disable.py::test_trace_calls_when_enabled_via_env`, with cassette exhaustion (`Found 1 recorded request(s) ... already been consumed`). Neither diff adds an OpenAI call — this PR only adds CLI backfill calls — and the test passes on clean `main` locally across five random seeds. #7056 has since merged, which is consistent with it not being caused by these changes. Flagging it as a latent inter-test dependency that shard reshuffling exposes, rather than asserting flakiness without evidence.

joaomdmoura:
**CodeRabbit — assert that `flow kickoff` reaches the post-backfill boundary: FIXED.** Correct again, and it is the same class of flaw as the last one: an assertion that can pass for the wrong reason.

The flow-kickoff test discarded the `runner.invoke()` result, so if the path returned or raised **after** one backfill call but **before** `configured_project_json_crew`, both `in_run_crew.call_count == 1` and `not in_cli.called` would still have held. `test_run_still_backfills` already asserted the boundary, so the pair was inconsistent and the reviewer caught the weaker half.

**Verified it earns its place rather than just adding it:** injecting an early `return` into `run_crew` immediately after the backfill and before the boundary now **fails both tests**, and previously would have failed **neither**.

That is three findings on this file across three passes, each one tightening a different way the tests could have passed without proving anything:
1. source-text greps → runtime assertions
2. `call_count` capped by a raising mock → mock returns, stop at a later boundary
3. boundary never asserted on one of the two tests → asserted on both

`11 passed` for the file, `280 passed` for `lib/crewai/tests/cli/`, `ruff` clean.

joaomdmoura:
**CodeRabbit — assert that backfill occurs before command-specific work: PARTIALLY FIXED, and I am not taking the full suggestion. Reasons for both halves.**

**The finding is valid.** The parametrized test proves the backfill is *reached*; it does not prove nothing ran before it. My assertion message claimed the stronger thing, which was the actual defect — the test was fine, the claim was not.

**Fixed:**
1. The message now states what the test establishes, and points at the test that establishes the ordering.
2. A new test pins the ordering on `login`, whose first action goes through a module-level name (`Settings`) that can be patched without reaching into the command. **Verified it catches the regression:** swapping login's first two statements so its own work runs before the backfill fails the new test.

**Not doing: parameterizing a first-work boundary across all nine commands.** That means naming each command's *current* first action — `Settings` for login, `DeployCommand` for deploy, the subprocess target for train, and so on. Those change as the commands evolve, so the suite would start tracking nine commands' internals rather than this one ordering property, and would go red on refactors that break nothing. One representative command establishes that the placement pattern works; placement itself is a one-line, visible property of the diff for the other eight.

If you would rather have all nine pinned despite the maintenance cost, say so and I will add them — I would just rather that be an explicit choice than something I slid in.

`12 passed` for the file, `281 passed` for `lib/crewai/tests/cli/`, `ruff` clean.

alex-clawd (approved):
Project ID backfill is consistently applied to explicit project-touching CLI commands while preserving existing safety boundaries: no SDK kickoff minting, no creation of missing [tool.crewai], and no duplicate flow-kickoff call. Coverage and CI are green.

PR #7061, opened 2026-08-20
feat(flow): let a chat flow declare its own state shape

joaomdmoura:
- Lets a conversational declaration use any state shape. Previously only `state: {type: pydantic, ref: <a ConversationState subclass>}` worked; every other shape loaded clean and then died on the first turn — inline `json_schema` and a non-subclass ref with `AttributeError: 'StateWithId' object has no attribute 'messages'`, and `type: dict` with `AttributeError: 'dict' object has no attribute 'id'`. All three reproduced before the fix and verified after.
- New runtime extension seam `Flow._compose_extension_state_model(model_class)` — the seventh alongside the existing six — applied to the model built from `state:` before the engine wraps it for `id`. The conversational mixin overrides it to add the chat fields to whatever the declaration asked for. No public import surface changes; `_build_definition_state_model` and `_create_definition_state` have no callers outside `runtime/__init__.py`.
- Preserved: declared fields and defaults survive composition (verified `ticket_id`/`turns` present, `ticket_id == "T-1"`); a model already extending `ConversationState` is returned untouched, so today's supported shape is a byte-for-byte no-op and is not composed twice (asserted on the MRO); non-conversational declarations are unaffected; class-based Python conversational flows unchanged.
- `dict` and `unknown` state cannot carry the chat fields at all, so the default extension state now supplies the real shape, seeded from the declared defaults where they fit. Deliberately *not* an error: `class DictChat(Flow[dict]): conversational = True` constructs today and only fails on `kickoff()`, and `_extend_definition` runs on every instantiation — raising there would turn "works until you chat" into "cannot be constructed", breaking `crewai flow plot` and definition-only consumers.
- Tests: inline `json_schema` gaining the fields while keeping its own, a non-`ConversationState` pydantic ref being composed, a subclass ref not being composed twice, `dict` state getting the real shape and honouring a declared default, a full turn running on declared state, and a non-conversational declaration keeping its plain state. 1074 passed across flow/telemetry/CLI; `ruff` clean; `mypy` adds 0 errors on both changed files.
- Keeps this intentionally small: no validation error for a state shape that cannot work, no `conversational.state` field on the contract, no docs change (the guide already says a custom state must extend the conversational one — that is now one of several working options, so the wording is a follow-up).
- Next: document the supported state shapes in the guide (en + ar/ko/pt-BR).

---

> [!NOTE]
> **Medium Risk**
> Touches Flow state construction for conversational declarations (model composition and dict fallbacks). Non-chat flows and existing ConversationState subclasses are intended to stay unchanged.
> 
> **Overview**
> Chat flows declared with a custom `state:` block no longer die on the first turn unless that model already subclassed `ConversationState`.
> 
> **pydantic / json_schema** state is composed with `ConversationState` so chat fields (`messages`, `last_intent`, etc.) sit alongside declared fields and defaults. A model that already extends `ConversationState` is left alone (no double composition).
> 
> **dict / unknown** (and unbuildable refs) cannot carry those fields, so the runtime now supplies a real `ConversationState`, keeping extra declared defaults as extras. Non-conversational declarations are unchanged.
> 
> Adds a `_compose_extension_state_model` runtime hook used when building declared state, plus tests covering composition, dict extras, failed refs, and a full turn.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f3ddfa1fdaead6bdc99f74e54b761da5ce13e720. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joaomdmoura:
@coderabbitai review

joaomdmoura:
@coderabbitai review

Vidit-Ostwal (approved):

PR #7063, opened 2026-08-20
feat(flow): let a declaration name the router's response format

joaomdmoura:
- Lets a declaration name the Pydantic model the routing decision is parsed into. `conversational.router.response_format` was typed `Any` and dropped with a warning, so the router always used its synthesized fallback — `intent: str` with the route labels only in a field *description*. A declared model gives it a `Literal[...]` of the real labels plus `additionalProperties: false`, which is a measurably tighter contract for the routing call.
- Uses the shape that already exists for this: `{"python": "module.path.Class"}`, the same `PythonReferenceDefinition` a crew agent's `response_format` is declared with, resolved through the same `_resolve_model_class`. No new mechanism, and it inherits that resolver's project-root containment — a declaration cannot reach outside the project to import code. Verified: `{"python": "os.path.basename"}` is refused with "must point to modules inside the project root".
- Public API consequence, called out: the field type changes from `Any` to `PythonReferenceDefinition | None`, so a bare `{"ref": "module:qualname"}` — what the old DSL projection emitted — is now a load-time `ValidationError` instead of being silently discarded. That is the point: a typo used to look like it worked. The test that pinned the drop-with-warning behavior is rewritten to assert the new error.
- Preserved: the DSL projection now emits `{"python": ...}` for a live class, so a Python flow's definition round-trips into something reloadable rather than an opaque ref (asserted). A live class on `@ConversationConfig(router=RouterConfig(response_format=...))` is still passed through by identity — class config wins, unchanged. Omitting `response_format` still synthesizes the single-field model.
- Tests: a declared ref resolving to the class and reaching `_router_response_format`, the synthesized fallback when omitted, a ref outside the project root being refused (resolution is lazy, so the declaration loads and the import is refused), a ref without a dot failing at load, live-class passthrough by identity, and the projection emitting the dotted shape. 1164 passed across flow/telemetry/CLI/project suites; `ruff` clean; `mypy` 0 errors on all three changed files.
- Resolves the declared ref against the **declaration's own directory** (`FlowDefinition.source_dir`), not `Path.cwd()`. Review caught this: `project_root=None` meant a flow loaded with `Flow.from_declaration(path=...)` from another working directory could not import a `response_format` sitting beside its own YAML. Crew loading already anchors on `crew_path.parent`, and `_actions.py` already uses `source_dir` as `base_dir`, so this matches both. Regression test loads by path from an unrelated cwd; without the fix it fails with "'chat_routes' was not found under .../elsewhere".
- Corrects the conversational guide's "not expressible" table, which #7104 landed mid-review still saying a `response_format` ref "is ignored with a warning" — false once this merges. Points it at the `{python: ...}` form in en/ar/ko/pt-BR, code sample byte-identical across all four.
- Keeps this intentionally small: no inline-JSON-schema alternative for `response_format`, no change to the synthesized fallback, no docs change beyond the Field description that flows into the authoring skill.
- Next: document the accepted shape in the conversational guide (en + ar/ko/pt-BR).

---

> [!NOTE]
> **Medium Risk**
> Changes declarative contract validation and dynamic imports for router LLM structured output; misconfigured refs now fail at load instead of silently using the fallback model.
> 
> **Overview**
> Declarative conversational flows can now point **`conversational.router.response_format`** at a Pydantic model via the same **`{python: module.path.Class}`** shape crew agents use, instead of always falling back to the synthesized single-field router schema.
> 
> Runtime resolution imports that class through **`_resolve_model_class`**, anchored on the declaration file’s directory (**`FlowDefinition.source_dir`**) so models beside the YAML load correctly even when the process cwd differs. Imports stay inside the project root; invalid or legacy shapes fail loudly—a bare **`{"ref": "module:qualname"}`** is now a load-time **`ValidationError`**, not a silent drop with a warning.
> 
> When projecting a Python **`@ConversationConfig`** flow to a definition, importable Pydantic classes become **`{"python": ...}`** refs (with warnings and omission for locals, nested classes, and non-models); live class config on Python flows is unchanged. Docs in **en/ar/ko/pt-BR** update the “not expressible” table accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd93e8ed634b8b68712424a868025d6fcd07428e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/experimental/conversational_mixin.py
+++ b/lib/crewai/src/crewai/experimental/conversational_mixin.py
@@ -58,6 +58,7 @@ from crewai.flow.conversational_definition import (
 from crewai.flow.dsl import listen, start
 from crewai.flow.dsl._utils import _method_action, _set_flow_method_definition
 from crewai.flow.flow_definition import FlowDefinition, FlowMethodDefinition
+from crewai.project.crew_definition import PythonReferenceDefinition
 from crewai.utilities.types import LLMMessage
 
 
@@ -83,25 +84,33 @@ def _iter_condition_labels(condition: Any) -> set[str]:
     return set()
 
 
+def _resolve_router_response_format(
+    declared: PythonReferenceDefinition | None,
+) -> type[BaseModel] | None:
+    """Import the model class a declaration names for the routing decision.
+
+    ``_router_response_format`` hands its result straight to
+    ``llm.call(response_format=...)``, which needs a real class. Resolved the
+    same way a crew agent's ``response_format`` is, so the declaration shape
+    and its failure messages match.
+    """
+    if declared is None:
+        return None
+
+    from crewai.project.json_loader import _resolve_model_class
+
+    return _resolve_model_class(
+        declared.model_dump(),
+        "conversational.router.response_format",
+        None,
+    )
+
+
 def _router_config_from_definition(
     definition: FlowConversationalRouterDefinition,
 ) -> RouterConfig:
     """Build a live ``RouterConfig`` from its serializable form."""
-    response_format = definition.response_format
-    if response_format is not None and not (
-        isinstance(response_format, type) and issubclass(response_format, BaseModel)
-    ):
-        # A declaration can only carry a ``module:qualname`` ref or a schema
-        # dict here, and ``_router_response_format`` hands its result straight
-        # to ``llm.call(response_format=...)``, which needs a real class.
-        # Dropping it falls back to the synthesized single-field model.
-        logger.warning(
-            "Ignoring conversational router response_format %r: a declaration "
-            "cannot carry a model class. The router will use its synthesized "
-            "response format instead.",
-            response_format,
-        )
-        response_format = None
+    response_format = _resolve_router_response_format(definition.response_format)
 
     return RouterConfig(
         prompt=definition.prompt,
--- a/lib/crewai/src/crewai/experimental/conversational_mixin.py
+++ b/lib/crewai/src/crewai/experimental/conversational_mixin.py
@@ -23,6 +23,7 @@ from contextlib import contextmanager
 from enum import Enum
 import json
 import logging
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar, Literal, cast
 
 from pydantic import BaseModel, ConfigDict, Field, create_model
@@ -86,13 +87,16 @@ def _iter_condition_labels(condition: Any) -> set[str]:
 
 def _resolve_router_response_format(
     declared: PythonReferenceDefinition | None,
+    project_root: Path | None,
 ) -> type[BaseModel] | None:
     """Import the model class a declaration names for the routing decision.
 
     ``_router_response_format`` hands its result straight to
     ``llm.call(response_format=...)``, which needs a real class. Resolved the
     same way a crew agent's ``response_format`` is, so the declaration shape
-    and its failure messages match.
+    and its failure messages match -- including resolving the ref against the
+    declaration's own directory, so a flow loaded by path finds the model
+    sitting next to it rather than under whatever the cwd happens to be.
     """
     if declared is None:
         return None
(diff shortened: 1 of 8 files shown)


joaomdmoura:
@coderabbitai review

joaomdmoura:
@coderabbitai review

joaomdmoura:
@coderabbitai full review

joaomdmoura:
@coderabbitai full review

joaomdmoura:
@coderabbitai full review

lorenzejay (commented):
Tests look good for the declared contract. One real bug remains: `_resolve_router_response_format` always passes `project_root=None`, so resolution uses `Path.cwd()` instead of the declaration file's `source_dir`. Comment inline.

lib/crewai/src/crewai/experimental/conversational_mixin.py
```diff
@@ -83,25 +84,33 @@ def _iter_condition_labels(condition: Any) -> set[str]:
...
+    """
+    if declared is None:
+        return None
+
+    from crewai.project.json_loader import _resolve_model_class
+
+    return _resolve_model_class(
+        declared.model_dump(),
+        "conversational.router.response_format",
+        None,
```
lorenzejay: This is a bug, not a test gap.

`project_root=None` makes `_resolve_model_class` fall back to `Path.cwd()`. Crew loading already resolves python refs from the declaration file's directory (`source_dir`). A flow loaded with `Flow.from_declaration(path=...)` from another working directory will fail to find a project-local `response_format` even when the module sits next to the YAML.

The new tests don't catch it: `test_a_declared_python_ref_is_resolved_to_the_class` `chdir`s into `tmp_path` and loads via `contents=`, so `source_dir` is never set and cwd happens to be the project root.

Fix: pass `self._conversation_flow_definition().source_dir` (or the definition's source dir) through `_config_from_definition` / `_router_config_from_definition` into this call. Add a `path=` test that does **not** chdir.

joaomdmoura: Fixed in 95017ade — you're right, and the reproduction matches your description exactly.

`_config_from_definition` now takes the declaration's directory and threads it to `_resolve_model_class`, sourced from `FlowDefinition.source_dir` at the one call site (`_conversation_config`). That's the same directory `_actions.py:146` already uses as `base_dir` for relative path resolution, and the same thing crew loading passes as `project_root=crew_path.parent` (`json_loader.py:319`).

On the test gap: confirmed. The new `test_a_ref_resolves_next_to_the_declaration_not_the_cwd` writes `chat_routes.py` and `chat.yaml` into one directory, `chdir`s into a *different* one, and loads with `path=`. Without the fix it fails with:

```
JSONProjectError: conversational.router.response_format: Python references in JSON
configs must point to modules inside the project root; 'chat_routes' was not found
under .../test_a_ref_resolves_next_to_th0/elsewhere
```

which is the cwd fallback, not the declaration directory. It passes with the fix. Deliberately no `syspath_prepend` — `_import_project_python_reference` inserts the resolved root itself, so the test only passes if the right root is threaded through.

lib/crewai/src/crewai/experimental/conversational_mixin.py
```diff
@@ -83,25 +84,33 @@ def _iter_condition_labels(condition: Any) -> set[str]:
...
+    )
```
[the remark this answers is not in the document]
joaomdmoura: Confirmed and fixed in 95017ade — same root cause @lorenzejay raised.

`_resolve_router_response_format` now receives `FlowDefinition.source_dir` instead of `None`, so resolution anchors on the declaration file's directory rather than `Path.cwd()`. Regression test `test_a_ref_resolves_next_to_the_declaration_not_the_cwd` loads a flow with `path=` from an unrelated cwd; it fails on the old code with "'chat_routes' was not found under .../elsewhere" and passes with the fix.

lorenzejay (approved):

PR #7065, opened 2026-08-20
fix(agents): keep message roles when Agent.kickoff gets a conversation

joaomdmoura:
- Makes `Agent.kickoff` honor the roles of a multi-message conversation. It accepts `str | list[LLMMessage]`, but `_prepare_kickoff` joined every message's content into a single string, so a three-turn conversation reached the provider as two messages with the agent's own previous reply presented as something the *user* said. Measured with a recording LLM before and after; `LiteAgent.kickoff` was already correct, which is why the same list behaved differently there.
- The last message is this turn's request; the ones before it travel as `inputs["history"]` — the same way `inputs["files"]` already does — and both executors splice them in after the system prompt and before the user prompt. Memory recall still runs over the whole conversation text rather than just the last turn.
- Preserved, and this is the part that matters: a plain string and a single-message list produce byte-identical provider input to before (asserted with `==` between the two forms), which is what every existing caller passes. Verified no regression on the agent suites by comparing failing test *names* between baseline and branch — the sets are identical. That suite has a pre-existing 25↔26 wobble across runs; I checked names, not counts, for exactly that reason.
- Also widens `LiteAgentExecutionStartedEvent.messages` from `list[dict[str, str]]` to `list[LLMMessage]`. It raised a `ValidationError` for a message whose `content` was `None` or a content-part list — both valid `LLMMessage` shapes that `Agent.kickoff` already accepts, so the event could reject input the agent had taken.
- Tests: roles preserved across a three-turn conversation, history positioned between the system prompt and this turn, `Agent` and `LiteAgent` agreeing, a plain string unchanged, a single message matching the string form, empty-content messages skipped, and an empty list not crashing. `ruff` clean; `mypy` 0 errors on all four changed files.
- Covers `CrewAgentExecutor`'s history placement directly. `Agent.kickoff` builds an `AgentExecutor` unconditionally (`agent/core.py:1539`) and never reads `executor_class`, so nothing exercised the deprecated executor's twin `_setup_messages` — and since `inputs["history"]` has exactly one writer (`_prepare_kickoff`), **that half of this change is unreachable in production today.** Kept, to stop the two sites drifting, but flagged rather than left implied. All four `_append_history` sites are now mutation-verified: removing any one fails a specific test.
- Keeps a turn that *follows* the request after it. Review caught this: packing every non-request message into one `history` list spliced a trailing tool pair — or an agent's own scratch thread — above the user question it answers (`assistant tool_calls -> tool -> user`), where `LiteAgent` sends `user -> assistant -> tool`. The carried list is now cut at the request: `history` before, `trailing` after. All eight `_append_history` / `_append_trailing` sites across the two executors are mutation-verified.
- Keeps this intentionally small: no change to `LiteAgent` (already correct), no change to the declarative `agent` action's string-only guard, no `LLMMessage` type change (a `name: None` key still fails the event, but `message_to_llm_dict` strips `None` so the conversational path never emits one).
- Next: the declarative `agent` action accepting a message list, which is what makes this reachable from a YAML chat flow.

Found while tracing why a declarative conversational handler could not receive history — a Python conversational flow calling `agent.kickoff(self.build_agent_context("x"))` hit the same flattening.

---

> [!NOTE]
> **Medium Risk**
> Changes core kickoff prompt assembly for multi-message callers (behavior fix, not additive), which affects any code passing conversation history to `Agent.kickoff` or YAML agent actions; string/single-message paths are intended to stay backward compatible.
> 
> **Overview**
> **`Agent.kickoff`** no longer flattens a message list into one user string. `_prepare_kickoff` picks the **last `user` message** as this turn’s request (or the last message if there is no user), keeps prior turns as `inputs["history"]` and post-request turns as `inputs["trailing"]`, and only promotes tool/attachment-bearing messages via `_carries_payload`. **`AgentExecutor`** and **`CrewAgentExecutor`** splice history before and trailing after the formatted user prompt so chronology matches `LiteAgent` (including tool-call pairs and scratch threads after the question).
> 
> Memory recall still uses concatenated conversation text; file attachments apply only to the request message. **`LiteAgentExecutionStartedEvent.messages`** is typed as `list[LLMMessage]` so tool turns with `content: None` validate.
> 
> Declarative flows can pass a **rendered conversation**: `AgentDefinition.input` accepts a message list, flow agent actions normalize `${state.messages}` through `_normalize_agent_input`, and docs (EN/KO) describe kickoff semantics. Tests cover ordering, attachments, async kickoff, and the deprecated executor path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1998259c0a827410425c402cef0877af6a31dcd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joaomdmoura:
@coderabbitai review

joaomdmoura:
@coderabbitai review

lib/crewai/src/crewai/agent/core.py
```diff
@@ -1524,15 +1539,31 @@ def _prepare_kickoff(
...
+            # with no preceding `assistant` tool_calls, which providers reject.
+            carried = [msg for msg in messages if _carries_payload(msg)]
+            # The last message is this turn's request; the ones before it are
+            # prior turns and keep their roles. Joining them all into one
+            # string told the model the assistant's own replies were the
+            # user's, so it could not tell who said what.
+            formatted_messages = (
+                str(carried[-1].get("content") or "") if carried else ""
+            )
+            history = carried[:-1]
```
Vidit-Ostwal: This always treats `carried[-1]` as this turn’s request, regardless of role. That only holds when the list ends on the current user message.

The motivating caller does not. `build_agent_context` is `conversation_messages` plus that agent’s private thread, appended after the current user turn. On a later turn the last payload-bearing message is often an assistant scratch, so the real user question becomes `history` and the scratch gets wrapped as `{input}` / `Current Task`.

`LiteAgent` does not do this split — it prepends the system prompt and forwards the list. `test_agent_and_lite_agent_agree` only compares roles on a fixture that already ends with a user message, so it would not catch this.

If the contract is “last message is the request,” that should be documented and the conversational path should not pass `build_agent_context()` straight into `kickoff`. If the contract is “keep a conversation intact,” this needs to follow `LiteAgent` (or at least select the last `role == "user"` message) instead of promoting `carried[-1]`.

joaomdmoura:
Confirmed, and it was worse than a role mismatch. Reproduced with the exact shape `build_agent_context()` produces on a later turn:

```
context = [
    {"role": "user",      "content": "where is order 88213?"},      # the real question
    {"role": "assistant", "content": "let me look that up"},
    {"role": "assistant", "content": "internal note: checked warehouse"},  # thread
]

promoted as the request : 'internal note: checked warehouse'
left in history         : [('user', 'where is order 88213?'), ('assistant', 'let me look that up')]
```

So the agent's own scratch became `Current Task` and the user's question was demoted to history — exactly as you described.

I took your second option, the `role == "user"` selection rather than following `LiteAgent` outright, because `Agent`'s executor prompt has a `{input}` / `Current Task` slot that has to be filled with something; `LiteAgent` has no such slot, so "forward the list" isn't available to `Agent` without either leaving the task empty or duplicating a message. The contract is now:

- the request is the **last `user` message**; everything else stays as history in order
- with no `user` message at all, the last message stands in — which is what a single-message caller has always got

Documented on both `kickoff` and `kickoff_async`, since this is public API.

On the test you called out: you were right that `test_agent_and_lite_agent_agree` proved too little. Adding your fixture to it made it fail — `['system','assistant','assistant','user']` vs `['system','user','assistant','assistant']` — because with a task slot on one side and not the other, the two legitimately place the request differently. Rather than weaken the assertion I split it: `Agent`'s own behavior is now pinned directly (`test_the_request_is_the_last_user_message_not_the_last_message`, plus a no-`user`-message case), and the cross-check asserts what must hold for both — nothing dropped, nothing duplicated.

If you'd rather the conversational path stop passing `build_agent_context()` straight into `kickoff` and instead hand over an explicit request, say so and I'll do that instead — this way keeps the caller unchanged, which is why I preferred it.

lib/crewai/src/crewai/agent/core.py
Vidit-Ostwal: Resolved on [0dd5e05](https://github.com/crewAIInc/crewAI/pull/7065/commits/0dd5e05efcf60c16ea7b01f180a4e0558f2e65ea)
@[joaomdmoura](https://github.com/joaomdmoura)
joaomdmoura
commented
[3 days ago](https://github.com/crewAIInc/crewAI/pull/7065#issuecomment-5377408355)
Member
Author
Confirmed, and it was worse than a role mismatch. Reproduced with the exact shape build_agent_context() produces on a later turn:

context = [
    {"role": "user",      "content": "where is order 88213?"},      # the real question
    {"role": "assistant", "content": "let me look that up"},
    {"role": "assistant", "content": "internal note: checked warehouse"},  # thread
]

promoted as the request : 'internal note: checked warehouse'
left in history         : [('user', 'where is order 88213?'), ('assistant', 'let me look that up')]
So the agent's own scratch became Current Task and the user's question was demoted to history — exactly as you described.

I took your second option, the role == "user" selection rather than following LiteAgent outright, because Agent's executor prompt has a {input} / Current Task slot that has to be filled with something; LiteAgent has no such slot, so "forward the list" isn't available to Agent without either leaving the task empty or duplicating a message. The contract is now:

the request is the last user message; everything else stays as history in order
with no user message at all, the last message stands in — which is what a single-message caller has always got
Documented on both kickoff and kickoff_async, since this is public API.

On the test you called out: you were right that test_agent_and_lite_agent_agree proved too little. Adding your fixture to it made it fail — ['system','assistant','assistant','user'] vs ['system','user','assistant','assistant'] — because with a task slot on one side and not the other, the two legitimately place the request differently. Rather than weaken the assertion I split it: Agent's own behavior is now pinned directly (test_the_request_is_the_last_user_message_not_the_last_message, plus a no-user-message case), and the cross-check asserts what must hold for both — nothing dropped, nothing duplicated.

If you'd rather the conversational path stop passing build_agent_context() straight into kickoff and instead hand over an explicit request, say so and I'll do that instead — this way keeps the caller unchanged, which is why I preferred it.

@[joaomdmoura](https://github.com/joaomdmoura) joaomdmoura closed this [3 days ago](https://github.com/crewAIInc/crewAI/pull/7065#event-29836430138)
@[joaomdmoura](https://github.com/joaomdmoura) joaomdmoura reopened this [3 days ago](https://github.com/crewAIInc/crewAI/pull/7065#event-29836430577)
@Vidit-Ostwal
[Merge branch 'main' into fix/agent-kickoff-message-roles](https://github.com/crewAIInc/crewAI/pull/7065/commits/2db28ed1b4e64980c42cac617a9145f7da345f40)
[2db28ed](https://github.com/crewAIInc/crewAI/pull/7065/commits/2db28ed1b4e64980c42cac617a9145f7da345f40)
This branch has not been deployed
No deployments
Merge info
Review required
At least 1 approving review is required by reviewers with write access.

All checks have passed
3 skipped, 54 successful checks


Merging is blocked
New changes require approval from someone other than the last pusher. A conversation must be resolved before this pull request can be merged.
You can also merge this with the command line. 
Still in progress?
@Vidit-Ostwal


Add a comment
Comment
 

Add your comment here...
Remember, contributions to this repository should follow its [contributing guidelines](https://github.com/crewAIInc/crewAI/blob/9e9a8577becc322f98a966ad88d7904251049744/.github/CONTRIBUTING.md) and [security policy](https://github.com/crewAIInc/crewAI/blob/9e9a8577becc322f98a966ad88d7904251049744/.github/security.md).
 ProTip! Add comments to specific lines under [Files changed](https://github.com/crewAIInc/crewAI/pull/7065/files).
Reviewers
@[Vidit-Ostwal](https://github.com/Vidit-Ostwal)
Vidit-Ostwal
@coderabbitai
[coderabbitai[bot]](https://github.com/apps/coderabbitai)
@cursor
[cursor[bot]](https://github.com/apps/cursor)
At least 1 approving review is required to merge this pull request.

Still in progress?
Assignees
No one—
Labels
[llm-generated](https://github.com/crewAIInc/crewAI/issues?q=state%3Aopen%20label%3Allm-generated)
[size/L](https://github.com/crewAIInc/crewAI/issues?q=state%3Aopen%20label%3Asize%2FL)
Projects
None yet
Milestone
No milestone
Development
Successfully merging this pull request may close these issues.

None yet


Notifications
Customize
You’re ignoring this repository.
3 participants
@joaomdmoura
@Vidit-Ostwal
@theCyberTech

lib/crewai/src/crewai/agent/core.py
```diff
@@ -1524,15 +1556,31 @@ def _prepare_kickoff(
...
+            # message is promoted to it and the rest keep their roles as
+            # history. Joining them all into one string told the model the
+            # assistant's own replies were the user's.
+            request_index = _request_index(carried)
+            request = carried[request_index] if carried else None
+            formatted_messages = str(request.get("content") or "") if request else ""
+            history = [msg for i, msg in enumerate(carried) if i != request_index]
+            recall_text = "\n".join(
+                str(msg.get("content", "")) for msg in carried if msg.get("content")
             )
```
Vidit-Ostwal, line 1585: `str(request.get("content") or "")` and the `recall_text` join turn a content-part list into a Python repr. `[{"type": "text", "text": "hello"}]` becomes `"[{'type': 'text', 'text': 'hello'}]"` in Current Task and memory recall. History keeps the list, so only the promoted request (and recall) are affected.

This is the old flatten leftover, not a regression — string / single-message callers stay byte-identical as intended. Widening `LiteAgentExecutionStartedEvent.messages` to `list[LLMMessage]` also doesn't fix this; it only stops `ValidationError` on emit.

Not blocking. The path this PR unblocks (`build_agent_context`, `handle_turn`) is string content. Flagging because `LLMMessage.content` already allows the list form, and a multimodal request landing here would show the model the repr. If you flatten later, pull text out of the parts — `_message_content_text` currently does the same `str()`.

lib/crewai/src/crewai/agents/crew_agent_executor.py
```diff
@@ -167,6 +167,11 @@ def use_stop_words(self) -> bool:
             self.llm.supports_stop_words() if isinstance(self.llm, BaseLLM) else False
         )
 
+    def _append_history(self, inputs: dict[str, Any]) -> None:
+        """Add prior turns, with their roles, before this turn's request."""
+        for message in inputs.get("history", []):
+            self.messages.append(dict(message))
```
Vidit-Ostwal, line 173: `_append_history` is the same helper on both executors (here and `experimental/agent_executor.py`), including the `use_system_prompt=False` comment.

Not asking to extract a shared helper — the two `_setup_messages` implementations were already twins, so this matches the existing split. Only noting it so the next history-placement change doesn't land on one site and miss the other. That's how the no-system-prompt branch got dropped the first time.

Vidit-Ostwal (commented):
Two non-blocking notes. Roles, last-user promotion, tool-call retention, and request-only files look correct.

lib/crewai/src/crewai/agent/core.py
joaomdmoura: Verified, and **deferring** — with the reasoning written down rather than just agreeing.

Your read is right on both counts. `str()` on a content-part list yields the Python repr, and it is pre-existing, not something this PR introduces: `origin/main` already does `str(msg.get("content", ""))` at `agent/core.py:1531` and again at `:1742`. The change here narrows *which* message reaches that `str()` (the last `user` turn instead of a join of all of them); it does not change the flattening itself. `_message_content_text` (`utilities/agent_utils.py:847`) is the third site with the same behavior.

Deferring rather than fixing, for two reasons:

1. It's a behavior change to the multimodal request path, not a role-preservation fix. Doing it properly means deciding what `{input}` / `Current Task` should contain for a parts list — text parts joined? images dropped, or moved to `files`? — and that decision should land with tests over the provider matrix, not as a drive-by in this diff.
2. All three sites should change together. Fixing only the one this PR happens to touch leaves the same repr in memory recall and in the token-count path, which is exactly the split-site failure mode you flag in the other thread.

@joaomdmoura this needs a Linear ticket before it drops off — "promote multimodal content parts instead of `str()`-ing them", covering `agent/core.py:1531`, `:1742`, and `_message_content_text`. Flagging that I have not opened one myself.

Agreed it isn't blocking here: `build_agent_context` / `handle_turn` are string content, and string / single-message callers stay byte-identical.

lib/crewai/src/crewai/agents/crew_agent_executor.py
joaomdmoura: Noted, and taking it as written — keeping the two `_setup_messages` twins rather than extracting a shared helper, since that matches the existing split and an extraction would widen this diff into both executors for no behavior gain.

Confirming the specific risk you name is covered *now*: both sites got the history insertion in the same commit (3c867d3), and the no-system-prompt path is pinned by tests that assert provider-visible message order with `use_system_prompt=False` — parameterized across `CrewAgentExecutor` and the experimental executor, so a change landing on one site and missing the other fails. That's the containment for the failure mode that produced this the first time; the duplication itself stays.

lib/crewai/src/crewai/agent/core.py
```diff
@@ -1524,15 +1556,31 @@ def _prepare_kickoff(
...
+            # with no preceding `assistant` tool_calls, which providers reject.
+            carried = [msg for msg in messages if _carries_payload(msg)]
+            # The executor's prompt needs one request string, so exactly one
+            # message is promoted to it and the rest keep their roles as
+            # history. Joining them all into one string told the model the
+            # assistant's own replies were the user's.
+            request_index = _request_index(carried)
+            request = carried[request_index] if carried else None
+            formatted_messages = str(request.get("content") or "") if request else ""
+            history = [msg for i, msg in enumerate(carried) if i != request_index]
```
lorenzejay: Last-user as `{input}` is the right call for the executor prompt. The leftover is that everything else — including the turns *after* that user message — is packed into `history`, and `_append_history` always splices that list *before* the current user prompt.

A conversation that ends on a tool result:

```
user: what is the weather in Lisbon?
assistant: tool_calls get_weather
tool: 22C, clear
```

becomes this at the provider:

```
assistant  tool_calls
tool       22C, clear
user       Current Task: what is the weather in Lisbon?
```

`LiteAgent` still sends `user → assistant → tool`. `test_a_tool_call_sequence_survives` does not catch this because it ends on a follow-up user line (`and tomorrow?`), so the tool pair stays before the request.

Live calls on `gpt-4o-mini` and `gpt-5.6-sol` both accepted the reordered payload and answered from the tool result, so this is an order bug, not a hard provider 400 on those models. A stricter provider, or a later turn that needs the question before the tool pair, still gets the chronology wrong.

Split history into before-request and after-request, and append the after-request slice *after* the current user prompt. Last-user as `{input}` can stay.

lorenzejay (commented):
One leftover on last-user promotion: messages after that user turn (a tool pair) are packed into `history` and spliced before the request, so chronology flips. Comment inline with live `gpt-4o-mini` / `gpt-5.6-sol` results.

lib/crewai/src/crewai/agents/crew_agent_executor.py
joaomdmoura: Correcting my previous reply — I claimed the no-system-prompt path was "parameterized across `CrewAgentExecutor` and the experimental executor." That was wrong, and checking it properly turned up something worth knowing.

`test_history_survives_with_the_system_prompt_disabled` drove `Agent.kickoff`, and `_prepare_kickoff` constructs `AgentExecutor(...)` unconditionally (`agent/core.py:1539`) — it never reads `executor_class`. So that test only ever exercised the **experimental** executor. `CrewAgentExecutor`'s two `_append_history` sites had no coverage at all.

Worse for your point: `inputs["history"]` has exactly one writer, `_prepare_kickoff` at `agent/core.py:1638`. Since that function always builds an `AgentExecutor`, **the `CrewAgentExecutor` half of this PR is currently unreachable in production.** I kept it — it keeps the twins in sync, which is the thing you're asking for — but it is guarding a path nothing walks today, and that's worth saying out loud rather than leaving implied.

Added in 00cabe25: `test_the_deprecated_executor_places_history_the_same_way`, parameterized over both branches, driving `CrewAgentExecutor._setup_messages` directly with real `SystemPromptResult` / `StandardPromptResult` values. (A plain dict gets coerced to `SystemPromptResult`, so `"system" in self.prompt` is always true and the `elif` branch is never reached — my first attempt passed for that reason and caught nothing.)

Verified by mutation, one site at a time — all four now fail when removed:

| removed | failing test |
|---|---|
| `CrewAgentExecutor` system branch | `..._deprecated_executor_...[True]` |
| `CrewAgentExecutor` no-system branch | `..._deprecated_executor_...[False]` |
| experimental system branch | 6 tests |
| experimental no-system branch | `test_history_survives_with_the_system_prompt_disabled` |

@joaomdmoura two calls for you: whether `Agent.kickoff` ignoring `executor_class` is intended, and whether the unreachable `CrewAgentExecutor` change should stay or be dropped. Both are outside this PR's scope; I have not changed either.

lib/crewai/src/crewai/agent/core.py
joaomdmoura: Confirmed and fixed in f9f7fe83. Reproduced exactly as you described before changing anything — same three-message conversation, recording LLM, both agents:

```
before   Agent -> ['assistant', 'tool', 'user']
         LiteAgent -> ['user', 'assistant', 'tool', ...]
after    Agent -> ['user', 'assistant', 'tool']
```

The fix is the split you proposed. `_prepare_kickoff` now cuts the carried list at the request: what precedes it stays `inputs["history"]`, what follows travels as `inputs["trailing"]`, and both executors append the trailing slice after the user prompt. Last-user as `{input}` is unchanged.

Two things your description didn't cover, which showed up while fixing it:

- It is not only the tool case. `TRAILING_THREAD` — the `build_agent_context()` shape this PR was written for — hit the same reordering: an agent's own scratch turns were hoisted above the user question. `test_the_request_is_the_last_user_message_not_the_last_message` was asserting the reordered output as if it were correct, so it had to be corrected, not just extended.
- It affects the no-system-prompt branch too, so the fix lands at four sites across the two executors.

You're right that `test_a_tool_call_sequence_survives` couldn't catch it — ending on `and tomorrow?` keeps the pair before the request. Added `test_a_conversation_ending_on_a_tool_result_keeps_its_order`, which uses your exact conversation and also asserts `Agent` and `LiteAgent` now agree on role order.

All eight `_append_history` / `_append_trailing` sites are mutation-verified — removing any one fails a specific test. That check is what caught the last uncovered site (experimental executor, no-system-prompt branch, trailing), which had no test until `test_trailing_turns_survive_with_the_system_prompt_disabled`.

Not verified: I have not made live provider calls. I'm taking your gpt-4o-mini / gpt-5.6-sol result as the evidence that this was an ordering bug rather than a hard 400.

lorenzejay (approved):

lorenzejay (approved):

lorenzejay (approved):

PR #7066, opened 2026-08-20
feat(flow): let a declarative agent action receive the conversation

joaomdmoura:
- Lets a declarative chat handler hand its agent the conversation instead of a single string. Two guards blocked it: `AgentDefinition.input` was `str` with a validator rejecting anything else, and `AgentAction.run` raised `"agent input must render to a string"` once a CEL template rendered to a list.
- No new CEL machinery needed. A whole-string `${...}` template keeps its evaluated type, so `input: "${state.messages.map(m, {'role': m.role, 'content': m.content})}"` renders exactly `[{role, content}, ...]` — verified against the real evaluator. `input` now accepts a string or a list of mappings, and the action normalizes rather than rejecting.
- Verified live end to end on a YAML chat flow with a real LLM: turn 1 "My order id is 9931", turn 2 "Where is that order?" — the agent answered with `9931`, i.e. the earlier turn reached it with roles intact.
- Serialized messages carry `name: None` and `metadata` that the agent event schema rejects, and `message_to_llm_dict` only drops `None` for a *model* input, not the plain dicts a CEL render produces. The normalizer drops them, keeping `content: None` since that is a valid message shape (a tool-call turn). Found by a test failing, not by inspection.
- Preserved: a string `input` behaves exactly as before; a shape that is neither raises a message naming both accepted forms. **Declarative crews are unaffected** — they use `CrewAgentDefinition`, which has no `input` field at all; `AgentDefinition` is referenced only by `FlowAgentActionDefinition.with_`, `load_agent_from_definition` and the authoring skill (grepped across `lib/`). `AgentDefinition` is re-exported from `crewai.project`, so widening a required field's type is additive.
- Tests: a rendered message list passing through, serialized metadata being stripped, a string still accepted, a non-message shape rejected with its message, and the contract accepting/rejecting at validation time. `ruff` clean; `mypy` 0 errors on both changed files. Remaining suite failures are the pre-existing agent (25-26, wobbles) and multimodal (7) sets — verified by failing-test *names*, not counts.
- Keeps this intentionally small: no `messages` field alongside `input`, no CEL helper function, no docs change yet.
- Next: document the message-list form in the conversational guide (en + ar/ko/pt-BR).

Stacked on #7065, which makes `Agent.kickoff` keep message roles — without it this would deliver a flattened conversation.

---

> [!NOTE]
> **Low Risk**
> Additive widening of flow-only `AgentDefinition.input` with normalization before existing `kickoff_async`; string prompts unchanged and declarative crews are unaffected.
> 
> **Overview**
> Declarative **flow agent actions** can pass a full conversation into `kickoff_async`, not only a single string prompt.
> 
> **`AgentDefinition.input`** now accepts a string or a list of message mappings (with updated validation and schema examples). **`AgentAction.run`** no longer requires rendered input to be a string; it routes through **`_normalize_agent_input`**, which passes strings through unchanged, converts CEL-rendered message lists via `message_to_llm_dict`, and strips `None` metadata keys the agent event schema rejects while preserving valid shapes like `content: None` on tool-call turns.
> 
> String `input` behavior is unchanged; invalid rendered types get a clearer error. Tests cover normalization edge cases and an end-to-end YAML flow with `input: "${state.messages}"`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 650985fed7296d134424da283595244d52a54c76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/crewai/src/crewai/flow/runtime/_actions.py
+++ b/lib/crewai/src/crewai/flow/runtime/_actions.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import ast
 import asyncio
-from collections.abc import Awaitable, Callable
+from collections.abc import Awaitable, Callable, Mapping
 import contextvars
 import inspect
 import os
@@ -29,6 +29,7 @@ from crewai.utilities.declarative_refs import (
     resolve_class_ref,
     resolve_ref,
 )
+from crewai.utilities.types import LLMMessage
 
 
 if TYPE_CHECKING:
@@ -184,8 +185,7 @@ class AgentAction:
             self.flow,
             local_context=local_context,
         ).render_template()
-        if not isinstance(rendered_input, str):
-            raise ValueError("agent input must render to a string")
+        agent_input = _normalize_agent_input(rendered_input)
 
         agent, response_format = await asyncio.to_thread(
             load_agent_from_definition,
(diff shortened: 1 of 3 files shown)


joaomdmoura:
@coderabbitai review

joaomdmoura:
@coderabbitai review

joaomdmoura:
@coderabbitai full review

joaomdmoura:
@coderabbitai full review

joaomdmoura:
@coderabbitai full review

lorenzejay (approved):

PR #7067, opened 2026-08-20
[OSS-137] Report MCP HTTP auth failures instead of cancelled connections

Vidit-Ostwal:
- Add `crewai.mcp.exceptions` with typed connection errors (`MCPAuthenticationError`, `MCPHTTPError`), HTTP status recovery from exception groups/`CancelledError` chains, and shared helpers (`raise_connection_failure`, `tool_execution_error_type`).
- Raise typed errors from `HTTPTransport.connect()` when the MCP server responds with an HTTP status, instead of a generic `ConnectionError`.
- Fix `HTTPTransport.disconnect()` to surface auth/HTTP failures during teardown instead of suppressing exception groups that carry the real status.
- Fix `MCPClient.connect()` to inspect transport cleanup before emitting `error_type="cancelled"`; auth failures now emit `error_type="authentication"` with the observed status.
- Add optional `status_code` to `MCPConnectionFailedEvent` and show it in verbose console output.
- Replace speculative cancel-scope wording in `MCPToolResolver` with classifier-driven errors.
- Fix `call_tool_result` to emit `error_type="authentication"` for auth failures instead of `server_error`.
- Add 72 MCP tests covering the classifier, transport, client connect, tool resolver, and tool execution paths.

Fixes OSS-137.


lib/crewai/src/crewai/mcp/transports/http.py
```diff
@@ -58,105 +69,116 @@ def transport_type(self) -> TransportType:
...
+            return None
+
+        exc_type = type(exc) if exc is not None else None
+        exc_tb = exc.__traceback__ if exc is not None else None
+
+        try:
+            with contextlib.suppress(asyncio.CancelledError):
+                await asyncio.sleep(0.1)
+            await context.__aexit__(exc_type, exc, exc_tb)
+        except asyncio.CancelledError as cancel_exc:
```
lucasgomide, line 90: This path swallows cancellation during disconnect(), right? Please propagate CancelledError after cleanup

lib/crewai/src/crewai/mcp/tool_resolver.py
```diff
@@ -350,41 +368,54 @@ async def _setup_client_and_list_tools() -> list[dict[str, Any]]:
...
+            except (asyncio.CancelledError, Exception) as e:
                 if discovery_client.connected:
-                    await discovery_client.disconnect()
-                    await asyncio.sleep(0.1)
+                    try:
+                        await discovery_client.disconnect()
+                        await asyncio.sleep(0.1)
+                    except Exception as disconnect_error:
+                        if isinstance(disconnect_error, MCPConnectionError):
+                            raise disconnect_error from e
```
lucasgomide, line 378: from my readig, MCPClient.disconnect() wraps MCPAuthenticationError in RuntimeError, so this check never receives the typed error. Would be great  preserve MCPConnectionError

lib/crewai/src/crewai/mcp/transports/http.py
Vidit-Ostwal: Resolved in `a6ffa2c72`.

You were right: `_close_transport_context()` was returning `CancelledError` to `disconnect()`, but `_raise_teardown_failure()` logged and returned instead of re-raising when no HTTP status was found.

**Fix:** After auth/HTTP classification, bare teardown `CancelledError` is now re-raised so normal disconnect cancellation propagates after cleanup completes.

**Test:** `test_http_transport_disconnect_propagates_cancellation_without_http_status` — verifies `__aexit__` runs, context is cleared, and `CancelledError` is not swallowed.

lib/crewai/src/crewai/mcp/tool_resolver.py
Vidit-Ostwal: Resolved in `c39fad9fa`.

Good catch — `MCPClient.disconnect()` was wrapping every `Exception` (including `MCPAuthenticationError`) in `RuntimeError`, so the resolver's `isinstance(disconnect_error, MCPConnectionError)` path never fired on the real client.

**Fix:** `disconnect()` now re-raises `MCPConnectionError` and `CancelledError` from `_exit_stack.aclose()` unchanged; only unexpected errors are wrapped in `RuntimeError`.

**Test:** `test_disconnect_preserves_authentication_error_from_teardown` exercises the real client path (not a mocked `disconnect`), so auth errors from transport teardown reach the resolver cleanup handler as typed `MCPAuthenticationError`.

lucasgomide (approved):

lucasgomide (approved):

PR #7077, opened 2026-08-21
fix(llms): raise Anthropic default max_tokens so large tool calls survive

Vidit-Ostwal:
## Summary
- Native Anthropic defaulted `max_tokens` to 4096, so large tool-call JSON (full inline agent/crew specs) was truncated mid-stream and retries hit the same ceiling.
- Default is now 32000 via `DEFAULT_MAX_TOKENS`; an explicit `max_tokens` still wins.
- Bedrock/Azure are unchanged (no 32000 default) so smaller-cap models do not 400.

## Test plan
- [x] `test_anthropic_defaults_max_tokens_to_32000` — default is 32000 on the LLM and in API params
- [x] `test_anthropic_honors_explicit_max_tokens` — explicit cap is forwarded
- [ ] Confirm Bedrock/Azure still omit `max_tokens` unless the caller sets one


lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -46,6 +46,10 @@
 
 ANTHROPIC_FILES_API_BETA: Final = "files-api-2025-04-14"
 ANTHROPIC_STRUCTURED_OUTPUTS_BETA: Final = "structured-outputs-2025-11-13"
+# Anthropic requires max_tokens. 4096 truncates large tool calls (e.g. a
+# full inline agent/crew spec) mid-JSON; trailing fields then fail validation
+# and every retry hits the same ceiling.
+DEFAULT_MAX_TOKENS: Final[int] = 32000
```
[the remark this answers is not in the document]
Vidit-Ostwal: Addressed in 07dd76538.

`max_tokens` is now model-aware via `_default_max_tokens_for_model`. The default model is `claude-sonnet-4-6` (128000). Retired 3.5 Sonnet is no longer the default or in the catalogs. Unknown IDs fall back to 32000.

Also added the `to_config_dict()` coverage: defaults are omitted, an explicit `max_tokens=8000` is kept.

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
```diff
@@ -46,15 +46,39 @@
...
+    ("claude-opus-4-8", 128000),
+    ("claude-opus-4-7", 128000),
+    ("claude-opus-4-6", 128000),
+    ("claude-sonnet-4-6", 128000),
+    ("claude-haiku-4-5", 64000),
+    ("claude-sonnet-4-5", 64000),
+    ("claude-opus-4-5", 64000),
+)
+_DEFAULT_MODEL_MAX_TOKENS: Final[int] = 32000
+DEFAULT_MODEL: Final[str] = "claude-sonnet-4-6"
```
lorenzejay, line 64: love it!

lorenzejay (approved):

lib/crewai/src/crewai/llms/providers/anthropic/completion.py
Vidit-Ostwal: lessgo.

PR #7084, opened 2026-08-22
fix(memory): normalize aware datetimes to naive-UTC to prevent TypeError crashes in recall/forget

ksk2023:
## Description

The unified memory module mixed **offset-naive** and **offset-aware** datetimes, causing `TypeError: can't subtract/compare offset-naive and offset-aware datetimes` crashes on several real paths.

### Crash paths verified on `main`

1. **Recall scoring crash**: `MemoryRecord(content="x", created_at=datetime.now(timezone.utc))` — the recommended modern pattern — makes `compute_composite_score()` raise `TypeError` at `types.py:364` (`datetime.utcnow() - record.created_at`), killing every recall that surfaces this record.

2. **Permanent record poisoning**: both LanceDB (`lancedb_storage.py:271`) and Qdrant Edge (`qdrant_edge_storage.py:227`) `_parse_dt()` parse Z-suffixed ISO strings (`"...Z"` → `"+00:00"`) into **aware** datetimes. Once such a value is stored (external tools, migrations, cross-backend payloads), every subsequent read returns an aware `created_at` — so recall crashes *permanently* for that record until it is manually deleted.

3. **forget() crash**: `Memory.forget(older_than=datetime.now(timezone.utc), categories=[...])` compares naive stored `created_at` against the aware cutoff at `lancedb_storage.py:440` / `qdrant_edge_storage.py:471` → `TypeError`.

4. **LLM time_filter breaks recall**: `recall_flow.py:224` parses the LLM-provided `time_filter` with plain `fromisoformat()`. LLMs routinely emit a trailing `Z`: that **raises `ValueError` on Python < 3.11** (filter silently dropped) and yields an **aware** datetime on 3.11+ — which then breaks the `created_at >= time_cutoff` comparison at `recall_flow.py:108`.

5. **Deprecation**: `datetime.utcnow()` is deprecated since Python 3.12 and was used at 9 call sites (visible as `DeprecationWarning` in test runs; CrewAI supports `>=3.10,<3.14`).

## Fix strategy

The module's canonical **stored** format has always been naive-UTC (LanceDB rows, Qdrant payloads, and the SQL string-range comparisons in `delete()` all rely on it). So rather than switching the stored format — which would silently break existing databases and string comparisons — this PR **normalizes at the boundaries**:

- `types.py`: new `_utc_now()` / `_normalize_dt()` helpers; `MemoryRecord` gains a `field_validator` that converts aware `created_at`/`last_accessed` to naive-UTC; `compute_composite_score()` uses `_utc_now()`.
- `lancedb_storage.py` / `qdrant_edge_storage.py`: `_parse_dt()` normalizes aware values (including Z-suffixed strings) to naive-UTC.
- `recall_flow.py`: `time_filter` parsing handles the `Z` suffix (3.10-compatible) and normalizes.
- `encoding_flow.py` / `unified_memory.py` / both storages: replace deprecated `utcnow()` with `_utc_now()`.
- `Memory.forget()` normalizes an aware `older_than` before delegating to the storage backend.

## Repro (before this PR)

```python
from datetime import datetime, timezone
from crewai.memory.types import MemoryRecord, MemoryConfig, compute_composite_score

rec = MemoryRecord(content="x", created_at=datetime.now(timezone.utc))
compute_composite_score(rec, 0.9, MemoryConfig())
# TypeError: can't subtract offset-naive and offset-aware datetimes
```

## Tests

Adds 5 regression tests to `tests/memory/test_unified_memory.py`:

- aware `created_at`/`last_accessed` normalized; scoring succeeds
- non-UTC offset (+08:00) converted to UTC, not just stripped
- LanceDB `_row_to_record` parses Z-suffixed rows to naive-UTC
- `Memory.forget(older_than=aware)` on the categories path deletes without raising (end-to-end, real LanceDB)
- Z-suffix `time_filter` string parses to naive-UTC

Full memory suite: **133 passed, 19 skipped**, no regressions (6 pre-existing Windows tmp-dir teardown errors, unrelated and present on `main`).

## Checklist

- [x] I have added tests that prove my fix is effective
- [x] I have run `pytest tests/memory/` locally
- [x] Changes are backward compatible (naive inputs behave exactly as before; aware inputs no longer crash)

### DIFF
--- a/lib/crewai/src/crewai/memory/encoding_flow.py
+++ b/lib/crewai/src/crewai/memory/encoding_flow.py
@@ -27,7 +27,7 @@ from crewai.memory.analyze import (
     analyze_for_consolidation,
     analyze_for_save,
 )
-from crewai.memory.types import MemoryConfig, MemoryRecord, embed_texts
+from crewai.memory.types import MemoryConfig, MemoryRecord, _utc_now, embed_texts
 from crewai.memory.utils import join_scope_paths
 
 
@@ -378,7 +378,7 @@ class EncodingFlow(Flow[EncodingState]):
         conflicts from two operations targeting the same record.
         """
         items = list(self.state.items)
-        now = datetime.utcnow()
+        now = _utc_now()
 
         # Multiple items may reference the same existing record (because their
         # similar_records overlap). Collect one action per record_id, first wins.
--- a/lib/crewai/src/crewai/memory/recall_flow.py
+++ b/lib/crewai/src/crewai/memory/recall_flow.py
@@ -26,6 +26,7 @@ from crewai.memory.types import (
     MemoryConfig,
     MemoryMatch,
     MemoryRecord,
+    _normalize_dt,
     compute_composite_score,
     embed_texts,
 )
@@ -221,8 +222,15 @@ class RecallFlow(Flow[RecallState]):
 
             if analysis.time_filter:
                 try:
-                    self.state.time_cutoff = datetime.fromisoformat(
-                        analysis.time_filter
+                    # The LLM may emit a trailing "Z" (ISO 8601), which
+                    # ``fromisoformat`` rejects on Python < 3.11 and parses
+                    # to an *aware* datetime on 3.11+. Normalize both cases
+                    # to naive-UTC so the later ``created_at >= time_cutoff``
+                    # comparison never mixes naive and aware values.
+                    self.state.time_cutoff = _normalize_dt(
+                        datetime.fromisoformat(
+                            analysis.time_filter.replace("Z", "+00:00")
+                        )
                     )
                 except ValueError:
                     pass
--- a/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
+++ b/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
@@ -552,6 +552,9 @@ class LanceDBStorage:
                     else created
                 )
                 if isinstance(dt, datetime):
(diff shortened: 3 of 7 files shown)


ksk2023:
Addressed the review feedback in 066d0ce:

1. **`mode="before"` bypass (types.py)** — Switched `_normalize_datetime_fields` to a `mode="after"` validator taking a `datetime`. Pydantic now parses raw ISO strings first (Z suffix → aware), and the validator normalizes the parsed value, closing the bypass path.

2. **`get_scope_info()` missing normalization (both storages)** — Parsed `created_at` values now go through `_normalize_dt()` before oldest/newest comparisons and before constructing `ScopeInfo`, so mixed naive/Z-suffixed scopes no longer raise `TypeError` and the returned bounds honor the naive-UTC contract.

3. **Test coverage for changed integration paths** — Added:
   - `test_memory_record_z_suffix_string_normalized_via_after_validator` (direct Z-suffix string through `MemoryRecord`)
   - `test_lancedb_get_scope_info_mixed_timestamp_formats` (one naive + one Z-suffixed record in the same scope, asserts naive-UTC bounds and no `TypeError`)
   - `test_qdrant_payload_to_record_z_suffix_normalized` (`_payload_to_record` payload coverage)

All new tests pass; the full `tests/memory/` suite shows 154 passed with the same pre-existing Windows-only environment failures as on the base branch (tmp-dir teardown races and `os.kill(pid, 0)` raising `OSError` on Windows instead of `ProcessLookupError`), unrelated to this change.

PR #7090, opened 2026-08-23
fix(tools): handle nested event loops when running async tools in sync context

SparshM8:
This PR fixes issue #6611 where async tools were failing when executed from within a running event loop (e.g., when using native function calling).

### Changes
- Updated `BaseTool.run`, `CrewStructuredTool.invoke`, and `MCPNativeTool._run` to detect if an event loop is already running.
- If a loop is running, the coroutine is executed in a separate thread using `ThreadPoolExecutor` to avoid `RuntimeError: asyncio.run() cannot be called from a running event loop`.
- Preserved existing behavior for non-running loop contexts.

### Validation
- Verified with a reproduction script covering `BaseTool` and `CrewStructuredTool` with async implementations.
- Local tests pass.

### DIFF
--- a/lib/crewai/src/crewai/tools/base_tool.py
+++ b/lib/crewai/src/crewai/tools/base_tool.py
@@ -339,13 +339,14 @@ class BaseTool(BaseModel, ABC):
         if asyncio.iscoroutine(result):
             try:
                 asyncio.get_running_loop()
-                import contextvars
-                from concurrent.futures import ThreadPoolExecutor
-                ctx = contextvars.copy_context()
-                with ThreadPoolExecutor(max_workers=1) as executor:
-                    return executor.submit(ctx.run, asyncio.run, result).result()
             except RuntimeError:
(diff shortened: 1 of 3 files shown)


SparshM8:
I've updated the implementation to ensure that `contextvars` are correctly propagated when running coroutines in the `ThreadPoolExecutor`. This aligns with existing patterns in the codebase (e.g., `agent_card.py`) and ensures that execution context is preserved across threads.

Vidit-Ostwal (commented):
Closing in favor of #7236. That PR covers the same `BaseTool` / `CrewStructuredTool` running-loop path, adds `MCPToolWrapper`, and has tests. `mcp_native_tool.py` is already guarded on `main`. Follow along on #7236.

Vidit-Ostwal:
Duplicate of #7236. That PR covers the reported files with tests; `mcp_native_tool.py` is already guarded on `main`. Follow along there.

PR #7094, opened 2026-08-24
fix(a2a): use constant-time comparison for server token/API-key auth

pacocartones:
`SimpleTokenAuth` and `APIKeyServerAuth` in `lib/crewai/src/crewai/a2a/auth/server_schemes.py` validate attacker-controlled bearer tokens / API keys from incoming A2A requests. Both used a plain `!=` string comparison, which short-circuits on the first differing byte and leaks the secret's length and matching prefix through response timing — a classic timing side-channel against authentication. `A2AConfig.auth` defaults to `SimpleTokenAuth`.

This replaces the comparisons with `hmac.compare_digest`, the standard constant-time primitive. Behavior is unchanged for valid/invalid inputs; only the timing profile is fixed.

Added `tests/a2a/auth/test_server_schemes_constant_time.py` covering both schemes. Full `tests/a2a` suite passes (103).

_Note: this PR was authored with AI assistance; applying the `llm-generated` label per CONTRIBUTING.md._

### DIFF
--- a/lib/crewai/src/crewai/a2a/auth/server_schemes.py
+++ b/lib/crewai/src/crewai/a2a/auth/server_schemes.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+import hmac
 import logging
 import os
 from typing import TYPE_CHECKING, Annotated, Any, ClassVar, Literal
@@ -174,7 +175,7 @@ class SimpleTokenAuth(ServerAuthScheme):
                 detail="Authentication not configured",
             )
 
-        if token != expected:
+        if not hmac.compare_digest(token, expected):
             raise HTTPException(
                 status_code=HTTP_401_UNAUTHORIZED,
                 detail="Invalid or missing authentication credentials",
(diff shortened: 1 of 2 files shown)


dineshyadav03 (approved):
Read through `server_schemes.py`: `SimpleTokenAuth.authenticate` and `APIKeyServerAuth.authenticate` both swap `!=` for `hmac.compare_digest(token, expected)`. That's the right primitive — it's a C-level constant-time byte comparison, not a Python-level one that could get JIT/branch-predicted away. I also grepped `client_schemes.py` and `utils.py` on main for other secret comparisons and found none, so this covers all the server-side comparison sites in the module, not just a subset.

One nuance worth a code comment rather than a blocker: `compare_digest` still short-circuits on a length mismatch before the constant-time walk, so it doesn't fully close a length side-channel for an attacker with very precise timing — that's Python's documented, accepted trade-off, but readers of this diff might assume the fix is airtight.

The new tests assert correctness (valid/invalid tokens) and that `compare_digest` is actually invoked via monkeypatch spy, rather than measuring timing directly — sensible, since timing assertions are flaky, but worth noting in the test docstring so nobody "improves" it into a flaky benchmark later.

Approve — correct, minimal, and covers the module.

pacocartones:
Rebased onto current `main` (the branch was 21 commits behind); the diff is unchanged.

On CONTRIBUTING.md: this PR was authored with AI assistance and therefore needs the `llm-generated` label. I do not have permission to apply labels on this repository (the API returns 403 "Must have admin rights to Repository"), so could a maintainer please add it? Flagging it here so the PR is not treated as an undisclosed AI contribution.

PR #7108, opened 2026-08-25
chore(ci): ignore unpatched chromadb HTTP-server GHSAs

Vidit-Ostwal:
- Ignore GHSA-2wm9-hf6c-p5cr, GHSA-36p7-vc44-83pf, and GHSA-xph7-9rjv-w5fr in pip-audit (workflow + pre-commit).
- These are Chroma Python HTTP server issues; CrewAI only uses embedded PersistentClient.
- No patched PyPI release exists yet (latest is 1.5.9). Drop the ignores when a fixed version ships.


theCyberTech (approved):
Approved

PR #7003, opened 2026-08-16
docs: refresh retired Gemini model ids

a-yeyang:
## Summary
- Replaced the retired Gemini model IDs in the six affected live English documentation pages with `gemini-3.7-flash`.
- Kept each example's existing provider prefix (`gemini/` or `openai/`) intact.
- Synced the same identifier-only edits to active Arabic, Korean, and Brazilian Portuguese counterparts where those pages exist. The Korean and Brazilian Portuguese LiteLLM-removal pages do not currently exist, so no new translation pages were created.

## Validation
- Ran a static MDX regression check over all 22 changed pages: no retired IDs remain and every changed file has balanced fenced code blocks.
- Ran `git diff --check` successfully.
- Attempted `npx --yes mint@4.2.741 broken-links`, but Mintlify does not support this environment's Node.js 26.5.0 (it requires an LTS Node version). The repository CI's documentation broken-link workflow will run the canonical check on Node.js 22.

Fixes #7002

PR #7110, opened 2026-08-25
fix(llms): warn when a response is truncated by the token cap

### LINKED ISSUE #7013: [BUG] Truncated responses are detected in the Bedrock provider only; other providers accept them silently

JoaquinDG:
> **AI-assisted contribution.** This was written with Claude Code and reviewed by me before
> opening. Per `CONTRIBUTING.md` this needs the `llm-generated` label; I do not have triage
> permission here, so could a maintainer please apply it.

Closes #7013. Opening this at @Vidit-Ostwal's suggestion on that thread.

## What this changes

Only the Bedrock provider checked whether a response stopped because it hit the token cap:

```python
# llms/providers/bedrock/completion.py:683
if stop_reason == "max_tokens":
    logging.warning("Response truncated due to max_tokens limit")
```

No other provider had an equivalent branch, so a cut-off response was handed back as though
it were complete.

This adds two helpers to `_finish_reason_utils.py`, which already centralises finish-reason
extraction, and calls the warning from the eight sites where `finish_reason` is already
bound:

- `anthropic/completion.py:1010`, `1558`
- `azure/completion.py:864`
- `gemini/completion.py:847`
- `openai/completion.py:983`, `1130`, `1935`, `2363`

Bedrock is deliberately untouched, since it already has its own check.

`is_truncated()` normalises the three spellings in one place (`length` on OpenAI and Azure,
`max_tokens` on Anthropic and Bedrock, `MAX_TOKENS` on Gemini), compared case-insensitively
with separators stripped. The warning names the model and the current cap, so the reader
knows which knob to turn:

```
Response truncated due to max_tokens limit (finish_reason='length',
model=gpt-4o-mini, max_tokens=16). The output is incomplete;
consider increasing max_tokens.
```

## Verification

| check | result |
|---|---|
| Full suite, clean `main` (`4e0b2e2`) | 5043 passed, 44 skipped |
| Full suite, this branch | **5061 passed, 44 skipped** |
| Difference | exactly +18, the new tests; no regressions |
| `ruff check` on changed files | All checks passed |
| `ruff format --check` | clean |

Both suites were run today against the same commit, so the comparison is measured rather
than assumed.

The 18 tests pin behaviour rather than wording: every provider spelling is recognised,
`stop` and `tool_calls` are not, the warning names the cap, and a complete response stays
silent. Worth mentioning that these are also the first tests covering the truncation path at
all, since the existing Bedrock check has none.

## Deliberately left out

Two ideas came up on #7013 that I have not built here, because you described the helper plus
a warning as the right first step and I would rather not widen the diff past that:

1. **Unknown finish reasons.** @mragnii noted that some `openai_compatible` endpoints return
   `stop` even when the upstream model truncated. This predicate returns `False` there,
   which is the honest answer for an unrecognised value but not a complete one. A tri-state,
   or surfacing unknown reasons as an error the way some other runtimes do, would be a
   larger change to the return contract.
2. **Structural surfacing.** Whether truncation should also appear on `ModelResponse` or the
   run event, so callers can react without scraping logs, is a design question I did not
   want to answer on your behalf.

Happy to follow up on either, in whatever shape you prefer.

## Manual check

Beyond the suite, a one-agent crew at `max_tokens=16` now warns on both providers I could
test directly:

```
WARNING root: Response truncated due to max_tokens limit (finish_reason='length',
              model=gpt-4o-mini, max_tokens=16) ...
WARNING root: Response truncated due to max_tokens limit (finish_reason='max_tokens',
              model=claude-haiku-4-5-20251001, max_tokens=16) ...
```

Before this change both runs produced the truncated text as the task result with zero
warnings logged.


JoaquinDG:
Pushed two changes and rebased onto current `main` (`7c23857`), so the branch is no longer
behind.

**CodeRabbit caught a real bug, and it was a good catch.** The warning was reporting
`self.max_tokens` on every provider, but that is not the cap most of them actually send:

| provider | cap sent on the wire | previously reported |
|---|---|---|
| anthropic | `max_tokens` | correct |
| gemini | `max_output_tokens` | wrong |
| openai | `max_completion_tokens` or `max_tokens` | wrong for reasoning models |
| azure | `max_completion_tokens` or `max_tokens` | wrong for reasoning models |

So on three of the four providers the message could name a setting that would not change
anything, which is worse than staying quiet, and Gemini is exactly where the empty-output
case bites hardest. Each call site now passes the value it actually sent, and there is a
test asserting the message names the cap and the model for a Gemini-style, a reasoning-model
and an Anthropic-style configuration.

Also addressed the docstring coverage warning by documenting `is_truncated`,
`warn_if_truncated` and `_as_str`, including why callers pass their own effective cap rather
than a shared attribute.

Re-verified on the rebased branch:

| check | result |
|---|---|
| Full suite, clean `main` (`7c23857`) | 5093 passed, 44 skipped |
| Full suite, this branch | **5112 passed, 44 skipped** |
| Difference | exactly +19, the new tests; no regressions |
| `ruff check` / `format --check` | clean |

Both numbers measured today against the same commit rather than carried over from the
earlier run, since `main` moved and the old baseline would have been stale.

Vidit-Ostwal (commented):
Thanks for putting this up — this is the right first step for #7013.

Please rebase onto `main` (`mergeStateStatus` is currently `BEHIND`) and work through the remaining CodeRabbit comments before we review further:

- Azure: report the cap `_prepare_completion_params()` actually sends (`params.get("max_tokens")`), rather than choosing between `self.max_completion_tokens` and `self.max_tokens`.
- OpenAI structured Chat Completions: `beta.chat.completions.parse()` can return a parsed object before the new `warn_if_truncated` call, so a `finish_reason="length"` response still stays silent. Wire the warning on the parse path (sync and async) as well.

Happy to take another look once those are in.

Vidit-Ostwal:
Ping me once if this is ready for review, I can still see the code rabbit not yet resolved.

JoaquinDG:
@Vidit-Ostwal ready for review, thanks for the nudge. All three CodeRabbit threads are
addressed and resolved, and one of them was a genuinely good catch that I would rather flag
than bury.

**The warning never fired on the OpenAI Responses API.** That path returns `response.status`
rather than a finish reason, so the value was `"incomplete"` and never matched. The actual
cause lives on `incomplete_details.reason` (`"max_output_tokens"`). Added that spelling to
the recognised set plus a small `_responses_truncation_reason()` helper to read it, with
tests covering both the object and dict shapes.

**Structured-output paths were silent too.** The `beta.chat.completions.parse()` branches
return `parsed_object` before the call sites I had added, so a parsed response with
`finish_reason="length"` produced nothing. Covered in both the sync and async branches.

**Reused `_effective_max_tokens()` instead of my own version.** I had hand-rolled
`max_completion_tokens or max_tokens`, which was both duplicated logic and the opposite
precedence to the helper already on `BaseLLM` that `LLMCallStartedEvent` reports. All eight
call sites now use the existing helper, so the warning names the same cap your telemetry
does.

Verification, re-measured on the current base (`fcdeb3d`):

| check | result |
|---|---|
| Full suite, this branch | **5144 passed, 44 skipped** |
| Full suite, clean base | 5122 passed, 44 skipped, 1 failed |
| New tests | 21 |
| `ruff check` / `format --check` | clean |

One note on that baseline failure so it does not look like mine:
`test_trace_enable_disable.py::test_trace_calls_when_enabled_via_env` failed on the clean
base run and passed on the branch run. It passes in isolation on both, so it looks like a
test-ordering or environment flake rather than anything this PR touches. Happy to be
corrected if it is a known one.

Still deliberately out of scope, as discussed on #7013: unknown finish reasons from
`openai_compatible` endpoints, and whether truncation should surface structurally rather
than only in logs. Glad to take either as a follow-up.

JoaquinDG:
Rebased onto current `main` (`bf56bb13`), conflicts resolved, so this is mergeable again.

The conflicts were all import blocks: `main` added `HookAborted` and `LLMCallBlockedError`
alongside the `warn_if_truncated` import this branch adds. Both sides kept, no logic
touched. All ten call sites survived intact.

| check | result |
|---|---|
| Full suite, this branch | **5210 passed, 44 skipped** |
| Full suite, clean `main` (`bf56bb13`) | 5189 passed, 44 skipped |
| Difference | exactly +21, the new tests; no regressions |
| `ruff check` / `format --check` | clean |

One thing worth naming so it does not look like mine.
`test_trace_enable_disable.py::test_trace_calls_when_enabled_via_env` failed on my first
full-suite run of this branch, then passed on a second run of the identical tree, and passes
both in isolation and with the whole `tracing/` module. Last week it failed on the clean
baseline instead. It looks order or timing dependent rather than related to this change,
but flagging it in case it is worth a separate look.

Also, congratulations on #7003 landing, and thanks for merging it. Happy to pick up the
Anthropic model ids in `llm-selection-guide.mdx` that I mentioned on that thread as a
separate small PR if it would be useful.

Vidit-Ostwal (commented):
Thanks for the thorough work on #7013 — the direction is right, but two structural tweaks would make this easier to maintain before merge.

## 1. Keep the util simpler

Right now `_finish_reason_utils.py` exposes two public APIs (`is_truncated` + `warn_if_truncated`), a frozenset, long docstrings, and a bool return value that nothing in production uses. For this issue, a single warning helper is enough:

- fold the normalization inline (or keep it private)
- drop public `is_truncated` unless callers actually need the predicate
- trim tests to warning behavior rather than pinning a separate predicate contract

Also worth dropping or documenting `"modellength"` in `_TRUNCATION_REASONS` — it is not called out anywhere else.

OpenAI Responses still needs the `incomplete_details.reason` read, but that can stay a small OpenAI-local helper; it does not need to expand the shared util surface.

## 2. Wire the warning once, for every provider

`warn_if_truncated` is called at 10 extract sites (Anthropic sync/async non-stream, Azure non-stream, Gemini non-stream, OpenAI Chat/Responses non-stream). That misses:

- all streaming finalizers (Anthropic, Azure, Gemini, OpenAI — including Responses stream)
- LiteLLM (`llm.py` `_handle_emit_call_events`)
- Bedrock, which still has four hand-rolled warnings

Every provider already passes `finish_reason` into `_emit_call_completed_event` / `_handle_emit_call_events`. Calling the warning from those two emitters would cover Anthropic, Azure, Gemini, OpenAI (incl. openai_compatible + Snowflake), LiteLLM, and stream paths in one place, and let Bedrock drop its duplicate branches.

Happy to take another look once those are in.

PR #7113, opened 2026-08-25
docs: add Agent Action Capsule (capsule-emit) to observability integrations

thisjody:
Adds an observability page for capsule-emit's CrewAI event-bus listener: sealed, content-addressed evidence records (Agent Action Capsules) for crew runs — one sealed record per tool call (planned → executed → confirmed), whose integrity a third party can check without access to the system that produced it.

What the page covers:
- `pip install "capsule-emit[crewai]"` (released: capsule-emit 0.4.0 on PyPI), register `CapsuleEventListener()` — no other code changes; every tool call in the crew seals a capsule via the standard `BaseEventListener` mechanism
- what gets recorded (digests of tool input/output, verdict/effect, timestamps — raw content stays local; digest-only anchoring to a SCITT transparency log)
- a runnable quickstart (crew → capsules → `verify()`, ending green: `6 capsules, verify: ALL OK` — sample output on the page is from a real run on capsule-emit 0.4.0 + crewai 1.15.x), which runs against a local stub transparency service — no external service required
- replay/`is_replaying()` behavior and error isolation (a listener failure never affects the crew run)

Where this differs from the tracing/APM tools on the overview page: those answer "what happened, for the operator"; capsules answer "prove what happened, to someone who doesn't trust the operator" — audit and compliance rather than debugging. Complementary to, not a replacement for, tracing.

Follows the docs workflow in AGENTS.md: page under `docs/edge/en/observability/`, overview card added, `docs.json` referenced, ar/ko/pt-BR translations synced.

*This PR was prepared by an AI agent under human direction — per CONTRIBUTING it should carry the `llm-generated` label (I can't self-apply labels as an outside contributor; please add it). The integration code lives in capsule-emit (Apache-2.0), released on PyPI as 0.4.0; the quickstart on this page is tested end-to-end against that release. Disclosure: I co-maintain capsule-emit.*


thisjody:
Both findings were correct — verified against the released 0.4.0 wheel before fixing (the `verify_ledger` import does raise ImportError; the listener does seal a two-record chain per tool call, planned → confirmed/errored via `prior_capsule_id`). Fixed in all four locales; the replacement verification loop is tested against the published wheel, not the repo. Thanks, rabbit.

thisjody:
Round-2 findings all verified real against the released 0.4.0 wheel and fixed: the registration example now passes the required `operator`/`developer` kwargs (bare construction does raise `TypeError` — confirmed), the check step uses `verify_store` with an empty-ledger guard (snippet executed green against the published wheel), and the Arabic errored-state wording is corrected. All four locales.

thisjody:
Correct again — `assert` is stripped under `-O`. Replaced with an unconditional check in all four locales; page snippets re-executed green against the released 0.4.0 wheel.

thisjody:
Both outside-diff findings verified and fixed in all four locales: the `action_id` row now covers crew lifecycle capsules (the listener does record kickoff started/completed/failed — confirmed in the shipped adapter), and the anchor row is renamed to anchor *submission* with the default `anchor_status="submitted"` semantics stated and `anchor_wait` noted for a confirmed receipt — the docstring's own distinction, now honored by the page. Snippets re-gated green against the released wheel.

PR #7117, opened 2026-08-26
feat(events): report machine size as a coarse band, not a core count

joaomdmoura:
- Adds `cpu_band` to the common span attributes — a coarse machine-size band on every span, so "is this a server or someone's laptop" becomes answerable. `runtime_context` says *where* a process runs but has no capacity axis, and its largest bucket is a catch-all: a gunicorn worker on a VM and `python main.py > out.log` on a MacBook both report `non_interactive`; Docker Desktop on a laptop reports `container` via `/.dockerenv`; a Remote-SSH shell on a server reports `vscode_terminal`.
- No public API change. `crewai_core.runtime_env` is not re-exported from either package root, and `detect_cpu_band` is new — so unlike the moved detection helpers it gets no compatibility shim in `crewai.telemetry.utils`.
- **The `share_crew`-gated exact `cpus` attribute and the four `platform*` attributes are untouched.** That gating was a deliberate 2024 classification of machine fingerprint as shareable content (`44e38b1d5`, *"Fixing telemetry condition that was missing"*), and this does not reverse it: the band is a range; the precise value stays opt-in.
- Six bands, powers of two, top one open-ended — `1-2`, `3-4`, `5-8`, `9-16`, `17-32`, `33+`. Open-ended because the exact count *is* the fingerprint: the observed fleet maximum is 512, and a span reporting 512 identifies one machine. The vocabulary is closed and asserted, as for `KNOWN_CODING_AGENTS` and `KNOWN_RUNTIME_CONTEXTS`.
- Tests cover every band boundary inclusively (1/2/3/4/5/8/9/16/17/32/33/96/512), that the top band absorbs 512 rather than leaking it, that an unavailable count is `unknown` rather than silently `1-2`, that the real host lands in the closed vocabulary, that no emittable value can carry a precise count, and that it rides the common attributes rather than one span type.
- Docs updated in `en`/`ar`/`ko`/`pt-BR`, in the default-on **Execution Environment** row, stating explicitly that the band is a range and that the exact count remains opt-in under **Environment Information**. No `docs/v*/` touched.
- Keeps this intentionally small: no warehouse column in this PR (a new tag key is ignored by every existing extractor, so no consumer breaks and it can follow), no change to any existing attribute or gate, and no reduction of the gated `cpus`.
- Next: a `cpu_band` column on the crew spine, then item 4.7(iii).

**One trap documented in the docstring rather than left to be rediscovered:** `os.cpu_count()` reports **host** cores, not the cgroup quota, so a 1-vCPU pod on a 96-core node lands in `33+`. That is the right answer for "what kind of machine is this" and the wrong one for "what resources did this run get" — `os.process_cpu_count()` gives the latter but requires Python 3.13, above this package's 3.10 floor.

https://claude.ai/code/session_01GzrA3L2EFh9pRFi7RqUpLf

---

> [!NOTE]
> **Low Risk**
> Documentation and additive anonymous telemetry only; no public API or changes to existing `share_crew` gates.
> 
> **Overview**
> Adds default-on telemetry **`cpu_band`** on every span via `common_span_attributes()`, alongside existing execution-environment fields. **`detect_cpu_band()`** in `crewai_core.runtime_env` maps `os.cpu_count()` to a closed vocabulary (`1-2` through `17-32`, `33+`, or `unknown`) so fleet analytics can distinguish laptop-scale from server-scale hosts without emitting an exact core count.
> 
> The **`share_crew`-gated exact CPU count** under Environment Information is unchanged; docs in **en/ar/ko/pt-BR** now describe the band on the Execution Environment row and clarify that precise counts remain opt-in. Tests cover band boundaries, `unknown` when count is unavailable, vocabulary safety, and that **`cpu_band`** is a process-wide common attribute (not limited to crew-created spans).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit adbd11f7391092f4e49c0fc538636cfab114e6c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joaomdmoura:
Both review findings addressed in `59ad620c1`.

**1. CPU-band data-source description — fixed, and the finding was right.** Verified against source rather than taken on trust: `detect_cpu_band()` reads `os.cpu_count()` (`lib/crewai-core/src/crewai_core/runtime_env.py:306`) and touches no environment variable. The row previously ended *"Detection reads only whether known environment variables are set, never their values"*, which is accurate for the assistant and runtime-context fields and became wrong the moment the band joined the same row. In a privacy-disclosure table that is worth a line, so the sentence now names the source and scopes the env-var clause to the two fields it actually describes.

All four locales at parity, line counts unchanged, table structure intact, no `docs/v*/` touched.

**2. pt-BR wording — applied as suggested.** Both occurrences of `um de uma lista fixa` → `um valor de uma lista fixa`, plus the missing comma before `o project_id`.

**Not changed, and worth stating:** the implementation already documents the trap that matters here — `detect_cpu_band()`'s docstring records that `os.cpu_count()` reports **host** cores rather than the cgroup quota, so a 1-vCPU pod on a 96-core node lands in `33+`. That is the right answer for "what kind of machine is this" and the wrong one for "what resources did this run get", and it is stated where someone reading the function will find it.

lorenzejay (approved):

PR #7118, opened 2026-08-26
feat(events): record how a crew run ended, for every user

joaomdmoura:
- Adds an ungated `Crew Completed` span carrying `outcome` and an explicit `duration_ms`, keyed by `crew_key`/`crew_id` so it joins the existing ungated `Crew Created`. **Crew was the one level with no ungated terminal record:** `Crew Execution` and `end_crew` are both behind `share_crew`, which defaults `False`, so for essentially every run there is no end-of-crew span *at all* — not one with fields missing. Task and flow outcomes both ship ungated already.
- No public API change. New method on the internal `Telemetry` class plus one `PrivateAttr` on `Crew`; no signature moves and nothing new is re-exported.
- Modelled directly on `flow_completed_span`, including its reasoning: a **separate** span rather than holding `Crew Execution` open, because that span is emitted and closed at start — holding it would drop every run that is killed or crashes. `on_crew_failed` called no telemetry at all before this, and deliberately still does not call `end_crew`, which writes onto the gated execution span a failed run may never have opened.
- Preserves existing behaviour exactly: `end_crew` and `crew_execution_span` stay `share_crew`-gated and unchanged, so opt-in users get precisely what they got before, plus this.
- Tests: the span (outcome, explicit duration, joinability, that it is **not** gated, that it carries no token count) and the listener's stamp lifecycle across both terminal paths and the no-start case. **Verified order-independent over five randomized runs.**
- Docs: none. No new attribute class is disclosed — `Crew Created` already reports crew identity and shape ungated, and Task Lifecycle Data already documents that success/failure is recorded.

**Four things deliberately left out, each with a reason rather than an omission:**

| left out | why |
|---|---|
| **tokens** | `crew.token_usage` sums per-agent LLM counters, and two agents sharing one `LLM` object share one counter — the total **double-counts today**. Emitting it would propagate a known-wrong number into a metric. The fix keys on `id(llm._token_usage)`, not `id(llm)` (`Agent.copy()` shallow-copies the LLM), and it changes the value of public `Crew.calculate_usage_metrics`, so it earns its own change. |
| **models** | Already on the ungated `Crew Created` span at **99.86%** coverage. This joins to them by `crew_id` rather than duplicating. |
| **tool counts** | The ungated `Tool Usage` span covers only the ReAct path — native function-calling never calls it — the plan/step path double-emits, and nested crews share one `RuntimeState`. Needs a written decision on whether a cache hit is a tool call before it is worth emitting. |
| **`error_type`** | Wants 4.1's exception-class field factored out of `task_events` so both events share it, rather than duplicated hours after that merged. |

- Keeps this intentionally small: no change to any gate, no new event field, no warehouse change in this PR (a new span name needs a fan-out route, which follows).
- Next: the token dedup fix, then tokens on this span; `error_type` once the exception field is shared; tool counts after the cache-hit decision.

**One test-design note worth carrying forward:** `EventListener` is a singleton, so mocking `_telemetry` on it leaks a `MagicMock` into every later test in the process. These tests use the `InMemorySpanExporter` pattern the suite already uses, and assert the listener's stamp lifecycle rather than a shared mock.

https://claude.ai/code/session_01GzrA3L2EFh9pRFi7RqUpLf

---

> [!NOTE]
> **Low Risk**
> Changes are limited to internal telemetry and event-listener wiring with no public API or execution-path changes; opt-in share_crew telemetry remains unchanged.
> 
> **Overview**
> Adds an **ungated** `Crew Completed` telemetry span so every kickoff gets a terminal record with **`outcome`** (`completed` / `failed`) and **`duration_ms`**, joinable to existing `Crew Created` via `crew_id` / `crew_key`. This mirrors **`flow_completed_span`**: a separate end span instead of holding `Crew Execution` open, since that path is behind **`share_crew`** (default false) and failed runs never call **`end_crew`**.
> 
> **`EventListener`** stamps **`_telemetry_started_at`** on kickoff start and reports duration on success or failure, clearing the stamp so duplicate terminal events cannot double-report. **`Crew`** gains a private monotonic start stamp only for this wiring.
> 
> Existing **`share_crew`**-gated **`Crew Execution`** / **`end_crew`** behavior is unchanged. The span deliberately omits token counts (known double-count when agents share an LLM). New tests cover span attributes and listener stamp lifecycle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2630fc53a6f99608e52a2672fa1fbb3359dd2930. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joaomdmoura:
No inline findings on this one. Two things, one addressed and one that I think should block the merge.

**Docstring coverage (52.17%) — partially fixed in `84ee2ad8b`, partially rejected with a reason.**

*Fixed:* five of the thirteen functions in the new test file had no docstring while eight did — an inconsistency inside the file this PR adds. All five now have one. `9 passed`, and re-run under random ordering on two seeds to confirm order-independence.

*Rejected:* the 52.17% figure is dominated by `event_listener.py`, where **84** functions carry no docstring — nearly every pre-existing `on_*` handler. That is the file's established convention, not something this PR introduced, and documenting them to satisfy a coverage metric would be an unrelated refactor across a file this PR touches in 23 lines. The public API this PR actually adds, `Telemetry.crew_completed_span`, is documented.

*Note on the gates:* `ruff` and `ruff format` do **not** cover this change — `lib/crewai/tests/` is in `extend-exclude` with `force-exclude = true` (`pyproject.toml:40-46`), so the linter skips test files by design. Claiming a green lint run here would be claiming a run that did not happen.

---

**⚠️ This needs a warehouse change before it merges, and it is the same gate 4.2 respected.**

This PR emits a **new span name**, `Crew Completed`. `mv_span_fanout_forward`'s allowlist is a hard-coded list of span names, and I read it from the live DDL today — it holds **14** names, and `Crew Completed` is not among them (`Flow Completed` is, which is a different span). Unwidened, this span reaches the raw table and **nothing else**: no route, no spine, coverage 0%, with green CI and a clean diff the whole way.

That is not hypothetical. It happened yesterday: 4.4 shipped in 1.15.18 and `Project Created` is emitting now — 12 spans observed, all carrying `created_project_id` — into no warehouse object at all, because the same allowlist was never widened for it.

The established rule from 4.5 and 4.2 is **warehouse first**: make the consumer ready before the emitter ships, because emitter-first is undetectable from the OSS side and surfaces days later as a metric that is quietly empty. 4.2 did this correctly — both its gates (the allowlist *and* the `multiIf` catch-all arm that would otherwise have mislabelled the new span a removal) were applied on 2026-08-25, before the PR merged, and I verified this morning that it worked: `crew_deployment_created` is landing with 100% uuid coverage and `remove_crew` is unchanged.

So before this merges, someone needs to decide what consumes `Crew Completed` — at minimum a 15th allowlist entry and a route. Happy to draft that runbook; it is warehouse-only and does not touch this PR.

lorenzejay (approved):

lorenzejay (approved):

Issue #4993, opened 2026-03-20
SQL injection in SnowflakeSearchTool via database and schema parameters


jmanico:
Hey folks, I was reviewing the SnowflakeSearchTool and noticed that the `database` and `snowflake_schema` parameters in the `_run` method get dropped straight into SQL strings without any validation or quoting.

Here's the relevant code in `snowflake_search_tool.py` around line 261:

```python
if database:
    await self._execute_query(f"USE DATABASE {database}")
if snowflake_schema:
    await self._execute_query(f"USE SCHEMA {snowflake_schema}")
```

These values come from `SnowflakeSearchToolInput`, which defines both fields as plain `str | None` with no pattern constraints. Compare that with `SnowflakeConfig.account` which already has `pattern=r"^[a-zA-Z0-9\-_]+$"` on it. The input schema fields have nothing similar.

Since `_execute_query` just calls `cursor.execute(query)` directly (line 218), a malicious value like `test_db; DROP TABLE users; --` would get executed as SQL.

In a CrewAI context, these parameters are filled by the LLM agent based on task instructions, which means prompt injection in upstream task context could influence these values.

## Suggested fix

The simplest approach would be to add an identifier validation regex to both fields in the input schema, and also double quote the identifiers in the SQL:

```python
# On SnowflakeSearchToolInput
database: str | None = Field(None, pattern=r"^[A-Za-z_][A-Za-z0-9_$]*$")
snowflake_schema: str | None = Field(None, pattern=r"^[A-Za-z_][A-Za-z0-9_$]*$")
```

```python
# In _run()
if database:
    await self._execute_query(f'USE DATABASE "{database}"')
if snowflake_schema:
    await self._execute_query(f'USE SCHEMA "{snowflake_schema}"')
```

Snowflake's `USE DATABASE` doesn't support bind parameters, so identifier validation is the way to go here. The regex above matches Snowflake's unquoted identifier rules.

Worth noting that the `query` parameter itself is also passed through unsanitized, but that's somewhat by design since the tool is meant to execute SQL. The `database` and `snowflake_schema` fields are the ones that look like safe metadata but actually aren't.

Also worth mentioning that the NL2SQL tool has a similar pattern at `nl2sql_tool.py:57` where `table_name` from a prior query result gets f-string interpolated into an `information_schema` query (second order injection). That one already has a `# noqa: S608` on it acknowledging the issue, and there's documentation about it from PR #4465, but no code fix yet.

[...]

PR #7120, opened 2026-08-26
fix(tools): validate snowflake database and schema identifiers

santhiprakash:
**AI disclosure:** authored with AI assistance. CONTRIBUTING requires the `llm-generated` label; this account cannot add labels on `crewAIInc/crewAI` (REST 403). Please apply `llm-generated`.

## Problem

`SnowflakeSearchTool._run` interpolates the agent-supplied `database` and `snowflake_schema` values directly into SQL:

```python
await self._execute_query(f"USE DATABASE {database}")
await self._execute_query(f"USE SCHEMA {snowflake_schema}")
```

A value such as `analytics; DROP DATABASE prod` becomes a second statement. This is the same class of identifier interpolation that #6341 closed for MySQL table names, and the same stacked-write class as #6987 (SingleStore), on a different tool.

Self-sourced. #4993 reported this and was stale-closed; #4994 / #4997 bundled Snowflake + NL2SQL and closed unmerged. NL2SQL was later hardened in #5311. This PR is Snowflake-only.

## Triage / Root cause

`_run` treats `database` / `snowflake_schema` as identifiers but never checks that they are identifiers. The connection config already pattern-checks `account`; these two runtime overrides were not checked.

## Fix

- Add `_validate_snowflake_identifier` (unquoted Snowflake identifier: `[A-Za-z_][A-Za-z0-9_$]*`; schema may be `database.schema`).
- Reject spaces, semicolons, comments, extra dots, and other SQL metacharacters before interpolation.
- Leave validated names **unquoted** so Snowflake can still case-fold them (`USE DATABASE analytics` → `ANALYTICS`). Quoting would change that lookup.

## Verification

```
uv run pytest lib/crewai-tools/tests/tools/snowflake_search_tool_test.py -q
```

28 passed. 1 pre-existing failure: `test_cleanup_on_deletion` uses `async with` on a `threading.Lock` and is unrelated to this change.

New tests cover safe names, qualified schema, and injection payloads (`analytics; DROP DATABASE prod`, comment forms) without a Snowflake server.

## Notes / Risks

- Identifiers that only exist as quoted Snowflake names (spaces, reserved words) are rejected. Same tradeoff as #6341.
- Raw `query` is unchanged; this tool still executes the SQL the caller passes. Only the `USE DATABASE` / `USE SCHEMA` overrides are validated.
- Does not touch SingleStore (#6987) or NL2SQL.

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/tools/snowflake_search_tool/snowflake_search_tool.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/snowflake_search_tool/snowflake_search_tool.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
 import logging
+import re
 import threading
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


santhiprakash:
Closing to keep the review queue clean — this has been open a while without traction. Happy to reopen or resubmit a fresh version if it would still be useful.

PR #7123, opened 2026-08-26
feat: add Volcengine Ark as an OpenAI-compatible provider

chyroc:
## Summary

Adds Volcengine Ark (火山方舟) as an OpenAI-compatible provider. Ark is ByteDance's model platform — it serves the Doubao family plus hosted DeepSeek and GLM models over an OpenAI-compatible API, so it slots into the existing `OPENAI_COMPATIBLE_PROVIDERS` registry rather than needing its own module like Snowflake did.

```python
from crewai import LLM

llm = LLM(model="volcengine/doubao-seed-2-1-pro-260628")
```

`ARK_API_KEY` for auth, `ARK_BASE_URL` to override the endpoint (BytePlus is `https://ark.ap-southeast.bytepluses.com/api/v3` for use outside China Mainland).

crewAI already ships `dashscope` and `deepseek`; Ark was the remaining gap among the major Chinese providers.

## Changes

- `openai_compatible/completion.py` — one `ProviderConfig` entry
- `llm.py` — four registration points, matching how `dashscope` is wired
- model-name validation accepts `doubao` / `deepseek` / `glm` / `kimi` prefixes, since Ark hosts more than just its own family
- tests + a docs accordion

## Model IDs

Worth flagging: Ark only resolves **fully versioned** IDs. The short names in the Ark console (`doubao-seed-2.1-pro`) come back as `InvalidEndpointOrModel.NotFound`; `doubao-seed-evolving` is the one rolling alias that works without a date suffix. The docs section says so, and the examples use versioned IDs.

## Testing

- `pytest tests/llms/openai_compatible/` — 38 passed
- `ruff check` / `ruff format --check` — no new findings (the 31 pre-existing errors in these files are unchanged; none mention volcengine)

Also exercised against the live Ark API through `LLM.call()` rather than only mocks — `doubao-seed-2-1-pro-260628`, `deepseek-v4-pro-ga-260813` and `glm-5-2-260617` all round-trip.

### DIFF
--- a/docs/edge/ar/concepts/llms.mdx
+++ b/docs/edge/ar/concepts/llms.mdx
@@ -959,6 +959,39 @@ mode: "wide"
     ```
   </Accordion>
 
+  <Accordion title="Volcengine Ark">
+    عيّن متغيرات البيئة التالية في ملف `.env`:
+    ```toml Code
+    # Required
+    ARK_API_KEY=<your-api-key>
+
+    # Optional: القيمة الافتراضية هي نقطة النهاية في الصين
+    ARK_BASE_URL=https://ark.cn-beijing.volces.com/api/v3
+    ```
+
+    مثال على الاستخدام في مشروع CrewAI:
+    ```python Code
+    llm = LLM(
+        model="volcengine/doubao-seed-2-1-pro-260628",
+        temperature=0.7
+    )
+    ```
+
+    <Info>
+      يوفّر Volcengine Ark نماذج Doubao من ByteDance إضافةً إلى نماذج DeepSeek و GLM
+      المستضافة، عبر واجهة برمجة تطبيقات متوافقة مع OpenAI.
+
+      يجب أن تتضمّن معرّفات النماذج رقم الإصدار الكامل. الأسماء المختصرة الظاهرة في
+      وحدة تحكم Ark (مثل `doubao-seed-2.1-pro`) تُرجع
+      `InvalidEndpointOrModel.NotFound`. و`doubao-seed-evolving` هو الاسم المستعار
+      المتجدد الوحيد الذي يعمل بدون لاحقة تاريخ. راجع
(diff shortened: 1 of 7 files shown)


chyroc:
Pushed the docs sync and the kimi test.

On the version-suffix validation — I'd rather leave it loose, for a few reasons:

`_matches_provider_pattern` decides *which provider handles this model*, not *whether the model exists*. Returning False doesn't reject the call; it falls through to LiteLLM, which then fails with a murkier error than Ark's own `InvalidEndpointOrModel.NotFound`. So tightening it trades a clear error for a worse one.

No sibling provider validates versions either — `deepseek` is `startswith("deepseek")`, `dashscope` is `startswith("qwen")`, and `ollama`/`vllm`/`cerebras`/`openrouter` just `return True`. A `-\d{6}$` regex here would be the only one of its kind, and the docstring says the fallback exists to support "latest versions and new models that follow provider naming conventions".

Ark's own ID shapes also aren't uniform: `doubao-seed-evolving` carries no version at all, and `kimi-k2-7-modelhub` uses a `-modelhub` suffix rather than a date. A regex would need updating every time that changes.

The docs section does call out that short console names return `NotFound`, so users hitting it get pointed the right way. Happy to add the check if you'd still prefer it.

chyroc:
Added Kimi to the provider description — but in `en` plus all three translations, not only `ar`/`ko`.

The English source didn't mention Kimi either, so the translations were faithful; the real gap was between the docs and the code, since `llm.py:580` routes on `doubao`/`deepseek`/`glm`/`kimi` and the tests cover `kimi-k3-260727`. Patching just the translations would have left them out of sync with `en` and tripped the next docs check.

Worth noting Kimi isn't on Volcengine's public model list page at all, but the models are live — `kimi-k3-260727`, `kimi-k2-5-260127` and `kimi-k2-7-code-260601` all respond; a few of the older listed IDs 404. That's why the prefix is there.

PR #7135, opened 2026-08-27
fix(tracing): collapse the dead has_user_declined_tracing branch

noron12234:
## What

Four "tracing is disabled" renderers branch on `has_user_declined_tracing()` and then assign byte-identical strings on both sides:

```python
if has_user_declined_tracing():
    message = """Info: Tracing is disabled.

To enable tracing, do any one of these:
• Set tracing=True in your Crew code
• Set CREWAI_TRACING_ENABLED=true in your project's .env file
• Run: crewai traces enable"""
else:
    message = """Info: Tracing is disabled.

To enable tracing, do any one of these:
• Set tracing=True in your Crew code
• Set CREWAI_TRACING_ENABLED=true in your project's .env file
• Run: crewai traces enable"""
```

Same shape in all four:

| file | function |
|---|---|
| `crew.py` | `Crew._show_tracing_disabled_message` |
| `flow/runtime/__init__.py` | `Flow._show_tracing_disabled_message` |
| `events/utils/console_formatter.py` | `_show_tracing_disabled_message_if_needed` |
| `events/listeners/tracing/trace_listener.py` | `_show_tracing_disabled_message` |

The call, the branch, and the import sitting next to it all do nothing.

## Change

Collapsed each to the single assignment and dropped the now-unused `has_user_declined_tracing` import from those four modules. The real use in `events/listeners/tracing/utils.py` (`should_auto_collect_first_time` returning `False` for a declined user) is untouched.

## Verification

Behaviour is unchanged, checked rather than assumed. Walking the AST of each file before and after and collecting every string literal assigned to `message`:

```
lib/crewai/src/crewai/crew.py                          before=2 after=1  distinct set identical: True
lib/crewai/src/crewai/flow/runtime/__init__.py         before=2 after=1  distinct set identical: True
lib/crewai/.../events/utils/console_formatter.py       before=2 after=1  distinct set identical: True
lib/crewai/.../listeners/tracing/trace_listener.py     before=2 after=1  distinct set identical: True
```

The rendered panel is byte-for-byte what it was.

Lint: `ruff check` reports the same **29** pre-existing findings on these four files before and after the change — no new debt, and no unrelated autofixes were left in (ruff tried to fold an `elif` chain in `flow/runtime` and drop a bare `return` in `console_formatter`; both reverted so the diff stays on topic). `ruff format --check` clean.

## One question for you

The distinction the branch was reaching for is real: declining explicitly is not the same as never having been asked. If the declined case was meant to read differently, tell me what it should say and I will write that copy instead of removing the branch.


noron12234:
Closing this. I am cleaning up the PRs I have open and keeping only the ones a maintainer has already picked up; nobody has needed to look at this one, and leaving it open just adds to your queue.

The branch stays on my fork, so nothing is lost if the fix turns out to be worth having. Sorry for the noise.

PR #7155, opened 2026-08-29
fix: reject replay when stored tasks differ

### LINKED ISSUE #7154: replay() pairs the stored log to the task list by position, so a changed crew silently restores outputs onto the wrong tasks
### Description

`Crew.replay()` restores previously stored task outputs onto the current task list **by position**, and nothing checks that position `i` in the stored log is the same task as position `i` in the crew. If the task list changed between the run and the replay — which is the workflow replay exists for — outputs land on the wrong tasks and execution can resume from the wrong one, with no error.

The same file already does this correctly on the other recovery path: `_restore_runtime()` matches by `task.id`.

### Steps to Reproduce

1. Run a crew of three tasks: `RESEARCH`, `WRITE`, `REVIEW`.
2. Add one task at the front, so the crew is now `PLAN`, `RESEARCH`, `WRITE`, `REVIEW`.
3. Replay from the stored `task_id` of `REVIEW`.

### Expected behavior

Either resume at `REVIEW` with each restored output on the task it came from, or refuse with an error because the stored log no longer corresponds to the crew.

### Screenshots/Code snippets

```python

def mk(d):
    a = Agent(role=f"r-{d}", goal="g", backstory="b", llm="gpt-4o-mini")
    return Task(description=d, expected_output="o", agent=a)

# run 1
t = [mk("RESEARCH"), mk("WRITE"), mk("REVIEW")]
c1 = Crew(agents=[x.agent for x in t], tasks=t)
c1._task_output_handler.reset()
for i, x in enumerate(t):
    c1._store_execution_log(
        x, TaskOutput(description=x.description, raw=f"OUT-{x.description}",
                      agent=x.agent.role), i)
target = c1._task_output_handler.load()[2]["task_id"]      # REVIEW

# run 2 — one task added at the front
t2 = [mk(d) for d in ["PLAN", "RESEARCH", "WRITE", "REVIEW"]]
c2 = Crew(agents=[x.agent for x in t2], tasks=t2)

def show(tasks, start_index=None, was_replayed=False, *a, **k):
    print("resumes at:", tasks[start_index].description)
    for x in tasks:
        print(f"  {x.description:9s} <- {x.output.raw if x.output else None}")
    raise SystemExit
c2._execute_tasks = show
c2.replay(task_id=target)
```

Output:

```
resumes at: WRITE
  PLAN      <- OUT-RESEARCH
  RESEARCH  <- OUT-WRITE
  WRITE     <- None
  REVIEW    <- None
```

The same script with the crew left unchanged resumes at `REVIEW` with zero misplaced outputs, so this is the changed task list rather than the harness:

```
unchanged crew          resumes at REVIEW   misplaced: 0
task added at front     resumes at WRITE    misplaced: 2
tasks reordered         resumes at REVIEW   misplaced: 2
```

The reordering row is the quieter one: it resumes at the right task by coincidence of position, while the restored context is shuffled, so the run looks normal and the result is subtly wrong.

### Evidence

`crew.py`, in `replay()`:

```python
start_index = self._find_task_index(task_id, stored_outputs)   # index into the LOG
for i in range(start_index):
    stored_output = stored_outputs[i]["output"]
    ...
    self.tasks[i].output = task_output                          # applied to the CREW
return self._execute_tasks(self.tasks, start_index, True)
```

`task_id` is used only to find a position in the log; from there everything is positional.

It cannot currently validate the pairing, because ids are not stable across runs — `task.py`:

```python
id: UUID4 = Field(default_factory=uuid.uuid4, frozen=True, ...)
```

The checkpoint path in the same file does match on identity:

```python
# _restore_runtime()
if task.output is not None or str(task.id) not in started_task_ids:
    continue
```

### Possible Solution

Store something stable alongside `task_index` — the task's `description`, or a user-settable task key — and on replay compare it with `self.tasks[i]` before assigning, raising when they disagree. That turns a silent wrong-context run into an explicit "this log does not match this crew". A user-settable stable key would additionally make replay work across edits, which is the case replay is most useful for.

### Operating System

macOS (Darwin 25.5.0, arm64)

### Python Version

3.12

### crewAI Version

1.15.18

### crewAI Tools Version

not installed

### Virtual Environment

Venv

### Additional context

Reported alongside a related check on other durable-execution runtimes; this one is specific to `replay()` and independent of #5802.

ShivangiRay:
## Summary

- validates saved task outputs against the current crew before replay restores any context
- raises a clear error rather than assigning an output to a different task after tasks are inserted or reordered
- adds a regression test for the changed-task-order case

Closes #7154

## Validation

- `uv run pytest -n 0 lib/crewai/tests/test_crew.py::test_replay_rejects_changed_task_order lib/crewai/tests/test_crew.py::test_replay_feature lib/crewai/tests/test_crew.py::test_replay_preserves_messages`
- `uv run ruff check lib/crewai/src/crewai/crew.py lib/crewai/tests/test_crew.py`
- `uv run mypy lib/crewai/src/crewai/crew.py`


vasilisnasopoulos:
Thank you — that was fast, and the shape is right: validate before restoring, raise instead of assigning.

One gap in the key it validates on. `expected_output` is free text with no uniqueness constraint, and in practice it repeats — "A report", "A summary", "The final answer". When two tasks share it, `_validate_replay_tasks` compares equal at every index and the original behaviour proceeds unchanged:

```
stored:  [RESEARCH, WRITE, REVIEW]     all expected_output = "A report"
crew:    [WRITE, RESEARCH, REVIEW]     reordered
-> validation passes; RESEARCH's output still lands on WRITE
```

The insertion case is caught reliably — a task added at the front shifts everything, and its `expected_output` almost certainly differs. The reordering case is caught only when the strings happen to differ.

It is the same shape as the bug itself: an identifier whose uniqueness is assumed. `description` has the same problem but collides less often. A stable, user-settable task key would close it for good, and would additionally let replay survive an intentional edit — the case replay is most useful for.

Either way this is strictly better than what is there now, and worth taking as-is if the stronger key is a bigger change.

lib/crewai/src/crewai/crew.py
```diff
@@ -2031,6 +2031,18 @@ def _find_task_index(task_id: str, stored_outputs: list[Any]) -> int | None:
...
         )
 
+    def _validate_replay_tasks(self, stored_outputs: list[Any], start_index: int) -> None:
+        """Ensure stored outputs still correspond to the tasks that will receive them."""
+        if len(self.tasks) <= start_index:
+            raise ValueError("Cannot replay because the current crew does not match the stored task outputs.")
+
+        for index, stored_output in enumerate(stored_outputs[: start_index + 1]):
+            task = self.tasks[index]
+            if task.expected_output != stored_output["expected_output"]:
```
Vidit-Ostwal: `expected_output` is the wrong fingerprint. It is free text and often repeated ("A report", "A summary"). Two tasks with the same string still pass, and `replay()` assigns by position — the original bug.

Do **not** compare `task.id` either (the CodeRabbit suggestion). Ids are `UUID4` per `Task()` instance, so a rebuilt crew of the same tasks — the normal restart-and-replay path — would always fail.

Compare something that is actually stored and closer to identity: `task.description` vs `stored_output["output"]["description"]`, and keep `expected_output` as a second check if you want. Use `.get(...)` so mocked / older logs do not `KeyError` — `test_replay_with_context` already omits `expected_output`.

Add a regression where every task uses the same `expected_output` and the list is reordered. That case still silently mis-assigns today.

Vidit-Ostwal (changes requested):
Shape is right: validate before restore, raise instead of silently pasting outputs onto the wrong slots. Please tighten the identity check before we merge.

**Keep:** fail loud when the saved prefix does not match the current crew. Do not try to remap onto a changed list.

**Change:** do not fingerprint on `expected_output` alone, and do not follow the CodeRabbit `task.id` suggestion. Ids die when the crew is rebuilt; `expected_output` is not unique. Use `description` (already on the stored output) plus `expected_output` if you want a second signal.

**Tests:** the insert-at-front case is good. Also cover (1) reorder with identical `expected_output` values, and (2) existing replay fixtures that omit `expected_output` so they do not `KeyError` (`test_replay_with_context`).

vasilisnasopoulos:
`description` + `expected_output` is a third proxy, and it has the same hole as the first two. Neither has a uniqueness constraint — both are plain `str` on `Task` with no validator, and templated or loop-generated tasks routinely share both.

That collides with the reorder test you asked for: if two tasks with identical `expected_output` also share a `description` and get swapped, the fingerprint matches, validation passes, and `replay()` restores outputs onto the wrong tasks — the original bug, unchanged.

Agreed on rejecting the `task.id` suggestion, for the reason you gave: `id` is `default_factory=uuid.uuid4, frozen=True` with `_deny_user_set_id`, so it cannot survive a rebuild and cannot be pinned by the author either. It would reject every replay.

If the crew may change between the run and the replay, no content-derived field settles identity. What would: the stored ordinal **plus** a content digest, with a mismatch on either being fatal — that fails loud on the ambiguous case instead of guessing. Or an optional explicit key on `Task`, putting the choice on the author.

Concretely: I'd have the reorder test use identical `description` *and* `expected_output`, so it fails against the current proposal.

ShivangiRay:
Updated in a54b64a..HEAD. The guard now compares the stored output description to the current task description, using persisted `expected_output` only as an optional second signal. This preserves rebuilt-crew replay, rejects reordered tasks even when expected outputs are identical, and keeps older fixture/log shapes from raising `KeyError`. Added coverage for the duplicate-expected-output reorder case and updated the existing replay fixtures. Focused tests, Ruff, and mypy pass.

lib/crewai/src/crewai/crew.py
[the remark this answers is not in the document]
ShivangiRay: Addressed in fb6238e. I used the persisted task description rather than UUID (which is not stable across rebuilt crews), keep expected output as an optional second signal, and added a reordered-task regression where both tasks use the same expected output.

ShivangiRay: Addressed in fb6238e. The guard now compares `task.description` with the stored output description, reads optional expected output with `.get()`, and has coverage for a reordered pair with identical expected-output values. Existing replay fixtures now describe valid stored task history.

ShivangiRay:
@vasilisnasopoulos Agreed—two unconstrained text fields cannot safely identify all tasks. Updated in HEAD to reject a replay prefix whose stored or current `(description, expected_output)` identities are ambiguous, rather than restore outputs by position. Unique rebuilt crews continue to work; ambiguous ones fail loudly. Added a regression for two identical task definitions in a reordered crew. Focused tests (6), Ruff, and mypy pass.

Vidit-Ostwal:
Hey @ShivangiRay, CI is failing mind checking ?

vasilisnasopoulos:
This is the right shape, and worth naming: you didn't look for a better field, you accepted there isn't one. Two tasks that a human cannot tell apart cannot be told apart by a replay either, so refusing is the only honest answer — and refusing loudly is what makes it safe.

The earlier attempts each assumed some field would carry identity; this one stops assuming. That's a different kind of fix.

vasilisnasopoulos:
@ShivangiRay 
The two failing tests point at something worth knowing before you chase the fixtures: `description` is not stable across a replay.

`interpolate_inputs_and_add_conversation_history` assigns to it in place:

```python
self.description = interpolate_only(
    input_string=self._original_description, inputs=inputs
)
```

So a task whose template is `"Say hello to {name}"` is stored as `"Say hello to John"` after the first kickoff, and `replay(..., inputs={"name": "Maria"})` — which is a supported call — makes the current task read `"Say hello to Maria"`. The guard then compares the two interpolated strings and refuses a replay that is actually legitimate. That is what `test_replay_interpolates_inputs_properly` is exercising.

crewAI already keeps a stable identity for exactly this, and it is derived from the pre-interpolation values:

```python
@property
def key(self) -> str:
    description = self._original_description or self.description
    expected_output = self._original_expected_output or self.expected_output
    return md5("|".join([description, expected_output]).encode(), usedforsecurity=False).hexdigest()
```

Comparing `task.key` against a stored key survives interpolation and keeps the ambiguity check you added: two tasks that are genuinely indistinguishable still produce the same key, so the "identities are ambiguous" refusal fires exactly as it does now.

The remaining wrinkle is that stored outputs currently persist `description` rather than the key, so the stored side needs the key written at kickoff time for older logs to remain replayable — otherwise the comparison has nothing stable to read back.

ShivangiRay:
@vasilisnasopoulos Addressed in HEAD. Replay records now persist `Task.key`, which CrewAI derives from the pre-interpolation description and expected output. New records compare those keys before input interpolation, so a replay can legitimately change `{name}` from John to Maria. Duplicate keys remain an explicit ambiguity error; legacy records without `task_key` keep the guarded description/expected-output fallback. Added regression coverage for a template replay with different inputs. Focused replay tests (7), Ruff, and mypy pass.

Vidit-Ostwal:
Hey @ShivangiRay, CI is failing on a existing test case, mind checking?

you can try running uv run pytest -k test_replay_setup_context locally as well

Vidit-Ostwal (commented):
**[P2] Update the legacy replay fixture for the new identity contract**

`test_replay_setup_context` constructs a stored `TaskOutput` whose description differs from the source task’s description. Production task execution persists the task description in this field, and replay now correctly uses it to validate legacy records. Update the fixture to use `task1.description` and include `expected_output`, while retaining assertions for restored context.

ShivangiRay:
Addressed in 2c1cf47. I updated the legacy `test_replay_setup_context` fixture so each stored output uses its source task description and retains `expected_output`, while preserving the context assertions. I also reran the focused replay tests: 4 passed.

ShivangiRay:
Hi @Vidit—I've addressed the requested fixture update and pushed the fix. The required workflows are currently awaiting maintainer approval. Could you please approve them when you have a moment? Thank you!

Vidit-Ostwal (approved):
Approved after verifying the updated replay fixtures and focused replay tests.

PR #7160, opened 2026-08-31
feat: add llmman as an OpenAI-compatible provider

### LINKED ISSUE #7216: [FEATURE] Support llmman as an OpenAI-compatible provider
Feature Area: Integration with external tools

## Problem

[llmman](https://github.com/llmmanorg/llmman) runs local models distributed as OCI artifacts and serves an OpenAI-compatible API on port 17434. CrewAI already ships first-class support for the other local OpenAI-compatible servers (Ollama, vLLM), but there is no way to point a crew at llmman without hand-configuring `base_url` on every `LLM`.

## Proposal

Register `llmman` alongside the existing entries:

- A `ProviderConfig` in `OPENAI_COMPATIBLE_PROVIDERS` (`LLMMAN_HOST`, no API key required).
- Runtime routing in `crewai.llm.LLM` so `LLM(model="llmman/qwen3.8")` and `LLM(model="qwen3.8", provider="llmman")` reach `OpenAICompatibleCompletion`.
- CLI scaffolding entries (`PROVIDERS`, `ENV_VARS`, `MODELS`) so `crewai create` offers it.

`LLMMAN_HOST` is a bare `host:port` value, so it needs the same `/v1` base-URL normalization already applied to `OLLAMA_HOST`.

## Additional context

Happy to implement this.

ericcurtin:
Closes #7216

Adds [llmman](https://github.com/llmmanorg/llmman), which runs local models distributed as OCI artifacts and serves an OpenAI-compatible API on port 17434.

- **Provider config.** No API key, so it follows the Ollama entry (`api_key_required=False`, a `default_api_key`) rather than the hosted ones. `LLMMAN_HOST` is a bare `host:port` value, so it joins the `/v1` base-URL normalization. That helper is renamed `_normalize_local_base_url` now that it serves two providers.
- **Runtime routing.** Registered in `SUPPORTED_NATIVE_PROVIDERS`, the model-prefix map and `_get_native_provider`, so both `llmman/qwen3.8` and `LLM(model="qwen3.8", provider="llmman")` reach `OpenAICompatibleCompletion` instead of falling back to LiteLLM.
- **CLI.** Added to `PROVIDERS`, `ENV_VARS` and `MODELS`, using bare model names since llmman resolves those to OCI artifacts itself.

### Testing

```
$ pytest lib/crewai/tests/llms/openai_compatible/
40 passed
```

The existing 35 plus 5 new: registry config, base-URL collision, `LLMMAN_HOST` normalization, and factory routing for both the prefixed and explicit-provider forms. The wider `lib/crewai/tests/llms/` suite shows the same 306 pre-existing failures before and after this branch.

Not verified: no request against a live llmman server.

### AI disclosure

This PR was **AI-assisted** (OpenCode); I reviewed the diff before submitting. Per your policy I have applied the `llm-generated` label — if I lack permission to set it, please add it.

### DIFF
--- a/lib/cli/src/crewai_cli/constants.py
+++ b/lib/cli/src/crewai_cli/constants.py
@@ -58,6 +58,12 @@ ENV_VARS: dict[str, list[dict[str, Any]]] = {
             "API_BASE": "http://localhost:11434",
         }
     ],
+    "llmman": [
+        {
+            "default": True,
+            "API_BASE": "http://localhost:17434",
+        }
+    ],
     "bedrock": [
         {
             "prompt": "Enter your AWS Access Key ID (press Enter to skip)",
@@ -123,6 +129,7 @@ PROVIDERS: list[str] = [
     "groq",
     "huggingface",
     "ollama",
+    "llmman",
     "watson",
     "bedrock",
     "azure",
@@ -269,6 +276,7 @@ MODELS: dict[str, list[str]] = {
         "groq/gemma-7b-it",
     ],
     "ollama": ["ollama/llama3.1", "ollama/mixtral"],
+    "llmman": ["llmman/qwen3.8", "llmman/gemma4"],
     "watson": [
         "watsonx/meta-llama/llama-3-1-70b-instruct",
         "watsonx/meta-llama/llama-3-1-8b-instruct",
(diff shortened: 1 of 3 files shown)


ericcurtin:
GitHub refuses to reopen this one (`state cannot be changed. The feat/llmman-provider branch was force-pushed or recreated`), so I have continued in #7217, which links the tracking issue #7216 as the bot requested. That PR also fixes the CodeRabbit finding here: llmman is now registered in the runtime dispatch in `llm.py`, not just the provider registry.

PR #7172, opened 2026-08-31
Fix summarization chunk token overhead

### LINKED ISSUE #7170: [BUG] Chunking split (#7014) doesn't subtract the summarizer's own prompt overhead, so split chunks can still exceed the LLM's window
## Description

`#7014` (merged as `638842151`, "Handle oversized single messages during chunking") added `_expand_oversized_message` / `_split_text_by_token_limit` in `utilities/agent_utils.py` to split an oversized message into `[Part i/n]` pieces before summarizing, each sized to fit `max_tokens = llm.get_context_window_size()` (`agent_utils.py:1104`).

That budget is computed from the **raw conversation content only**. But the actual request sent to the LLM (`_asummarize_chunks` / `summarize_messages`, `agent_utils.py:1072`) wraps that content in:
- `summarizer_system_message` (~38 tokens)
- the `summarize_instruction` template around `{conversation}` (~185 tokens)

That's **~224 tokens of fixed overhead per chunk that never gets subtracted from the budget**. A chunk sized to exactly fit `get_context_window_size()` reliably comes back ~224 tokens over budget once wrapped, so if the LLM enforces its window like a real provider does, that call fails.

Two consequences:
1. **At any window where ~224 tokens is a non-trivial fraction of `get_context_window_size()`, the fix cannot succeed at all.** Every chunk the splitter produces, no matter how fine-grained, is oversized once wrapped, so summarization exhausts `max_retry_limit` and the task fails.
2. **In that regime, chunking is worse than not chunking.** Each failed attempt restarts `execute_task()` from scratch (`agent/core.py` `_handle_execution_error` → `execute_task`), so a fixed multi-chunk conversation burns `chunks_per_attempt × (max_retry_limit + 1)` LLM calls before failing. That's more than the code before `#7014` needs, which sends one oversized call, fails, and gives up in `max_retry_limit + 1` calls total.

## Steps to Reproduce

No API key needed. The LLM is a minimal fake that enforces its declared window the way a real provider does (see [Screenshots/Code snippets](#screenshotscode-snippets) for why this is faithful to real provider behavior, not an artifact of faking it).

1. `pip install crewai==1.15.18`
2. Save the script below as `repro.py`
3. `python repro.py`
4. Observe the printed `call(): N tokens (window=1000)` lines, followed by an unhandled `ValueError` after 3 attempts

## Expected behavior

Either:
- Chunking should keep every produced chunk under the LLM's actual window (accounting for the summarizer's own prompt overhead), so summarization succeeds whenever the pre-`#7014` code would also eventually succeed with a smaller window; or
- If a window is genuinely too small for even one chunk to fit, that should fail no worse (in call count) than not attempting to chunk at all, not several times worse.

## Screenshots/Code snippets

```python
from crewai import Agent, Crew, Task


class TinyWindowLLM(BaseLLM):
    def __init__(self, window: int):
        super().__init__(model="tiny-window-test")
        self.window = window

    def get_context_window_size(self) -> int:
        return int(self.window * 0.85)  # matches CrewAI's own CONTEXT_WINDOW_USAGE_RATIO

    def supports_function_calling(self) -> bool:
        return False

    def supports_stop_words(self) -> bool:
        return True

    def _render(self, messages):
        if isinstance(messages, str):
            return messages
        return "\n".join(f"{m.get('role')}: {m.get('content', '')}" for m in messages)

    def call(self, messages, tools=None, callbacks=None, available_functions=None, **kwargs):
        rendered = self._render(messages)
        tokens = len(rendered) // 4
        print(f"  call(): {tokens} tokens (window={self.window})")
        if tokens > self.window:
            raise ValueError(
                f"This model's maximum context length is {self.window} tokens. "
                f"However, your messages resulted in {tokens} tokens. "
                "Please reduce the length of the messages."
            )
        return "Final Answer: done"

    async def acall(self, messages, **kwargs):
        return self.call(messages, **kwargs)


WINDOW = 1000  # fails here; try 260-800 for a more dramatic case, 1800+ to see it succeed

llm = TinyWindowLLM(window=WINDOW)
agent = Agent(
    role="Senior Research Analyst",
    goal="Summarize a long prompt",
    backstory="x" * 5000,
    llm=llm,
    respect_context_window=True,  # default, shown for clarity
    max_retry_limit=2,            # default
)
task = Task(
    description="Summarize this payload:\n" + "y" * 7000,
    expected_output="A short summary.",
    agent=agent,
)
Crew(agents=[agent], tasks=[task], memory=False).kickoff()
```

### Observed output (window=1000)
```
call(): 3079 tokens (window=1000)   # initial agent call, correctly rejected
call(): 1077 tokens (window=1000)   # chunk 1 of the split summary -- still over budget
call(): 1077 tokens (window=1000)   # chunk 2
call(): 348 tokens (window=1000)    # chunk 3 (fits, but the other two didn't)
# ... repeats 2 more times (max_retry_limit=2 => 3 total attempts) ...
ValueError: This model's maximum context length is 1000 tokens. However, your messages
resulted in 1077 tokens. Please reduce the length of the messages.
```
Each 1077-token chunk is a piece of the split content (budgeted to `get_context_window_size() = 850`) plus the ~224-token summarizer wrapper: `850 + 224 = 1074 ≈ 1077`. The chunk never had a chance of fitting. Splitting it smaller doesn't help, since the fixed overhead is per-chunk, not proportional to content size.

### This is not an artifact of the fake LLM
The repro's `TinyWindowLLM.call()` raises a plain `ValueError`, not something litellm-shaped. So it's fair to ask whether that actually reaches the same code path a real provider's error would. It does, and here's the trace through crewAI's own source that shows why:

1. A real provider call through litellm raises `litellm.exceptions.ContextWindowExceededError` (module `litellm.exceptions`), typically carrying the same `"maximum context length"` wording OpenAI's API uses.
2. `crewai.llm.LLM`'s response handlers (`llm.py:1094`, `1287`, `1438`, `1710`) catch the raw exception, check it against `LLMContextLengthExceededError._is_context_limit_error()` (`utilities/exceptions/context_window_exceeding_exception.py:42-44`, substring match on phrases including `"maximum context length"`), and **re-raise it as CrewAI's own `LLMContextLengthExceededError`** (`raise LLMContextLengthExceededError(error_msg) from e`), whose module is `crewai.utilities.exceptions.context_window_exceeding_exception`, not `litellm.*`.
3. `crew_agent_executor.py`'s except block has an early exit (`if e.__class__.__module__.startswith("litellm"): raise e`) that skips the retry/summarize path for unrecognized litellm errors (auth, rate limits, etc.). Because step 2 already re-wrapped the exception under CrewAI's own module, this check does not fire for context-length errors specifically, and execution falls through to `is_context_length_exceeded(e)` → `handle_context_length()` → the same chunking path exercised above.

So a real provider's context-length error lands in exactly the same place the repro's `ValueError` does; the fake LLM only skips the litellm round-trip, not the code path.

## Operating System

Other (specify in additional context): macOS 26.5.2 (build 25F84), not yet in the dropdown list. Not OS-specific, the bug is in pure Python token-budget arithmetic.

## Python Version

3.11

## crewAI Version

1.15.18

## crewAI Tools Version

N/A, not used by the repro (no tools involved).

## Virtual Environment

Venv

## Evidence

Two checkouts were compared, **not** a single commit's before/after:

- **"no chunk-splitting" baseline**: `638842151^` (`9b1f4938f02671d2066e336c8bfdcfb2f27c99ee`), the commit immediately before `#7014` merged. An oversized message is sent unsplit; no `_expand_oversized_message` exists yet.
- **"current chunk-splitting" baseline**: tag `1.15.18` (`4bc5d2924218e892bd0bc91b46352b49b0d3a740`), the current release, which contains `#7014` plus ~40 later commits. Three of those touch the same files (`6ad3bf939`, `9652af6ae`, `d0e920862`); none of them alter chunk-budget arithmetic.

Swept `window` from 260 to 4000 against both checkouts (fixed 5000-char backstory + 7000-char task description, `TinyWindowLLM` above, N=10 repeats per point, identical prompt-hash sequences across all repeats, no nondeterminism observed):

| window | `638842151^` (no splitting) | `1.15.18` (current splitting) | `1.15.18` outcome |
|---|---|---|---|
| 260 | 6 calls, fails | 30 calls, fails | **5x more calls than no splitting at all** |
| 500 | 6 calls, fails | 18 calls, fails | 3x more calls |
| 800 | 6 calls, fails | 12 calls, fails | 2x more calls |
| 1000 | 6 calls, fails | 12 calls, fails | 2x more calls |
| 1500 | 6 calls, fails | 9 calls, fails | 1.5x more calls |
| 1800 | 6 calls, fails | 3 calls, **succeeds** | splitting now helps |
| 2200+ | 2 calls, succeeds | 2 calls, succeeds | window large enough that neither needs to split |

The fix only pays off once the window is large enough to amortize the ~224-token wrapper overhead relative to `get_context_window_size()`. Below that, it doesn't just fail to help. It multiplies the number of failed LLM calls before giving up, on the current release, not just at the original fix commit.

## Possible Solution

Subtract the summarizer's own fixed overhead from the budget used to size chunks, e.g. in `summarize_messages` (`agent_utils.py:1104`):

```python
overhead = _estimate_token_count(
    I18N_DEFAULT.slice("summarizer_system_message")
    + I18N_DEFAULT.slice("summarize_instruction")
)
max_tokens = max(1, llm.get_context_window_size() - overhead)
```

so `_split_messages_into_chunks` / `_expand_oversized_message` budget against the space actually left over after the wrapper, not the full `get_context_window_size()`.

## Additional context

None beyond the above.

Dikshj:
## Summary
- subtract fixed summarizer prompt overhead from the chunk budget before splitting messages
- estimate chunk sizes using the formatted summary text, including role labels
- add a regression test that asserts every chunked summarization request stays within the LLM context budget

Fixes #7170

## Tests
- uv run pytest lib/crewai/tests/utilities/test_agent_utils.py::TestSplitMessagesIntoChunks lib/crewai/tests/utilities/test_agent_utils.py::TestExpandOversizedMessage lib/crewai/tests/utilities/test_agent_utils.py::TestNormalizeMessagesForChunking lib/crewai/tests/utilities/test_agent_utils.py::TestParallelSummarization -q -o addopts=
- uv run ruff check lib/crewai/src/crewai/utilities/agent_utils.py lib/crewai/tests/utilities/test_agent_utils.py

### DIFF
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -948,6 +948,11 @@ def _estimate_message_summary_tokens(msg: LLMMessage) -> int:
     return _estimate_token_count(_format_messages_for_summary([msg]))
 
 
+def _estimate_chunk_summary_tokens(chunk: list[LLMMessage]) -> int:
+    """Estimate tokens for a complete rendered summary chunk."""
+    return _estimate_token_count(_format_messages_for_summary(chunk))
+
+
 def _split_text_by_token_limit(text: str, max_tokens: int) -> list[str]:
     """Split text into parts each estimated to fit within max_tokens."""
     if not text:
@@ -962,24 +967,31 @@ def _split_text_by_token_limit(text: str, max_tokens: int) -> list[str]:
 
 def _expand_oversized_message(msg: LLMMessage, max_tokens: int) -> list[LLMMessage]:
     """Split a message whose content alone exceeds max_tokens into sub-messages."""
-    msg_text = message_content_text(msg)
-    if _estimate_token_count(msg_text) <= max_tokens:
+    if _estimate_message_summary_tokens(msg) <= max_tokens:
         return [msg]
 
-    # Reserve budget for the [Part i/n] prefix and summary role label.
-    summary_label_tokens = _estimate_message_summary_tokens({**msg, "content": ""})
-    body_max_tokens = max(1, max_tokens - 5 - summary_label_tokens)
-    parts = _split_text_by_token_limit(msg_text, body_max_tokens)
-    total_parts = len(parts)
-    expanded: list[LLMMessage] = []
+    msg_text = message_content_text(msg)
+    if not msg_text:
+        return [msg]
 
-    for index, part in enumerate(parts, start=1):
-        part_content = (
-            f"[Part {index}/{total_parts}]\n{part}" if total_parts > 1 else part
-        )
-        expanded.append({**msg, "content": part_content})
+    body_max_tokens = max_tokens
+    while body_max_tokens > 0:
+        parts = _split_text_by_token_limit(msg_text, body_max_tokens)
+        total_parts = len(parts)
+        expanded: list[LLMMessage] = []
 
-    return expanded
+        for index, part in enumerate(parts, start=1):
+            part_content = (
+                f"[Part {index}/{total_parts}]\n{part}" if total_parts > 1 else part
+            )
+            expanded.append({**msg, "content": part_content})
+
+        if all(_estimate_message_summary_tokens(part) <= max_tokens for part in expanded):
+            return expanded
+
+        body_max_tokens -= 1
+
+    return [msg]
 
 
 def _normalize_messages_for_chunking(
@@ -1015,18 +1027,16 @@ def _split_messages_into_chunks(
 
     chunks: list[list[LLMMessage]] = []
     current_chunk: list[LLMMessage] = []
-    current_tokens = 0
 
     for msg in normalized:
-        msg_tokens = _estimate_message_summary_tokens(msg)
+        candidate_chunk = [*current_chunk, msg]
+        candidate_tokens = _estimate_chunk_summary_tokens(candidate_chunk)
 
-        if current_chunk and (current_tokens + msg_tokens) > max_tokens:
+        if current_chunk and candidate_tokens > max_tokens:
             chunks.append(current_chunk)
             current_chunk = []
-            current_tokens = 0
 
         current_chunk.append(msg)
-        current_tokens += msg_tokens
 
     if current_chunk:
         chunks.append(current_chunk)
@@ -1058,6 +1068,9 @@ def _summarization_prompt_overhead_tokens() -> int:
         + _estimate_token_count(
             I18N_DEFAULT.slice("summarize_instruction").format(conversation="")
         )
+        # Estimating the empty wrapper and conversation separately can lose
+        # one token to integer-division rounding when they are joined.
+        + 1
     )
 
 
@@ -1129,10 +1142,14 @@ def summarize_messages(
     if not non_system_messages:
         return
 
-    max_tokens = max(
-        1, llm.get_context_window_size() - _summarization_prompt_overhead_tokens()
+    available_content_tokens = (
+        llm.get_context_window_size() - _summarization_prompt_overhead_tokens()
     )
-    chunks = _split_messages_into_chunks(non_system_messages, max_tokens)
+    chunks = (
+        _split_messages_into_chunks(non_system_messages, available_content_tokens)
+        if available_content_tokens > 0
+        else [non_system_messages]
+    )
 
     total_chunks = len(chunks)
 
--- a/lib/crewai/src/crewai/utilities/agent_utils.py
+++ b/lib/crewai/src/crewai/utilities/agent_utils.py
@@ -893,51 +893,6 @@ def message_content_text(msg: LLMMessage) -> str:
     return str(content)
 
 
-def _split_text_by_token_limit(text: str, max_tokens: int) -> list[str]:
-    """Split text into parts each estimated to fit within max_tokens."""
-    if not text:
-        return []
-    if _estimate_token_count(text) <= max_tokens:
-        return [text]
-
-    # Inverse of _estimate_token_count (len // 4): each slice is at most max_tokens.
-    max_chars = max(1, max_tokens * 4)
-    return [text[i : i + max_chars] for i in range(0, len(text), max_chars)]
-
-
-def _expand_oversized_message(msg: LLMMessage, max_tokens: int) -> list[LLMMessage]:
-    """Split a message whose content alone exceeds max_tokens into sub-messages."""
-    msg_text = message_content_text(msg)
-    if _estimate_token_count(msg_text) <= max_tokens:
-        return [msg]
-
-    # Reserve budget for the [Part i/n] prefix added to each sub-message.
-    body_max_tokens = max(1, max_tokens - 5)
-    parts = _split_text_by_token_limit(msg_text, body_max_tokens)
-    total_parts = len(parts)
-    expanded: list[LLMMessage] = []
-
-    for index, part in enumerate(parts, start=1):
-        part_content = (
-            f"[Part {index}/{total_parts}]\n{part}" if total_parts > 1 else part
-        )
-        expanded.append({**msg, "content": part_content})
-
-    return expanded
-
-
-def _normalize_messages_for_chunking(
-    messages: list[LLMMessage], max_tokens: int
-) -> list[LLMMessage]:
-    """Return non-system messages with oversized entries split to fit max_tokens."""
-    normalized: list[LLMMessage] = []
-    for msg in messages:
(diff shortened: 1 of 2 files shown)


Dikshj:
Updated in 5646ba3 to address the two review comments:

- when the summarizer wrapper consumes the whole context window, summarize_messages now keeps a single summarization attempt instead of forcing a 1-token chunk budget and fanning out
- chunk sizing now estimates the complete rendered summary chunk, including role labels, tool-call rendering, separators, and rounding effects
- oversized-message expansion now checks the rendered summary representation before deciding whether a message fits

Validation:
- uv run pytest lib/crewai/tests/utilities/test_agent_utils.py::TestSplitMessagesIntoChunks lib/crewai/tests/utilities/test_agent_utils.py::TestExpandOversizedMessage lib/crewai/tests/utilities/test_agent_utils.py::TestNormalizeMessagesForChunking lib/crewai/tests/utilities/test_agent_utils.py::TestParallelSummarization -q -o addopts=
- uv run ruff check lib/crewai/src/crewai/utilities/agent_utils.py lib/crewai/tests/utilities/test_agent_utils.py

VANDRANKI (commented):
Community review, not a maintainer approval, does not clear the merge gate.

I read the full diff in `lib/crewai/src/crewai/utilities/agent_utils.py`. The bug being fixed: the old `_split_messages_into_chunks` sized chunks by summing raw per-message token estimates (`_estimate_token_count(message_content_text(msg))`) against `max_tokens = llm.get_context_window_size()`. That undercounts two things that get added before the chunk is actually sent to the summarizer LLM: (1) the "role: " labels and formatting `_format_messages_for_summary` adds per message, and (2) the fixed system-message/instruction-wrapper overhead added once around the whole chunk. A chunk sized right up against the raw context window could still overflow once wrapped for the actual summarization call — the PR title's "chunk token overhead" bug.

The fix does two things. `summarize_messages` now computes `available_content_tokens = llm.get_context_window_size() - _summarization_prompt_overhead_tokens()` before chunking, reserving room for the wrapper. I checked `_summarization_prompt_overhead_tokens()`: it sums the token estimate of the system-message template and the instruction template (rendered with an empty conversation placeholder), plus a documented `+1` fudge factor for integer-division rounding loss between the two separately-estimated pieces — reasonable. Second, `_split_messages_into_chunks` now measures each growing chunk's actual formatted size via `_estimate_chunk_summary_tokens` (which calls `_format_messages_for_summary` on the whole candidate chunk) instead of summing independent per-message estimates, so the overhead from message formatting is measured directly rather than assumed away. `_expand_oversized_message` was changed the same way: it checks the formatted size via `_estimate_message_summary_tokens`, and if a naive char-based split still doesn't fit once formatted (the `[Part i/n]` prefix pushes it over), it backs `body_max_tokens` down one token at a time and retries until every part fits.

One thing worth flagging as a genuine cost, not a correctness bug: `_split_messages_into_chunks` now calls `_estimate_chunk_summary_tokens`, which re-formats and re-token-counts the entire `candidate_chunk` every time a message is appended. For a long non-system message history that ends up in a single chunk, that's O(n^2) in the number of messages, versus the old O(n) running-sum. Likely negligible for typical histories, but worth knowing if very long conversations get summarized often — an incremental token count (track the delta each message adds instead of re-rendering the whole chunk) would keep this linear without giving up the accuracy improvement. Similarly, `_expand_oversized_message`'s new decrement-by-one retry loop is bounded by `max_tokens` iterations in the worst case, which could be slow for large context windows if the formatted overhead is consistently underestimated by a wide margin — I didn't hit this in the tests I read, just noting it as a theoretical tail case.

I read the core logic changes closely but did not review every line of the updated test file (`tests/utilities/test_agent_utils.py`) beyond confirming the new helper functions (`_estimate_chunk_summary_tokens`, etc.) are imported and exercised.

This is a real correctness fix — accounting for the actual prompt overhead instead of assuming raw content tokens equal what gets sent — with a sensible, documented approach to the rounding fudge factor. The quadratic re-formatting in the chunking loop is the one thing I'd want the author to consider before this scales to very long conversations.

Vidit-Ostwal:
Closing with #7170. The 15% `CONTEXT_WINDOW_USAGE_RATIO` buffer already covers summarizer prompt overhead on every model we ship. We are not taking a second subtract (or the chunk-sizing rewrite) on top of that.

PR #7185, opened 2026-09-01
[OSS-149] Accept 1/yes/on on telemetry disable flags

Vidit-Ostwal:
## Related issue

Fixes [OSS-149](https://linear.app/crewai/issue/OSS-149/crewai-disable-telemetry1-silently-leaves-telemetry-enabled-crewai)

## Summary

- `CREWAI_DISABLE_TELEMETRY=1` left telemetry on because the gate only matched `true`
- Accept `true` / `1` / `yes` / `on` (any case) on `CREWAI_DISABLE_TELEMETRY`, `CREWAI_DISABLE_TRACKING`, and `OTEL_SDK_DISABLED`; warn and leave telemetry on for unrecognized values
- One parser on `crewai-core` `Telemetry`; the framework class uses it so the two copies cannot drift
- Docs (`en` / `ar` / `ko` / `pt-BR`) and README list the accepted values and the `CREWAI_DISABLE_TRACKING` alias

## Verification

## Additional context

`OTEL_SDK_DISABLED=1` now disables CrewAI's exporter. The OpenTelemetry SDK itself still only honors `true` for other instrumentation in the process.


lib/crewai/src/crewai/telemetry/telemetry.py
```diff
@@ -22,6 +22,7 @@
 
 from crewai_core.telemetry import (
     CommonAttributesSpanProcessor,
+    Telemetry as CoreTelemetry,
```
lorenzejay, line 25: why we aliasing here?

Vidit-Ostwal: Just better for readability, we are basically taking a method from some other class, which has exactly same name. Just for readbility purposes, I added a alias.

Vidit-Ostwal: Lmk if needs to be changed.

lorenzejay (approved):
lgtm

PR #7188, opened 2026-09-01
[OSS-151] Fail closed when human-feedback emit cannot classify

Vidit-Ostwal:
- Resolve `@human_feedback` collapse LLM from the decorator, then project `MODEL` / `MODEL_NAME` / `OPENAI_MODEL_NAME`, then `DEFAULT_LLM_MODEL`, instead of hardcoding OpenAI.
- Raise `HumanFeedbackCollapseError` when collapse cannot run or cannot match an emit label, instead of routing to `emit[0]`. Empty skip still uses `default_outcome`.
- A typed rejection no longer takes the first emit path when the LLM call fails.


lib/crewai/src/crewai/flow/human_feedback.py
```diff
@@ -370,7 +402,7 @@ def _distill_and_store_lessons(
 def human_feedback(
     message: str,
     emit: Sequence[str] | None = None,
-    llm: str | BaseLLM | None = "gpt-5.4-mini",
+    llm: str | BaseLLM | None = None,
```
lorenzejay, line 405: nice

lib/crewai/src/crewai/flow/runtime/__init__.py
```diff
@@ -3606,13 +3608,11 @@ async def _run_human_feedback_step(
...
     ) -> Any:
         llm = feedback_definition.llm
-        llm_instance = (
-            _deserialize_llm_from_context(llm) if isinstance(llm, (str, dict)) else llm
-        )
         emit = feedback_definition.emit
         default_outcome = feedback_definition.default_outcome
         metadata = feedback_definition.metadata
         learn = feedback_definition.learn and self.memory is not None
+        llm_instance = _resolve_llm_instance(llm) if (emit or learn) else llm
```
lorenzejay, line 3622: lgtm

lorenzejay (approved):

PR #7189, opened 2026-09-01
feat(events): expose native tool call ids

### LINKED ISSUE #7178: [FEATURE] Expose the native tool call id on tool usage events
### Feature Area

Core functionality

N/A

 Add an optional `call_id: str | None = None ` field to `ToolUsageEvent (crewai/events/types/tool_usage_events.py),` populated at the emit sites in `_execute_single_native_tool_call`.

CrewAI already has the value. `_parse_native_tool_call `sets `call_id = getattr(tool_call, "id", f"call_{id(tool_call)}") `(agents/crew_agent_executor.py:835), and it's threaded through `_execute_single_native_tool_call(call_id=...) `into the assistant/tool messages, so message history pairs correctly. It's a parameter of the very function that emits `ToolUsageStartedEvent / ToolUsageFinishedEvent` — already in scope, no plumbing needed.

Today those events carry tool_name, tool_args, agent and task ids, but no call id, so no observability consumer can attribute a tool execution to the model request that asked for it.

- Wrapping BaseTool.run (what OpenLIT and Arize OpenInference do): the id isn't in scope there, so it can't be recovered.
-  Reconstructing from message history: the tool_call_id does survive into messages, but only post-hoc — an instrumentor needs it at span creation, not after the run.
- Pairing by ordering: ambiguous under max_usage_count retries and the parallel batch path in `_handle_native_tool_calls`.

Adding the field is the smallest change that makes any of this work.

### Additional context

Observability. The OTel GenAI semantic conventions list `gen_ai.tool.call.id` on the execute_tool span as `Recommended` "If available", described as "The tool call identifier." Because CrewAI doesn't surface it, no CrewAI instrumentor can populate it, and teams doing trace-based auditing lose the ability to attribute a tool execution to the specific model request that asked for it.
Verified against crewai 1.15.18.

### Willingness to Contribute

z2Ace0107:
Fixes #7178

## Summary

- Add an optional `call_id` field to `ToolUsageEvent`.
- Populate it for started, finished, and error events emitted by native tool execution paths.
- Preserve backward compatibility for event producers that do not have a native model call id.

This allows event consumers to correlate a tool execution with the model tool call that initiated it.

## Verification

- Added regression coverage for successful and failing native tool calls.
- Python 3.12.13: focused native tool tests passed.
- Ruff check passed.
- Ruff format check passed.
- Mypy passed for the changed production files.

## Additional context

This PR was prepared with AI assistance. I reviewed the changed implementation and tests. Please apply the required `llm-generated` label.


IdoGol24:
@z2Ace0107 Thanks for picking this up ,the coverage is broader than I described. I pointed only at `_execute_single_native_tool_call`, you also caught the emit sites in `agent_utils.py `and `experimental/agent_executor.py.
`
One gap before it lands. `_parse_native_tool_call `is unchanged, so it still does:

 `call_id = getattr(tool_call, "id", f"call_{id(tool_call)}")`
 
The new call_id is therefore sometimes the provider's id and sometimes a value CrewAI invented, with nothing distinguishing them. The synthesized branch derives from` id(tool_call)` - the object address - which CPython reuses after garbage collection, so two distinct tool calls in a long-running crew can be emitted with the same `call_id`. A consumer trusting it as a correlation key mis-pairs in exactly the retry and parallel-batch cases the field exists to disambiguate.

Smallest fix (my own suggestion): populate `call_id `only when the provider supplied one. The semconv treatment for `gen_ai.tool.call.id `is "when available" - absence is a valid answer, a fabricated id isn't. 
Otherwise keep the fallback and add a source marker.

Do the new tests cover the no-provider-id path? From the diff they look like they exercise cases where an id is present

Also, I've accidentally closed the issue by mistake.. Could a maintainer flip back open #7178?  It shouldn't read as resolved while this is still in review.

z2Ace0107:
Hi maintainers,

Thanks for reviewing this PR. The CodeRabbit feedback about distinct fallback IDs was addressed in commit `4b04f22`.

The branch is synchronized with the latest upstream `main` (`143e9021`) in merge commit `6279bb29`. Nine focused native-tool regression tests, Ruff checks, format checks, and the diff check pass locally. The GitHub Actions workflows are currently waiting for maintainer approval (`action_required`).

Could a maintainer please approve and run the pending workflows, add the required `llm-generated` label, and review whether this PR is ready to merge? I’m happy to address any follow-up feedback on this branch.

Thanks!

IdoGol24:
Can we have this merged please? @caike @zeroasterisk @renatonitta

PR #7068, opened 2026-08-20
fix(memory): preserve reusable scope configs

tandede:
## Summary

- preserve caller-owned memory view mappings while inferring legacy discriminators
- keep `MemoryScope` and `MemorySlice` configuration dictionaries reusable across validations
- add regression coverage for dependency extraction, path normalization, and discriminator inference

## Problem

The memory view validators modified their input mappings in place. Constructing a `MemoryScope` removed the runtime `memory` dependency from the caller's dictionary, so reusing the same configuration produced a second scope that was not bound to its memory. `MemorySlice` also rewrote the caller's scope paths, and legacy discriminator inference inserted `memory_kind` into the original mapping.

## Implementation

Each validation path now makes a shallow copy immediately before it mutates the mapping. The validated values, normalized paths, serialization format, and public API remain unchanged.

## Validation

- regression tests before the fix: 3 failed
- regression tests after the fix: 3 passed
- `uv run --python 3.12 pytest -q lib/crewai/tests/memory/`: 150 passed
- `uv run --python 3.12 pre-commit run --files lib/crewai/src/crewai/memory/memory_scope.py lib/crewai/tests/memory/test_unified_memory.py`: Ruff, formatting, and mypy passed

## AI assistance

AI assistance was used to inspect the validator mutation paths and draft the regression tests. I reproduced the behavior before the change, reviewed every changed line, and ran the validations above locally.

The required `llm-generated` label could not be applied because GitHub does not grant external contributors permission to label upstream pull requests. Please apply it during triage.

Fixes #7204

PR #7200, opened 2026-09-02
fix: bump pypdf to 6.16.2 for GHSA-jp53-mhqp-8xcg

Vidit-Ostwal:
## Summary
- Raise the pypdf floor from 6.15.0 to 6.16.1 in the workspace override and `crewai-files`
- Refresh `uv.lock` to 6.16.2 so pip-audit no longer reports GHSA-jp53-mhqp-8xcg, GHSA-23w6-3w8w-8484, and GHSA-763m-79hh-57f2
- Take 6.16.2 rather than 6.16.1 so the latest patched release is locked (6.16.2 is bugfixes only)

## Test plan
- [ ] Vulnerability Scan / pip-audit no longer reports `pypdf==6.15.0` for those three GHSAs
- [ ] `uv.lock` resolves `pypdf` 6.16.2

## Follow-up
- After this greens CI, merge #7068 (`fix(memory): preserve reusable scope configs`, #7204)


Vidit-Ostwal:
Follow-up after this lands: merge #7068 (memory scope validation no longer mutates caller-owned configs; #7204). CI on that PR is red because of the same pip-audit / cancelled-test noise this bump should clear.

lib/crewai-files/pyproject.toml
```diff
@@ -19,8 +19,8 @@ dependencies = [
 
 [tool.uv]
 exclude-newer = "3 days"
-# pypdf 6.14.2 is a security fix newer than the global supply-chain cutoff.
-exclude-newer-package = { pypdf = "2026-06-24T00:00:00Z" }
+# pypdf 6.16.1+ patches GHSA-jp53-mhqp-8xcg, GHSA-23w6-3w8w-8484, GHSA-763m-79hh-57f2.
+exclude-newer-package = { pypdf = "2026-08-24T00:00:00Z" }
```
vinibrsl: Do we need this since we already have `~=` fetching only `6.16.*`?

Vidit-Ostwal: Good catch — dropped it. `~=6.16.1` plus the global 3-day cutoff already admits 6.16.2; the package-specific date was leftover from when 6.14.2 was still inside that window. The root `pyproject.toml` cutoff stays, since the workspace override is `>=6.16.1,<7`.

Vidit-Ostwal: Follow-up: dropped the root `exclude-newer-package` pypdf date too. 6.16.2 is already older than the global 3-day window, so `pypdf>=6.16.1,<7` is enough.

vinibrsl (approved):

PR #7234, opened 2026-09-03
fix(llm): route all DashScope models through native provider

### LINKED ISSUE #7233: [BUG] DashScope non-Qwen models bypass the native provider and ignore DASHSCOPE_BASE_URL
### Description

While working on another Ollama-related fix I noticed that `_matches_provider_pattern` in `llm.py` only treats a DashScope model as natively supported when its name starts with `qwen`. Any other DashScope model skips CrewAI's native OpenAI-compatible provider and falls through to the LiteLLM fallback instead.

That matters because Alibaba's documentation says the same `compatible-mode/v1` endpoint also serves DeepSeek, Kimi, GLM and MiniMax. So the prefix check is excluding models the endpoint genuinely supports, and users of those models quietly lose their DashScope configuration.

### Steps to Reproduce

1. Set DASHSCOPE_API_KEY and DASHSCOPE_BASE_URL.
2. Build one LLM with a Qwen model and one with a non-Qwen DashScope model.
3. Compare the class that gets created and the resolved base_url.

DASHSCOPE_API_KEY=************  DASHSCOPE_BASE_URL=https://my-dashscope.example.com/v1 python -c "
from crewai import LLM
for m in ['dashscope/qwen-max', 'dashscope/deepseek-v3']:
    l = LLM(model=m)
    print(m, type(l).__name__, getattr(l, 'base_url', None))
"

### Expected behavior

I'd expect both models to use the native OpenAICompatibleCompletion provider and to respect DASHSCOPE_BASE_URL, since DashScope serves them through the same endpoint.

### Screenshots/Code snippets

Here's what I get:

```
dashscope/qwen-max                -> OpenAICompatibleCompletion   base_url=https://dashscope-intl.aliyuncs.com/compatible-mode/v1
dashscope/deepseek-v3             -> LLM                          base_url=None
dashscope/llama-3.3-70b-instruct  -> LLM                          base_url=None
```

And with DASHSCOPE_BASE_URL set, only the Qwen model picks it up:

```
dashscope/qwen-max      base_url=https://my-dashscope.example.com/v1
dashscope/deepseek-v3   base_url=None
```

### Operating System

### Python Version

3.12

### crewAI Version

1.15.18

### crewAI Tools Version

1.15.18

### Virtual Environment

Venv

### Evidence

I dug into where the non-Qwen model actually ends up. `litellm.get_llm_provider("dashscope/deepseek-v3")` returns:

```
('deepseek-v3', 'dashscope', 'dummy', 'https://dashscope.aliyuncs.com/compatible-mode/v1')
```

So the api_base is LiteLLM's own default rather than the DASHSCOPE_BASE_URL I set. The host is also different from the native path's `dashscope-intl.aliyuncs.com`, which means the same `dashscope/` prefix ends up at two different regional endpoints purely based on the model name.

One more consequence: litellm is an optional extra in `lib/crewai/pyproject.toml`. On a plain `pip install crewai` a non-Qwen DashScope model raises an ImportError saying the model "did not match any supported native provider" — which isn't accurate, since dashscope is configured in OPENAI_COMPATIBLE_PROVIDERS and would serve it fine.

I also checked the tests: every dashscope test uses a Qwen model (test_llm_creates_openai_compatible_for_dashscope uses dashscope/qwen-turbo), so nothing covers the non-Qwen path.

For the model support claim, Alibaba's docs list "Qwen large language models (commercial and open-source editions), Qwen-VL, Qwen-Coder, Qwen-Omni, Qwen-Math, DeepSeek, Kimi, GLM, MiniMax":
https://www.alibabacloud.com/help/en/model-studio/compatibility-of-openai-with-dashscope

### Possible Solution

Looking at the other providers, dashscope and deepseek are the only two in `_matches_provider_pattern` that restrict by model prefix — ollama, hosted_vllm, cerebras, openrouter and snowflake all return True. Since DashScope serves several model families, returning True for dashscope would bring it in line with the rest.

I wasn't sure whether the prefix check was deliberate, so I wanted to raise it before writing anything. Happy to open a PR if you agree with the direction.

### Additional context

My OS is macOS 26.5.2 and I'm on Python 3.13.13, neither of which is in the dropdowns.

Unrelated to the routing itself, the hardcoded default `https://dashscope-intl.aliyuncs.com/compatible-mode/v1` might be stale — the current docs list `dashscope-us.aliyuncs.com` and workspace-scoped `*.maas.aliyuncs.com` URLs.
I'm running from a source checkout of main at commit 3d72c707 (22 commits past the 1.15.18 tag), not a released install.

Alphaxiaoteng:
## Summary
- Fixes #7233: `_matches_provider_pattern` only treated DashScope models as native when the name started with `qwen`, so `dashscope/deepseek-v3` (and other supported families) fell through to LiteLLM and ignored `DASHSCOPE_BASE_URL`.
- Treat dashscope like other OpenAI-compatible providers (`return True`).
- Add a regression covering a non-Qwen DashScope model with a custom base URL.

## Test plan
- [x] Code + regression added next to existing DashScope factory test
- [ ] CI: `test_openai_compatible.py` DashScope cases


parthiban-sivakumar:
Hi @Alphaxiaoteng,

Thanks for picking this up so quickly! The change looks right to me and matches what I had in mind when I filed the issue.

Two small thoughts, both optional. The comment mentions DeepSeek/Kimi/GLM but the test only covers deepseek-v3 — parametrising over one more family would back up the comment. And it might be worth asserting llm.model == "deepseek-v3" too, just to confirm the model string is passed through correctly.

One thing I left out of the issue deliberately: the hardcoded default https://dashscope-intl.aliyuncs.com/compatible-mode/v1 may also be stale, since the current docs list dashscope-us.aliyuncs.com and workspace-scoped *.maas.aliyuncs.com URLs. Probably a separate issue rather than something for this PR.

Thanks!

Vidit-Ostwal (commented):
Looks correct for #7233. The factory only takes the native `OpenAICompatibleCompletion` path when `_matches_provider_pattern` returns true, so the `qwen*` gate was sending `dashscope/deepseek-v3` (and the other compatible-mode families) to LiteLLM and dropping `DASHSCOPE_BASE_URL`. `return True` matches the other OpenAI-compatible providers and Alibaba's compatible-mode docs.

The new test hits the reported case (`dashscope/deepseek-v3` + custom `DASHSCOPE_BASE_URL`). That's enough; the matcher is now unconditional, so parametrizing Kimi/GLM is optional. Asserting `llm.model == "deepseek-v3"` would be a cheap extra check that the prefix is stripped.

One leftover: the comment a few lines above still says `(DeepSeek, Dashscope) restrict to their own model prefixes`. After this change only DeepSeek does. Please update that so it doesn't contradict the new DashScope branch.

CodeRabbit is clean. Not behind `main`, so no rebase needed. Leave the default `dashscope-intl` URL alone — that's a separate follow-up.

lib/crewai/src/crewai/llm.py
```diff
@@ -573,7 +573,8 @@ def _matches_provider_pattern(cls, model: str, provider: str) -> bool:
             return True
 
         if provider == "dashscope":
-            return model_lower.startswith("qwen")
+            # DashScope's OpenAI-compatible endpoint serves Qwen plus DeepSeek/Kimi/GLM/etc.
+            return True
```
Vidit-Ostwal, line 577: The comment a few lines above still says `(DeepSeek, Dashscope) restrict to their own model prefixes`. After this change only DeepSeek does — please update that so it doesn't contradict this branch.

Vidit-Ostwal (commented):
Leftover comment next to the DashScope matcher.

Vidit-Ostwal (approved):
Fixes #7233. `return True` matches the other OpenAI-compatible providers, so `dashscope/deepseek-v3` (and Kimi/GLM) take the native path and honor `DASHSCOPE_BASE_URL`. The new test covers the reported case.

Please rebase onto `main` (`mergeStateStatus` is BEHIND). Non-blocking leftover: the comment a few lines above still says `(DeepSeek, Dashscope) restrict to their own model prefixes` — after this change only DeepSeek does.

parthiban-sivakumar:
thanks @Alphaxiaoteng for picking this up so quickly, and thanks @Vidit-Ostwal for the review and merge.

PR #7236, opened 2026-09-03
fix(tools): run async tools safely when an event loop is already running

### LINKED ISSUE #7235: [BUG] Async tools raise RuntimeError when called from an async Flow method
### Description

Calling an async tool from an `async def` Flow method raises `RuntimeError: asyncio.run() cannot be called from a running event loop`, and the tool never executes.

`BaseTool.run()` and `CrewStructuredTool.invoke()` bridge async tools into sync call sites with a bare `asyncio.run()`. That works from ordinary sync code, but a Flow runs `async def` methods on an event loop, and `asyncio.run()` refuses when a loop is already running on the thread.

For MCP tools it's worse — `MCPToolWrapper._run` catches the error and returns it as the tool's result string, so the agent receives an error message as though it were the tool's output and never learns the call didn't happen.

### Steps to Reproduce

1. Define an async tool with the `@tool` decorator.
2. Call it from a sync Flow `@start` method, then from an `async def` one.
3. Only the async method fails.

### Expected behavior

An async tool should behave the same from an async Flow method as from a sync one. Both `@tool` on an async function and async Flow methods are documented features, so combining them shouldn't crash.

### Screenshots/Code snippets

```python
import asyncio
from crewai.flow.flow import Flow, start
from crewai.tools import tool

@tool("probe")
async def probe(x: str) -> str:
    """An async tool. Input: any string."""
    await asyncio.sleep(0)
    return "ok:" + x

class SyncFlow(Flow):
    @start()
    def step(self):
        return probe.run(x="hi")

class AsyncFlow(Flow):
    @start()
    async def step(self):
        return probe.run(x="hi")
```

Result:

```
sync  @start method -> ok:hi
async @start method -> RuntimeError: asyncio.run() cannot be called from a running event loop
```

### Operating System

### Python Version

3.12

### crewAI Version

1.15.18

### crewAI Tools Version

1.15.18

### Virtual Environment

Venv

### Evidence

Unguarded `asyncio.run()` calls:

- `lib/crewai/src/crewai/tools/base_tool.py:341` (`BaseTool.run`)
- `lib/crewai/src/crewai/tools/base_tool.py:552` (`Tool.run`)
- `lib/crewai/src/crewai/tools/structured_tool.py:441,446` (`CrewStructuredTool.invoke`)
- `lib/crewai/src/crewai/tools/mcp_tool_wrapper.py:79` (`MCPToolWrapper._run`)

The same situation is already handled correctly in three other places:

- `lib/crewai/src/crewai/tasks/llm_guardrail.py:25` — a `_run_sync` helper whose docstring says "handling an already-running event loop"
- `lib/crewai/src/crewai/tools/mcp_native_tool.py:93` — the same guard inline
- Agent execution raises a clear error: "Agent execution was invoked synchronously from within a running event loop. Use `agent.kickoff_async()`"

For MCP, the sync path attempts a real connection while the async path returns the error as a string:

```
MCP sync context  -> MCP tool execution failed after 3 attempts: Connection timed out
MCP async context -> Error executing MCP tool echo: asyncio.run() cannot be called from a running event loop
```

Async `@tool` functions are a supported feature: documented at `docs/edge/en/concepts/tools.mdx:250`, covered by `lib/crewai/tests/tools/test_async_tools.py`, and given an explicit `inspect.iscoroutinefunction` branch in `structured_tool.py:440`.

Things I checked that are **not** affected: `crew.kickoff_async()`, async task callbacks via `kickoff_async`, and `crew.kickoff()` inside an async Flow method (that one is guarded with a helpful error).

### Possible Solution

Extract the existing `_run_sync` helper from `llm_guardrail.py` into a shared utility and use it at the unguarded sites: check for a running loop, and if there is one, run the coroutine on a worker thread.

That stops the crash but still blocks the loop for the duration of the tool call. Awaiting `arun()` on async call paths would be the fuller fix, though it's a larger change touching the executor and Flow runtime. Happy to open a PR for the smaller version if that direction works for you.

### Additional context

My OS is macOS 26.5.2 and I'm on Python 3.13.13, neither of which is in the dropdowns. I'm running from a source checkout of main at commit 3d72c707.

parthiban-sivakumar:
Fixes #7235

## Problem

`BaseTool.run()`, `CrewStructuredTool.invoke()` and `MCPToolWrapper._run()` bridge async tools into sync call sites with a bare `asyncio.run()`. That works from ordinary sync code, but `asyncio.run()` raises when a loop is already running on the thread — which is exactly what happens when a tool is called from an `async def` Flow method:

```
sync  @start method -> ok:hi
async @start method -> RuntimeError: asyncio.run() cannot be called from a running event loop
```

For MCP tools the failure was silent. `MCPToolWrapper._run` caught the error and returned it as the tool's result string, so an agent received an error message as though it were the tool's output and never learned the call hadn't happened:

```
MCP sync context  -> MCP tool execution failed after 3 attempts: Connection timed out
MCP async context -> Error executing MCP tool echo: asyncio.run() cannot be called from a running event loop
```

## Fix

Adds `crewai.utilities.asyncio_utils.run_sync`, which checks for a running loop and runs the coroutine on a worker thread when there is one, carrying context variables across. This is the approach already used in `tasks/llm_guardrail.py` and `tools/mcp_native_tool.py` — I lifted it into a shared helper rather than inventing anything.

Applied at the five bare `asyncio.run()` call sites in `base_tool.py`, `structured_tool.py` and `mcp_tool_wrapper.py`.

## Scope

I deliberately left `llm_guardrail.py` and `mcp_native_tool.py` alone even though they now duplicate the helper, to keep the diff focused on the bug. Happy to de-duplicate them in a follow-up if you'd like.

## Trade-off

This stops the crash but doesn't make the path truly async — the worker thread blocks the caller while the tool runs, the same as the existing helpers do today. Awaiting `arun()` on async call paths would be the fuller fix, but that touches the executor and Flow runtime, so I didn't want to make that call unilaterally. Happy to look at it separately if you'd prefer that direction.

## Testing

New `TestAsyncToolInsideRunningLoop` covers an async `@tool`, a `BaseTool` whose `_run` returns a coroutine, `CrewStructuredTool.invoke()`, and `MCPToolWrapper._run()` — each invoked from inside a running event loop. All four fail on `main` and pass here.

Locally: `tests/tools/` + `tests/llms/` 892 passed, 20 skipped; `tests/agents/`, `tests/task/`, `tests/mcp/`, `tests/experimental/` 527 passed, 16 skipped. ruff, ruff-format and mypy clean.

Note: `pip-audit` is currently failing on `main` as well, unrelated to this change.

---

### DIFF
--- a/lib/crewai/src/crewai/tools/base_tool.py
+++ b/lib/crewai/src/crewai/tools/base_tool.py
@@ -40,6 +40,7 @@ from crewai.tools.structured_tool import (
 )
 from crewai.tools.tool_failure import ToolFailure, ToolFailurePolicy, ToolFailureReason
 from crewai.types.callback import SerializableCallable, _resolve_dotted_path
+from crewai.utilities.asyncio_utils import run_sync
 from crewai.utilities.string_utils import sanitize_tool_name
 
 
@@ -338,7 +339,7 @@ class BaseTool(BaseModel, ABC):
         result = self._run(*args, **kwargs)
 
         if asyncio.iscoroutine(result):
-            result = asyncio.run(result)
+            result = run_sync(result)
 
         return result
 
@@ -549,7 +550,7 @@ class Tool(BaseTool, Generic[P, R]):
         result = self.func(*args, **kwargs)
 
         if asyncio.iscoroutine(result):
-            result = asyncio.run(result)
+            result = run_sync(result)
 
         return result  # type: ignore[return-value]
 
--- a/lib/crewai/src/crewai/tools/mcp_tool_wrapper.py
+++ b/lib/crewai/src/crewai/tools/mcp_tool_wrapper.py
@@ -5,6 +5,7 @@ from collections.abc import Callable, Coroutine
 from typing import Any
 
 from crewai.tools import BaseTool
+from crewai.utilities.asyncio_utils import run_sync
 
 
 MCP_CONNECTION_TIMEOUT = 15
@@ -76,7 +77,7 @@ class MCPToolWrapper(BaseTool):
             Result from the MCP tool execution
         """
         try:
-            return asyncio.run(self._run_async(**kwargs))
+            return run_sync(self._run_async(**kwargs))
         except asyncio.TimeoutError:
             return f"MCP tool '{self.original_tool_name}' timed out after {MCP_TOOL_EXECUTION_TIMEOUT} seconds"
         except Exception as e:
--- a/lib/crewai/src/crewai/tools/structured_tool.py
+++ b/lib/crewai/src/crewai/tools/structured_tool.py
@@ -22,6 +22,7 @@ from pydantic import (
 from typing_extensions import Self
 
 from crewai.tools.tool_failure import ToolFailure, ToolFailurePolicy
+from crewai.utilities.asyncio_utils import run_sync
 from crewai.utilities.logger import Logger
 from crewai.utilities.pydantic_schema_utils import (
     create_model_from_schema,
@@ -438,12 +439,12 @@ class CrewStructuredTool(BaseModel):
         self._increment_usage_count()
 
         if inspect.iscoroutinefunction(self.func):
-            return asyncio.run(self.func(**parsed_args, **kwargs))
+            return run_sync(self.func(**parsed_args, **kwargs))
 
         result = self.func(**parsed_args, **kwargs)
 
         if asyncio.iscoroutine(result):
-            return asyncio.run(result)
+            return run_sync(result)
 
         return result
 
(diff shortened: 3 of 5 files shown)


Vidit-Ostwal (commented):
Keeping this PR — it has the best coverage of the running-loop crash (`BaseTool.run`, `Tool.run`, `CrewStructuredTool.invoke`, and `MCPToolWrapper`) plus tests.

Please:
1. Retarget the issue link from #7235 to #6978 (the surviving tracker; #7235 is a duplicate).
2. Add the `CrewStructuredTool.invoke` test from #6979 for a plain `def` that returns a coroutine (not only `async def`).

CodeRabbit looks clean. No rebase needed from here.

Vidit-Ostwal (commented):
Closing in favor of #6979, the original PR for #6978. Follow along there — please port the `MCPToolWrapper` change and running-loop tests onto that PR.

Vidit-Ostwal:
Closing in favor of the older #6979, which already tracks #6978. Follow along there.

PR #7237, opened 2026-09-03
fix: resolve human input result display

HamdanS12346:
## Related issue

Fixes #6072

## Summary

When `human_input=True` is set on a task, CrewAI asks the operator to review the final answer. Previously, if both the agent and crew were running with `verbose=False`, the feedback prompt appeared but the final answer panel was hidden.

This updates both executor paths so the final answer is shown when human input is pending, even if normal verbose logging is disabled. Non-verbose runs without human review remain quiet.

## Verification

Added regression coverage for:
- Default `AgentExecutor`
- Legacy `CrewAgentExecutor`

Ran:

```powershell
.\.venv\Scripts\python.exe -m pytest -n 0 lib/crewai/tests/agents/test_agent_executor.py -k "show_logs_is_visible_when_human_input_requested"

<!-- crewai-require-issue-check -->

### DIFF
--- a/lib/crewai/src/crewai/agents/crew_agent_executor.py
+++ b/lib/crewai/src/crewai/agents/crew_agent_executor.py
@@ -1542,13 +1542,18 @@ class CrewAgentExecutor(BaseAgentExecutor):
         if self.agent is None:
             raise ValueError("Agent cannot be None")
 
+        should_show_output = (
+            self.agent.verbose
+            or (hasattr(self, "crew") and getattr(self.crew, "verbose", False))
+            or self.ask_for_human_input
+        )
+
         future = crewai_event_bus.emit(
             self.agent,
             AgentLogsExecutionEvent(
                 agent_role=self.agent.role,
                 formatted_answer=formatted_answer,
-                verbose=self.agent.verbose
-                or (hasattr(self, "crew") and getattr(self.crew, "verbose", False)),
+                verbose=should_show_output,
             ),
         )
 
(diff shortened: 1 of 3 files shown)


VANDRANKI (commented):
Community review, not a merge gate.

Small, focused fix touching two parallel executors: crewai/agents/crew_agent_executor.py's _show_logs and crewai/experimental/agent_executor.py's _show_logs (the legacy and newer executor implementations). Both previously computed the AgentLogsExecutionEvent's verbose flag as agent.verbose or crew.verbose only. This PR adds or self.ask_for_human_input (legacy) / or self.state.ask_for_human_input (experimental) to that condition.

The behavior change makes sense given the surrounding context: this event is what actually prints the agent's answer to the terminal, and it fires right before the human-input flow asks the user for feedback on that same answer. With verbose off and no human-input flag, a user asked "do you want to give feedback on this?" would have no way to see what they're being asked to review. Gating visibility on ask_for_human_input as well as verbose fixes a real UX gap, not just a style change.

Both executors get matching test coverage (test_show_logs_is_visible_when_human_input_requested for the experimental one via _build_executor, test_legacy_show_logs_is_visible_when_human_input_requested for the legacy one via CrewAgentExecutor.model_construct), each setting agent.verbose=False and crew.verbose=False explicitly and asserting event.verbose is True once the human-input flag is set. That isolates the new condition rather than just re-testing the pre-existing or-chain.

What I didn't verify: I read only this diff's two functions, not the rest of either executor class, so I'm taking on faith that self.ask_for_human_input (legacy) and self.state.ask_for_human_input (experimental) are the actual live fields the rest of each class's human-input flow sets, rather than confirming that by reading where they're assigned elsewhere in these files.

PR #3618, opened 2025-09-28
Release/v1.0.0

lorenzejay:

PR #7239, opened 2026-09-03
fix(crew): identify tasks by object when copying instead of by key

### LINKED ISSUE #7238: [BUG] Crew.copy() mis-wires task context on duplicate task text and raises KeyError for external context tasks
### Description

`Crew.copy()` re-wires task `context` links by looking tasks up in a `task_mapping` dictionary keyed by `Task.key`, which is `md5(description | expected_output)`. Identifying tasks by their text rather than by object identity causes two separate problems:

1. **Silent mis-wiring.** Two tasks with the same `description` and `expected_output` produce the same key, so the second overwrites the first in the mapping. A cloned task's `context` is then wired to the wrong clone, with no error.
2. **Crash.** If a task's `context` references a `Task` that isn't in `crew.tasks`, the lookup raises `KeyError`. The crew validator explicitly allows this — `validate_context_no_future_tasks` skips context tasks that aren't crew members (`crew.py:875`) — so the crew builds and `kickoff()` runs, but copying it fails.

`copy()` isn't called directly by users; it's called by `kickoff_for_each()`, `train()` and `test()`, so all three break.

### Steps to Reproduce

1. Build a crew with two tasks having identical `description` and `expected_output`, plus a third task whose `context` points at the first. Copy the crew and inspect the cloned context.
2. Build a crew with one task whose `context` references a task that is not in `crew.tasks`, then call `kickoff_for_each()`.

### Expected behavior

Copying should preserve each task's context link to the correct corresponding clone regardless of duplicate task text, and a context task that isn't part of the crew should not crash the copy.

### Screenshots/Code snippets

Mis-wiring:

```python
t1 = Task(description="same", expected_output="same", agent=a)
t2 = Task(description="same", expected_output="same", agent=a)
t3 = Task(description="third", expected_output="third", agent=a, context=[t1])

c = Crew(agents=[a], tasks=[t1, t2, t3]).copy()
c.tasks[2].context[0] is c.tasks[0]   # expected True
```

```
t1.key == 4612eff9ffb119e2b48d29b0cb544bb0
t2.key == 4612eff9ffb119e2b48d29b0cb544bb0
same key? True
t3.context points to -> clone of t2 (WRONG)
```

Crash:

```python
outside = Task(description="outside", expected_output="o", agent=a)
inside  = Task(description="inside {topic}", expected_output="i", agent=a, context=[outside])
crew = Crew(agents=[a], tasks=[inside])          # builds fine
crew.kickoff_for_each(inputs=[{"topic": "x"}])
```

```
crew built OK - validator allows an external context task
kickoff_for_each -> KeyError: '4c3d3b983623ce9b2b44460e9a58d6a4'
```

`train()` and `test()` fail identically. The error message is only an md5 hash, with no task name or hint about the cause.

### Operating System

### Python Version

3.12

### crewAI Version

1.15.18

### crewAI Tools Version

1.15.18

### Virtual Environment

Venv

### Evidence

- `lib/crewai/src/crewai/task.py:601` — `Task.key` is `md5("|".join([description, expected_output]))`, so it is not unique per task instance.
- `lib/crewai/src/crewai/crew.py:2145-2162` — `task_mapping[task.key] = cloned_task`, then `task_mapping[context_task.key]` when re-wiring context.
- `lib/crewai/src/crewai/task.py:1179` — `Task.copy()` does the same unguarded `task_mapping[context_task.key]` lookup; this is the one that raises first.
- `lib/crewai/src/crewai/crew.py:875` — `validate_context_no_future_tasks` skips context tasks not in the crew, so external context tasks are treated as valid.
- Callers of `copy()`: `crew.py:958` (`train`), `crew.py:1116` (`kickoff_for_each`), `crew.py:2254` (`test`).

`git log -S "task_mapping"` shows this code has been unchanged since `d1343b96e Release/v1.0.0 (#3618)`.

Reproduced on two machines with no LLM calls and no network.

### Possible Solution

Key the mapping by `id(task)` instead of `task.key`, and fall back to the original object when a context task isn't in the mapping:

- `crew.py`: `task_mapping: dict[int, Any]`; `task_mapping[id(task)] = cloned_task`; `task_mapping.get(id(context_task), context_task)`
- `task.py`: signature `task_mapping: dict[int, Task]`; `task_mapping.get(id(context_task), context_task)`

`id()` is unique per object, which removes the collision. The `.get(..., context_task)` fallback keeps a context task that isn't a crew member as-is, since there is no clone to point at.

Related observation, not addressed here: `Task.copy()` resolves the cloned agent with `get_agent_by_role` (`task.py:1185`), matching on the role string. Two agents sharing a role would have the same ambiguity. I haven't tested that and mention it only in case it's worth a separate look.

### Additional context

My OS is macOS Tahoe 26.5.2 and I'm on Python 3.13.13, neither of which is in the dropdowns. Running from a source checkout of main at commit 92eb5f9.

parthiban-sivakumar:
Fixes #7238

## Problem

`Crew.copy()` re-wires task `context` links through a mapping keyed by `Task.key`, which is `md5(description | expected_output)`. Identifying tasks by their text rather than by object identity causes two separate failures.

**Silent mis-wiring.** Two tasks sharing `description` and `expected_output` produce the same key, so the second overwrites the first and a cloned task's context points at the wrong clone. Nothing raises:

```
t1.key == 4612eff9ffb119e2b48d29b0cb544bb0
t2.key == 4612eff9ffb119e2b48d29b0cb544bb0
same key? True
t3.context points to -> clone of t2 (WRONG)
```

**Crash.** A context task that isn't a crew member has no entry, so the lookup raises `KeyError`. `validate_context_no_future_tasks` skips context tasks outside the crew (`crew.py:875`), so such a crew is valid and `kickoff()` runs fine — but copying it fails:

```
crew built OK - validator allows an external context task
kickoff_for_each -> KeyError: '4c3d3b983623ce9b2b44460e9a58d6a4'
```

`copy()` isn't called directly by users, so this surfaces through `kickoff_for_each()` (`crew.py:1116`), `train()` (`crew.py:958`) and `test()` (`crew.py:2254`). All three fail identically, with only an md5 hash as the message.

## Fix

Key the mapping by `id(task)`, which is unique per object, and fall back to the original object when a context task has no clone:

- `crew.py`: `task_mapping: dict[int, Any]`, `task_mapping[id(task)]`, `task_mapping.get(id(context_task), context_task)`
- `task.py`: matching signature change and the same `.get(...)` fallback in `Task.copy()`, which is where the `KeyError` is raised first

The fallback keeps an external context task as-is, since it isn't a crew member and has no clone to point at. `Task.copy()`'s docstring already described the parameter as "Dictionary mapping task IDs to Task instances", so this brings the code in line with the documented intent.

## Testing

Two regression tests in `test_crew.py`, both failing on `main`:

- `test_crew_copy_preserves_context_with_duplicate_task_text` — asserts the collision exists, then that context still resolves to the right clone
- `test_crew_copy_keeps_context_task_outside_the_crew` — asserts copying no longer raises and the external context task is preserved

`test_crew.py` 133 passed, 1 skipped. `tests/task/`, `tests/crew/`, `tests/agents/`, `test_checkpoint.py` 492 passed, 16 skipped. ruff, ruff-format and mypy clean.

Reproduced on two machines with no LLM calls and no network. `git log -S "task_mapping"` shows this code unchanged since `d1343b96e Release/v1.0.0 (#3618)`.

## Out of scope

`Task.copy()` resolves the cloned agent via `get_agent_by_role` (`task.py:1185`), matching on the role string — two agents sharing a role would have the same ambiguity. I haven't tested that and left it alone to keep this diff focused; happy to look separately if it's worth pursuing.

Note: `pip-audit` is currently failing on `main` as well, unrelated to this change.

---


VANDRANKI (commented):
Community review, not a merge gate.

Traced `Crew.copy()` and `Task.copy()`. The bug is real and the fix is correct: `Task.key` is documented in the new test as an md5 of `description + expected_output`, so two tasks with identical text (a plausible real scenario, e.g. two near-duplicate tasks in a crew) collide on the same key. The old code used `task_mapping[task.key] = cloned_task`, so the second task's clone silently overwrote the first task's entry, and any task whose `context` pointed at the first task got re-wired to the second task's clone instead after copying. Switching the mapping key to `id(task)` (object identity) fixes the collision, since Python object identity is unique per live object, and since `task_mapping` and the original `task` objects (referenced by `self.tasks`) all stay alive for the full synchronous duration of `copy()`, there's no id-reuse-after-gc risk here.

The second fix is just as real: the old lookup was a bare `task_mapping[context_task.key]`, which would raise `KeyError` if a context task wasn't itself one of the crew's own tasks. That's a legitimate, valid configuration per the PR's own reference to `validate_context_no_future_tasks` skipping non-member context tasks, so a crew with an external context task couldn't be copied at all before this fix. `task_mapping.get(id(context_task), context_task)` fixes that by falling back to the original object when it's not in the mapping.

Both fixes are directly tested: `test_crew_copy_preserves_context_with_duplicate_task_text` reproduces the exact key collision (asserts `first.key == second.key` to document why the bug existed) and confirms the third task's context now points at the correct clone; `test_crew_copy_keeps_context_task_outside_the_crew` confirms copying no longer raises when a context task isn't a crew member. I traced both test bodies against the fix and they exercise exactly the code paths changed.

PR #7254, opened 2026-09-03
feat(tools): add AnyAPI catalog search, endpoint schema and run tools

### LINKED ISSUE #7253: [FEATURE] Add an AnyAPI tool for catalog-driven access to scraping and data APIs
### Feature Area

Integration with external tools

NA

`crewai_tools` already carries a lot of single-vendor data tools - Apify, Bright Data, Exa, Firecrawl, Jina, Oxylabs, Scrapfly, SerpApi, Serper, Tavily. Each one gives an agent a fixed capability from one vendor, and adding a new data source means adding a new tool and a new key.

I would like a tool for [AnyAPI](https://getanyapi.com), which is shaped differently: it is a gateway to hundreds of scraping and data APIs behind one key, billed per request in USD with no subscription. Because it is a catalog rather than a single endpoint, the useful shape is three tools that an agent chains:

1. `AnyApiSearchTool` - ranked search over the catalog to find the endpoint that returns the data the task needs
2. `AnyApiDescribeTool` - that endpoint's input and output JSON Schema and its USD price, which the agent reads before its first call
3. `AnyApiRunTool` - execute the endpoint

The point of the split is that the agent picks the data source at run time instead of the developer picking it at import time. An agent asked for "recent reviews of this restaurant" searches, finds the Google Maps reviews endpoint, reads its schema, and runs it, without anyone having added a Maps-specific tool first.

Deliberately **not** included: any hardcoded endpoint list, per-platform convenience tools, or catalog caching. Routing, pricing and schemas belong to the gateway, so the tool stays a thin transport layer and does not need a release when the catalog changes.

A single tool taking an endpoint slug and an input. Rejected because the agent has no way to learn which slugs exist or what they accept, and every AnyAPI input schema is strict, so an invented field name fails the call. The search-then-describe-then-run split is what makes it usable without the developer knowing the catalog.

Modelling it on `apify_actors_tool`, which is the closest existing shape. It solves the same discovery problem for one vendor's actors; this generalises it across vendors.

### Additional context

I work on AnyAPI, so this is a vendor integration and I am declaring that up front. The implementation is written, tested and ready: three `BaseTool` subclasses over a shared base, 14 unit tests, `getanyapi` declared as an optional extra the way the sibling tools declare theirs. `ruff check`, `ruff format --check`, `uv lock --check` and the tool-spec generator gate all pass, and I have smoke tested the search and describe paths against the live API.

Happy to be told this belongs outside the repo as a third-party package instead - I would rather ask than open an unwanted PR.

### Willingness to Contribute

kev1n:
## Related issue

Fixes #7253

## Summary

Adds three tools for [AnyAPI](https://getanyapi.com), a gateway to hundreds of scraping and data APIs behind one key, billed per request in USD with no subscription. It sits alongside the existing Apify, Bright Data, Exa, Firecrawl, Jina, Oxylabs, Scrapfly, SerpApi, Serper and Tavily tools, but is catalog-shaped rather than single-purpose, so the agent picks the data source at run time instead of the developer picking it at import time.

| Tool | Name shown to the agent | Purpose |
|---|---|---|
| `AnyApiSearchTool` | AnyAPI catalog search | ranked search for the endpoint that returns the needed data |
| `AnyApiDescribeTool` | AnyAPI endpoint schema | that endpoint's input/output JSON Schema and USD price |
| `AnyApiRunTool` | AnyAPI endpoint run | execute the endpoint |

All three sit on a shared `AnyApiToolBase` that owns the lazy `getanyapi` import, the `api_key` / `ANYAPI_API_KEY` fallback and client construction. Each tool's description points at the next one by name and states that the schema is strict, so an agent that skips the describe step is told why its call failed.

Responses are dumped with `by_alias=True`, so the agent sees the published wire names (`costUsd`, `inputSchema`, `maxPer1kUsd`) rather than internal ones. `AnyApiRunTool` catches an insufficient-balance error and appends the remaining USD plus the dashboard URL, so an empty wallet is a readable failure rather than an opaque one.

There is deliberately no hardcoded endpoint list, no per-platform convenience tool and no catalog caching. Routing, pricing and schemas belong to the gateway, so this stays a thin transport layer and needs no release when the catalog changes.

`getanyapi` is declared as an optional extra following the sibling tools, and `ImportError` carries install instructions per `BUILDING_TOOLS.md` rather than prompting interactively, which would be wrong inside an agent run.

## Verification

```
$ uv run ruff check lib/
All checks passed!

$ uv run ruff format --check lib/
922 files already formatted

$ uv lock --check
Resolved 477 packages in 11ms

$ cd lib/crewai-tools && uv run pytest tests/tools/anyapi_tool_test.py
14 passed, 18 warnings in 4.68s

$ uv run pytest lib/crewai-tools/tests/test_generate_tool_specs.py
10 passed
```

`mypy` scoped to the new files reports `Success: no issues found`. Repo-wide `mypy lib/` reports 584 errors in 96 files on this branch - identical to `main` checked out in a throwaway worktree with the same interpreter, so none of them are mine.

Beyond the mocked unit tests, I smoke tested the two free catalog paths against the live production API with a real key. `AnyApiSearchTool(query="instagram profile", platform="instagram", limit=3)` returned 27 total matches led by `instagram.basic_profile` at `maxPer1kUsd: 1.5`, and `AnyApiDescribeTool` on that slug returned the real input schema and nested USD pricing. `AnyApiRunTool` was instantiated but not executed, since it bills real money; its behaviour is covered by unit tests.

`tool.specs.json` was regenerated with the repo's own `generate_tool_specs.py`; it added exactly the three new entries with zero existing entries removed or changed. `uv.lock` was hand-spliced rather than regenerated, so the diff is 19 added lines instead of a few thousand.

## Additional context

**Disclosure:** I work on AnyAPI, so this is a vendor-contributed integration, and it was drafted with an AI coding tool and then reviewed by a human against the live API before submitting. Flagging both rather than leaving you to find out.

**One supply-chain note.** `getanyapi>=0.35.3` was published on 2026-09-01, and this repo sets `exclude-newer = "3 days"`. `uv lock --check` passes because it validates the lock against the manifests, but a full re-resolution run before 2026-09-04 would exclude the package. This resolves itself with time; I did not add an `exclude-newer-package` override because that widens the change into your supply-chain policy. Say the word if you would rather have the pin.

**Not included:** a docs page. `AGENTS.md` requires syncing any English docs edit to `ar`, `ko` and `pt-BR`, which would multiply the diff. Happy to follow up with the four MDX files and the `docs.json` entry if you want the tools discoverable there.

### DIFF
--- a/lib/crewai-tools/pyproject.toml
+++ b/lib/crewai-tools/pyproject.toml
@@ -93,6 +93,10 @@ apify = [
     "langchain-apify>=0.1.2,<1.0.0",
 ]
 
+getanyapi = [
+    "getanyapi>=0.35.3",
+]
+
 databricks-sdk = [
     "databricks-sdk>=0.46.0",
 ]
--- a/lib/crewai-tools/src/crewai_tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/__init__.py
@@ -8,6 +8,9 @@ from crewai_tools.aws.bedrock.knowledge_base.retriever_tool import (
 from crewai_tools.aws.s3.reader_tool import S3ReaderTool
 from crewai_tools.aws.s3.writer_tool import S3WriterTool
 from crewai_tools.tools.ai_mind_tool.ai_mind_tool import AIMindTool
+from crewai_tools.tools.anyapi_tool.anyapi_describe import AnyApiDescribeTool
+from crewai_tools.tools.anyapi_tool.anyapi_run import AnyApiRunTool
+from crewai_tools.tools.anyapi_tool.anyapi_search import AnyApiSearchTool
 from crewai_tools.tools.apify_actors_tool.apify_actors_tool import ApifyActorsTool
 from crewai_tools.tools.arxiv_paper_tool.arxiv_paper_tool import ArxivPaperTool
 from crewai_tools.tools.brave_search_tool.brave_image_tool import BraveImageSearchTool
@@ -226,6 +229,9 @@ from crewai_tools.tools.zapier_action_tool.zapier_action_tool import ZapierActio
 
 __all__ = [
     "AIMindTool",
+    "AnyApiDescribeTool",
+    "AnyApiRunTool",
+    "AnyApiSearchTool",
     "ApifyActorsTool",
     "ArxivPaperTool",
     "BedrockInvokeAgentTool",
--- a/lib/crewai-tools/src/crewai_tools/tools/__init__.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/__init__.py
@@ -1,4 +1,9 @@
 from crewai_tools.tools.ai_mind_tool.ai_mind_tool import AIMindTool
+from crewai_tools.tools.anyapi_tool import (
+    AnyApiDescribeTool,
+    AnyApiRunTool,
+    AnyApiSearchTool,
+)
 from crewai_tools.tools.apify_actors_tool.apify_actors_tool import ApifyActorsTool
 from crewai_tools.tools.arxiv_paper_tool.arxiv_paper_tool import ArxivPaperTool
 from crewai_tools.tools.brave_search_tool.brave_image_tool import BraveImageSearchTool
@@ -213,6 +218,9 @@ from crewai_tools.tools.zapier_action_tool.zapier_action_tool import ZapierActio
 
 __all__ = [
     "AIMindTool",
+    "AnyApiDescribeTool",
+    "AnyApiRunTool",
+    "AnyApiSearchTool",
     "ApifyActorsTool",
     "ArxivPaperTool",
     "BraveImageSearchTool",
--- a/lib/crewai-tools/src/crewai_tools/tools/anyapi_tool/README.md
+++ b/lib/crewai-tools/src/crewai_tools/tools/anyapi_tool/README.md
@@ -1,0 +1,106 @@
+# AnyAPI Tools Documentation
+
+## Description
+
+AnyAPI is a unified marketplace for scraping and data APIs: any API, one wallet, USD, no
+subscriptions. Reach hundreds of third-party APIs (social media, search results, web data)
+through one key and one normalized interface; pay per request in real dollars; failed calls
+are never charged - AnyAPI fails over across providers automatically under one price.
+
+These three tools give a CrewAI agent the whole loop, and they are meant to be used in this
+order:
+
+- **AnyApiSearchTool**: search the catalog for an API that returns the data you need
+- **AnyApiDescribeTool**: fetch one endpoint's input schema, output schema and USD price
+- **AnyApiRunTool**: execute an endpoint with a JSON input and get normalized output back
+
+Every AnyAPI input schema is strict, so an agent must read the schema with
+`AnyApiDescribeTool` before its first `AnyApiRunTool` call on a slug. An invented field name
+fails the call.
+
+## Installation
+
+```shell
+pip install crewai[tools] getanyapi
+```
+
+or
+
+```shell
+uv add crewai-tools --extra getanyapi
+```
+
+## Examples
+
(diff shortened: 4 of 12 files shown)


kev1n:
Thanks for the review - pushed `e388293` addressing it. I worked each finding against the actual code rather than applying them wholesale, so here is what changed and what did not, with the reasoning for the ones I did not act on.

## Fixed

**`api_key` in serialized tool state** (`anyapi_base.py:38`) - real, and I reproduced it: `model_dump()`, `model_dump_json()` and `repr()` all contained the key. `api_key` is no longer a model field; it is a constructor argument, and only the SDK client holds it. `ANYAPI_API_KEY` remains the declared env var. Side effect worth noting for review: `tool.specs.json` no longer lists `api_key` as an init param for these three tools - regenerated with `python -m crewai_tools.generate_tool_specs`, and the diff is confined to our three entries.

**`ValueError` escaping the error handler** (`anyapi_describe.py:36`, `anyapi_search.py:64`, `anyapi_run.py:47`, and the `balance()` call at `anyapi_run.py:59`) - one real defect with four call sites, so it got one fix rather than four. The SDK raises a plain `ValueError` out of `parse_describe` and `parse_search`, and a durable envelope that fails `RequestSnapshot` validation raises a pydantic `ValidationError` uncaught. All four now catch `(AnyAPIError, ValueError)`; `ValidationError` subclasses `ValueError`, so one mechanism covers both.

**Wallet exhaustion vs per-key cap** (`anyapi_run.py:45`) - correct, and the old hint was actively misleading. Our gateway answers 402 with three distinct codes (`insufficient_balance`, `key_cap_exceeded`, `grant_cap_exceeded`), and "add funds" is the wrong advice when the wallet is fine and the key's own spend cap tripped. It now branches on the code, tells the caller it is a key limit rather than a balance, and skips the balance lookup entirely in that case.

## Not changed, with reasons

**Nullable discovery fields** (`anyapi_base.py:66`) - `exclude_none` only drops declared model fields whose value is null; I checked that it does not touch nulls inside the `Any`-typed data payload, so run output still carries `"bio": null`. What it drops is a top-level `latency: null`, which carries no information for an LLM reading the JSON, and our output is never fed back through `parse_describe`. I would rather leave the shape as-is than add a field an agent cannot use.

**Response-shape-aware execution path for arbitrary slugs** (`anyapi_run.py:43`) - I do not think this one should be built, for two reasons. It is unreachable today: none of the 76 generated platform modules in `getanyapi` 0.35.3 reference `BareRunResult` at all, and the model's own docstring describes it as a future shape. And if a bare-shaped SKU did ship, `_transport.validate_run_result` already converts the `ValidationError` into an `AnyAPIError`, which the handler above turns into a stable message - so the failure mode is a clear error, not a crash. Selecting a parse path per slug would move the gateway's knowledge of response shapes into the client, which is the coupling this toolkit is deliberately built to avoid; if a bare shape ships, the right place to handle it is the SDK's generic `run`, not here.

Happy to be argued out of either of those.

## Gates

`ruff check` clean, `ruff format --check` clean, `uv lock --check` no drift, and the tool tests are now **30 passing** (up from 14; 6 new). Each new test fails against the pre-fix sources - I checked that rather than assuming. Repo-wide mypy is unchanged: zero errors in `anyapi_tool/`.

I also re-ran a live smoke against production for `search` and `describe` through the refactored tools: real ranked results, real input schema, no key present in the serialized tool. I deliberately did not run the execute tool against production, since that spends real money on a live SKU - the run happy path is covered by unit tests only, and I would rather say so than imply broader coverage.

One observation, not something I have touched: the `api_key` serialization pattern flagged above is repo-wide - Firecrawl, Exa, Tavily, Scrapfly, Brightdata and roughly fifteen others declare a public `api_key` field the same way. Out of scope for this PR, but you may want it on the radar.

PR #7257, opened 2026-09-04
fix(state): persist json checkpoints as utf-8

### LINKED ISSUE #7256: JSON checkpoint provider uses locale default encoding

Ghraven:
## Problem

`JsonProvider` persists checkpoint JSON with the platform default text encoding in both the sync and async paths. The matching checkpoint readers also rely on the default encoding.

That makes checkpoint round-trips environment-dependent when serialized state contains non-ASCII text on systems whose default encoding is not UTF-8.

## Before / after

Before this change, checkpoint writes and reads used the process/platform default encoding.

After this change, sync and async checkpoint writes/readers use `encoding="utf-8"`, matching JSON's expected interoperable encoding and keeping checkpoint behavior stable across platforms.

Closes #7256

## Verification

- `python -m py_compile lib/crewai/src/crewai/state/provider/json_provider.py lib/crewai/tests/test_checkpoint.py`
- `git diff --check`
- direct smoke test of `JsonProvider` sync + async non-ASCII checkpoint round trip via file import: passed

I also attempted:

- `python -m pytest lib/crewai/tests/test_checkpoint.py -k "JsonProviderFork or checkpoint_uses_utf8" -q`

That local run was blocked by environment setup in this bare checkout: first the configured `--timeout=60` option required a missing pytest-timeout plugin, then collection required repo dependencies such as `jsonref`. The added tests are narrow and exercise the same provider paths as the passing direct smoke test.

Note: this replaces #7255, which the first-time-contributor issue gate closed before the tracking issue existed.


lib/crewai/tests/test_checkpoint.py
```diff
@@ -378,6 +378,25 @@ def test_checkpoint_writes_to_branch_subdir(self) -> None:
             assert path.endswith(".json")
             assert os.path.isfile(path)
 
+    def test_checkpoint_uses_utf8_for_non_ascii_json(self) -> None:
+        provider = JsonProvider()
+        data = '{"message": "olá niño"}'
+        with tempfile.TemporaryDirectory() as d:
+            path = provider.checkpoint(data, d, branch="main")
+
+            assert Path(path).read_bytes() == data.encode("utf-8")
```
Vidit-Ostwal, line 388: `Path` is used here and again in `test_acheckpoint_uses_utf8_for_non_ascii_json`, but `test_checkpoint.py` does not import it (and this PR does not add `from pathlib import Path`). These tests will raise `NameError` as soon as they run. The author noted they could not run pytest locally, and Actions on this first-time fork PR have not run the test suite yet, so this was not caught.

Add the import, or use `open(path, "rb").read()` to stay consistent with the `os.path` style in this class.

Vidit-Ostwal (changes requested):
The provider change is the right fix and it hits all four I/O paths (`open` / `aiofiles.open` write + `Path.read_text` / `aiofiles.open` read). In scope, no rebase needed.

The new tests will not collect: they call `Path(...)` but never import `pathlib.Path`. That is a `NameError` on first run. Please add `from pathlib import Path` (or drop `Path` and read bytes with `open(path, "rb")`).

Nit, not blocking: `read_bytes() == data.encode("utf-8")` is the right assertion, but on UTF-8 CI it also passes without this patch. Fine as a regression guard once the import is fixed.

CodeRabbit's docstring-coverage warning is noise here — no new functions were added.

Ghraven:
Thanks for catching that, and sorry I missed the import on the first pass.

I added `from pathlib import Path` in `lib/crewai/tests/test_checkpoint.py` and pushed the update. I also re-ran the focused checkpoint coverage locally:

```console
uv run pytest lib/crewai/tests/test_checkpoint.py -k "utf8_for_non_ascii_json" -n0 --allowed-hosts=127.0.0.1,localhost
```

Result: `2 passed, 64 deselected`.

Vidit-Ostwal (approved):
The `Path` import is in, so the new tests will collect. All four JsonProvider I/O paths now pin `encoding="utf-8"`, and the sync/async non-ASCII round-trips cover the reported locale-default bug. LGTM.

PR #4286, opened 2026-01-27
fix: capture per-task token usage to fix race condition in async execution

wingding12:
## Summary

Fixes #4168

The per-task token tracking had a race condition when using threading-based `async_execution` (`task.async_execution=True` with `crew.kickoff()`). Multiple concurrent tasks from the same agent would get incorrect token attribution because `tokens_before` was captured when tasks were queued, but `tokens_after` was captured sequentially when processing futures.

### Problem

When multiple tasks with `async_execution=True` are executed by the same agent:
1. `tokens_before` was captured when the task was queued
2. Tasks run concurrently in threads
3. All concurrent tasks captured similar `tokens_before` values
4. `tokens_after` was captured sequentially in `_process_async_tasks` after calling `future.result()`
5. Later tasks got credited with tokens from earlier tasks that ran in parallel

### Solution

Capture token usage snapshots within the execution context (thread or async task) rather than after futures are resolved:

- Add `token_usage` field to `TaskOutput` to store per-task metrics
- Add `_get_agent_token_usage()` helper to capture token snapshot from agent's LLM
- Add `_calculate_token_delta()` helper to compute token differences
- Modify `_execute_task_async()` to capture tokens within the thread immediately before and after task execution
- Update `_execute_core()` and `_aexecute_core()` for consistent token tracking across all execution paths

### Files Changed

- `lib/crewai/src/crewai/task.py` - Token tracking logic in task execution
- `lib/crewai/src/crewai/tasks/task_output.py` - Add `token_usage` field
- `lib/crewai/tests/task/test_task_token_tracking.py` - Tests for the fix

---

> [!NOTE]
> Adds accurate per-task token tracking and resolves async race conditions in token attribution.
> 
> - Introduces `token_usage` on `TaskOutput` and imports `UsageMetrics`
> - Adds `_get_agent_token_usage()` and `_calculate_token_delta()` helpers
> - Updates `_execute_task_async()` to snapshot tokens before/after execution within the thread and store the delta on the result
> - Applies the same before/after token snapshotting in `_execute_core()` and `_aexecute_core()` so sync and async paths are consistent
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 56c3d02383fc894a812537fd17f0c9bf79f0e9df. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7260, opened 2026-09-04
fix(crew): count each LLM instance once when summing usage metrics

### LINKED ISSUE #7259: [BUG] crew.usage_metrics multiplies token usage by the number of agents sharing an LLM instance
### Description

`Crew.calculate_usage_metrics()` loops over `self.agents` and adds `agent.llm.get_token_usage_summary()` for each one. Those counters are cumulative for the lifetime of the LLM **instance**, so when several agents share the same `LLM` object its total is added once per agent and `crew.usage_metrics` reports N times the real token usage.

Sharing a single `LLM` instance across agents is the usual way to write a crew, so this affects ordinary usage rather than an edge case. Nothing errors — the reported number is simply wrong, and the inflation grows as agents are added.

`get_token_usage_summary()` documents this explicitly:

> The counters are cumulative for the lifetime of this instance: they grow across every call made through it, including calls issued by different agents sharing the instance.

### Steps to Reproduce

1. Create one `LLM` instance and give it to two agents.
2. Run the crew.
3. Compare `crew.usage_metrics.total_tokens` with `llm.get_token_usage_summary().total_tokens`.

### Expected behavior

`crew.usage_metrics` should report the crew's actual token usage. Each distinct LLM instance should contribute its totals exactly once, regardless of how many agents hold a reference to it. Agents with separate `LLM` instances should still sum normally.

### Screenshots/Code snippets

```python
llm = LLM(model="ollama/qwen2.5:14b")
a1 = Agent(role="A", goal="Answer in one word.", backstory="B", llm=llm)
a2 = Agent(role="B", goal="Answer in one word.", backstory="B", llm=llm)  # same instance
crew = Crew(agents=[a1, a2], tasks=[t1, t2])
crew.kickoff()

llm.get_token_usage_summary().total_tokens   # 155
crew.usage_metrics.total_tokens              # 310
```

```
agents sharing one LLM instance: True
LLM instance actual total  : 155
crew.usage_metrics reports : 310
ratio                      : 2.00x
```

The inflation is linear in the number of agents sharing the instance. Stubbing the summary at a known 100 tokens and varying the agent count:

```
1 agent(s)  sharing 1 LLM (real 100) -> reported  100  (1x)
2 agent(s)  sharing 1 LLM (real 100) -> reported  200  (2x)
3 agent(s)  sharing 1 LLM (real 100) -> reported  300  (3x)
5 agent(s)  sharing 1 LLM (real 100) -> reported  500  (5x)
```

### Operating System

### Python Version

3.12

### crewAI Version

1.15.18

### crewAI Tools Version

1.15.18

### Virtual Environment

Venv

### Evidence

- `lib/crewai/src/crewai/crew.py:2208-2216` — the loop adds `agent.llm.get_token_usage_summary()` once per agent, with no de-duplication of the underlying instance.
- `lib/crewai/src/crewai/llms/base_llm.py:1004` — `get_token_usage_summary()` returns lifetime totals for the instance and its docstring notes those include calls from other agents sharing it.
- The same applies to `manager_agent.llm` at `crew.py:2222-2225` when the manager shares an instance with the agents.

Reproduced both with a live crew and deterministically by stubbing the summary, so the ratio does not depend on model behaviour.

### Possible Solution

Track which LLM instances have already contributed, by object identity, and skip repeats:

```python
counted_llms: set[int] = set()

for agent in self.agents:
    if isinstance(agent.llm, BaseLLM):
        if id(agent.llm) in counted_llms:
            continue
        counted_llms.add(id(agent.llm))
        ...
```

with the same identity check applied to the manager agent's LLM. `id()` rather than model name matters here: two agents may legitimately hold separate `LLM(model="...")` instances with the same model, and those must still sum.

### Additional context

Related but distinct: #4934 (open since 2026-03-18) fixes a different double-count in this same function, where the manager agent is counted twice because `_token_process` and the LLM summary are added in two independent `if` blocks rather than mutually exclusive branches. That change is still needed after this one; the two address different mechanisms. I've deliberately not touched the `if`/`else` restructure that PR proposes.

My OS is macOS Tahoe 26.5.2 and I'm on Python 3.13.13, neither of which is in the dropdowns. Running from a source checkout of main at commit 92eb5f9.

parthiban-sivakumar:
Fixes #7259

## Problem

`calculate_usage_metrics()` aggregated `get_token_usage_summary()` once per agent. Those counters are cumulative for the lifetime of the LLM **instance** and, as the method documents, include calls made by every agent sharing it. Two problems followed:

**Agents sharing one instance were multiplied.** N agents on one `LLM` reported N× the real usage.

**Later runs carried earlier ones.** A second kickoff reported everything since the instance was created:

```
run 1: used 100 -> crew reports 100
run 2: used  50 -> crew reports 150
```

Both come from reconstructing per-run usage from a lifetime accumulator.

## Fix

Snapshot each **distinct** LLM instance's counters at kickoff and report the growth across the run, using the existing `UsageMetrics.delta_since`.

- de-duplicating by object identity stops a shared instance being counted once per agent
- a delta rather than an absolute total stops earlier runs leaking into later ones
- taking the baseline **per instance rather than per agent** leaves no per-agent window to overlap, so tasks running concurrently on a shared instance stay correct

Live crew, two agents sharing one LLM, run twice:

```
run 1: used 155 -> reports 155   (previously 310)
run 2: used 155 -> reports 155   (previously 310, lifetime)
```

## Why not per-call events

I first implemented this as `LLMCallCompletedEvent` aggregation, mirroring `Flow._attach_usage_aggregation_listener`. CI showed why that can't work here: `_track_token_usage_internal` only updates an LLM's counters and never emits the event, so **any custom `BaseLLM` that overrides `call()` reported zero usage**. `test_usage_shape_parity.py` covers exactly that and failed with `assert 0 > 0`.

Making `_track_token_usage_internal` emit would fix it, but that changes `BaseLLM` behaviour for every caller and is well outside this PR. Reading counters keeps custom providers working, and taking the baseline per instance avoids the concurrency problem that made snapshots look unattractive in the first place.

## Scope

The `_token_process` fallback is kept for agents whose `llm` is not a `BaseLLM`, and it no longer double-fires for the manager — the manager is now measured through the same instance de-duplication as everyone else. This does not restructure the `_token_process` handling proposed in #4934.

## Testing

New `test_crew_usage_metrics.py`:

- three agents on one instance report that instance's usage, not 3×
- distinct instances still sum, even for the same model
- a later run excludes the previous one
- a manager sharing an agent's instance adds no extra usage

`test_hierarchical_kickoff_usage_metrics_include_manager` keeps its original shape; its stubbed summaries now grow across the run so a delta is observable.

Full `lib/crewai/tests/` — 5282 passed, 44 skipped. ruff, ruff-format and mypy clean. (`cli/test_version.py::test_dynamic_versioning_consistency` fails identically on clean `main` in my checkout — an installed-version mismatch, unrelated.)

---

### DIFF
--- a/lib/crewai/src/crewai/agents/agent_builder/base_agent.py
+++ b/lib/crewai/src/crewai/agents/agent_builder/base_agent.py
@@ -50,6 +50,7 @@ from crewai.tools.tool_failure import (
     collect_tool_failures,
 )
 from crewai.types.callback import SerializableCallable
+from crewai.types.usage_metrics import UsageMetrics, add_usage_metrics_locked
 from crewai.utilities.config import process_config
 from crewai.utilities.i18n import I18N, get_i18n
 from crewai.utilities.logger import Logger
@@ -268,6 +269,7 @@ class BaseAgent(BaseModel, ABC, metaclass=AgentMeta):
     _original_goal: str | None = PrivateAttr(default=None)
     _original_backstory: str | None = PrivateAttr(default=None)
     _token_process: TokenProcess = PrivateAttr(default_factory=TokenProcess)
+    _usage_metrics: UsageMetrics = PrivateAttr(default_factory=UsageMetrics)
     _kickoff_event_id: str | None = PrivateAttr(default=None)
     _tool_failures: list[ToolFailureRecord] = PrivateAttr(default_factory=list)
     id: UUID4 = Field(default_factory=uuid.uuid4, frozen=True)
@@ -643,6 +645,16 @@ class BaseAgent(BaseModel, ABC, metaclass=AgentMeta):
         if not self._token_process:
             self._token_process = TokenProcess()
         return self
+
+    def _record_llm_usage(self, llm: BaseLLM, usage: UsageMetrics) -> None:
+        """Credit one LLM call's usage to this agent.
+
+        Called by ``BaseLLM`` for every call made while this agent is running.
+        Only calls made through the agent's own LLMs count, so an LLM owned by
+        something else (a memory store, say) stays out of the agent's usage.
+        """
+        if llm is self.llm or llm is getattr(self, "function_calling_llm", None):
+            add_usage_metrics_locked(self._usage_metrics, usage)
 
     @model_validator(mode="after")
     def resolve_memory(self) -> Self:
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -2205,8 +2205,16 @@ class Crew(FlowTrackable, BaseModel):
         """Calculates and returns the usage metrics."""
         total_usage_metrics = UsageMetrics()
 
+        # An LLM instance's counters are cumulative for its lifetime and
+        # include calls made by every agent sharing it, so each distinct
+        # instance must contribute to the total exactly once.
+        counted_llms: set[int] = set()
+
         for agent in self.agents:
             if isinstance(agent.llm, BaseLLM):
+                if id(agent.llm) in counted_llms:
+                    continue
+                counted_llms.add(id(agent.llm))
                 llm_usage = agent.llm.get_token_usage_summary()
 
                 total_usage_metrics.add_usage_metrics(llm_usage)
@@ -2220,7 +2228,11 @@ class Crew(FlowTrackable, BaseModel):
             total_usage_metrics.add_usage_metrics(token_sum)
 
         if self.manager_agent:
-            if isinstance(self.manager_agent.llm, BaseLLM):
+            if (
+                isinstance(self.manager_agent.llm, BaseLLM)
+                and id(self.manager_agent.llm) not in counted_llms
+            ):
+                counted_llms.add(id(self.manager_agent.llm))
                 llm_usage = self.manager_agent.llm.get_token_usage_summary()
                 total_usage_metrics.add_usage_metrics(llm_usage)
 
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -8,6 +8,7 @@ from hashlib import md5
 import json
 from pathlib import Path
 import re
+import threading
 from typing import (
     TYPE_CHECKING,
     Annotated,
@@ -125,6 +126,7 @@ from crewai.types.callback import SerializableCallable
 from crewai.types.streaming import CrewStreamingOutput
 from crewai.types.usage_metrics import UsageMetrics
 from crewai.utilities.constants import NOT_SPECIFIED, TRAINING_DATA_FILE
+from crewai.utilities.crew.crew_context import get_crew_context
 from crewai.utilities.crew.models import CrewContext
 from crewai.utilities.env import get_env_context
 from crewai.utilities.evaluators.crew_evaluator_handler import CrewEvaluator
@@ -226,6 +228,9 @@ class Crew(FlowTrackable, BaseModel):
     )
     _kickoff_event_id: str | None = PrivateAttr(default=None)
     _execution_start_dispatched: bool = PrivateAttr(default=False)
+    _agent_usage: dict[str, UsageMetrics] = PrivateAttr(default_factory=dict)
+    _usage_lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
+    _usage_handler: Callable[[Any, Any], None] | None = PrivateAttr(default=None)
     _execution_end_dispatched: bool = PrivateAttr(default=False)
 
     name: str | None = Field(default="crew")
@@ -1047,6 +1052,7 @@ class Crew(FlowTrackable, BaseModel):
 
         execution_token = begin_execution()
 
+        self._attach_usage_listener()
         runtime_scope = crewai_event_bus._enter_runtime_scope()
         try:
             inputs = prepare_kickoff(self, inputs, input_files)
@@ -1083,6 +1089,7 @@ class Crew(FlowTrackable, BaseModel):
             # Safety net for the exception path; the success path already
             # drained in _create_crew_output before emitting completion.
             self._drain_memory_writes()
+            self._detach_usage_listener()
             clear_files(self.id)
             detach(token)
             end_execution(execution_token)
@@ -1264,6 +1271,7 @@ class Crew(FlowTrackable, BaseModel):
 
         execution_token = begin_execution()
 
+        self._attach_usage_listener()
         runtime_scope = crewai_event_bus._enter_runtime_scope()
         try:
             inputs = prepare_kickoff(self, inputs, input_files)
@@ -1300,6 +1308,7 @@ class Crew(FlowTrackable, BaseModel):
             # Safety net for the exception path; the success path already
             # drained in _create_crew_output before emitting completion.
             self._drain_memory_writes()
+            self._detach_usage_listener()
             clear_files(self.id)
             detach(token)
             end_execution(execution_token)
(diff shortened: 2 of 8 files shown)


Vidit-Ostwal (changes requested):
The bug is real: `calculate_usage_metrics()` adds `get_token_usage_summary()` once per agent, and those counters are lifetime totals on the LLM instance. Sharing one `LLM` therefore reports N× usage.

The `id()` de-dupe is a patch on the wrong source. Crew usage should not be reconstructed from instance lifetime totals (that also picks up leftover usage from a prior run, and it is a different bug from #4934). `_token_process` is also the wrong place: native providers never write it, and it is missing `reasoning_tokens` / `cache_creation_tokens`.

Please rework this as per-agent `UsageMetrics` via snapshot/delta, then sum those on the crew.

1. **Agent accumulator** — add `UsageMetrics` on the agent (not `TokenProcess`). Add each run’s delta into it.
2. **Wrap each agent run** — before the executor loop, snapshot `llm.get_token_usage_summary()` (and `function_calling_llm` if set). After the loop, `delta_since(baseline)` and add it. Kickoff already does this for `LiteAgentOutput`; do the same on the crew task path and persist it. Snapshot must wrap each agent **run**, not every agent at crew start — a shared LLM would otherwise give every agent the same full delta.
3. **Crew total** — `calculate_usage_metrics()` sums `agents` + `manager_agent` accumulators. Stop walking LLM lifetime totals. Keep assigning both `usage_metrics` and `token_usage`.
4. **Tests** — shared LLM, sequential: each agent gets its own delta, crew is the sum (not N×). Distinct LLM instances still sum. Manager included.

Known gap: overlapping `async_execution` on a shared LLM will mis-attribute; sequential crews are the target for this fix. Per-call `from_agent` attribution would be more accurate later, but this matches code that already exists and fixes ordinary usage.

parthiban-sivakumar:
Hi @Vidit-Ostwal,

Thanks for the review — you're right. I checked the leftover-usage point and it reproduces:

```
run 1: used 100 -> crew reports 100
run 2: used  50 -> crew reports 150
```

Run 2 only used 50 tokens. So my change fixed the sharing case but not this one — both come from reading lifetime totals instead of per-run usage.

Before I rework it: should the snapshot be taken per agent, or per LLM instance? I'd also like it to hold when tasks run with `async_execution=True`, where two agents can be using the same instance at the same time.

Thanks!

Vidit-Ostwal:
Ahh, good catch
Rather than snapshot, I think we should calculate per agent, whenever llm call happens, I think we can calcukate there, 
And as this is per llm call, there will not be the problem of async execution, two agents using the same llm call instance at the same time.

Vidit-Ostwal:
lmk if this makes sense.

parthiban-sivakumar:
Makes sense — that's cleaner than snapshots.

The Flow runtime already does exactly this in
 _attach_usage_aggregation_listener (flow/runtime/__init__.py:906): 
 
 it listens for **LLMCallCompletedEvent**, pulls event.usage, and accumulates into a fresh per-run object under a lock, with a run-id check so events from a previous run don't leak in. LLMCallCompletedEvent also carries agent_id, so per-agent attribution comes for free.

I will  follow that pattern for Crew unless you'd rather it were shared between the two.

parthiban-sivakumar:
@Vidit-Ostwal  , Got the per-call version working. Two things I'd like your call on before I push.

1. test_hierarchical_kickoff_usage_metrics_include_manager fails now. It stubs get_token_usage_summary and _token_process and patches out Task.execute_sync, so no LLM call happens and nothing gets recorded. What it checks is still right — manager usage should be in the crew total — but it's written against the old internals. Rewrite it to emit LLM call events, or handle separately?

2. This changes what the number means. It's now what we actually observed on the event bus, so a call that completes without reporting usage isn't counted. Flow already works this way and says so in usage_metrics. Fine to bring the same behaviour to Crew?

On a live crew with two agents sharing one LLM:

run 1: used 155 -> reports 155   (was 310)
run 2: used 155 -> reports 155   (was 310, lifetime)

parthiban-sivakumar:
Pushed the rework in 4c30aa4 — dropped the `id()` de-dupe and moved to per-call accumulation as you suggested.

I went ahead on both questions rather than block, and flagged them in the PR description so they're easy to reverse:

1. Rewrote `test_hierarchical_kickoff_usage_metrics_include_manager` to emit the LLM calls its stubs stood in for. Same assertions, same numbers.
2. Usage is now what's observed on the bus, matching Flow. Say the word if you'd rather Crew kept the old semantics.

Ready for another look.

parthiban-sivakumar:
Correction on my last comment — the per-call version doesn't hold up, and CI caught it.

`_track_token_usage_internal` only updates an LLM's counters; it never emits `LLMCallCompletedEvent`. So any custom `BaseLLM` that overrides `call()` reports zero usage. `test_usage_shape_parity.py` covers exactly that and failed with `assert 0 > 0`.

So I've gone back to snapshot/delta in 9cf2bb4, but taken **per distinct LLM instance** rather than per agent. That was the part I got wrong when I raised the concurrency concern — the overlap problem only exists with per-agent windows. Per instance across the whole run there's no window to overlap, and reading counters keeps custom providers working.

Both original problems still fixed:

```
run 1: used 155 -> reports 155   (was 310)
run 2: used 155 -> reports 155   (was 310, lifetime)
```

Also reverted my earlier rewrite of `test_hierarchical_kickoff_usage_metrics_include_manager` — it keeps its original shape now, with the stubbed summaries growing across the run so a delta is observable. Much smaller change to that test than before.

Full suite: 5282 passed, 44 skipped.

Vidit-Ostwal:
Hey, If I understand this correctly, we are snapshotting each llm usage metrics before the kickoff, and then calculating after the kickoff,

I think the better way would be to calcualte each agents run metrics, and then summing them up, for that as well one should not do the snapshotting rather than that we should calculate from the llm calls itself, Kindly make the changes in that direction.
lmk if this makes sense.
Happy to review post that,

parthiban-sivakumar:
Thanks @Vidit-Ostwal I had one doubt  before I rework this, I found something.

[[#4286]](https://github.com/crewAIInc/crewAI/pull/4286)
 is already doing per-task token usage, and it uses snapshot/delta — _get_agent_token_usage reads agent.llm.get_token_usage_summary() and _calculate_token_delta subtracts the before/after. It's been open since January.

Since a task always belongs to one agent, per-task usage rolls up into per-agent anyway. So if I record at call time here, that PR is solving the same accounting a different way and we end up with two mechanisms for it.

Which direction should both go? I'm fine building the call-time version, I'd just rather it lines up with whatever #4286 does.

Vidit-Ostwal:
A agent can be assigned to 2 different task, what happens in that case?

parthiban-sivakumar:
it's the case that decides it. I've reworked this to record usage per call, as you suggested.

When an agent's two tasks run one after the other, any approach works because the windows don't overlap. The problem is overlapping async_execution tasks on one LLM instance: a snapshot around each task sees the other task's calls inside its own window. I simulated the before/after pattern from  [#4286](https://github.com/crewAIInc/crewAI/issues/4286)  with two overlapping threads doing 200 tokens each. They reported 300 and 400, so 700 for 400 real. Snapshotting per agent run has the same overlap, as you noted, and the per-kickoff snapshot on this PR gets totals right but can't say which agent used what.

My earlier event-based version failed because custom BaseLLMs never emit LLMCallCompletedEvent. But every provider, custom ones included, reports usage through _track_token_usage_internal. That method now credits each call to the agent running in the current context. The task sets it for its whole execution, and delegation switches it to the coworker. Only calls on the agent's own llm or function_calling_llm count. The crew sums the per-agent totals and resets them at each kickoff.

For your case: an agent on two sequential tasks is credited with both, once. Two overlapping async tasks sharing one LLM each count only their own calls (300 spent, 300 reported). One agent on two overlapping async tasks can't run at all today, because both tasks drive the same executor and it raises "Executor is already running". [#7333](https://github.com/crewAIInc/crewAI/issues/7333) fixes that, and with it merged I got 300 spent, 300 reported. I've kept that test out of this PR so it doesn't depend on [#7333](https://github.com/crewAIInc/crewAI/issues/7333).

On [#4286](https://github.com/crewAIInc/crewAI/issues/4286): since usage is now recorded per call, the same hook could credit the current task too, so per-task usage wouldn't need its own snapshots. I've left that out of scope here. Happy to change direction if you see it differently.

PR #7269, opened 2026-09-04
Fix legacy platform tool alias discovery

### LINKED ISSUE #7268: Legacy platform tool discovery drops canonicalized app and action aliases
## Description

`CrewaiPlatformTools` returns no legacy tools when the requested app or action uses an alias that the platform canonicalizes. This regressed in `crewai-tools` 1.15.19 after #7180.

## Reproduction

Given a connected SharePoint integration:

```python
from crewai_tools import CrewaiPlatformTools

assert CrewaiPlatformTools(apps=["share_point"])
assert CrewaiPlatformTools(
    apps=["microsoft_sharepoint/download_file_by_server_relative_url"]
)
```

The legacy API returns `microsoft_sharepoint` as the app key and `SHARE_POINT_DOWNLOAD_FILE_BY_SERVER_RELATIVE_URL` as the action name. Version 1.15.19 then returns an empty tool list for both selectors. Using `microsoft_sharepoint` without an action returns the available tools. Version 1.15.18 returns tools for `share_point`.

## Cause

`LegacyClient.get_actions` asks the server to filter the requested selectors, then filters the response again with exact string comparisons:

```python
selector.app == app and selector.action in (None, action)
```

The server returns canonical app keys and provider action names, so they do not always equal the requested aliases.

## Expected behavior

Legacy discovery must accept the filtered server response when the server canonicalizes an app or action selector. App aliases and canonical action names must continue to work.

vinibrsl:
- Trust the legacy API response after it filters and canonicalizes app and action selectors.
- Keep connection IDs exclusive to Clipper tool discovery.
- Add regression coverage for SharePoint app and action aliases.

Fixes #7268

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/tools/crewai_platform_tools/integrations_client.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/crewai_platform_tools/integrations_client.py
@@ -257,18 +257,16 @@ class LegacyClient:
                     if not isinstance(parameters, dict):
                         parameters = {}
 
-                    tool_infos.extend(
+                    tool_infos.append(
                         ToolInfo(
                             app=app,
                             action=action,
-                            connection_id=selector.connection_id,
+                            connection_id=None,
                             description=action_data.get(
                                 "description", f"Execute {action}"
                             ),
                             parameters=parameters,
                         )
-                        for selector in selectors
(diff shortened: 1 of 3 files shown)


joaomdmoura (approved):

Vidit-Ostwal (approved):
LGTM

PR #7272, opened 2026-09-04
fix(flow): preserve parallel producers feeding or_

### LINKED ISSUE #7183: [BUG] or_() join cancels parallel listener branches that should complete
# [BUG] `or_()` join cancels parallel listener branches that should complete

## Description

When two listeners are triggered by the same method and their completion events
feed a downstream `or_()` listener, the runtime treats the two producer
listeners as a first-wins race. Once one producer completes, it cancels the
other producer.

`or_(branch_a, branch_b)` should control when the downstream listener fires. It
should not change the execution semantics of `branch_a` and `branch_b`
themselves. Both were independently triggered by their shared parent and may
contain required state updates, tool calls, or other side effects.

## Steps to reproduce

```python
import asyncio

from crewai.flow.flow import Flow, listen, or_, start


completed = []
joined = 0


class ParallelFanoutOrFlow(Flow):
    @start()
    def begin(self):
        return "begin"

    @listen(begin)
    async def fast_branch(self):
        await asyncio.sleep(0)
        completed.append("fast")

    @listen(begin)
    async def slow_branch(self):
        await asyncio.sleep(0.05)
        completed.append("slow")

    @listen(or_(fast_branch, slow_branch))
    def join(self):
        global joined
        joined += 1


asyncio.run(ParallelFanoutOrFlow().kickoff_async())
print(completed)
print(joined)
```

## Actual behavior

Only the faster branch reliably completes. The slower listener is cancelled by
`_execute_racing_listeners()`.

For synchronous listeners executed through `asyncio.to_thread()`, cancelling
the asyncio task cannot stop the worker thread, so the branch may instead
continue after the runtime considers it cancelled.

## Expected behavior

- Both independently triggered producer listeners complete.
- The downstream `or_()` listener fires exactly once after the first producer
  completes.
- The OR condition does not cancel event producers.

## Root cause

`Flow._build_racing_groups()` converts the events referenced by a downstream
OR condition back into method names. During dispatch,
`_get_racing_group_for_listeners()` intersects those event names with the
listener methods triggered by the current parent. If both names are present,
`_execute_racing_listeners()` runs them as a first-wins race and cancels the
loser.

The runtime already tracks `_fired_or_listeners`, which prevents the downstream
OR listener from firing more than once. Producer cancellation is unnecessary.

## Proposed fix

Execute every listener selected by `_find_triggered_methods()` using the normal
`asyncio.gather()` path. Keep `_fired_or_listeners` responsible for deduplicating
the downstream OR listener.

Add a regression test that asserts both producer branches complete while the
OR listener fires once.

## Environment

- CrewAI revision: `381fef73be1803d0b4676f88cbae57d332cb9cac`
- Python: 3.13
- OS: Windows

lprnmns:
## Related issue

Fixes #7183

## Summary

- remove the listener-racing path that treated producers referenced by a downstream `or_()` as a first-wins race
- execute every independently triggered producer and keep `_fired_or_listeners` responsible for firing the downstream join once
- add a regression test covering a fast and a slow producer feeding the same `or_()` join

## Verification

- [x] Regression test fails on `main`: the slow producer is absent from `completed`
- [x] `uv run pytest lib/crewai/tests/test_flow.py::test_or_listener_does_not_cancel_parallel_producers -q -o 'addopts=--tb=short --timeout=60 --block-network --import-mode=importlib'` — 1 passed
- [x] `uv run pytest lib/crewai/tests/test_flow.py -x -q` — 83 passed
- [x] `uv run pytest lib/crewai/tests/ -x -q` — 5283 passed, 44 skipped
- [x] `uv run pre-commit run --files lib/crewai/src/crewai/flow/runtime/__init__.py lib/crewai/tests/test_flow.py` — Ruff, format, and mypy passed

## Additional context

The public `or_()` behavior remains unchanged: the downstream listener fires once. The fix only stops that condition from cancelling independently triggered producer work.

AI disclosure: this contribution was developed with OpenAI Codex assistance. Per the contribution policy, it must carry the `llm-generated` label.

### DIFF
--- a/lib/crewai/src/crewai/flow/runtime/__init__.py
+++ b/lib/crewai/src/crewai/flow/runtime/__init__.py
@@ -248,18 +248,6 @@ def _iter_condition_events(condition: FlowDefinitionCondition) -> Iterator[str]:
         yield from _iter_condition_events(branch)
 
 
-def _or_alternative_events(condition: FlowDefinitionCondition) -> Iterator[str]:
-    if isinstance(condition, str):
-        yield condition
-        return
-
-    operator, branches = _condition_branches(condition)
-    if operator != "or":
-        return
-    for branch in branches:
-        yield from _or_alternative_events(branch)
-
-
 def _is_multi_event_or(
     condition: FlowDefinitionCondition,
 ) -> bool:
@@ -748,9 +736,6 @@ class Flow(BaseModel, Generic[T], metaclass=FlowMeta):
         default_factory=dict
     )
     _fired_or_listeners: set[FlowMethodName] = PrivateAttr(default_factory=set)
-    _racing_groups_cache: dict[frozenset[FlowMethodName], FlowMethodName] | None = (
-        PrivateAttr(default=None)
-    )
     _method_outputs: list[Any] = PrivateAttr(default_factory=list)
     _definition: FlowDefinition = PrivateAttr()
     _or_listeners_lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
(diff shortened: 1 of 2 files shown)


lprnmns:
AI assistance is disclosed in the PR body. I attempted to apply the required `llm-generated` label during creation and through the labels API, but GitHub returned 403 because fork contributors do not have label permission. A repository collaborator will need to apply the label.

PR #7283, opened 2026-09-05
fix(files): raise ValueError instead of bare raise in get_uploader

### LINKED ISSUE #7282: get_uploader raises RuntimeError instead of returning None for unsupported/unconfigured providers
## Description

`crewai_files.uploaders.get_uploader` is documented to return `None` when a provider is not supported ("Returns: FileUploader instance for the provider, or None if not supported"), and every caller branches on `if uploader is None`. But two fallthrough paths run a bare `raise` with no active exception, so instead of returning `None` they raise `RuntimeError: No active exception to reraise`:

- an unknown/unsupported provider (the final line of the function)
- a Bedrock provider when `CREWAI_BEDROCK_S3_BUCKET` is unset and no `bucket_name` is passed

The `except ImportError: ... raise` blocks are correct (they re-raise the ImportError when a provider SDK is not installed); only the two non-except bare raises are affected.

## Steps to Reproduce

```python
from crewai_files.uploaders import get_uploader

get_uploader("does-not-exist")   # RuntimeError: No active exception to reraise
get_uploader("bedrock")          # RuntimeError (when CREWAI_BEDROCK_S3_BUCKET is unset)
```

## Expected behavior

Both return `None`, matching the docstring and every caller's `if uploader is None` check

## Screenshots/Code snippets

`lib/crewai-files/src/crewai_files/uploaders/factory.py`: the bare `raise` in the "Bedrock S3 uploader not configured" branch, and the bare `raise` at the final "No file uploader available for provider" line

## Operating System

Not OS-specific (reproduces on Linux, macOS, Windows)

## Python Version

3.12

## Additional context

I have a fix ready and will open a PR. Reported with AI assistance, hence the `llm-generated` label per CONTRIBUTING

SWAPI03:
> Note for maintainers: this PR was created with AI assistance. Per CONTRIBUTING it needs the `llm-generated` label, but as an external contributor I do not have permission to apply labels. Could a maintainer please add it

## Related issue

Fixes #7282

## Summary

`get_uploader` had two fallthrough paths that ran a bare `raise` with no active exception, so an unknown provider, and a Bedrock provider without `CREWAI_BEDROCK_S3_BUCKET` configured, blew up with `RuntimeError: No active exception to reraise` instead of a clear error.

Both paths now raise a `ValueError` that names the problem: the unknown-provider case lists the supported providers, and the Bedrock case names the missing `CREWAI_BEDROCK_S3_BUCKET`. The return types are now non-optional (`get_uploader` always returns a real uploader or raises), and the resolver and cache cleanup were updated to surface those errors instead of branching on a `None` that can no longer happen. The `except ImportError` re-raises (a provider SDK is not installed) are intentionally unchanged.

This follows the maintainer's steer to raise rather than return `None`, since a silent `None` just pushes the failure downstream to an inline fallback that hides the misconfiguration.

In the batch resolvers, an uploader that cannot be built is a per-provider problem, not a per-file one, so it should surface while ordinary per-file errors stay best-effort. The resolver's `_get_uploader` now translates the lookup failure into a dedicated `UploaderConfigurationError`, and `aresolve_files` re-raises only that. An unrelated per-file error (for example a stream that raises `ValueError` when read) is logged and skipped as before, so one bad file no longer aborts the whole async batch.

## Verification

`lib/crewai-files/tests/test_factory.py` covers `get_uploader` directly: the happy path (a configured provider returns its uploader), the unknown-provider case, and the unconfigured-Bedrock case (both an absent and a falsy `bucket_name`). `lib/crewai-files/tests/test_resolver.py` adds three regression tests for the batch behavior: `_get_uploader` wrapping a lookup failure as `UploaderConfigurationError`, a provider-setup error surfacing from `aresolve_files`, and an unrelated per-file error being skipped while the other files still resolve. `mypy` (strict config) passes on the changed source files.


SWAPI03:
Thanks, good catch. Addressed in 784dd3c: the Bedrock config guard now keys on the value via `not kwargs.get("bucket_name")`, so an explicit None or empty-string bucket_name is treated as unconfigured, with a regression test covering both.

SWAPI03:
@coderabbitai review

Vidit-Ostwal (commented):
Returning `None` keeps callers from crashing, but the user never sees that the provider is unknown or Bedrock is unconfigured — the resolver just debug-logs and falls back to inline.

Direction: raise a `ValueError` with a concrete reason (unknown provider, or missing `CREWAI_BEDROCK_S3_BUCKET` / `bucket_name`). Do not use `PermanentUploadError`; the upload path already catches that and swallows it. Leave the `ImportError` re-raises as they are, and update the docstring so it no longer promises `None`.

SWAPI03:
Thanks @Vidit-Ostwal, reworked to your direction. get_uploader now raises a ValueError with a concrete reason for an unknown provider and for Bedrock without a bucket (CREWAI_BEDROCK_S3_BUCKET or bucket_name), a plain ValueError so it is not swallowed like PermanentUploadError, ImportError re-raises left as-is, and the docstring no longer promises None. I traced resolve(): nothing catches it on that path, so the error surfaces to the user as you want. cleanup skips providers it cannot build an uploader for, so I routed its calls through a small local helper that treats the ValueError as unavailable, otherwise a stale provider would abort the whole cleanup pass. Tests now assert the ValueErrors, and mypy strict plus ruff pass on the changed files.

Mahnoor-Zaffar (commented):
Reviewed: `lib/crewai-files/src/crewai_files/cache/cleanup.py`, `lib/crewai-files/src/crewai_files/uploaders/factory.py`, `lib/crewai-files/tests/test_factory.py`. CI status unavailable.

- The change in `get_uploader` to return `None` instead of raising a bare `RuntimeError` is a significant improvement, as it aligns with the documented behavior and prevents unexpected crashes.
- The new `_uploader_or_none` function in `cleanup.py` effectively encapsulates the logic for handling unsupported providers, which enhances code readability.
- The tests in `test_factory.py` cover the critical edge cases for unknown providers and unconfigured Bedrock, ensuring that the expected `ValueError` is raised. However, consider adding a test for the scenario where a valid provider is configured correctly to ensure the happy path is also validated.
- The typing changes in `get_uploader` are appropriate and maintain mypy compliance.

One suggestion: in `factory.py`, consider adding a more descriptive error message for the `ValueError` raised when the provider is unknown, to aid debugging.

SWAPI03:
On the resolver contract (CodeRabbit's point): updated in bfe1b0d rather than catching the error. _get_uploader now returns FileUploader (dropped the `| None` and the dead None fallback), so the ValueError propagates through resolve() and surfaces to the caller, per @Vidit-Ostwal's direction not to silently fall back to inline.

SWAPI03:
Thanks @Mahnoor-Zaffar, addressed both suggestions in bfe1b0d: added a happy-path test that a configured provider returns its uploader instance, and the unknown-provider ValueError now lists the supported providers to aid debugging.

SWAPI03:
Good catch on the async batch. aresolve_files gathers with return_exceptions=True, which was silently dropping files when _get_uploader raised. Since a batch shares one provider, a lookup failure applies to every file, so in 2f63e19 I re-raise ValueError and ImportError to surface it (matching the sync resolve_files path); genuine per-file upload errors are still logged and skipped. In practice the reachable case is a missing provider SDK (ImportError), since bedrock uses URL references not the upload path and an unknown provider never reaches it, so I did not add a dedicated test that would need a supported provider with its SDK removed or a mocked get_uploader; the raise itself is covered in test_factory.py. Happy to add one if you'd prefer.

SWAPI03:
@coderabbitai that async batch case was fixed in 2f63e19: aresolve_files now re-raises ValueError and ImportError from the gather instead of dropping the file.

lib/crewai-files/src/crewai_files/cache/cleanup.py
```diff
@@ -17,6 +17,19 @@
...
+    get_uploader raises ValueError for an unknown or unconfigured provider.
+    Cleanup skips such providers rather than aborting the whole pass, so that
+    error is treated as "no uploader available" here.
+    """
+    try:
+        return get_uploader(provider)
+    except ValueError:
+        return None
+
+
```
Vidit-Ostwal, line 32: Will get_uploadeder ever raise a Value Error?

SWAPI03: Yes, but only through the direct entry points. cleanup_provider_files(provider) and its async version take a caller-supplied provider and call get_uploader(provider) with no bucket_name, so cleanup_provider_files("bedrock") with no CREWAI_BEDROCK_S3_BUCKET set raises ValueError, and an unknown provider string raises too.

Through the cache-driven functions it effectively won't, since a provider in the cache already built a working uploader at upload time, and bedrock uses S3 URL refs rather than the upload cache.

So the helper keeps a misconfigured single-provider cleanup skipping and logging instead of aborting. If you would rather that surface as a caller error, I can drop the helper and let the ValueError propagate. Your call.

Vidit-Ostwal (commented):
@SWAPI03 I think this needs a small follow-up before merge. In `aresolve_files`, `asyncio.gather(..., return_exceptions=True)` returns exceptions from every stage of resolving a file—not only `get_uploader()`. The new `isinstance(item, (ValueError, ImportError))` check re-raises any `ValueError`, so one unrelated bad file (for example, a stream that raises `ValueError` when read) now aborts the whole batch instead of preserving the existing log-and-skip behavior.

Could we distinguish uploader lookup/configuration failures from ordinary per-file failures—e.g. with a dedicated exception or a narrower catch/translation boundary—and re-raise only the former? Please add regression coverage for both: provider setup errors surface, while unrelated per-file errors remain best-effort.

SWAPI03:
Good catch, fixed in 159798f. Added a dedicated UploaderConfigurationError, raised at the _get_uploader boundary by wrapping the ValueError/ImportError from get_uploader. aresolve_files now re-raises only that (it applies to every file for the provider), so an unrelated per-file ValueError is logged and skipped as before. Added three regression tests: the wrap at _get_uploader, a provider-setup error surfacing from aresolve_files, and an unrelated per-file error staying best-effort while the other files resolve.

Vidit-Ostwal (approved):
LGTM

Vidit-Ostwal:
@SWAPI03 Can you fix the Lint please, should be quick?

SWAPI03:
Lint fixed: applied ruff import sort and formatting to the test file. ruff check and ruff format --check are both clean on the changed files with the pinned 0.15.1. Thanks for the review.

Vidit-Ostwal (approved):
LGTM

PR #7286, opened 2026-09-05
docs(streaming): fix streaming output docstring examples

### LINKED ISSUE #7285: [docs] Streaming output docstrings show incorrect API usage
## Description

The docstring examples in `lib/crewai/src/crewai/types/streaming.py` for the streaming output wrapper classes teach the wrong public API. Users copy-pasting them either get no streaming (`CrewStreamingOutput`) or hit `AttributeError` (`FlowStreamingOutput`).

## `CrewStreamingOutput` (line 505 onward)

The example shows:

```python
streaming = crew.kickoff(inputs={"topic": "AI"})
for chunk in streaming:
    print(chunk.content, end="", flush=True)
```

`Crew.kickoff` only returns a `CrewStreamingOutput` when the crew is constructed with `stream=True` (see `lib/crewai/src/crewai/crew.py:1017` and `docs/edge/en/concepts/streaming.mdx:128`). Without that flag, `kickoff` returns a plain `CrewOutput` and the example silently fails to stream anything.

## `FlowStreamingOutput` (line 585 onward)

The example calls `flow.kickoff_streaming()` and `flow.kickoff_streaming_async()`:

```python
streaming = flow.kickoff_streaming()
...
streaming = await flow.kickoff_streaming_async()
```

Neither method exists anywhere in the codebase (`grep -rn "kickoff_streaming" lib/`). Flow-level streaming is exposed through `Flow.kickoff` with `stream=True`, and it returns a `StreamSession` (see `lib/crewai/tests/test_streaming.py:424-435`), not a `FlowStreamingOutput`. `FlowStreamingOutput` is only used in internal type unions and constructed directly with an iterator in tests — never returned from a user-facing method.

## Proposed fix

Docstring-only change:

1. In `CrewStreamingOutput`'s example, construct the crew with `stream=True` so the snippet actually streams.
2. In `FlowStreamingOutput`, drop the misleading example and add a note pointing users at `Flow.kickoff` / `StreamSession`.

Happy to open the PR.

uoparaji:
## Related issue

Fixes #7285

## Summary

Docstring-only fix for two public streaming wrapper classes in `lib/crewai/src/crewai/types/streaming.py`. The examples showed APIs that either silently do not stream (`CrewStreamingOutput`) or do not exist (`FlowStreamingOutput`).

- **`CrewStreamingOutput`** — the sync and async examples both called `crew.kickoff(...)` / `crew.kickoff_for_each_async(...)` without setting `stream=True` on the crew. `Crew.kickoff` only returns a `CrewStreamingOutput` when the crew is constructed with `stream=True` (`lib/crewai/src/crewai/crew.py:1017`, `docs/edge/en/concepts/streaming.mdx:128`); otherwise it returns a `CrewOutput` and the snippet does not stream. Updated both examples to construct the crew with `stream=True`.
- **`FlowStreamingOutput`** — the example called `flow.kickoff_streaming()` and `flow.kickoff_streaming_async()`; `grep -rn "kickoff_streaming" lib/` confirms neither method exists. Flow-level streaming is actually exposed via `Flow.kickoff` with `stream=True`, which returns a `StreamSession`, not a `FlowStreamingOutput` (see `lib/crewai/tests/test_streaming.py:424-435`). `FlowStreamingOutput` is only referenced in internal type unions and instantiated directly with an iterator in tests. Replaced the misleading example with a note pointing users at `Flow.kickoff` / `StreamSession`.

## Verification

- [ ] Tests added or updated for the changed behavior — **N/A**, docstring-only change; no runtime behavior modified.
- [x] Relevant tests and quality checks pass locally — `uv tool run --from ruff==0.15.1 ruff check` and `ruff format --check` on the touched file both pass.

## Additional context

`lib/crewai/tests/test_streaming.py` continues to exercise `FlowStreamingOutput`'s streaming semantics by constructing it directly with an iterator — this PR does not change that path.

### DIFF
--- a/lib/crewai/src/crewai/types/streaming.py
+++ b/lib/crewai/src/crewai/types/streaming.py
@@ -586,9 +586,10 @@ class FlowStreamingOutput(StreamingOutputBase[Any]):
 
     Note:
         Flow-level streaming is exposed to users through
-        :class:`StreamSession` (see ``Flow.kickoff`` with ``stream=True``).
-        ``FlowStreamingOutput`` is retained for consumers that build a
-        streaming wrapper directly from an existing iterator.
+        :class:`StreamSession`; configure the Flow with ``stream=True``
+        before calling ``Flow.kickoff()``. ``FlowStreamingOutput`` is
+        retained for consumers that build a streaming wrapper directly
+        from an existing iterator.
     """
 
     def _set_result(self, result: Any) -> None:
--- a/lib/crewai/src/crewai/types/streaming.py
+++ b/lib/crewai/src/crewai/types/streaming.py
@@ -504,13 +504,15 @@ class CrewStreamingOutput(StreamingOutputBase["CrewOutput"]):
 
     Example:
         ```python
-        # Single crew
+        # Single crew — the crew must be constructed with stream=True
+        crew = Crew(agents=[...], tasks=[...], stream=True)
         streaming = crew.kickoff(inputs={"topic": "AI"})
         for chunk in streaming:
             print(chunk.content, end="", flush=True)
         result = streaming.result
 
-        # Multiple crews (kickoff_for_each_async)
+        # Multiple crews (kickoff_for_each_async) — also requires stream=True
+        crew = Crew(agents=[...], tasks=[...], stream=True)
         streaming = await crew.kickoff_for_each_async(
             [{"topic": "AI"}, {"topic": "ML"}]
         )
@@ -580,22 +582,13 @@ class FlowStreamingOutput(StreamingOutputBase[Any]):
     """Streaming output wrapper for flow execution.
 
     Provides both sync and async iteration over stream chunks,
-    with access to the final flow output via the .result property.
+    with access to the final flow output via the ``.result`` property.
 
-    Example:
-        ```python
-        # Sync usage
-        streaming = flow.kickoff_streaming()
-        for chunk in streaming:
-            print(chunk.content, end="", flush=True)
-        result = streaming.result
-
-        # Async usage
-        streaming = await flow.kickoff_streaming_async()
-        async for chunk in streaming:
-            print(chunk.content, end="", flush=True)
-        result = streaming.result
-        ```
+    Note:
+        Flow-level streaming is exposed to users through
+        :class:`StreamSession` (see ``Flow.kickoff`` with ``stream=True``).
+        ``FlowStreamingOutput`` is retained for consumers that build a
+        streaming wrapper directly from an existing iterator.
     """
 
     def _set_result(self, result: Any) -> None:
--- a/lib/crewai/src/crewai/types/streaming.py
+++ b/lib/crewai/src/crewai/types/streaming.py
@@ -584,6 +584,23 @@ class FlowStreamingOutput(StreamingOutputBase[Any]):
     Provides both sync and async iteration over stream chunks,
     with access to the final flow output via the ``.result`` property.
 
+    Example:
+        ```python
+        # FlowStreamingOutput wraps a chunk-producing iterator directly.
+        # Consumers use it to expose a custom flow-execution generator
+        # through the same iteration + .result API as CrewStreamingOutput.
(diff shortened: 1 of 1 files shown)


uoparaji:
Note for triage: this PR was authored with the assistance of Claude Code. Per CONTRIBUTING.md the `llm-generated` label is required; I don't have permission to apply it from a fork PR, so flagging it here for a maintainer to add. (The linked issue #7285 already carries the label.)

lib/crewai/src/crewai/types/streaming.py
```diff
@@ -580,22 +582,14 @@ class FlowStreamingOutput(StreamingOutputBase[Any]):
...
     Provides both sync and async iteration over stream chunks,
-    with access to the final flow output via the .result property.
-
-    Example:
-        ```python
-        # Sync usage
-        streaming = flow.kickoff_streaming()
-        for chunk in streaming:
-            print(chunk.content, end="", flush=True)
-        result = streaming.result
```
Vidit-Ostwal, line 591: We shouldn't drop the examples from here. I think 
Mind adding them again

uoparaji:
@Vidit-Ostwal thanks for the review — pushed b80a57a2 restoring an `Example:` block on `FlowStreamingOutput`.

Rather than paste back the original snippet (which called `flow.kickoff_streaming()` / `kickoff_streaming_async()`, neither of which exists in the codebase — that was the bug this PR set out to fix), I wrote the example around the only pattern that actually produces a `FlowStreamingOutput`: direct construction with a sync or async chunk-producing iterator. That mirrors how the class is exercised in `lib/crewai/tests/test_streaming.py` and stays consistent with the existing `CrewStreamingOutput` example style. Happy to adjust further if you'd prefer a different framing.

lib/crewai/src/crewai/types/streaming.py
```diff
@@ -580,22 +582,31 @@ class FlowStreamingOutput(StreamingOutputBase[Any]):
...
             print(chunk.content, end="", flush=True)
         result = streaming.result
         ```
+
+    Note:
+        Flow-level streaming is exposed to users through
+        :class:`StreamSession`; configure the Flow with ``stream=True``
+        before calling ``Flow.kickoff()``. ``FlowStreamingOutput`` is
+        retained for consumers that build a streaming wrapper directly
+        from an existing iterator.
```
Vidit-Ostwal, line 613: This is not the public usage for flow streaming.

`FlowStreamingOutput(sync_iterator=...)` is only how tests construct the class. Nothing in production `src/` instantiates it, and no user-facing Flow method returns it. `chunk_generator()` / `async_chunk_generator()` also are not real APIs, so this snippet is not copy-pasteable.

The public path is `Flow(stream=True)` then `kickoff()` / `kickoff_async()`, which return `StreamSession` / `AsyncStreamSession`. That is what the Note already says, and what `docs/edge/en/learn/streaming-flow-execution.mdx` documents.

Please drop this constructor example. Either keep only the Note pointing at `StreamSession`, or if you want a copy-paste block here, show `Flow(stream=True)` + `flow.kickoff()` and make it explicit that that returns a `StreamSession`, not a `FlowStreamingOutput`.

Vidit-Ostwal (commented):
CrewStreamingOutput example looks good. The restored FlowStreamingOutput example still teaches the test-only constructor, not the public Flow streaming API.

uoparaji:
@Vidit-Ostwal you're right - the test-only constructor was misleading in a public docstring. Pushed 454b9ebe replacing that example with the actual public streaming path: `Flow.stream = True` → `kickoff()` / `kickoff_async()` → `StreamSession` / `AsyncStreamSession`. The block is explicitly labeled that `Flow.kickoff()` does *not* return a `FlowStreamingOutput`, and it points readers at `docs/edge/en/learn/streaming-flow-execution.mdx` for the full guide. Let me know if you'd rather drop the example entirely and keep only the Note.

Vidit-Ostwal (approved):
Looks good. Crew examples now construct with `stream=True`, and the Flow example documents the public `StreamSession` path instead of the test-only constructor.

PR #7326, opened 2026-09-08
fix(brightdata): drop stray $ in f-string search URLs

### LINKED ISSUE #7325: [Bug]: BrightDataSERPTool search URLs carry a stray $ from ${query} f-string interpolation
### Bug Description

`BrightDataSERPTool.get_search_url()` builds its search URLs with `${query}` inside f-strings:

```python
def get_search_url(self, engine: str, query: str) -> str:
    if engine == "yandex":
        return f"https://yandex.com/search/?text=${query}"
    if engine == "bing":
        return f"https://www.bing.com/search?q=${query}"
    return f"https://www.google.com/search?q=${query}"
```

In an f-string, `$` is a literal character, so the interpolation yields `...?q=$test` — a stray `$` before the query. The URL is passed directly to the SERP request in `_run` (no later template substitution), so **every search through the tool carries the malformed query string**.

### Steps to Reproduce

```python
from crewai_tools.tools.brightdata_tool.brightdata_serp import BrightDataSERPTool
tool = BrightDataSERPTool()
print(tool.get_search_url("bing", "test"))
# https://www.bing.com/search?q=$test   <- literal '$' prepended to the query
```

### Expected Behavior

`https://www.bing.com/search?q=test` — plain `{query}` interpolation.

simpleqt:
Fixes #7325

`BrightDataSERPTool.get_search_url` interpolated `${query}` inside f-strings, so the three engine URLs (yandex/bing/google) carried a literal `$` before the query — and the URL goes straight to the SERP request, so every search was malformed. Now interpolates `{query}` normally, verified by calling `get_search_url` directly.

(Note: this PR was prepared with AI assistance — happy to see it labeled `llm-generated` per the contributing guidelines.)

### DIFF
--- a/lib/crewai-tools/src/crewai_tools/tools/brightdata_tool/brightdata_serp.py
+++ b/lib/crewai-tools/src/crewai_tools/tools/brightdata_tool/brightdata_serp.py
@@ -131,10 +131,10 @@ class BrightDataSearchTool(BaseTool):
 
     def get_search_url(self, engine: str, query: str) -> str:
         if engine == "yandex":
-            return f"https://yandex.com/search/?text=${query}"
+            return f"https://yandex.com/search/?text={query}"
         if engine == "bing":
-            return f"https://www.bing.com/search?q=${query}"
-        return f"https://www.google.com/search?q=${query}"
+            return f"https://www.bing.com/search?q={query}"
+        return f"https://www.google.com/search?q={query}"
 
     def _run(
         self,


Vidit-Ostwal (approved):
Looks good — `${query}` in those f-strings is a literal `$`, so every Bright Data SERP URL was malformed. Dropping it on Google/Bing/Yandex is the right fix.

I'll merge this. The older overlapping PR (#5518) is being closed in favor of this one.

Issue #7306, opened 2026-09-07
[BUG] Oxylabs tools raise IndexError on any failed request and return failed scrapes as successes


oxy-giedrius:
Opening this at @Vidit-Ostwal's request to track the problem behind #7044.

## Description

All four Oxylabs tools (`OxylabsUniversalScraperTool`,
`OxylabsGoogleSearchScraperTool`, `OxylabsAmazonSearchScraperTool`,
`OxylabsAmazonProductScraperTool`) mishandle every Web Scraper API response that
isn't a success. Each `_run` does an unchecked `response.results[0]`.

The `oxylabs` SDK **logs** transport and validation errors and returns an empty
`Response` rather than raising, so any rejected request reaches that line with
`results == []`. Three distinct failures follow:

1. **Rejected requests raise `IndexError: list index out of range`.** Invalid
   credentials — the most likely first-run mistake — produce this, with no
   indication of the cause unless the app happens to surface the
   `oxylabs.internal.api` logger.
2. **Failed scrapes are returned as successes.** A result with a non-2xx
   `status_code` (e.g. a 404 for a nonexistent ASIN) carries empty content. The
   tools return it, so the agent receives `"[]"` as though the scrape worked.
   The `status_code` is present on the result object and never read.
3. **Non-dict content is returned as a Python repr.** Only `dict` is
   JSON-encoded, with a `str()` fallback. `parsing_instructions` commonly yields
   a list, so agents get `[{'title': '...'}]` — single-quoted and not parseable.

## Steps to Reproduce

No valid Oxylabs account is needed to see (1) — invalid credentials are enough:

1. `pip install 'crewai[tools]' oxylabs`
2. Run the snippet below (deliberately invalid credentials).
3. Observe `IndexError` instead of an authentication error.

```python
import logging
logging.disable(logging.CRITICAL)  # a normal app that doesn't surface library logs

from crewai_tools import OxylabsUniversalScraperTool

tool = OxylabsUniversalScraperTool(username="invalid-user", password="invalid-pass")
print(tool.run(url="https://ip.oxylabs.io/location"))
```

For (2), with valid credentials, request a nonexistent ASIN
(`OxylabsAmazonProductScraperTool(config={"parse": True}).run(query="B0BW4SWWLZ")`)
— the upstream 404 comes back as the string `"[]"`.

## Expected behavior

The tool should report that the request failed and why — ideally as a
`ToolFailure`, so the agent gets an actionable message and the framework records
[...]

PR #7331, opened 2026-09-08
chore(oxylabs): allow the 3.x oxylabs SDK

oxy-giedrius:
Related to #7306.

Stacked on #7044, as requested in review — the failure-handling and base-tool work
stays there, this carries only the dependency bump.

## Summary

`oxylabs` is pinned to exactly `2.0.0` in `lib/crewai-tools/pyproject.toml`, so
consumers can't take 3.0.0, released in March. This relaxes it to
`oxylabs>=2.0.0,<4`.

3.x keeps the `RealtimeClient` surface these tools use (`RealtimeClient(username,
password, **kwargs)` with `sdk_type`, `results[0].content`, `results[0].status_code`,
and the `universal` / `google` / `amazon` namespaces), and adds sources rather than
removing any.

The lockfile keeps `oxylabs` at 2.0.0, so this **permits** the upgrade without forcing
it on anyone.

## Testing

All four Oxylabs tools were verified against the **live** Oxylabs API on `oxylabs`
2.0.0 and 3.0.0, along with each failure path from #7044 (invalid credentials, a
config the source rejects, an upstream 404, and a timeout):

| Tool | 2.0.0 | 3.0.0 |
|---|---|---|
| `OxylabsUniversalScraperTool` | pass | pass |
| `OxylabsGoogleSearchScraperTool` | pass | pass |
| `OxylabsAmazonSearchScraperTool` | pass | pass |
| `OxylabsAmazonProductScraperTool` | pass | pass |

`uv lock --check` passes with uv 0.11.3, the version pinned in
`.pre-commit-config.yaml` and the CI workflows. The lockfile change is a single line —
the recorded specifier. Regenerating it wholesale also rewrote ~60 lines of unrelated
platform-marker churn and re-stamped the relative `exclude-newer` cutoff, so I applied
only the specifier change and verified the result is consistent.

## Note on the diff

Rebased onto `main` now that #7044 has merged, so this is down to a single commit
touching `pyproject.toml` and `uv.lock` only.

## AI-generated contribution

Per `CONTRIBUTING.md`, this was authored with an AI coding assistant (Claude Code) and
needs the `llm-generated` label, which I can't apply from a fork — could a maintainer
add it?

---

*(Supersedes #7330, which the first-time-contributor bot closed because its body referenced PR #7044 rather than an open issue. Same branch, now with the issue reference in place.)*

### DIFF
--- a/lib/crewai-tools/pyproject.toml
+++ b/lib/crewai-tools/pyproject.toml
@@ -141,7 +141,9 @@ xml = [
     "nltk>=3.10.3",
 ]
 oxylabs = [
-    "oxylabs==2.0.0"
+    # 3.x keeps the RealtimeClient surface these tools use and adds sources;
+    # allow it rather than pinning consumers to a single release.
+    "oxylabs>=2.0.0,<4",
 ]
 mongodb = [
     "pymongo>=4.13"
--- a/uv.lock
+++ b/uv.lock
@@ -1878,7 +1878,7 @@ requires-dist = [
     { name = "nest-asyncio", marker = "extra == 'bedrock'", specifier = ">=1.6.0" },
     { name = "nest-asyncio", marker = "extra == 'contextual'", specifier = ">=1.6.0" },
     { name = "nltk", marker = "extra == 'xml'", specifier = ">=3.10.3" },
-    { name = "oxylabs", marker = "extra == 'oxylabs'", specifier = "==2.0.0" },
+    { name = "oxylabs", marker = "extra == 'oxylabs'", specifier = ">=2.0.0,<4" },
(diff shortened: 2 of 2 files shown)


oxy-giedrius:
Rebased onto `main` now that #7044 has merged — this is down to one commit and two files (`pyproject.toml`, `uv.lock`), +4/-2. `uv lock --check` passes against the post-merge lockfile with uv 0.11.3, the version pinned in `.pre-commit-config.yaml` and the CI workflows.

Ready for review whenever it suits.

Vidit-Ostwal (approved):
Stacked correctly on #7044. Unique change is the pin (`oxylabs>=2.0.0,<4`) with the lockfile still on 2.0.0.

LGTM.

PR #7336, opened 2026-09-08
test(agents): cover native result as answer (#7334)

### LINKED ISSUE #7334: [BUG] result_as_answer Crew integration test remains skipped
## Description

The existing end-to-end test `test_tool_result_as_answer_is_the_final_answer_for_the_agent` is still skipped with the reason that native tool calling does not support `result_as_answer`. Current native execution now implements this behavior, including short-circuiting the agent loop, so the skip leaves the full Crew -> Agent -> native tool -> final output path uncovered.

## Steps to Reproduce

1. Check current `main` in `lib/crewai/tests/agents/test_agent.py`.
2. Locate `test_tool_result_as_answer_is_the_final_answer_for_the_agent`.
3. Observe that the test is marked skipped and still depends on a VCR recording.
4. Remove only the skip and run it: the stale recording no longer matches the current native request shape, so it cannot provide deterministic coverage.

## Expected behavior

The integration test should run offline and verify that a native tool with `result_as_answer=True`:

- returns its result as the Crew's final output;
- stops the native agent loop without a second LLM call; and
- does not depend on provider credentials or a VCR cassette.

## Screenshots/Code snippets

A deterministic test can patch the native executor's `get_llm_response` with one OpenAI-style tool call and assert both `result.raw == "Howdy!"` and a single LLM invocation.

## Operating System

Windows 11

## Python Version

3.13

## crewAI Version

Current `main`

## crewAI Tools Version

Current `main`

## Virtual Environment

Venv (`uv` workspace)

## Evidence

With the LLM response patched at the native executor boundary, the existing Crew-level assertion passes. On Windows, pytest then reports an unrelated teardown error because `latest_kickoff_task_outputs.db` remains open; the test behavior itself passes before teardown.

## Possible Solution

Replace the VCR/skip decorators with a deterministic mocked native tool call and assert the LLM response helper is called once.

## Additional context

I am happy to submit the focused test-only pull request. This issue and the proposed change were prepared with AI coding assistance.

liang0417:
## Related issue

Fixes #7334

## Summary

- replace the skipped VCR-backed Crew-level `result_as_answer` test with a deterministic native tool call
- verify the tool output becomes the Crew's final result
- verify the native loop stops after one LLM response instead of requesting another completion

## Verification

- [ ] Relevant tests and quality checks pass locally

The focused test reaches `1 passed` for its assertions on Windows/Python 3.13. Pytest then reports an unrelated existing teardown error because `latest_kickoff_task_outputs.db` remains locked while the temporary test directory is removed. `git diff --check` passes, and the staged redaction scan reports 0 blockers and 0 advisories.

## Additional context

This pull request was prepared with AI coding assistance. I attempted to apply the repository's required `llm-generated` label, but GitHub does not grant external contributors the `AddLabelsToLabelable` permission. Maintainers: please apply the label during triage.


Vidit-Ostwal (approved):
Looks good. Replaces the skipped VCR `result_as_answer` test with a deterministic native tool-call mock and asserts the tool output is the Crew result after a single LLM response. Approving.

PR #7340, opened 2026-09-08
refactor(llms): single source of truth for context windows

### LINKED ISSUE #7303: [BUG] o1, o1-pro, and o3 reasoning models fallback to 8k default context window
### Description

OpenAI's flagship reasoning models (`o1`, `o1-pro`, `o3`) and Azure OpenAI deployments of `o1`, `o1-mini`, and `o3-mini` are missing from `LLM_CONTEXT_WINDOW_SIZES` and native provider context window tables.

Because prefix lookup fails (`"o1".startswith("o1-preview")` is false), querying with `model="o1"` or `model="o3"` falls back to `DEFAULT_CONTEXT_WINDOW_SIZE` (8,192 tokens, or 6,144 usable tokens with ratio) instead of their official 200,000 tokens. This can trigger false `LLMContextLengthExceededError` or premature prompt truncation.

### Proposed Solution
Register official 200k context windows for `o1`, `o1-pro`, and `o3` across `llm.py`, `openai/completion.py`, and `azure/completion.py`, while preserving the 128k windows for `o1-preview` and `o1-mini`.

### Steps to Reproduce

1. Initialize an LLM instance with an o-series model:
   llm = LLM(model="o1", is_litellm=True)
2. Query its context window:
   size = llm.get_context_window_size()
3. Check the returned size:
   Expected ~150,000 (200k * 0.75 ratio), but returns 6,144 (8k * 0.75 ratio).

### Expected behavior

`o1`, `o1-pro`, and `o3` should recognize the official 200,000 context window (150,000 usable after ratio), while `o1-preview` and `o1-mini` retain their 128,000 window.

### Screenshots/Code snippets

from crewai.llm import LLM

llm = LLM(model="o1")
print(llm.get_context_window_size())
# Outputs: 6144 (Incorrect: default 8192 * 0.75)
# Expected: 150000 (Official 200000 * 0.75)

### Operating System

macOS Sonoma

### Python Version

3.12

### crewAI Version

1.15.20

### crewAI Tools Version

1.15.20

### Virtual Environment

Venv

### Evidence

In lib/crewai/src/crewai/llm.py:
LLM_CONTEXT_WINDOW_SIZES contains "o1-preview": 128000 and "o1-mini": 128000, but omits "o1", "o1-pro", and "o3".

In lib/crewai/src/crewai/llms/providers/openai/completion.py:
context_windows omits "o1", "o1-pro", and "o3".

In lib/crewai/src/crewai/llms/providers/azure/completion.py:
context_windows omits all o-series models.

### Possible Solution

1. Add "o1": 200000, "o1-pro": 200000, "o3": 200000 to LLM_CONTEXT_WINDOW_SIZES in lib/crewai/src/crewai/llm.py.
2. Add "o1-pro", "o1", "o3" to context_windows in openai/completion.py with longest-prefix-first ordering.
3. Add o-series models to azure/completion.py context_windows.

### Additional context

I have already developed and tested a patch with unit tests in test_llm.py, test_openai.py, and test_azure.py, and will submit a PR linking to this issue immediately.

### LINKED ISSUE #7304: [IMPROVEMENT] Single source of truth for LLM context window sizes
### Feature Area

Core functionality

### Is your feature request related to an existing bug? Please link it here.

Related: #7303 (o1 / o1-pro / o3 fall back to the 8k default because they are missing from some of the duplicated tables). That class of bug is what this issue is meant to stop.

Context window sizes are hardcoded in **six places**. Adding or updating a model means editing `llm.py` **and** the native provider (and sometimes Azure or Bedrock too). The copies have already drifted.

**Current maps**

| Location | Used by |
|---|---|
| `lib/crewai/src/crewai/llm.py` — `LLM_CONTEXT_WINDOW_SIZES` | LiteLLM fallback |
| `lib/crewai/src/crewai/llms/providers/openai/completion.py` | Native OpenAI |
| `lib/crewai/src/crewai/llms/providers/azure/completion.py` | Native Azure |
| `lib/crewai/src/crewai/llms/providers/gemini/completion.py` | Native Gemini |
| `lib/crewai/src/crewai/llms/providers/anthropic/completion.py` | Native Anthropic |
| `lib/crewai/src/crewai/llms/providers/bedrock/completion.py` | Native Bedrock |

OpenAI / Azure / Gemini import `LLM_CONTEXT_WINDOW_SIZES` only to validate bounds, then ignore it and use a local dict.

**Known drift (fix these as part of the merge)**

- Native OpenAI has `gpt-5` / `gpt-5-mini` / `gpt-5-nano` (1,047,576). LiteLLM `gpt-5` falls through to **8192**.
- Native Anthropic has 1M models (`claude-sonnet-4-6`, `claude-opus-5`, …) that `llm.py` does not list as bare Claude ids.
- Native Bedrock prefix `anthropic.claude-sonnet-4` is 200k, so `anthropic.claude-sonnet-4-6` (1M) gets **200k** on the native path.
- OpenAI-compatible (`deepseek/…`, `openrouter/…`) inherits OpenAI’s local map, so LiteLLM-only keys like `deepseek-chat` become **8192**.

Lookup rules also disagree: native providers first-match (keys inserted longest-first); `llm.py` last-match-wins.

### Suggested implementation (good first issue)

One module owns lookup + constants. Provider files only declare **their** prefixes.

1. **Add** `lib/crewai/src/crewai/llms/context_window.py`
   - `CONTEXT_WINDOW_USAGE_RATIO` (`0.85`), `DEFAULT_CONTEXT_WINDOW_SIZE` (`8192`), `MIN_CONTEXT`, `MAX_CONTEXT`
   - `resolve_context_window_size(model, sizes, *, default, extra_names=())` — **longest prefix wins**, then apply the 0.85 ratio; validate bounds once
   - Family maps (source of truth for sizes):
     - `OPENAI_CONTEXT_WINDOWS` — GPT / o-series (OpenAI + Azure)
     - `AZURE_CONTEXT_WINDOWS` — Azure-only extras (`gpt-35-turbo`, `text-embedding`)
     - `ANTHROPIC_CONTEXT_WINDOWS` — bare `claude-*` prefixes
     - `GEMINI_CONTEXT_WINDOWS`
     - `BEDROCK_CONTEXT_WINDOWS` — Titan, Nova, Llama, etc.
     - `LITELLM_CONTEXT_WINDOWS` — Groq / Mistral / Together / other LiteLLM-only ids currently in `llm.py`
   - Expand Bedrock Claude ids from `ANTHROPIC_CONTEXT_WINDOWS` (`anthropic.`, `us.anthropic.`, `eu.anthropic.`, `apac.anthropic.`, `global.anthropic.`) so a new Claude window is added once
   - `LLM_CONTEXT_WINDOW_SIZES` = merge of the above (LiteLLM + OpenAI-compatible)

2. **Rewire providers** — delete the local `context_windows` dicts. `get_context_window_size` becomes a one-liner against that provider’s map. Keep today’s defaults: OpenAI/Azure/Bedrock 8192, Anthropic 200k, Gemini 1M.

3. **Re-export from** `lib/crewai/src/crewai/llm.py`
   - `LLM_CONTEXT_WINDOW_SIZES`, `CONTEXT_WINDOW_USAGE_RATIO`, `DEFAULT_CONTEXT_WINDOW_SIZE` so existing tests (`patch.dict("crewai.llm.LLM_CONTEXT_WINDOW_SIZES", …)`) keep working.
   - LiteLLM `get_context_window_size` calls the shared resolver with the merged map and `_context_window_model_name()`.

4. **Do not change** `base_llm.DEFAULT_CONTEXT_WINDOW_SIZE = 4096` (custom LLM fallback). That is a different default.

After this, adding `gpt-5.7` is one entry in `OPENAI_CONTEXT_WINDOWS`. Native OpenAI, Azure, LiteLLM, and OpenAI-compatible all pick it up. Adding `claude-sonnet-4-7` is one Anthropic entry; Bedrock regional ids are generated.

### Acceptance criteria

- [ ] One module owns raw window sizes and longest-prefix lookup. Provider completion files do not copy the full list.
- [ ] Adding a GPT / Claude / Gemini model is a single map entry (plus Bedrock prefix expansion for Claude).
- [ ] `crewai.llm` still exports `LLM_CONTEXT_WINDOW_SIZES`, `CONTEXT_WINDOW_USAGE_RATIO`, and `DEFAULT_CONTEXT_WINDOW_SIZE`.
- [ ] Existing tests stay green: `lib/crewai/tests/test_llm.py`, `llms/openai/test_openai.py`, `llms/azure/test_azure.py`, `llms/anthropic/test_anthropic.py`.
- [ ] New unit tests for `resolve_context_window_size` (longest prefix, bounds, default).
- [ ] Parity: same family id via native vs `is_litellm=True` returns the same window (`gpt-5`, `claude-sonnet-4-6`, `gemini-2.0-flash`).
- [ ] `LLM(model="bedrock/anthropic.claude-sonnet-4-6").get_context_window_size()` is `int(1_000_000 * CONTEXT_WINDOW_USAGE_RATIO)` (not 200k).
- [ ] LiteLLM `gpt-5` uses the native 1,047,576 window, not 8192. `gpt-5.4-mini` stays 200k.

- Keep one giant dict in `llm.py` and have every provider call it. That is still one source of truth, but provider files would not own only the models they need, and Azure/Bedrock extras get mixed into the OpenAI/Anthropic lists.
- Fetch windows at runtime from LiteLLM’s `model_prices_and_context_window.json`. The CLI already references that URL; it is not used at runtime today and would add a network dependency. Out of scope.

### Additional context

- Do **not** change frozen docs under `docs/v*/`.
- Follow `AGENTS.md`: keep the diff small; write tests for behavior (lookup + parity), not implementation details.

Yao-Y-B:
## Summary
- changes for #7304 
- centralize LLM context window definitions and resolution
- use longest-prefix matching consistently across providers
- preserve compatibility for existing context-window mappings
- include the related changes for #7303

## Testing
- Keep the existing tests stay green
- Same family id via native vs 'is_litellm' returns the same window
- Add 'o1' , 'o1-pro', and 'o3' in 'OPENAI_CONTEXT_WINDOW' and can be accessed by 'LLM_CONTEXT_WINDOW'.
- Full test suite: 383 passed, 4 skipped
- Ruff check passed
- Ruff format check passed

Closes #7304
Closes #7303


Yao-Y-B:
Hi @Vidit-Ostwal — could you re-run the failed tests job when you get a chance?

The only failing test is tests/telemetry/test_task_failure_instrumentation.py::test_on_task_failed_closes_the_span_as_error_either_way[with-crew], and it looks like an order/isolation flake rather than a regression from this PR:
 - The same commit passed on Python 3.10 in this same run. The suite uses pytest-randomly, so each job gets a different order.
- The test passes in isolation (33 passed when the file is run alone).
- The assertion is len(handlers) == 1 for TaskFailedEvent on the shared event bus — a previous test left an extra EventListener handler registered, which is independent of these context-window changes.

 If it reoccurs I'm happy to open a small separate PR to make that telemetry fixture use a fresh bus / clean up handlers.

Yao-Y-B:
@Vidit-Ostwal  CI is green and all checks pass. Could you review and merge when convenient?  This PR covers #7304 and #7303 and supersedes #7329 / #7323.

Vidit-Ostwal:
Hey do you mind, making a check to remove the modles which are depreciated and retired?
For example  "gemini-1.0-pro": 32768" has been retired and removed, do you mind droppping this and rest of the other models as well.

Yao-Y-B:
> Hey do you mind, making a check to remove the modles which are depreciated and retired? For example "gemini-1.0-pro": 32768" has been retired and removed, do you mind droppping this and rest of the other models as well.

@Vidit-Ostwal Thanks!  I will check all the models in their official sites and remove the deprecated/retired ones like gemini-1.0-pro. After this I will make a new commit.  **But I have a question**: some models like 'gpt-3.5-turbo' are in the future deprecation lists according to official docs but they are still in use (there's still some time before the shutdown date), should I remove them as well or just keep them in the maps???

Vidit-Ostwal:
Only remove the ones we have depriciated.

Yao-Y-B:
> Only remove the ones we have depriciated.

I'll remove entries whose models are fully retired/removed from the provider that owns the map (e.g. gemini-1.0-pro). Models that are merely deprecated but still served, or still live on another provider (e.g. OpenAI OK / Azure deprecated), I keep so lookups don't silently fall back to the default window. Let me know if my understanding is correct.

Yao-Y-B:
@Vidit-Ostwal
 Done as requested — I swept the context-window maps and removed entries whose models are retired/removed per the official pages: [OpenAI,](https://developers.openai.com/api/docs/deprecations#past-deprecations) [Azure,](https://learn.microsoft.com/en-us/azure/foundry/openai/concepts/model-retirement-schedule) [Gemini,](https://ai.google.dev/gemini-api/docs/models#model_deprecations) [Bedrock.](https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html)

**A few judgment calls I'd like your sign-off on:**
1. Entries in Bedrock that **I could not find if models are retired** from the official pages are kept as live keys (an extra entry beats an accidental 8k fallback). In LiteLLM, the models which are described as retired in [groq docs ](https://console.groq.com/docs/deprecations)are deleted but I'm not sure if it is the right approach — tell me if you'd rather I delete them outright.
 2. I kept open-weight PascalCase ids (e.g. Meta-Llama-3.3-70B-Instruct) because I can't find their hosts (Together/Fireworks-class), which I couldn't confirm have dropped them.

**Maybe we should purpose a new issue about the entires in LiteLLM? And only make changes to achieve the single source of truth for context windows in this PR?**

 Happy to any of idea from you.

Vidit-Ostwal:
@Yao-Y-B CI is failing mind checking

lib/crewai/src/crewai/llms/context_window.py
```diff
@@ -0,0 +1,250 @@
...
+    "anthropic.claude-sonnet-4-6": 1000000,
+    "anthropic.claude-opus-4-6-v1": 1000000,
+    "anthropic.claude-opus-4-5-20251101-v1:0": 200000,
+    "anthropic.claude-haiku-4-5-20251001-v1:0": 200000,
+    "anthropic.claude-sonnet-4-5-20250929-v1:0": 200000,
+    "anthropic.claude-opus-4-1-20250805-v1:0": 200000,
+    "anthropic.claude-opus-4-20250514-v1:0": 200000,  # TODO(verify)
+    "anthropic.claude-sonnet-4-20250514-v1:0": 200000,  # TODO(verify)
+    "anthropic.claude-3-5-sonnet-20240620-v1:0": 200000,  # TODO(verify)
+    "anthropic.claude-3-5-sonnet-20241022-v2:0": 200000,  # TODO(verify)
```
Vidit-Ostwal, line 190: TODO (verify)
We need to make sure that we are not adding new bugs with this PR?

lib/crewai/src/crewai/llms/context_window.py
```diff
@@ -0,0 +1,250 @@
...
+    "claude-sonnet-5": 1000000,
+    "claude-opus-4-8": 1000000,
+    "claude-opus-4-7": 1000000,
+    "claude-opus-4-6": 1000000,
+    "claude-sonnet-4-6": 1000000,
+    "claude-opus-4-5": 200000,
+    "claude-sonnet-4-5": 200000,
+    "claude-haiku-4-5": 200000,
+}
+
```
Vidit-Ostwal, line 88: We dropped previously supportede gemini models

        context_windows = {
            "gemini-3-pro-preview": 1048576,  # 1M tokens
            "gemini-2.0-flash": 1048576,  # 1M tokens
            "gemini-2.0-flash-thinking": 32768,
            "gemini-2.0-flash-lite": 1048576,
            "gemini-2.5-flash": 1048576,
            "gemini-2.5-pro": 1048576,
            "gemini-1.5-pro": 2097152,  # 2M tokens
            "gemini-1.5-flash": 1048576,
            "gemini-1.5-flash-8b": 1048576,
            "gemini-1.0-pro": 32768,
            "gemma-3-1b": 32000,
            "gemma-3-4b": 128000,
            "gemma-3-12b": 128000,
            "gemma-3-27b": 128000,
        }

Vidit-Ostwal (commented):
We need to see how we can regularly update, this Ideal way would be to add the docs from official website where we can read this, and next time any coding agent would be able to do it automatically.

lib/crewai/src/crewai/llms/context_window.py
Yao-Y-B: I know that now some deleted gemini/gemma ids will fallback now, but I removed the retired models according to [deperation doc](https://ai.google.dev/gemini-api/docs/deprecations).  If you want me to keep previously supported models or something else?

lib/crewai/src/crewai/llms/context_window.py
Yao-Y-B: I checked that the window values  of these models are as same as origin/main. I added #TODO(verify) showing that I couldn't confirm whether the models are retired now or not in [BEDROCK MODEL CARDS](https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html)

lib/crewai/src/crewai/llms/context_window.py
Yao-Y-B: Makes sense — right now `context_window.py` has no trace of where these numbers
came from, so the next person (or agent) has to redo the research. Let me add
that to this PR, since the file is meant to be the single source of truth anyway:

- a module docstring listing the official pages per provider (OpenAI / Azure /
  Gemini / Bedrock / the LiteLLM-hosted ids), plus the refresh procedure;
- a source link **above each map**.

I plan to also write down the decision rule explicitly, so it's machine-followable:
remove an entry only when the owning provider has *fully retired* the model, and
never drop one that is merely deprecated-but-still-served, because a missing id
falls back to the default (8192 on the LiteLLM path) instead of erroring.

Do you think I should implement this in this PR?

Yao-Y-B:
@Vidit-Ostwal

Explanation for [the last push](https://github.com/crewAIInc/crewAI/pull/7340/commits/f990bb595237715fc0be7a1c2a5bcd4c248a5801)

**Fixed — the CI red was deterministic, not flaky.**

Root cause. The failing step is Run tests (group 3 of 8). Three assertions contradict the maps, all introduced by 040fcf4f ("drop deprecated model windows"), which removed keys the tests still required. That commit's own run never executed (conclusion action_required), so the first real run was the Sep 9 merge of main — every run since has been red.

 Note the 30 cancelled jobs are not separate failures: fail-fast: true cancels the rest of the matrix as soon as one job fails. Only 2 jobs actually failed.

**Changes**
 1. test_o_series_reasoning_models — o1-preview / o1-mini are retired and their 128k entries were removed, so the 128k assertions are dropped. They now match the shorter o1 prefix and resolve to 200k rather than falling back to the default; harmless for a retired id, but a prefix collision I left a comment about.
 2. test_context_window_parity_native_vs_litellm — swapped the retired gemini-2.0-flash case for the live gemini-2.5-flash. I replaced rather than deleted it on purpose: this case guards the PR's core "native vs is_litellm return the same window" invariant, and without a Gemini entry the LiteLLM path (default 8192) loses coverage.
 3. test_google.py::test_gemini_context_window_size — this is a pre-existing upstream test that this PR had turned red: it asserted google/gemini-1.5-pro > 1,000,000, and with that id removed the lookup falls back to Gemini's 1M default (891,289). Retargeted to the live google/gemini-2.5-pro. Flagging this explicitly since I'm modifying an upstream assertion, not one this PR added — happy to keep gemini-1.5-pro in the map instead if you'd rather not touch that test.

 I re-verified every get_context_window_size() assertion site under lib/crewai/tests against the maps, so these three are the complete set — not just the first three that --maxfail=3 surfaced.

lib/crewai/src/crewai/llms/context_window.py
Vidit-Ostwal: Yes that would be great

Vidit-Ostwal:
## Context-window audit findings

I compared the provider catalogs in `lib/crewai/src/crewai/llms/constants.py` with the refactored maps in `lib/crewai/src/crewai/llms/context_window.py`, and checked the supported IDs through `LLM._context_window_model_name()` / `resolve_context_window_size()`.

### Anthropic
- The direct Anthropic catalog is unchanged from `main`; all 13 listed IDs resolve in the new map.
- `claude-mythos-5` is an orphaned map entry (not in `ANTHROPIC_MODELS`).
- The current Anthropic Fable ID is `claude-fable-5-1`; the `claude-fable-5` prefix happens to cover it, but it is not catalogued exactly.

### Azure
- The catalog is unchanged, but generic prefix matching gives incorrect limits: GPT-3.5 `-0301`/`-0613` resolve to 16,385 instead of 4,096; GPT-4 32K and preview/vision variants resolve to 8,192 instead of their larger limits.
- `gpt-5` inherits 1,047,576 from the shared OpenAI map, while Azure documents 400K.

### OpenAI
- 20 catalogued IDs fall back to 8,192: all GPT-3.5, audio, image, realtime, and `whisper-1` families.
- GPT-4 32K / preview / vision variants resolve through the generic 8,192 `gpt-4` prefix.
- `o1-preview` and `o1-mini` changed from `main`'s explicit 128K entries to the generic 200K `o1` prefix.
- GPT-5 resolves to 1,047,576, but current OpenAI docs state 400K. GPT-4.1 and GPT-4o are correctly mapped.

### Gemini
- This is the main regression: 29 catalogued Gemini IDs now fall back to 8,192.
- The refactor removed `main`'s explicit windows for Gemini 3 Preview, Gemini 2.0 Flash/Flash-Lite, Gemini 1.5 Pro/Flash/Flash-8B, and many Live/experimental/latest aliases. Examples: Gemini 3 Pro Preview and Gemini 2.0 Flash are 1M; Gemini 1.5 Pro is 2,097,152.

### Bedrock
- 28 of 95 catalogued IDs fall back to 8,192.
- `amazon.nova-premier-v1:0` (and variants) is unmapped despite a 1M window; `anthropic.claude-3-5-haiku-20241022-v1:0` regressed from `main`'s 200K entry to the default.
- Llama 4, several Llama 3.2/3.3 IDs, AI21 Jamba, Cohere Command R, Mistral/Pixtral, Qwen, GPT-OSS, TwelveLabs, and `global.anthropic.claude-sonnet-4-20250514-v1:0` are unmapped.
- The generated Claude 4.5/4.6 regional/global prefixes are an improvement over `main`; they should be retained while restoring the missing coverage.

### Sources of truth to use for corrections
- Anthropic: Models API (`max_input_tokens`) and model overview: https://platform.claude.com/docs/en/models/overview
- Azure: Azure Foundry model catalog / deployment-version pages: https://learn.microsoft.com/en-us/azure/ai-foundry/azure-openai-in-ai-foundry
- OpenAI: per-model API catalog pages, e.g. https://developers.openai.com/api/docs/models/gpt-5
- Gemini: Models API (`inputTokenLimit`) or the model pages: https://ai.google.dev/api/models and https://ai.google.dev/gemini-api/docs/models
- Bedrock: per-model AWS model cards: https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html

The provider APIs are preferable where available, because limits can vary by snapshot, region, deployment tier, or rollout.

Yao-Y-B:
Thanks for the audit — this was exactly the right check to run. I reproduced it
independently before writing any code: comparing `constants.py` against the maps
gives **29 catalogued Gemini ids and 28 of 95 Bedrock ids** falling back, which
matches your numbers.

Ids that `main` resolved correctly and that `040fcf4f` left falling back, plus the
`o1` prefix collision you spotted:

| Group | Ids |
|---|---|
| Gemini (14) | `gemini-3-pro-preview`, `gemini-2.0-flash` and its `-001` / `-exp` / `-exp-image-generation` / `-lite` / `-lite-001` / `-lite-preview` / `-lite-preview-02-05` / `-preview-image-generation` / `-live-001` variants, `gemini-1.5-pro`, `gemini-1.5-flash`, `gemini-1.5-flash-8b` |
| Bedrock (1) | `anthropic.claude-3-5-haiku-20241022-v1:0` (your example — it regressed from main's 200K to the default) |
| OpenAI (2) | `o1-preview`, `o1-mini` — these were **not** falling back but resolving to the wrong window: the shorter `o1` prefix was matching, giving 200k instead of 128k |

Nine map entries cover all of them, because longest-prefix matching then supplies
the variant ids.

**Documentation (what you approved)**

`context_window.py` now documents itself as the single source of truth: the
per-provider pages to re-check (OpenAI / Azure / Anthropic / Gemini / Bedrock /
LiteLLM), the refresh procedure, and the rule written down explicitly so it is
machine-followable —

> Drop an entry only when the owning provider has *fully retired* the model **and**
> the id is removed from the matching `*_MODELS` catalog in the same change.
> Removing the map entry while the id stays catalogued is a bug: the lookup falls
> back to the default instead of erroring.
LiteLLM), the refresh procedure, and the rule written down explicitly so it is
machine-followable —

> Drop an entry only when the owning provider has *fully retired* the model **and**
> the id is removed from the matching `*_MODELS` catalog in the same change.
> Removing the map entry while the id stays catalogued is a bug: the lookup falls
> back to the default instead of erroring.

**Reverted the earlier test edits**

`f990bb595` had relaxed three assertions to match the maps. That was the wrong
direction — with the maps fixed, `tests/llms/google/test_google.py` is
byte-identical to `main` again and `test_context_window.py` is back to its
pre-fix state.

PR #7342, opened 2026-09-08
fix(openai): surface gateway errors reported inside an HTTP 200

joaomdmoura:
OpenAI-compatible gateways commit `200 OK` as soon as a provider accepts a request, so a later provider failure arrives in the body as an `error` object with no `choices`. That reached the SDK's parse helper and surfaced as `TypeError: 'NoneType' object is not iterable`, naming neither the provider, the status, nor the timeout that caused it. Non-streaming calls now raise the exception the upstream status maps to — a masked 504 is catchable exactly like an honest one.

- Covers all four non-streaming paths (sync/async × structured/plain); streaming already had this guard inside the SDK. 504 → `openai.InternalServerError`, 429 → `RateLimitError`, malformed 200 → `APIResponseValidationError`. No new public API.
- Compatibility: these paths now call `chat.completions.with_raw_response.create/parse`, so test doubles patching `chat.completions.create` need `with_raw_response.create` — four in-repo doubles updated.
- 39 tests for the gateway paths: happy path, token usage, tool execution and the follow-up turn, `LengthFinishReasonError`, the 404 → Responses fallback, and that the message carries only the host so URL credentials are never logged.
- Also fixes pre-existing test pollution that was blocking this branch: `TraceCollectionListener` is a singleton, and a leaked `trace_batch_id` redirected later trace POSTs until an unrelated cassette missed. An autouse fixture drops the cached instance; reproducible on `main` without this branch.
- Docs: `concepts/llms.mdx` in `en`, `ar`, `ko`, `pt-BR`. Out: the Responses API path and `crewai-tools` platform actions share the gateway defect and need tickets.

---

> [!NOTE]
> **Medium Risk**
> Changes core LLM error handling on all non-streaming OpenAI-compatible completion paths; behavior is more correct but callers mocking only `create` must update, and new exception types may alter retry/fallback flows.
> 
> **Overview**
> Fixes **masked gateway failures** on non-streaming OpenAI Chat Completions: when a gateway returns HTTP `200` with an `error` body and no `choices`, the provider now inspects the raw response and raises the same `openai.*` exception type the upstream status would have used (e.g. 504 → `InternalServerError`), instead of a cryptic `TypeError` during parse. Sync/async plain and structured paths all go through `with_raw_response` plus `_raise_for_upstream_error`; malformed 200s without a proper error envelope become `APIResponseValidationError`.
> 
> Adds broad **MockTransport** coverage and updates in-repo fakes that patched `chat.completions.create` to expose `with_raw_response.create`. Docs add a **Gateway Errors** tab under common LLM issues (`en`, `ar`, `ko`, `pt-BR`).
> 
> Separately, an autouse **`reset_tracing_state`** fixture clears the `TraceCollectionListener` singleton and tracing context vars after each test, with canary tests—addressing suite pollution that leaked batch IDs and broke unrelated tracing/telemetry tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e334022fecfa695c0c151af92b78cf0e0c9a0402. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joaomdmoura:
Fixed — you're right that the assertion was vacuous, and chasing it down found two further problems.

**Confirmed:** `_listeners_setup` is assigned on the instance at `trace_listener.py:229`, and nothing in `lib/` ever assigns it at class level, so `TraceCollectionListener._listeners_setup` is permanently `False` and the canary could never fail.

**The suggested fix doesn't hold either.** `BaseEventListener.__init__` calls `self.setup_listeners(crewai_event_bus)` at `base_event_listener.py:16`, so construction itself sets the flag — `TraceCollectionListener()._listeners_setup` is always `True`. I applied it and it failed immediately, including with the isolation fixture active:

```
FAILED tests/tracing/test_trace_listener_isolation.py::test_listener_setup_flag_starts_clean
```

So the class read is always `False` and the instance read is always `True`; neither observes a leak. The flag lives on the instance the fixture deletes, so it cannot outlive a test — that's guaranteed by construction, not assertable. I removed the canary and recorded the reasoning in the module docstring so nobody re-adds it in either form.

**Knock-on fix:** the same fact meant the fixture was clearing `_initialized` and `_listeners_setup` at class level, where they are never set. Only dropping `_instance` is load-bearing, so the fixture is now one line.

Verified: the surviving canary still fails without the fixture (`AssertionError: assert 'debug-trace-batch' is None`), so it isn't vacuous; and three full runs of `lib/crewai/tests/` give 438 failures, byte-identical to base `a68b5e903`.

joaomdmoura:
Fixed. You're right, and my note overstated it.

`setup_listeners` returns at `trace_listener.py:213-220` when `should_enable_tracing()`, `is_tracing_enabled_in_context()`, `should_auto_collect_first_time_traces()` and `is_tui_mode()` are all false, so the assignment at `:229` never runs. Confirmed:

```
tracing disabled -> _listeners_setup: False   (flag not even in instance __dict__)
tracing enabled  -> _listeners_setup: True
```

So construction sets it only when registration actually runs. The docstring now says that.

This makes the case for deleting the canary stronger rather than weaker: the instance read reports whether tracing happened to be enabled, i.e. ambient configuration, not whether state leaked between tests. Combined with the class read being permanently `False`, neither form can observe a leak — so the flag's isolation is guaranteed by dropping the instance, not by an assertion.

Docstring-only change; `tests/tracing/` is 58 passed.

joaomdmoura:
Fixed — and this was a good catch, high severity was right.

**Confirmed the mechanism.** After a tracing-enabled run the flag survives on the instance, `cleanup_event_handlers` wipes the bus, and re-running setup restores nothing:

```
after tracing-enabled construction: _listeners_setup = True
event types registered: 119
after bus wipe, event types: 0
after re-setup, event types: 0     <- silently registers nothing
```

So tracing collected no events for the rest of the worker. This predates the PR — the probe above touches no fixture — but this fixture is the right place to fix it, and my previous revision made it worse by explicitly declining to clear the flag.

**Why the earlier attempt at this regressed.** Dropping the singleton is the fix, but it broke `test_task_failure_instrumentation` with two `on_task_failed` handlers. Chasing that found a third leak: the `_tracing_enabled` context var stayed set, so the replacement listener still believed tracing was on and registered its handlers next to telemetry's. Clearing the context vars is what makes replacing the listener safe — with `_tracing_enabled` reset, the replacement sees tracing disabled and registers nothing.

The fixture now drops `_instance` and resets `_tracing_enabled` / `_tui_mode` together, which covers all three leaks: batch state, the setup flag, and tracing context.

**Verified:**
- Your case, as a canary — fails without the drop (`assert True is False`), so it isn't vacuous.
- `test_task_failure_instrumentation` after a tracing-enabled test: 34 passed, matching base.
- The original cassette-miss repro still fixed.
- `tracing/` + `telemetry/` + `test_flow_conversation`: 527 passed, twice, random order.
- Full suite: 392 failed / 4602 passed against base's 392 / 4561 — failure sets byte-identical, no regressions.

This also means the canary you and CodeRabbit originally asked for is now writable, because the context reset makes the expected value well-defined. It's back in, as `test_listener_can_register_handlers_again`.

lorenzejay (approved):

lorenzejay (approved):

PR #7346, opened 2026-09-09
Fix/agent executor async llm path

### LINKED ISSUE #7339: [Reopen] AgentExecutor async path still routes through LLM.call
## Summary
Issue `#6683` was previously closed by inactivity, but the behavior is still present and impacts async pipelines.

In `AgentExecutor.ainvoke()`/`run_async` flow, LLM calls are routed through the shared synchronous path (`get_llm_response()`), even when the provider implements an async path (`LLM.acall()` / `aget_llm_response()`).

As a result, async workloads can be constrained by thread-pool capacity and lose the expected async execution benefits.

## Current behavior
- `AgentExecutor` methods `call_llm_and_parse` and `call_llm_native_tools` call `get_llm_response()`.
- `get_llm_response()` invokes `LLM.call()`.
- `LLM.call()` uses sync execution path, requiring threadpool offload for providers that are actually async-capable.

## Expected behavior
- Async execution paths should use `aget_llm_response()` and `LLM.acall()` when available.
- Preserve compatibility for custom LLMs that only implement sync behavior by falling back to `asyncio.to_thread(...)` only on `NotImplementedError`.

## Suggested fix
1. Convert the two router methods to `async` and `await aget_llm_response()`.
2. In `aget_llm_response()`, implement narrow fallback:
   - try async path first
   - on `NotImplementedError`, delegate to `asyncio.to_thread(llm.call, ...)` (or equivalent)
3. Keep sync behavior unchanged for non-async contexts.

## Repro (minimal)
1. Create a custom LLM double that records calls for both `call()` and `acall()`.
2. Attach it to an `Agent` and invoke `await agent.kickoff_async("probe")`.
3. Validate that `acall()` is invoked when available.

If needed, I can provide a tiny regression test draft.

## Context
This is a re-open-style report derived from `#6683` and current source inspection.

stud925:
## Related issue

Fixes #7339

## Summary

The experimental AgentExecutor ran every LLM turn through the synchronous get_llm_response() → LLM.call(), even under  ainvoke()/kickoff_async(). Because the router methods were sync, the Flow runtime offloaded each one to a worker thread, so async pipelines were bounded by thread-pool capacity and providers' native async clients (LLM.acall()) were never used. The fix makes call_llm_and_parse and call_llm_native_tools async and has them await a new aget_llm_response_with_fallback() helper, which calls aget_llm_response() (LLM.acall()) and only drops to a threaded get_llm_response() on NotImplementedError, so custom sync-only BaseLLM subclasses keep working.

## Verification

- Added in tests/utilities/test_agent_utils.py, 
  - class TestAgetLlmResponseWithFallback:

  - test_awaits_acall_when_available: LLM double implementing both
    call and acall; asserts the helper awaits acall and never touches
    call. The core behavior of the change.
  - test_falls_back_when_acall_not_implemented: LLM double with only
    call (inherits BaseLLM.acall, which raises NotImplementedError);
    asserts the helper runs call and returns the result.
  - test_falls_back_when_acall_raises_not_implemented: LLM double
    that defines acall but raises NotImplementedError inside it
    (Bedrock-without-aiobotocore shape); asserts it still falls back
    to call. This is the case that justifies try/except over an
    up-front capability check.
  - test_propagates_non_not_implemented_errors: LLM double whose
    acall raises RuntimeError; asserts the error propagates and
    there's no fallback to call (the except is intentionally narrow to
    NotImplementedError).

- Updated in tests/agents/test_agent_executor.py:

  - test_call_llm_and_parse_does_not_pass_response_model_with_tools: 
     router is now async; wrapped call in asyncio.run, patch target moved to aget_llm_response_with_fallback.
  - test_call_llm_native_tools_does_not_pass_response_model_with_tools: downgrades to text tool-calling.
  - test_call_llm_parser_error: same; still asserts OutputParserError
    → "parser_error".
  - test_call_llm_context_error: same; still asserts context-length
    error → "context_error".

- `ruff check` on the two changed source files → no new violations; error count identical with and without the diff (all pre-existing).
- `ruff format --check` on the changed files → clean.

### DIFF
--- a/lib/crewai/src/crewai/experimental/agent_executor.py
+++ b/lib/crewai/src/crewai/experimental/agent_executor.py
@@ -86,13 +86,13 @@ from crewai.tools.tool_failure import (
 )
 from crewai.utilities.agent_utils import (
     _llm_stop_words_applied,
+    aget_llm_response_with_fallback,
     build_text_tool_calling_fallback_message,
     check_native_tool_support,
     enforce_rpm_limit,
     extract_tool_call_info,
     format_message_for_llm,
     format_native_tool_output_for_agent,
-    get_llm_response,
     handle_agent_action_core,
     handle_context_length,
     handle_max_iterations_exceeded,
@@ -1453,7 +1453,9 @@ class AgentExecutor(Flow[AgentExecutorState], BaseAgentExecutor):
         return "agent_finished"
 
     @router("continue_reasoning")
-    def call_llm_and_parse(self) -> Literal["parsed", "parser_error", "context_error"]:
+    async def call_llm_and_parse(
+        self,
+    ) -> Literal["parsed", "parser_error", "context_error"]:
         """Execute LLM call with hooks and parse the response.
 
         Returns routing decision based on parsing result.
@@ -1468,7 +1470,7 @@ class AgentExecutor(Flow[AgentExecutorState], BaseAgentExecutor):
                 None if self.original_tools else self.response_model
             )
 
-            answer = get_llm_response(
+            answer = await aget_llm_response_with_fallback(
                 llm=self.llm,
                 messages=list(self.state.messages),
                 callbacks=self.callbacks,
@@ -1526,7 +1528,7 @@ class AgentExecutor(Flow[AgentExecutorState], BaseAgentExecutor):
             raise
 
     @router("continue_reasoning_native")
-    def call_llm_native_tools(
+    async def call_llm_native_tools(
         self,
     ) -> Literal[
         "native_tool_calls",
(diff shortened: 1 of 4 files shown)


Vidit-Ostwal (commented):
Thanks for wiring the experimental routers through an async path — making `call_llm_and_parse` / `call_llm_native_tools` `async` is the right Flow change so `kickoff_async` awaits them instead of `to_thread`-ing the whole router.

Please drop `aget_llm_response_with_fallback` and put the fallback inside `aget_llm_response`. A second public helper that re-enters `get_llm_response` is the wrong seam:

- It duplicates the full signature for a 4-line `try`/`except`.
- On `NotImplementedError`, `_prepare_llm_call` / before-hooks run twice (once in `aget_llm_response`, again in `get_llm_response`).
- `crew_agent_executor` already calls `aget_llm_response` and would not get the fallback.

Keep prepare/validate once; only the LLM call should switch:

```python
with _prepare_llm_call(...) as prepared_messages:
    try:
        answer = await llm.acall(prepared_messages, ...)
    except NotImplementedError:
        answer = await asyncio.to_thread(llm.call, prepared_messages, ...)
```

Catch only `NotImplementedError`. Thread `llm.call`, not the whole sync helper.

Then:

- Routers `await aget_llm_response(...)`.
- Move the new tests onto `aget_llm_response`.
- Point executor patches at `aget_llm_response`.

Sync `get_llm_response` / `invoke` should stay unchanged.

PR #7350, opened 2026-09-09
chore(tools): make the vision_tool more dynamic

danielfsbarreto:
## Summary

Makes `VisionTool` more dynamic and configurable.
- Adds a `query` argument so agents can ask a custom question about an image (defaults to `"What's in this image?"` instead of being hardcoded).
- Adds a `complexity_level` argument (`easy` / `medium` / `hard`) that automatically selects the model:
  - `easy` → `gpt-5.6-luna`
  - `medium` → `gpt-5.6-terra` (default)
  - `hard` → `gpt-5.6-sol`
- Changes the default model from `gpt-4o-mini` to `gpt-5.6-terra`.
- An explicitly provided `llm` or `model` still takes precedence over complexity-based selection.
### Docs
- Updates `visiontool.mdx` (en) with the new arguments and complexity-to-model table.
- Syncs translations for `ar`, `ko`, and `pt-BR`.

## Verification

- [ ] Tests added or updated for the changed behavior
- [ ] Relevant tests and quality checks pass locally


Vidit-Ostwal:
## Review

### 1. Bug: `complexity_level` gets silently ignored after the LLM gets cached
`vision_tool.py:129` — `_llm_for_complexity` uses `if self._llm is not None: return self._llm` as its precedence check. This can't distinguish two different reasons `self._llm` is non-`None`:
- an explicit `llm=` was passed to `__init__` (should always win — correct), vs
- `self._llm` merely got cached as a side effect of something reading the `tool.llm` property, or of `model=` being set (should **not** override `complexity_level` — currently does).

Once `self._llm` is populated by either path, every subsequent `_run` call reuses that cached model regardless of the `complexity_level` passed in that call.

**Repro:**
```python
tool = VisionTool()
_ = tool.llm  # caches medium-tier model as a side effect
tool._run(image_path_url=x, complexity_level="hard")  # still uses the cached medium model, not hard-tier
```

**Fix:** track "LLM explicitly supplied" with its own flag, set only inside `__init__` when `llm is not None`. Check that flag first in `_llm_for_complexity`, instead of `self._llm is not None`.

**Test to add:** read `tool.llm` (or otherwise populate `self._llm` outside the constructor), then call `_run` with a non-medium `complexity_level`, and assert the model matching that complexity level is used — not the cached one.

### 2. Inconsistent LLM caching across branches
`vision_tool.py:127-137` — the explicit-`llm` and explicit-`model` paths cache the resolved `LLM` onto `self._llm`; the complexity-map branch builds a fresh `LLM` on every `_run` call and never caches it. Not a correctness bug by itself, but worth resolving consistently once the fix above reshapes this method — otherwise it's needless re-instantiation on every call.

### 3. Missing test: invalid `complexity_level`
No test exercises a bad/unexpected value (e.g. an agent passing `"very hard"`). `ImagePromptSchema`'s `Literal` validation will raise, but it currently gets swallowed by the blanket `except Exception` in `_run` into a generic `"An error occurred: ..."` string. Add a test asserting this is the intended behavior (or handle it more explicitly if not).

### 4. Missing test: same tool instance, two different complexity levels
No existing test reuses one `VisionTool` instance across two `_run` calls with different `complexity_level` values. This is exactly the scenario bug #1 breaks — add a test asserting each call resolves its own model independently, not just the single-call-per-instance tests currently present.

lib/crewai-tools/src/crewai_tools/tools/vision_tool/vision_tool.py
```diff
@@ -10,10 +10,34 @@
...
+# Model used when no complexity level or explicit model/LLM is provided.
+DEFAULT_MODEL: str = COMPLEXITY_MODEL_MAP["medium"]
+
+
 class ImagePromptSchema(BaseModel):
     """Input for Vision Tool."""
 
     image_path_url: str = "The image path or URL."
+    query: str = Field(
+        default="What's in this image?",
```
lorenzejay, line 31: should we drop this now that we should assume agents provide this?

lorenzejay (approved):

lib/crewai-tools/src/crewai_tools/tools/vision_tool/vision_tool.py
Vidit-Ostwal: Let's keep it for fallback purpose.

danielfsbarreto: My thoughts exactly. Wanted to impact existing behavior as little as possible.

PR #7361, opened 2026-09-09
fix(tasks): clean up execution_spans on async CancelledError

### LINKED ISSUE #7351: [BUG] Cancelled async tasks remain in execution_spans and retain task graphs
### Description

Cancelling an async task after `TaskStartedEvent` leaves the task in the global `EventListener.execution_spans` dictionary. The dictionary keeps a strong reference to the `Task`, which can retain its `Agent`, `Crew`, LLM clients, tools, and other request-scoped objects.

The completed and ordinary-failure paths remove entries with `pop()`, following the fix for #4222 / #4161. The cancellation path remains uncovered because `Task._aexecute_core()` catches `Exception`, while `asyncio.CancelledError` inherits directly from `BaseException`.

Current `main` at `a53ecc17f1829354e2d2ff58bbbad1d84ce568b2` still has this flow:

```python
crewai_event_bus.emit(
    self, TaskStartedEvent(context=context, task=self)
)

try:
    ...
except Exception as e:
    crewai_event_bus.emit(
        self,
        TaskFailedEvent(error=str(e), error_type=type(e), task=self),
    )
    raise e
```

`CancelledError` bypasses the `TaskFailedEvent`, so the listener never runs `self.execution_spans.pop(source, None)`.

### Steps to Reproduce

Run this with CrewAI 1.15.2 and Python 3.12. It does not call an LLM because `Agent.aexecute_task` is replaced with a deterministic cancelled coroutine.

```python
import asyncio
import gc
import weakref

from crewai import Agent, Crew, Task
from crewai.events.event_listener import event_listener


async def cancelled(self, *args, **kwargs):
    raise asyncio.CancelledError()


async def main():
    event_listener.execution_spans.clear()
    task_refs = []
    agent_refs = []
    crew_refs = []

    with patch.object(Agent, "aexecute_task", cancelled):
        for _ in range(5):
            agent = Agent(
                role="test",
                goal="test",
                backstory="test",
                llm="gpt-4o-mini",
            )
            task = Task(
                description="test",
                expected_output="test",
                agent=agent,
            )
            crew = Crew(agents=[agent], tasks=[task], tracing=False)
            task_refs.append(weakref.ref(task))
            agent_refs.append(weakref.ref(agent))
            crew_refs.append(weakref.ref(crew))

            try:
                await crew.akickoff()
            except asyncio.CancelledError:
                pass

            del crew, task, agent

    gc.collect()
    print(
        {
            "execution_spans": len(event_listener.execution_spans),
            "alive_tasks": sum(ref() is not None for ref in task_refs),
            "alive_agents": sum(ref() is not None for ref in agent_refs),
            "alive_crews": sum(ref() is not None for ref in crew_refs),
        }
    )


asyncio.run(main())
```

Observed result:

```text
{'execution_spans': 5, 'alive_tasks': 5, 'alive_agents': 5, 'alive_crews': 5}
```

The same accumulation occurs under repeated request cancellation: after 4, 8, and 12 cancellations, `execution_spans` contains 4, 8, and 12 task entries respectively. A full `gc.collect()` does not remove them. Removing those dictionary entries releases the associated object graphs.

### Expected behavior

Every `TaskStartedEvent` should have a terminal cleanup path. Async cancellation should remove the task from `execution_spans`, close or mark the telemetry span appropriately, and allow the task object graph to be garbage-collected.

### Screenshots/Code snippets

The minimal reproduction and observed output are included above.

### Operating System

macOS (Apple Silicon)

### Python Version

3.12

### crewAI Version

1.15.2; current `main` source at `a53ecc17f1829354e2d2ff58bbbad1d84ce568b2` has the same exception boundary

### crewAI Tools Version

Not installed / not required

### Virtual Environment

Venv

### Evidence

- Successful async executions leave `execution_spans` at zero.
- Ordinary exceptions leave `execution_spans` at zero because `TaskFailedEvent` is emitted.
- `CancelledError` executions add one retained task entry per cancellation.
- Full garbage collection leaves those entries and object graphs reachable.
- Clearing the entries releases the retained graphs.

This is related to #4222, but it is a separate terminal-path gap. #4222 fixed completed and failed tasks by replacing assignment-to-`None` with `pop()`. Async cancellation emits neither of those terminal events.

### Possible Solution

Handle `asyncio.CancelledError` explicitly in `_aexecute_core()` before `except Exception`, emit a terminal task event, and immediately re-raise the cancellation. Reusing `TaskFailedEvent` would be the smallest behavioral change; a dedicated cancellation event could preserve cancellation semantics for telemetry consumers.

Add a regression test asserting that cancellation leaves no entry in `event_listener.execution_spans` and that repeated cancellations do not retain prior task graphs.

### Additional context

This affects long-running async services where client disconnects, request timeouts, shutdown, or orchestration cancellation can propagate into `crew.akickoff()`.

devtechedge:
## Summary

Fixes #7351.

When an async task is cancelled after `TaskStartedEvent`, `asyncio.CancelledError` was skipping the `except Exception` path in `_aexecute_core()`. That meant no terminal event was emitted, so `EventListener.execution_spans` kept a strong reference to the `Task` (and through it the Agent / Crew graph).

This change catches `CancelledError` explicitly, emits `TaskFailedEvent` so the existing listener can pop the span entry and close telemetry the same way ordinary failures do, then re-raises to preserve cancellation semantics.

## Changes

- `lib/crewai/src/crewai/task.py`: handle `asyncio.CancelledError` before `except Exception` in `_aexecute_core()`
- `lib/crewai/tests/telemetry/test_task_cancellation_span_cleanup.py`: regression tests for span cleanup and `TaskFailedEvent` emission on cancel

## Why TaskFailedEvent

Reusing `TaskFailedEvent` is the smallest behavior change. The listener already pops `execution_spans` and routes failures through `task_failed`. A dedicated cancellation event would be nicer for telemetry consumers later, but it is not required to close the leak.

## Test plan

- [x] Added async regression covering repeated cancellations + weakref GC
- [x] Added producer-level check that `CancelledError` emits `TaskFailedEvent`
- [ ] CI: `uv run pytest lib/crewai/tests/telemetry/test_task_cancellation_span_cleanup.py -x -q`

## Notes for maintainers

Happy to adjust if you would rather introduce a dedicated cancellation event instead of reusing `TaskFailedEvent`.

I do not have permission to apply labels here - please add `llm-generated` when you can (required by the contributing guide).

Thanks for reviewing.

### DIFF
--- a/lib/crewai/src/crewai/task.py
+++ b/lib/crewai/src/crewai/task.py
@@ -795,6 +795,16 @@ class Task(BaseModel):
                 TaskCompletedEvent(output=task_output, task=self),
             )
             return task_output
+        except asyncio.CancelledError as e:
+            # CancelledError inherits from BaseException, not Exception, so it bypasses
+            # the ordinary failure path. Emit a terminal event so EventListener can pop
+            # execution_spans and release the retained task graph, then re-raise.
+            self.end_time = datetime.datetime.now()
+            crewai_event_bus.emit(
+                self,
+                TaskFailedEvent(error=str(e) or "CancelledError", error_type=type(e), task=self),
+            )
+            raise
         except Exception as e:
             self.end_time = datetime.datetime.now()
             crewai_event_bus.emit(
--- a/lib/crewai/tests/telemetry/test_task_cancellation_span_cleanup.py
+++ b/lib/crewai/tests/telemetry/test_task_cancellation_span_cleanup.py
@@ -1,0 +1,121 @@
+"""Cancelled async tasks must leave no execution_spans entries.
+
+TaskStartedEvent registers the Task in EventListener.execution_spans. Completed and
+ordinary-failure paths emit terminal events that pop that entry. asyncio.CancelledError
+inherits from BaseException, not Exception, so it used to bypass TaskFailedEvent and
+leak a strong reference to the Task (and its Agent/Crew graph).
+
+See https://github.com/crewAIInc/crewAI/issues/7351
+"""
+
+from __future__ import annotations
+
+import asyncio
+import gc
+import threading
+import weakref
+from unittest.mock import patch
+
+import pytest
+
+from crewai import Agent, Crew, Task
+from crewai.events.event_bus import crewai_event_bus
+from crewai.events.event_listener import EventListener
+from crewai.telemetry import Telemetry
+
+
+@pytest.fixture
+def clean_event_listener():
+    """Reset bus and listener singletons so span map state does not leak across tests."""
+
+    def _reset() -> None:
+        with crewai_event_bus._rwlock.w_locked():
+            crewai_event_bus._sync_handlers.clear()
+            crewai_event_bus._async_handlers.clear()
+        Telemetry._instance = None
+        EventListener._instance = None
+        if hasattr(Telemetry, "_lock"):
+            Telemetry._lock = threading.Lock()
+
+    _reset()
+    listener = EventListener()
+    yield listener
+    _reset()
+
+
+async def _cancelled_aexecute(self, *args, **kwargs):
+    raise asyncio.CancelledError()
+
+
+@pytest.mark.asyncio
+async def test_cancelled_async_task_clears_execution_spans(clean_event_listener):
+    """Cancellation must emit a terminal path and release retained task graphs."""
+    listener = clean_event_listener
+    listener.execution_spans.clear()
+
+    task_refs: list[weakref.ref] = []
+    agent_refs: list[weakref.ref] = []
+    crew_refs: list[weakref.ref] = []
+
+    with patch.object(Agent, "aexecute_task", _cancelled_aexecute):
+        for _ in range(5):
+            agent = Agent(
+                role="test",
+                goal="test",
+                backstory="test",
+                llm="gpt-4o-mini",
+            )
+            task = Task(
+                description="test",
+                expected_output="test",
+                agent=agent,
+            )
+            crew = Crew(agents=[agent], tasks=[task], tracing=False)
+            task_refs.append(weakref.ref(task))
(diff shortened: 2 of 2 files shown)


devtechedge:
Quick note for maintainers: I do not have permission to apply labels on this repo. Please add `llm-generated` when you get a chance - required by the contributing guide. Thanks.

devtechedge:
Hey - gentle ping on this when you have a moment.

Async cancellation after `TaskStartedEvent` was skipping the `Exception` path, so `execution_spans` kept a strong reference to the Task graph; this catches `CancelledError`, emits `TaskFailedEvent` for the existing listener cleanup, then re-raises (`Fixes #7351`).

There is a regression test for span cleanup and event emission on cancel, and the change stays inside `task.py` without inventing a new cancellation event type.

I can rebase, rename the event path, or add more telemetry notes if maintainers prefer a different shape.

No rush - just confirming it is still current and happy to iterate based on feedback.

gaoanze888 (approved):
Reviewed at exact head 1efb3378a6 (fixes #7351).

The fix is correct: `asyncio.CancelledError` derives from `BaseException`, so the existing `except Exception` never caught it, and `execution_spans` retained a strong reference to the Task/Agent/Crew graph. Emitting `TaskFailedEvent` before `raise` gives `EventListener` its terminal event to pop the span, and the surrounding `finally` still clears task files and resets the context token.

The tests are strong: the weakref/GC assertion across five `akickoff` cancellations proves the leak is actually closed (not just that an event fires), and the second test pins `error_type is asyncio.CancelledError` so downstream can distinguish cancellation from failure.

Minor non-blocking observation: reusing `TaskFailedEvent` for cancellation is slightly imprecise semantically, but it is the available terminal-event mechanism and `error_type` disambiguates it. Approving.

PR #7369, opened 2026-09-10
fix(memory): honor read_only on update and recall access times (#7368)

### LINKED ISSUE #7368: [BUG] Memory read_only still persists update and recall access-time writes
### Description

Memory.read_only blocks remember and remember_many, but two other implicit write paths still mutate the backing store: update rewrites a record, and recall refreshes last_accessed through touch_records. As a result, reading through a read-only Memory leaves a persistent trace.

### Steps to Reproduce

1. Create a Memory with local storage and store one record while writable.
2. Set memory.read_only = True.
3. Call update on that record; its content changes.
4. Or call recall; the stored last_accessed timestamp advances.

### Expected behavior

When read_only is true, implicit persistence paths leave stored records unchanged. Explicit deletion through forget and reset is a separate operation and can remain available.

### Screenshots/Code snippets

A focused regression is available in PR #7367. It asserts that read-only update returns the existing record unchanged and that read-only recall does not advance last_accessed, with writable positive controls.

### Operating System

Other: macOS Sequoia.

### Python Version

3.12

### crewAI Version

1.15.21

### crewAI Tools Version

Not used.

### Virtual Environment

Venv

### Evidence

The two regression tests fail on the current implementation and pass with PR #7367. The full memory test directory passes: 134 passed, 19 skipped. Ruff and the repository gates are clean.

### Possible Solution

Honor read_only before update writes and before recall invokes touch_records. Return the existing record from read-only update to match remember's silent no-op behavior.

### Additional context

This preserves recall results and leaves explicit forget/reset behavior unchanged. The patch is limited to two behavior checks, documentation, and regression tests.

roli-lpci:
Fixes #7368.

## Summary

Memory.read_only currently blocks remember and remember_many, but update still rewrites records and recall still persists last_accessed via touch_records. This change makes those two implicit mutation paths honor read_only while leaving explicit deletion through forget and reset unchanged.

## Why

A caller using Memory as a read-only retrieval surface should not leave a persistent trace merely by reading. The existing read barrier and reset guards show that the omitted checks were inconsistent with the flag rather than an intentional latency exemption.

## Tests

- Added regression coverage proving update is a no-op in read-only mode.
- Added positive coverage that writable update still works.
- Added regression coverage proving read-only recall does not refresh last_accessed, with a writable positive control.
- pytest lib/crewai/tests/memory: 134 passed, 19 skipped.
- ruff: clean.
- hermes-gate fast and review: PASS, zero findings.

### DIFF
--- a/lib/crewai/src/crewai/memory/unified_memory.py
+++ b/lib/crewai/src/crewai/memory/unified_memory.py
@@ -147,7 +147,12 @@ class Memory(BaseModel):
     )
     read_only: bool = Field(
         default=False,
-        description="If True, remember() and remember_many() are silent no-ops.",
+        description=(
+            "If True, stored records are left unchanged: remember() and "
+            "remember_many() store nothing, update() leaves the record "
+            "untouched, and recall() does not refresh access times. Explicit "
+            "deletion through forget()/reset() is unaffected."
+        ),
     )
     root_scope: str | None = Field(
         default=None,
@@ -781,7 +786,7 @@ class Memory(BaseModel):
                 )
                 results = flow.state.final_results
 
-            if results:
+            if results and not self.read_only:
                 try:
                     touch = getattr(self._storage, "touch_records", None)
                     if touch is not None:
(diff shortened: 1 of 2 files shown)


Vidit-Ostwal:
@lorenzejay Mind taking a look?

Vidit-Ostwal (approved):
Approved. This correctly makes Memory.read_only prevent update writes and recall access-time writes, with focused regression coverage.

PR #7372, opened 2026-09-10
fix(hooks): run tool call hooks on the flow tool action and adapters

impartshadow:
## Related issue

Related to #5888 (crew-level tool call authorization hook). This does not implement the requested contract type; it closes the paths where the existing `pre_tool_call` hook was skipped.

## Summary

A policy registered with `@on(InterceptionPoint.PRE_TOOL_CALL)` or `register_before_tool_call_hook` is dispatched by the ReAct, native function-calling and LiteAgent tool paths. Three call sites reached a tool body without it:

- `flow/runtime/_actions.py` — `ToolAction.run`, the declarative `do: call: tool` action
- `agents/agent_adapters/openai_agents/openai_agent_tool_adapter.py` — the `on_invoke_tool` wrapper
- `agents/agent_adapters/langgraph/langgraph_tool_adapter.py` — the `StructuredTool` wrapper

On those paths a deny was silently ignored and the tool ran. This is the tool-call counterpart of #7111 (model call hooks on every path).

Each site now builds a `ToolCallHookContext`, dispatches `PRE_TOOL_CALL` before the body and `POST_TOOL_CALL` after it, and reports a deny as the same `Tool execution blocked by hook. Tool: <name>` message the executor paths return. In-place mutation of `tool_input` by a hook is honoured: the body receives the dict the hooks saw. `agent`, `task` and `crew` are `None` on these paths, as they are for standalone LiteAgent tool calls. No behaviour change when no hooks are registered.

How the sites were found: a static walk of the installed package that requires every call site reaching a tool body to sit in a function that dispatches `run_before_tool_call_hooks`, or to be declared with a reason. At 1.15.16, 1.15.20 and 1.15.21 these three were the undeclared ones; with this change the walk reports every site covered.

## Verification

- [x] Tests added: `lib/crewai/tests/hooks/test_tool_call_hook_reach.py`. For each of the three paths: the hook sees the call exactly once with its input; a deny keeps the body from running and reaches the caller as the blocked message; a `post_tool_call` rewrite reaches the caller. All 9 fail on `main` before this change (the tool body runs under a deny) and pass after.
- [x] `pytest lib/crewai/tests/hooks lib/crewai/tests/test_flow_from_definition.py lib/crewai/tests/agents/agent_adapters` — 386 passed, 1 skipped
- [x] `ruff check`, `ruff format --check` and `mypy` on the changed files are clean

## Additional context

Authored by an AI agent (this account is an autonomous agent operated by a human). Please apply the `llm-generated` label; labels cannot be set from a fork.

The reducer's handling of unrecognised hook results (`0`, `"deny"`, `{"allow": False}` are currently read as allow) is discussed in #5888 and is deliberately not touched here. This PR only makes the existing hook reach every tool path.

### DIFF
--- a/lib/crewai/src/crewai/agents/agent_adapters/langgraph/langgraph_tool_adapter.py
+++ b/lib/crewai/src/crewai/agents/agent_adapters/langgraph/langgraph_tool_adapter.py
@@ -9,7 +9,13 @@ import inspect
 from typing import Any
 
 from crewai.agents.agent_adapters.base_tool_adapter import BaseToolAdapter
+from crewai.hooks.tool_hooks import (
+    ToolCallHookContext,
+    run_after_tool_call_hooks,
+    run_before_tool_call_hooks,
+)
 from crewai.tools.base_tool import BaseTool
+from crewai.utilities.string_utils import sanitize_tool_name
 
 
 class LangGraphToolAdapter(BaseToolAdapter):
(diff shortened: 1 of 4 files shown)


impartshadow:
Addressed the pre-hook input-rewrite review in b44bfddb2739b0d1e310d5b3556a0599b7b80bb8. A separate parametrized test now mutates `ctx.tool_input["text"]` and checks both the tool body and returned result on the Flow action, OpenAI Agents wrapper, and LangGraph wrapper, while retaining the original exactly-once/input assertions.

Validation: the focused reach suite passes 12 tests with no skips, with network blocked; Ruff lint/format and diff checks pass. This is test coverage only; hook/reducer behavior is unchanged.

PR #7375, opened 2026-09-10
test(bedrock): verify environment credentials

Vidit-Ostwal:
## Summary
- rewrite the Bedrock environment credential test to assert the exact boto3 Session arguments
- verify credentials, session token handling, and AWS region propagation

## Tests
- `UV_CACHE_DIR=/private/tmp/crewai-uv-cache uv run pytest -q lib/crewai/tests/llms/bedrock/test_bedrock.py::test_bedrock_environment_variable_credentials -s`
- 1 passed

### DIFF
--- a/lib/crewai/tests/llms/bedrock/test_bedrock.py
+++ b/lib/crewai/tests/llms/bedrock/test_bedrock.py
@@ -602,19 +602,23 @@ def test_bedrock_tool_conversion():
     assert "inputSchema" in bedrock_tools[0]["toolSpec"]
 
 
-def test_bedrock_environment_variable_credentials(bedrock_mocks):
+def test_bedrock_environment_variable_credentials():
     """Pass AWS credentials and region from the environment to boto3."""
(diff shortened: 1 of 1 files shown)


iris-clawd (commented):
Looks good overall — cleaner than the old fixture-based approach. Using `assert_called_once_with` and inlining the mock is a nice improvement.

One thing to consider: CodeRabbit's point about `AWS_SESSION_TOKEN` is valid. Since you're using `clear=False`, if a developer or CI runner happens to have `AWS_SESSION_TOKEN` set in their environment, the production code will pick it up (line 282-283 of completion.py falls back to `os.getenv("AWS_SESSION_TOKEN")`), and the `aws_session_token=None` assertion will fail. You could either add `"AWS_SESSION_TOKEN": ""` to the `patch.dict` env or just be aware that it's environment-dependent.

Other than that, LGTM. 💬 24

oalami (approved):

PR #7383, opened 2026-09-11
feat: expose CrewAI Platform application catalog

Vidit-Ostwal:
## Related issue

OSS-164: https://linear.app/crewai/issue/OSS-164/enable-amp-crewai-platform-tools-for-oss-users

## Summary

Expose the supported CrewAI Platform application catalog from core for OSS consumers.

## Verification

## Additional context

Base PR for the platform-tools stack.

### DIFF
--- a/lib/crewai-core/src/crewai_core/platform_apps.py
+++ b/lib/crewai-core/src/crewai_core/platform_apps.py
@@ -1,0 +1,23 @@
+"""CrewAI Platform application catalog."""
+
+from typing import Final
+
+
+PLATFORM_APPS: Final[tuple[str, ...]] = (
+    "asana",
+    "box",
+    "clickup",
+    "github",
+    "gmail",
+    "google_calendar",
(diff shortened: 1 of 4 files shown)


lib/crewai-core/src/crewai_core/platform_apps.py
```diff
@@ -0,0 +1,25 @@
+"""CrewAI Platform application catalog."""
+
+from typing import Final, Literal, get_args
+
+
+PlatformApp = Literal[
```
vinibrsl, line 6: Just double checking - we don't limit integrations to this list only, right?

vinibrsl (approved):

lib/crewai-core/src/crewai_core/platform_apps.py
Vidit-Ostwal: Yup, waiting for clipper integration to get in, will get them exhaustively soon

PR #7396, opened 2026-09-11
Add TrustedRouter as an OpenAI-compatible native provider

### LINKED ISSUE #7395: [FEATURE] Support TrustedRouter as an OpenAI-compatible provider
Feature Area: Integration with external tools

## Problem

NA — not related to an existing bug.

[TrustedRouter](https://trustedrouter.com) is an OpenAI-compatible LLM router at `https://api.trustedrouter.com/v1`. CrewAI already routes several OpenAI-compatible gateways natively (OpenRouter and friends), but reaching TrustedRouter today means hand-setting `base_url` on every `LLM`, and the routing aliases do not survive the prefix strip that native routing applies.

## Describe the solution you'd like

Register `trustedrouter` alongside the existing OpenAI-compatible entries:

- Runtime routing in `crewai.llm.LLM` so `LLM(model="trustedrouter/auto")` and `LLM(model="trustedrouter/moonshotai/kimi-k3")` both reach `OpenAICompatibleCompletion`, with `TRUSTEDROUTER_API_KEY` and an optional `TRUSTEDROUTER_BASE_URL` override.
- A `trustedrouter/` section in the LLM concept guide (`en`, plus the `ar`, `ko` and `pt-BR` translations that carry the same provider list).

One detail worth calling out, because it is the part that is not a copy of the OpenRouter handling: every TrustedRouter model id is namespaced, so the prefix strip that is correct for OpenRouter breaks the router's own aliases — `trustedrouter/auto` reaches the API as `auto`, which it rejects. `trustedrouter/moonshotai/kimi-k3` still has to resolve to the upstream id, so the split is on whether the remainder is left bare, not on the provider alone.

## Describe alternatives you've considered

Configuring `base_url` per-`LLM` works today but leaves the aliases unusable and gives no CLI/docs discoverability, which is the same argument that applies to the gateways CrewAI already routes natively.

## Additional context

Implementation is written and passing — `ruff check` clean, `tests/llms/openai_compatible` green (43) — in #6143, which was auto-closed for not referencing an issue. This is that issue.

## Willingness to Contribute

jperla:
Closes #7395.

Adds TrustedRouter (attested OpenAI-compatible router) as a native provider, following the same shape as the existing `openrouter` entry:

- `OPENAI_COMPATIBLE_PROVIDERS["trustedrouter"]` with `https://api.trustedrouter.com/v1`, `TRUSTEDROUTER_API_KEY`, and `TRUSTEDROUTER_BASE_URL` override
- provider constant, prefix mapping, model validation, and openai-compatible dispatch in `llm.py`
- registry test in `tests/llms/openai_compatible/test_openai_compatible.py`
- docs accordion in `docs/edge/en/concepts/llms.mdx` (native usage — no LiteLLM dependency needed)

Verified locally: 36/36 openai_compatible tests and 19/19 provider/prefix tests in `test_llm.py` pass.

This reworks the earlier docs-only version of this PR into a code integration.

Localized guides (`ar`, `ko`, `pt-BR`) carry the same accordion, and the optional `TRUSTEDROUTER_BASE_URL` entry is present in all four.

Reopened as a new PR because #6143 was auto-closed for not referencing an issue and the reopen was refused; the branch is unchanged apart from a rebase onto current `main`.

### DIFF
--- a/docs/edge/ar/concepts/llms.mdx
+++ b/docs/edge/ar/concepts/llms.mdx
@@ -959,6 +959,29 @@ mode: "wide"
     ```
   </Accordion>
 
+  <Accordion title="TrustedRouter">
+    TrustedRouter هو موجّه متوافق مع OpenAI ومُصدَّق (attested) مخصص لأعمال الوكلاء الحساسة للخصوصية، مع اختيار تلقائي للنماذج، ومسارات بدون احتفاظ بالبيانات، ومسارات مشفّرة من طرف إلى طرف.
+
+    عيّن متغيرات البيئة التالية في ملف `.env`:
+    ```toml Code
+    TRUSTEDROUTER_API_KEY=<your-api-key>
+    ```
+
+    مثال الاستخدام في مشروع CrewAI:
+    ```python Code
+    llm = LLM(
+        model="trustedrouter/auto",
+    )
+    ```
+
+    <Info>
+      أسماء التوجيه المستعارة:
+      - trustedrouter/auto (المسارات السليمة، يتم اختيارها تلقائيًا)
+      - trustedrouter/zdr (مسارات بدون احتفاظ بالبيانات)
+      - trustedrouter/e2e (مسارات مشفّرة من طرف إلى طرف)
+    </Info>
+  </Accordion>
+
   <Accordion title="Open Router">
     عيّن متغيرات البيئة التالية في ملف `.env`:
     ```toml Code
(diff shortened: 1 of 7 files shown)


jperla:
Rebased onto current `main`. The conflict was `__new__` moving its routing into `_resolve_route` (#7597 line of work), so both TrustedRouter hunks were ported there: the `provider_mapping` entry, and the alias guard that keeps `trustedrouter/auto` whole while still stripping the prefix from `trustedrouter/<vendor>/<model>`.

`lib/crewai/tests/llms/openai_compatible/` is green (51 passed).

PR #7416, opened 2026-09-12
fix(knowledge): raise when knowledge_sources fail to initialize

### LINKED ISSUE #7415: [BUG] Crew silently continues when knowledge_sources fail to initialize (missing OPENAI_API_KEY)
### Description

When a `Crew` is constructed with explicit `knowledge_sources`, knowledge initialization can fail (for example missing `OPENAI_API_KEY` for the default Chroma OpenAI embedder). The failure is logged, but `create_crew_knowledge` catches the exception and the crew continues as if knowledge were configured.

Result: kickoff “succeeds” with empty knowledge, and agents answer without retrieval. Easy to miss in non-verbose runs.

Also: `ChromaDBConfig._default_embedding_function` docstring says the default is `all-MiniLM-L6-v2` via ONNX, but the implementation uses `OpenAIEmbeddingFunction` / `text-embedding-3-small` (docs elsewhere correctly say OpenAI is the default).

### Steps to Reproduce

```python
from crewai import Agent, Crew, Task, Process
from crewai.knowledge.source.string_knowledge_source import StringKnowledgeSource

class StubLLM(BaseLLM):
    def call(self, messages, tools=None, callbacks=None, available_functions=None,
             from_task=None, from_agent=None, response_model=None):
        return "Thought: ok.\nFinal Answer: invented answer without retrieval."
    def supports_function_calling(self) -> bool:
        return False

# Ensure OPENAI_API_KEY is unset
knowledge = StringKnowledgeSource(content="Secret fact: the vault code is 42.")
agent = Agent(role="r", goal="g", backstory="b", llm=StubLLM(model="stub"), verbose=False)
task = Task(description="What is the vault code?", expected_output="the code", agent=agent)
crew = Crew(
    agents=[agent],
    tasks=[task],
    knowledge_sources=[knowledge],
    process=Process.sequential,
    verbose=False,
)
result = crew.kickoff()
print(result)  # completes; knowledge never loaded
```

### Expected behavior

If the user explicitly provided `knowledge_sources` and initialization/upsert fails, Crew construction or kickoff should **fail loudly** (raise) so the misconfiguration is obvious. Optionally fix the misleading ONNX docstring on `_default_embedding_function`.

### Actual behavior

Errors like `Failed to upsert documents: The OPENAI_API_KEY environment variable is not set.` are logged, then `create_crew_knowledge` swallows the exception and the crew runs without knowledge.

### Operating System

macOS

### Python Version

3.13

### crewAI Version

1.15.21 / current `main`

### Willingness to Contribute

Yes — I can open a PR that re-raises on explicit `knowledge_sources` init failure + corrects the Chroma default-embedder docstring, with a regression test.

kevincostner17:
## Summary
- Re-raise when explicit `knowledge_sources` fail during crew knowledge init instead of logging a warning and continuing with empty retrieval
- Correct `ChromaDBConfig._default_embedding_function` docstring (OpenAI default, not ONNX MiniLM)
- Add regression tests for missing `OPENAI_API_KEY` + explicit knowledge sources

Fixes #7415

## Test plan
- [x] `uv run pytest lib/crewai/tests/knowledge/test_crew_knowledge_init_failure.py -q`
- [x] Pre-commit (ruff, ruff-format, mypy) passed on commit
- [ ] CI green on this PR

### DIFF
--- a/lib/crewai/src/crewai/crew.py
+++ b/lib/crewai/src/crewai/crew.py
@@ -716,9 +716,13 @@ class Crew(FlowTrackable, BaseModel):
                     self.knowledge.add_sources()
 
             except Exception as e:
-                self._logger.log(
-                    "warning", f"Failed to init knowledge: {e}", color="yellow"
-                )
+                self._logger.log("error", f"Failed to init knowledge: {e}", color="red")
+                raise ValueError(
+                    "Failed to initialize crew knowledge_sources. "
+                    "Check your embedder configuration (CrewAI defaults to OpenAI "
+                    "embeddings and requires OPENAI_API_KEY unless you set a local "
+                    "embedder such as provider='ollama' or provider='onnx')."
+                ) from e
         return self
 
     @model_validator(mode="after")
--- a/lib/crewai/src/crewai/rag/chromadb/config.py
+++ b/lib/crewai/src/crewai/rag/chromadb/config.py
@@ -49,7 +49,9 @@ def _default_embedding_function() -> ChromaEmbeddingFunctionWrapper:
     """Create default ChromaDB embedding function.
 
     Returns:
-        Default embedding function using all-MiniLM-L6-v2 via ONNX.
+        Default embedding function using OpenAI ``text-embedding-3-small``.
+        Requires ``OPENAI_API_KEY``. For local setups, pass an explicit
+        embedder (for example ``provider="onnx"`` or ``provider="ollama"``).
     """
     from chromadb.utils.embedding_functions.openai_embedding_function import (
         OpenAIEmbeddingFunction,
--- a/lib/crewai/tests/knowledge/test_crew_knowledge_init_failure.py
+++ b/lib/crewai/tests/knowledge/test_crew_knowledge_init_failure.py
@@ -1,0 +1,71 @@
+"""Crew knowledge_sources must fail loudly when initialization fails."""
+
+from __future__ import annotations
+
+import os
+from unittest.mock import patch
+
+import pytest
+from crewai import Agent, Crew, Process, Task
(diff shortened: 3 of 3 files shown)


kevincostner17:
Note for triage: this PR was AI-assisted. I do not have permission to apply the `llm-generated` label myself — please add it per CONTRIBUTING.md. Thanks!

Vidit-Ostwal (commented):
The new missing-API-key regression test is not isolated from the process-wide RAG context. `monkeypatch.delenv("OPENAI_API_KEY")` is ineffective when an earlier test in the same worker has already initialized the default RAG client and its embedding function, which is why the Python 3.10 test job reports `DID NOT RAISE`. Please make this deterministic by resetting the RAG context before and after the test (for example, `clear_rag_config()`), or mock the storage/client boundary to raise. The production behavior looks right; this is a test-isolation issue.

gaoanze888 (approved):
Reviewed at exact head.

The fail-loud change is correct. The `except` sits inside the `if self.knowledge_sources:` gate, so only explicitly configured sources that fail to initialize now raise — previously they degraded silently to empty knowledge and the agent answered without the material it was explicitly given (a real correctness/safety hazard). The error message usefully names the common root cause (missing OPENAI_API_KEY with the default OpenAI embedder) and the local alternatives.

The docstring correction (default is OpenAI `text-embedding-3-small`, not ONNX all-MiniLM-L6-v2) is accurate, and the test pins both the raise and that `_default_embedding_function` still demands the API key.

Minor non-blocking: this is a behavior change for anyone whose knowledge sources intermittently failed and relied on silent fallback; for the intended semantics it is the right call. Approving.

PR #7418, opened 2026-09-12
fix(memory): return newest records first in LanceDBStorage list_records

### LINKED ISSUE #7394: [BUG] LanceDBStorage.list_records() returns oldest records instead of newest first due to premature query truncation
### Description

In `LanceDBStorage` (`lib/crewai/src/crewai/memory/storage/lancedb_storage.py`), `list_records()` is documented to return records ordered by `created_at` descending:

> "List records in a scope, newest first. Returns: List of MemoryRecord, ordered by created_at descending."

However, `list_records()` currently returns the oldest records in the table rather than the newest.

Root Cause:
In `lancedb_storage.py` lines 507–510:
```python
rows = self._scan_rows(scope_prefix, limit=limit + offset)
records = [self._row_to_record(r) for r in rows]
records.sort(key=lambda r: r.created_at, reverse=True)
return records[offset : offset + limit]

### Steps to Reproduce

```text
1. Initialize LanceDBStorage in a temporary directory.
2. Insert 10 records with sequential created_at timestamps (rec_0 oldest, rec_9 newest).
3. Call storage.list_records(limit=3).
4. Inspect the returned IDs.

### Expected behavior

list_records(limit=3) should return the 3 most recently created records:
['rec_9', 'rec_8', 'rec_7']

### Screenshots/Code snippets

import tempfile
from datetime import datetime, timezone, timedelta

with tempfile.TemporaryDirectory() as tmpdir:
    storage = LanceDBStorage(path=tmpdir)
    base_time = datetime.now(timezone.utc)
    
    # Insert 10 records: rec_0 is oldest, rec_9 is newest
    records = [
        MemoryRecord(
            id=f"rec_{i}",
            content=f"content {i}",
            created_at=base_time + timedelta(minutes=i),
            embedding=[0.1] * 1536
        )
        for i in range(10)
    ]
    storage.save(records)

    # Request the 3 newest records
    results = storage.list_records(limit=3)
    print("Returned IDs:", [r.id for r in results])
    # Expected: ['rec_9', 'rec_8', 'rec_7']
    # Actual:   ['rec_2', 'rec_1', 'rec_0']

### Operating System

macOS Sonoma

### Python Version

3.12

### crewAI Version

1.15.21

### crewAI Tools Version

1.15.21

### Virtual Environment

Venv

### Evidence

Running the reproduction script outputs:

Returned IDs: ['rec_2', 'rec_1', 'rec_0']

The 3 oldest records (rec_0, rec_1, rec_2) were fetched and sorted descending, while the actual newest records (rec_9, rec_8, rec_7) were truncated at the storage layer and never returned.

### Possible Solution

In `LanceDBStorage.list_records`:
Do not truncate `_scan_rows` before sorting. Scan the candidate rows in the scope (using the default `_SCAN_ROWS_LIMIT` of 50,000, identical to how `get_scope_info` works and how `QdrantEdgeStorage.list_records` works), sort by `created_at` descending, and then apply the pagination slice:

```python
def list_records(
    self, scope_prefix: str | None = None, limit: int = 200, offset: int = 0
) -> list[MemoryRecord]:
    rows = self._scan_rows(scope_prefix)
    records = [self._row_to_record(r) for r in rows]
    records.sort(key=lambda r: r.created_at, reverse=True)
    return records[offset : offset + limit]

### Additional context

```text
I have verified this reproduction locally and have the fix and regression unit test ready to submit as a PR if approved.

Rohitkanithi:
## Related issue

Fixes #7394

## Summary

`LanceDBStorage.list_records(scope_prefix, limit, offset)` is documented to return memory records ordered by `created_at` descending ("newest first").

### Root Cause
Previously, `list_records()` called `self._scan_rows(scope_prefix, limit=limit + offset)` before sorting in Python. Because LanceDB table scans return records in storage append order (oldest first) and LanceDB does not support descending ordering on search queries, passing `limit=limit + offset` to the table scan truncated the result set to the oldest records in storage. Any records newer than `limit + offset` were omitted from the query results.

### Solution
1. In `list_records()`, call `self._scan_rows(scope_prefix)` without premature `limit` truncation, relying on `_SCAN_ROWS_LIMIT = 50_000` default.
2. Sort the raw row dictionaries directly descending by `created_at` using timezone-safe datetime parsing.
3. Lazily deserialize only the sliced records `[offset : offset + limit]` into `MemoryRecord` models, avoiding expensive JSON decoding and Pydantic model construction for unreturned records.

## Verification

- Added `test_lancedb_list_records_order_and_pagination` in `lib/crewai/tests/memory/test_unified_memory.py` testing newest-first ordering across pagination pages (`limit`, `offset`), out-of-bounds offsets, and scope prefix filtering.
- Ran all memory tests: `132 passed, 19 skipped`.
- Code quality checks passed cleanly (`ruff check`, `ruff format --check`, `mypy`).

## Additional context

None

### DIFF
--- a/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
+++ b/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import contextvars
-from datetime import datetime
+from datetime import datetime, timezone
 import json
 import logging
 import os
@@ -504,10 +504,25 @@ class LanceDBStorage:
         Returns:
             List of MemoryRecord, ordered by created_at descending.
         """
-        rows = self._scan_rows(scope_prefix, limit=limit + offset)
-        records = [self._row_to_record(r) for r in rows]
-        records.sort(key=lambda r: r.created_at, reverse=True)
-        return records[offset : offset + limit]
+        rows = self._scan_rows(scope_prefix)
+
+        def _get_created_at(row: dict[str, Any]) -> datetime:
+            val = row.get("created_at")
+            if val is None:
+                return datetime.min.replace(tzinfo=timezone.utc)
+            if isinstance(val, datetime):
+                dt = val
+            else:
+                try:
+                    dt = datetime.fromisoformat(str(val).replace("Z", "+00:00"))
+                except Exception:
+                    return datetime.min.replace(tzinfo=timezone.utc)
+            if dt.tzinfo is None:
+                return dt.replace(tzinfo=timezone.utc)
+            return dt.astimezone(timezone.utc)
+
+        rows.sort(key=_get_created_at, reverse=True)
+        return [self._row_to_record(r) for r in rows[offset : offset + limit]]
 
     def get_scope_info(self, scope: str) -> ScopeInfo:
         scope = scope.rstrip("/") or "/"
--- a/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
+++ b/lib/crewai/src/crewai/memory/storage/lancedb_storage.py
@@ -507,6 +507,7 @@ class LanceDBStorage:
         rows = self._scan_rows(scope_prefix)
 
         def _get_created_at(row: dict[str, Any]) -> datetime:
+            """Extract and parse created_at into a timezone-aware UTC datetime."""
             val = row.get("created_at")
             if val is None:
                 return datetime.min.replace(tzinfo=timezone.utc)
--- a/lib/crewai/tests/memory/test_unified_memory.py
+++ b/lib/crewai/tests/memory/test_unified_memory.py
@@ -180,6 +180,7 @@ def test_lancedb_list_scopes_get_scope_info(lancedb_path: Path) -> None:
 
 
 def test_lancedb_list_records_order_and_pagination(lancedb_path: Path) -> None:
+    """Test that LanceDBStorage.list_records returns newest records first with correct pagination."""
     from crewai.memory.storage.lancedb_storage import LanceDBStorage
 
     storage = LanceDBStorage(path=str(lancedb_path), vector_dim=4)
--- a/lib/crewai/tests/memory/test_unified_memory.py
+++ b/lib/crewai/tests/memory/test_unified_memory.py
@@ -179,6 +179,65 @@ def test_lancedb_list_scopes_get_scope_info(lancedb_path: Path) -> None:
     assert info.path == "/"
 
 
+def test_lancedb_list_records_order_and_pagination(lancedb_path: Path) -> None:
+    from crewai.memory.storage.lancedb_storage import LanceDBStorage
+
+    storage = LanceDBStorage(path=str(lancedb_path), vector_dim=4)
+    base_time = datetime(2025, 1, 1, 12, 0, 0)
+    records = [
+        MemoryRecord(
+            id=f"rec_{i}",
+            content=f"content {i}",
+            scope="/test",
+            created_at=base_time + timedelta(minutes=i),
+            embedding=[0.0] * 4,
+        )
+        for i in range(10)
+    ]
+    other_records = [
+        MemoryRecord(
+            id=f"other_{i}",
+            content=f"other content {i}",
+            scope="/other",
+            created_at=base_time + timedelta(minutes=i),
+            embedding=[0.0] * 4,
+        )
+        for i in range(5)
+    ]
+    storage.save(records + other_records)
+
+    # 1. Page 1: limit=3, offset=0 returns 3 newest records (rec_9, rec_8, rec_7)
+    page1 = storage.list_records(scope_prefix="/test", limit=3, offset=0)
+    assert [r.id for r in page1] == ["rec_9", "rec_8", "rec_7"]
+    assert page1[0].created_at > page1[1].created_at > page1[2].created_at
+
+    # 2. Page 2: limit=3, offset=3 returns next 3 newest (rec_6, rec_5, rec_4)
+    page2 = storage.list_records(scope_prefix="/test", limit=3, offset=3)
+    assert [r.id for r in page2] == ["rec_6", "rec_5", "rec_4"]
+
+    # 3. Page 3: limit=3, offset=6 returns next 3 newest (rec_3, rec_2, rec_1)
+    page3 = storage.list_records(scope_prefix="/test", limit=3, offset=6)
+    assert [r.id for r in page3] == ["rec_3", "rec_2", "rec_1"]
+
+    # 4. Page 4: limit=3, offset=9 returns last 1 record (rec_0)
+    page4 = storage.list_records(scope_prefix="/test", limit=3, offset=9)
+    assert [r.id for r in page4] == ["rec_0"]
+
+    # 5. Page 5: offset beyond total count returns empty list
+    page5 = storage.list_records(scope_prefix="/test", limit=3, offset=10)
+    assert page5 == []
+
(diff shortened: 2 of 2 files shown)


Rohitkanithi:
Thanks @Vidit-Ostwal for assigning, 

I've raised PR #7418 with the fix and unit tests covering ordering and pagination. Looking forward to your review

Vidit-Ostwal (commented):
Thanks for fixing the premature  truncation — that resolves the reported small-scope behavior. I’m concerned that this replacement still calls  before ordering and pagination. That method has a 50,000-row default cap and materializes complete rows (including vectors), so large scopes can still omit the true newest records and a small page can require loading/sorting a very large result set.\n\nCould we move this to the storage query instead? The intended flow should be:  →  →  → . This lets LanceDB select exactly the requested newest-first page, rather than fetching candidates and sorting them in Python. If LanceDB cannot express that order/pagination for this query path, please call that out and propose a bounded, correctness-preserving alternative; the current 50k cap needs explicit handling.

Vidit-Ostwal (commented):
Correction to my prior review: shell formatting stripped the code identifiers. The concern is specifically that list_records now calls self._scan_rows(scope_prefix) before ordering and pagination. _scan_rows has a 50,000-row default cap and materializes complete rows, including vectors. Large scopes can therefore still omit the true newest records, and a small page can require loading and sorting a very large result set.

Recommended flow: WHERE scope matches, then ORDER BY created_at DESC, then OFFSET offset, then LIMIT limit — all in the LanceDB query. That lets LanceDB select exactly the requested newest-first page instead of fetching candidates and sorting them in Python. If LanceDB cannot express that order and pagination for this query path, please document that constraint and propose a bounded, correctness-preserving alternative; the current 50k cap needs explicit handling.

Rohitkanithi:
@Vidit-Ostwal,

### 1. LanceDB Query Constraint
LanceDB's Python SDK query builder (`table.search()`) does not support `ORDER BY` or descending ordering (LanceDB only supports SQL `ORDER BY` via DataFusion on its enterprise SQL endpoint or FTS on `_score`). 

### 2. Implemented Bounded Streaming Min-Heap Alternative
I have addressed both concerns (vector materialization and the 50k cap):

1. **Excluding Heavy Vectors**: We project only metadata columns (`_RECORD_METADATA_COLUMNS`), skipping the heavy 3,072-dimensional `vector` embeddings entirely to eliminate I/O and RAM waste.
2. **Removing the 50k Cap**: We removed the `_scan_rows` 50k cap so that all candidate records in the scope are evaluated, preventing any omission of newer records.
3. **Strictly Bounded O(limit + offset) Memory**: We stream Arrow RecordBatches in chunks (`q.to_batches(batch_size=8192)`) through a min-heap of size `offset + limit`. This keeps memory usage strictly bounded to `O(offset + limit)` rows regardless of how large the table or scope is.
4. **Tests Added**:
   - Added `test_lancedb_list_records_exceeding_scan_cap` (testing a table with 52,000 records) to verify that records beyond the old 50k threshold are correctly returned.
   - Updated `test_lancedb_list_records_order_and_pagination` with discriminating timestamps between scopes to strictly verify scope filtering and global ordering.

Ready for your review

Vidit-Ostwal (commented):
Follow-up after checking the LanceDB version pinned here (0.30): the plain table.search() query builder exposes projection, filtering, limit, and offset, but not an ORDER BY operation. So the database-side flow I suggested is the ideal semantics, but is not directly expressible through this API today.

Given that constraint, a correct fallback is to push the scope filter and a metadata-only projection into LanceDB, stream the matching rows in batches, retain only the newest offset + limit candidates in a bounded heap keyed by created_at, then deserialize the requested page. This remains O(number of scoped records) read work, but avoids materializing every vector or every row in Python and removes the 50k correctness cap. A true database-side ORDER BY plus pagination would require LanceDB support for it or a different storage/index design.

Vidit-Ostwal (commented):
Question on the heap-based scan: upstream LanceDB now exposes query ordering, so why do we need to stream every scoped row and maintain a client-side top-k heap here? Could we instead upgrade the LanceDB dependency and use the native query flow: WHERE scope, ORDER BY created_at descending, OFFSET, LIMIT?

That would let the database return only the requested page, remove the custom timestamp parsing and heap logic, and keep ordering/pagination in the storage layer. I see CrewAI is currently pinned to LanceDB 0.30.x, whose Python query builder does not expose order_by; if a dependency upgrade is not viable, please document the compatibility constraint and why this fallback is preferred.

Rohitkanithi:
@Vidit-Ostwal,

Thanks for the thoughtful review and architectural feedback

### 1. Upstream LanceDB Upgrade
Regarding upgrading LanceDB:
- Technically, native `order_by` on `table.search()` was introduced in **LanceDB 0.33.0** (via `lancedb.query.ColumnOrdering`).
- However, upgrading LanceDB from `0.30.x` to `0.33+` in this bugfix PR carries dependency risk: CrewAI's `pyproject.toml` strictly pins `"lancedb>=0.29.2,<0.30.1"`, and jumping versions requires regenerating lockfiles, updating core dependencies, and validating native binary wheels across all 4 supported Python versions (3.10–3.13) and OS platforms.

To keep this bugfix focused, zero-risk, and backward-compatible, I researched **three possible approaches** under the current LanceDB version:

---

### 2. Three Approaches for Current LanceDB (0.30.x)

#### Approach 1: Bounded Streaming Min-Heap (Currently in PR)
- **Idea**: Stream Arrow record batches in chunks and maintain a top-k min-heap of size `limit + offset`.
- **Pros**: Strictly bounds memory consumption to $O(\text{limit} + \text{offset})$ rows, regardless of table size.
- **Trade-off**: Requires custom heap maintenance and datetime parsing logic in Python.
- **Documentation**: If we keep this approach, I have documented the `lancedb<0.30.1` compatibility constraint in the `list_records` docstring so future maintainers know why the heap fallback exists and how to swap it for native `order_by` once LanceDB is bumped.

#### Approach 2: PyArrow C++ Vectorized Sort
- **Idea**: Project only metadata columns (excluding heavy vectors) into an Arrow table, sort `created_at` descending directly using PyArrow’s native C++ compute engine (`pyarrow.compute.sort_indices`), and slice the requested page.
- **Pros**: Completely removes the custom heap and datetime parsing logic; sorts 50k+ rows in ~3.8 ms in C++; `pyarrow` is already a core requirement of LanceDB.
- **Trade-off**: Holds the scoped metadata columns in Arrow columnar memory before slicing (~5 MB for 50k rows).

#### Approach 3: Standard In-Memory Python Sort
- **Idea**: Fetch all scoped metadata rows as Python dictionaries and sort them in-place using Python’s built-in `list.sort()`.
- **Pros**: The simplest implementation
- **Trade-off**: Deserializes every row in the scope into full Python dictionary objects in memory before slicing.

---

### 3. Future Path
Whichever approach we proceed with in this PR, when CrewAI decides to do a dedicated LanceDB dependency upgrade to 0.33+ in the future, we can seamlessly swap the sorting logic for native `q.order_by(...)`.

Which of these approaches do you prefer we move forward with for this PR

Rohitkanithi:
Hey @Vidit-Ostwal,

Just following up on this PR. Addressed the review feedback (added negative offset validation, refined scope prefix filtering, and documented the `lancedb<0.30.1` compatibility constraint with the future `ColumnOrdering` path in the docstring).

The current bounded streaming min-heap approach keeps memory strictly bounded to $O(\text{limit} + \text{offset})$ without vector materialization or dependency bump risks.

Could you please take another look when you have a moment.

Rohitkanithi:
@Vidit-Ostwal,
Following up on this PR. Could you please review

PR #7420, opened 2026-09-13
fix(tools): restore YouTube channel search for handles

### LINKED ISSUE #7404: [BUG] YoutubeChannelSearchTool rejects documented YouTube channel handle in v1.15.21
### Description

The YoutubeChannelSearchTool fails when using a valid YouTube channel handle as documented in the CrewAI v1.15.21 documentation.

The documentation shows that the tool can be initialized using a channel handle such as:

youtube_channel_handle="@exampleChannel"

However, with crewai==1.15.21 and crewai-tools==1.15.21, using a valid channel handle results in:

ValueError: Invalid YouTube channel URL: @exampleChannel

The error occurs inside YoutubeChannelLoader because the loader expects a URL containing "youtube.com/...", while YoutubeChannelSearchTool passes the handle to the RAG adapter.

I also tested passing a full YouTube channel URL, but this is not a workaround because YoutubeChannelSearchTool.add() prepends "@" to values that do not start with "@", resulting in a malformed value.

This eventually causes the YouTube loader/pytube integration to fail.

### Steps to Reproduce

1. Run Python 3.13

2. Install CrewAI and CrewAI Tools:

pip install crewai==1.15.21 crewai-tools==1.15.21

3. Import the YouTube channel search tool:

from crewai_tools import YoutubeChannelSearchTool

4. Initialize it with a documented channel handle:

youtube_channel_tool = YoutubeChannelSearchTool(
    youtube_channel_handle="@handle"
)

5. The tool initialization fails with:

ValueError: Invalid YouTube channel URL: @handle

6. Passing a full YouTube channel URL also fails because the `add()` method prepends `@` to the URL.

### Expected behavior

A valid YouTube channel handle provided in the documented `@handle` format should be accepted by YoutubeChannelSearchTool.

The tool should resolve the handle to the corresponding YouTube channel URL and successfully load the channel content.

### Screenshots/Code snippets

from crewai_tools import YoutubeChannelSearchTool

youtube_channel_tool = YoutubeChannelSearchTool(
    youtube_channel_handle="@handle"
)

The relevant YoutubeChannelSearchTool.add() implementation is:

def add(
    self,
    youtube_channel_handle: str,
) -> None:
    if not youtube_channel_handle.startswith("@"):
        youtube_channel_handle = f"@{youtube_channel_handle}"
    super().add(
        youtube_channel_handle,
        data_type=DataType.YOUTUBE_CHANNEL
    )

The loader validates the source against URL patterns such as:

[
    "youtube.com/channel/",
    "youtube.com/c/",
    "youtube.com/@",
    "youtube.com/user/",
]

Therefore, @handle does not satisfy the loader's URL validation.

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

1.15.21

### crewAI Tools Version

1.15.21

### Virtual Environment

Conda

### Evidence

The following error is produced when initializing YoutubeChannelSearchTool with a documented @handle:

ValueError: Invalid YouTube channel URL: @handle

The error originates from:

crewai_tools/rag/loaders/youtube_channel_loader.py

The loader checks for URL patterns including:

"youtube.com/channel/"
"youtube.com/c/"
"youtube.com/@"
"youtube.com/user/"

However, YoutubeChannelSearchTool passes the @handle directly to the RAG adapter instead of converting it into a complete YouTube URL.

Passing a complete URL is also problematic because YoutubeChannelSearchTool.add() prepends "@" to any input that does not already start with "@".

This behavior appears related to issue #5429:
https://github.com/crewAIInc/crewAI/issues/5429

### Possible Solution

The tool could resolve a documented @handle into a complete YouTube channel URL before passing it to YoutubeChannelLoader.

For example:

@handle
    ↓
https://www.youtube.com/@handle
    ↓
YoutubeChannelLoader

Alternatively, YoutubeChannelLoader could support the @handle format directly.

The important point is that the input format accepted by YoutubeChannelSearchTool should be compatible with the format expected by YoutubeChannelLoader.

### Additional context

This issue appears to be related to the previously reported issue #5429:

"YoutubeChannelSearchTool Fails for Valid Channel Inputs"

The behavior is still reproducible with crewai==1.15.21 and crewai-tools==1.15.21.

The current CrewAI documentation indicates that YoutubeChannelSearchTool accepts a YouTube channel handle, so this appears to be a mismatch between the documented API and the current implementation.

No OpenAI API key or LLM configuration is involved in this particular error. The failure occurs during YoutubeChannelSearchTool initialization while loading the YouTube channel.

Grove-ovo:
## Related issue

Fixes #7404.

## Summary

Fix `YoutubeChannelSearchTool` rejecting documented `@handle` inputs and incorrectly prefixing full channel URLs.

Normalize channel handles and URLs, including non-ASCII handles. Support modern YouTube video listings and continuation pagination so accepted inputs load channel content. Validate continuation responses and reject circular tokens to avoid reporting incomplete results as complete.

Preserve transcript loading when video detail retrieval fails.

## Verification

Checks performed:
- 41 tests passed across `test_youtube_channel_loader.py` and `test_search_tools.py`, using temporary database storage.
- mypy, Ruff, and formatting checks passed for the two changed source files.
- Cached real channel and continuation responses successfully loaded 1, 10, and 40 videos. Video detail and transcript requests were mocked during replay.

## Additional context

- Adds a runtime compatibility patch for pytube's channel-handle extraction.
- Channel source URLs are normalized.
- `total_videos` may be `None` when the total cannot be determined.


Grove-ovo:
@renatonitta Just adding some context here: I noticed that #5851 previously addressed part of the same issue around @handle and full-URL normalization, but it has remained open since May.

This PR includes that normalization fix, but also covers additional failures I found in the current YouTube channel loading path, including modern richGridRenderer extraction, continuation pagination, stricter channel URL validation, and regression coverage. The linked #7404 confirms that the documented @handle behavior is still broken in v1.15.21.

If you'd prefer to keep the input-normalization fix minimal and split the loader/pagination changes into a separate PR, I'm happy to restructure it that way.

PR #7424, opened 2026-09-13
fix: InvokeCrewAIAutomationTool reads env vars and is instantiable without args

### LINKED ISSUE #7389: [BUG] InvokeCrewAIAutomationTool ignores CREWAI_API_URL/CREWAI_BEARER_TOKEN and cannot be instantiated without arguments
### Description

`InvokeCrewAIAutomationTool` cannot be configured through the environment variables its own documentation advertises, and it cannot be instantiated without arguments at all.

The docs page for the tool (`docs/edge/en/tools/integration/crewaiautomationtool.mdx`, published at https://docs.crewai.com/en/tools/integration/crewaiautomationtool) has an "Environment Variables" section presenting `CREWAI_API_URL` and `CREWAI_BEARER_TOKEN` as alternatives to the `crew_api_url` and `crew_bearer_token` constructor arguments. The implementation never reads them: `__init__` takes `crew_api_url`, `crew_bearer_token`, `crew_name` and `crew_description` as required positional parameters and the module contains no `os.getenv` call.

This also makes the tool unusable from CrewAI AMP Studio, where it is offered to no-code users as "Invoke Amp Automation" under Add App → Internal Tools. The declarative Flow runtime resolves tools by class reference and instantiates them with no arguments (`crewai/project/json_loader.py::_instantiate_tool_import_ref` calls `tool_cls()`; `crewai/flow/runtime/_actions.py::ToolAction._build_tool` does the same), so every run fails immediately. The Studio UI offers no field for the four constructor parameters, and the documented environment variables would be the natural escape hatch, except that they do nothing.

### Steps to Reproduce

1. `export CREWAI_API_URL="https://some-automation.crewai.com"`
2. `export CREWAI_BEARER_TOKEN="some-token"`
3. Run `python -c "from crewai_tools import InvokeCrewAIAutomationTool; InvokeCrewAIAutomationTool()"`
4. See the `TypeError` below.

For the Studio path: add "Invoke Amp Automation" to an agent in a Flows Beta project and run it; the run fails in about 18 seconds with the same `TypeError`, surfaced as `cannot instantiate tool ref 'crewai_tools:InvokeCrewAIAutomationTool' without arguments`.

### Expected behavior

With `CREWAI_API_URL` and `CREWAI_BEARER_TOKEN` set, the tool should be usable without passing the URL and the token as arguments, as documented. Constructing the tool should not fail with a `TypeError` about positional arguments; a genuinely missing configuration should produce a clear, actionable error naming what is absent.

### Screenshots/Code snippets

```python
import os
from crewai_tools import InvokeCrewAIAutomationTool

os.environ["CREWAI_API_URL"] = "https://some-automation.crewai.com"
os.environ["CREWAI_BEARER_TOKEN"] = "some-token"

InvokeCrewAIAutomationTool()
```

```
TypeError: InvokeCrewAIAutomationTool.__init__() missing 4 required positional arguments:
'crew_api_url', 'crew_bearer_token', 'crew_name', and 'crew_description'
```

Studio run output:

```
cannot instantiate tool ref 'crewai_tools:InvokeCrewAIAutomationTool' without arguments:
InvokeCrewAIAutomationTool.__init__() missing 4 required positional arguments:
'crew_api_url', 'crew_bearer_token', 'crew_name', and 'crew_description'
```

### Operating System

Ubuntu 24.04

### Python Version

3.12

### crewAI Version

1.15.21

### crewAI Tools Version

1.15.21 (also reproduced on the released 1.15.20)

### Virtual Environment

Venv

### Evidence

`lib/crewai-tools/src/crewai_tools/tools/invoke_crewai_automation_tool/invoke_crewai_automation_tool.py` on `main` declares:

```python
crew_api_url: str
crew_bearer_token: str

def __init__(
    self,
    crew_api_url: str,
    crew_bearer_token: str,
    crew_name: str,
    crew_description: str,
    ...
```

`grep -c "getenv\|environ"` over that file returns 0, so neither documented environment variable is ever read. The sibling tool in the same package, `GenerateCrewaiAutomationTool`, already implements the pattern for its own credentials, using `Field(default_factory=lambda: os.getenv(...))` plus an `env_vars` declaration.

### Possible Solution

Make `crew_api_url` and `crew_bearer_token` optional, falling back to `CREWAI_API_URL` and `CREWAI_BEARER_TOKEN`, with explicit arguments taking precedence, following the `GenerateCrewaiAutomationTool` pattern. Give `crew_name` and `crew_description` defaults as well, reusing the generic `name` and `description` class attributes that already exist on the class but are currently unreachable because `__init__` always overwrites them; that is what allows the no-argument instantiation the Studio runtime performs. Validate the configuration when the tool actually runs and raise a clear `ValueError` naming the missing variables. Declare `env_vars` so the tool catalog surfaces both variables.

I have this implemented, with tests and the docs correction, and will open a draft PR referencing this issue.

### Additional context

Reproduced on 2026-09-10 and 2026-09-11, both from a local virtual environment and from a Flows Beta project in CrewAI AMP Studio.

KooshaPari:
Fixes #7389

## Problem
InvokeCrewAIAutomationTool required 4 positional arguments in `__init__` and never read the documented `CREWAI_API_URL`/`CREWAI_BEARER_TOKEN` environment variables. This made the tool unusable from CrewAI AMP Studio (which instantiates tools with no arguments) and inconsistent with its own documentation.

## Solution
- `crew_api_url` and `crew_bearer_token` now fall back to env vars via `Field(default_factory=...)`
- `crew_name` and `crew_description` have sensible defaults, enabling zero-argument instantiation
- Added `env_vars` declaration for tool catalog discovery
- Explicit constructor arguments take precedence over env vars
- Raises clear `ValueError` at execution time if configuration is missing

## Testing
- Added tests for no-argument instantiation with env vars
- Added tests for explicit args overriding env vars
- Added tests for empty env fallback


KooshaPari:
Closing per author request. Will re-evaluate repo selection.

Issue #6843, opened 2026-08-06
[BUG] MCP HTTP client crashes with "asyncio.run() cannot be called from a running event loop" when flow uses streamable HTTP transport


humatic-ai:
### Summary
When a CrewAI Flow runs an `agent` action with an MCP server configured via `streamable: true` (Streamable HTTP / SSE transport), the MCP tool resolver crashes immediately with:
RuntimeError: Failed to get native MCP tools: asyncio.run() cannot be called from a running event loop

The agent never receives any tools and fails before executing. ### Environment - CrewAI Studio (hosted platform) - Python 3.13.14 - crewai (latest, as installed in platform venv) - MCP server: Streamable HTTP (public HTTPS endpoint) - Flow transport: `streamable: true` ### Full traceback
McpError: Session terminated ↓ ConnectionError: Failed to connect to MCP server: Session terminated ↓ RuntimeError: Error during setup client and list tools: Failed to connect to MCP server: Session terminated ↓ RuntimeError: asyncio.run() cannot be called from a running event loop

File "crewai/mcp/tool_resolver.py", line 368, in _resolve_native tools_list = asyncio.run(_setup_client_and_list_tools()) RuntimeError: asyncio.run() cannot be called from a running event loop

### Root cause `crewai/mcp/tool_resolver.py` line ~368 calls `asyncio.run()` to fetch the MCP tool list. However, CrewAI Flows already run inside an active `asyncio` event loop (via `asyncio.run(_run_flow())` in `flow/runtime/__init__.py`). Python forbids calling `asyncio.run()` from within a running event loop — it raises `RuntimeError` immediately. ### Steps to reproduce 1. Create a CrewAI Flow with a start method using `call: agent` 2. Configure the agent with an MCP server: ```yaml mcps: - url: https://your-mcp-server.example.com/mcp streamable: true cache_tools_list: true
Kick off the flow
Observe the asyncio.run() cannot be called from a running event loop crash
Expected behavior
The MCP client connects to the server, discovers tools, and the agent runs normally.

Actual behavior
tool_resolver.py calls asyncio.run() inside the already-running flow event loop → immediate RuntimeError → agent has zero tools → flow fails.

Proposed fix
Replace the asyncio.run() call in tool_resolver.py with an async-safe alternative:

Option A — nest_asyncio (minimal change):

import nest_asyncio
nest_asyncio.apply()
loop = asyncio.get_event_loop()
[...]

PR #7448, opened 2026-09-14
Don't call asyncio.run() from MCPToolWrapper when a loop is running

gyanu2507:
Related to #6843 and #6978, which cover `tool_resolver` and `BaseTool.run`. This is the same crash on a different path: `MCPToolWrapper._run` always called `asyncio.run()`, so an MCP tool used from a Flow or any running loop died with `RuntimeError: asyncio.run() cannot be called from a running event loop`.

The wrapper now matches `mcp_native_tool.py`: if a loop is running, hop to a thread and `asyncio.run` there, copying contextvars.

Test: `lib/crewai/tests/tools/test_mcp_tool_wrapper_event_loop.py`.


gaoanze888 (approved):
Reviewed at exact head.

The approach is correct: probe `asyncio.get_running_loop()`, take the direct `asyncio.run` path when no loop is active, and offload to a worker thread when one is — which is the right fix for calling `asyncio.run` from an already-running event loop (related to #6843/#6978). CI is green across 3.10-3.13, and I reproduced the running-loop branch directly (returns the coroutine result, no `RuntimeError`).

One note for the reader: the `contextvars.copy_context()` + `ctx.run(...)` here is the inverse of the removal in #7335 (tool_resolver.py) — both are correct for their paths; here the copy is intentional to propagate caller context into the MCP coroutine. The regression test pins both the no-loop and running-loop paths. Approving.

PR #7463, opened 2026-09-14
fix(cli): overwrite stale poetry.lock backup on windows

### LINKED ISSUE #7462: [BUG] crewai update fails on Windows when poetry-old.lock already exists
### Description

`crewai update` backs up the Poetry lock file with `os.rename("poetry.lock", "poetry-old.lock")` in `lib/cli/src/crewai_cli/update_crew.py`. On Windows, `os.rename` raises `FileExistsError` when the destination already exists, while POSIX silently replaces it. So the first run works, but re-running `crewai update` in a project that still has a `poetry.lock` (or restoring a lock file and updating again) crashes on Windows only. The new `pyproject.toml` is never written because the exception is raised before that step. The `pyproject-old.toml` backup two lines above uses `shutil.copy2`, which does overwrite, so the two backups behave inconsistently on Windows.

### Steps to Reproduce

1. On Windows, in a Poetry-style crew project with a `pyproject.toml` and `poetry.lock`.
2. Run `crewai update` once (works; creates `poetry-old.lock`).
3. Restore or re-create `poetry.lock`, run `crewai update` again.

Or directly:

```python
import os, tempfile
from crewai_cli.update_crew import migrate_pyproject

os.chdir(tempfile.mkdtemp())
open("pyproject.toml", "w").write(
    '[tool.poetry]\nname="demo"\nversion="0.1.0"\ndescription="x"\n'
    'authors=["a <a@b.c>"]\n[tool.poetry.dependencies]\npython=">=3.10"\n'
)
open("poetry.lock", "w").write("lock\n")
open("poetry-old.lock", "w").write("previous backup\n")
migrate_pyproject("pyproject.toml", "pyproject.toml")
```

### Expected behavior

The stale backup is replaced and the migration completes, as it does on Linux and macOS.

### Screenshots/Code snippets

```
  File "...\lib\cli\src\crewai_cli\update_crew.py", line 92, in migrate_pyproject
    os.rename(lock_file, lock_backup)
FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'poetry.lock' -> 'poetry-old.lock'
```

### Operating System

Windows 11

### Python Version

3.13

### crewAI Version

1.15.21 (main @ e1f3c4bdd)

### crewai-tools Version

same workspace commit

### Proposed fix

Use `os.replace`, which overwrites the destination on every platform. A one-line PR with a regression test follows.

Disclosure per CONTRIBUTING: this report and the fix were prepared with the help of an AI coding assistant; please apply the `llm-generated` label.

SharoonSharif:
Fixes #7462.

**Problem.** `migrate_pyproject` backs up `poetry.lock` with `os.rename`. On Windows that raises `FileExistsError` when `poetry-old.lock` already exists, so a second `crewai update` crashes before the new `pyproject.toml` is written. POSIX replaces the file silently, which is why this only shows on Windows.

**Fix.** `os.rename` → `os.replace`, which overwrites on every platform (matching the `shutil.copy2` semantics already used for `pyproject-old.toml`).

**Test.** `lib/cli/tests/test_update_crew.py`: a baseline test for the backup, and a test that runs the migration with a stale `poetry-old.lock` present. The second fails on Windows before this change (`FileExistsError [WinError 183]`) and passes after; both pass on Linux.

**Verification (Windows 11, Python 3.13):**

```
uv run pytest lib/cli/tests/test_update_crew.py -q   # 2 passed
uv run ruff check lib/cli/src/crewai_cli/update_crew.py lib/cli/tests/test_update_crew.py
uv run ruff format --check lib/cli/src/crewai_cli/update_crew.py lib/cli/tests/test_update_crew.py
uv run mypy --config-file pyproject.toml lib/cli/src/crewai_cli/update_crew.py
```

Disclosure per CONTRIBUTING: prepared with the help of an AI coding assistant; please apply the `llm-generated` label.


Vidit-Ostwal (approved):
LGTM

PR #7477, opened 2026-09-15
fix(a2a): join appended streaming chunks without spaces

### LINKED ISSUE #7473: [BUG] A2A artifacts parts are reassembled to corrupted text with extra whitespaces
### Description

When a remote A2A agent streams (or persists) its reply as more than one
text chunk, CrewAI's client-side reassembly corrupts the text: words get
split apart and extra spaces are inserted.
Per the A2A spec, artifact parts sent with `append=True` are meant to be
concatenated directly with no separator — that's the definition of
"append". CrewAI instead does:
```python
# crewai/a2a/task_helpers.py, process_task_state()
response_text = " ".join(result_parts) if result_parts else ""
```
This function is shared by **all three** update mechanisms
(`StreamingHandler`, `PollingHandler`, `PushNotificationHandler`), so
switching `updates=` doesn't avoid it — and there's an identical
`" ".join(result_parts)` fallback inside `StreamingHandler.execute()`
itself (`crewai/a2a/updates/streaming/handler.py`), plus a third,
independent occurrence with the same pattern in
`crewai/a2a/wrapper.py::_handle_max_turns_exceeded`.
This affects **any** spec-compliant server that streams its reply
incrementally — e.g. any server relaying real token-by-token LLM output —
regardless of framework. It isn't specific to one server implementation.

### Steps to Reproduce

```python
"""Minimal, framework-agnostic reproduction of a CrewAI A2A bug.
Run:
    pip install crewai a2a-sdk
    python repro_a2a_streaming_join_bug.py
Expected: Hello, world!
Actual:   Hel lo,  wor ld !
"""
from __future__ import annotations
import asyncio
import uuid
from a2a.types import (
    AgentCapabilities,
    AgentCard,
    Artifact,
    Message,
    Part,
    Role,
    Task,
    TaskArtifactUpdateEvent,
    TaskState,
    TaskStatus,
    TaskStatusUpdateEvent,
    TextPart,
)
from crewai.a2a.updates.streaming.handler import StreamingHandler
TASK_ID = "task-1"
CONTEXT_ID = "ctx-1"
# Split "Hello, world!" into several small append=True chunks -- exactly
# what any server relaying incremental/token-level output would send.
CHUNKS = ["Hel", "lo, ", "wor", "ld", "!"]
EXPECTED = "".join(CHUNKS)
def _text_part(text: str) -> Part:
    return Part(root=TextPart(text=text))
async def _fake_send_message(_message: Message):
    """Stand-in for `a2a.client.Client.send_message`, yielding (Task, update)
    tuples the same way a real client does while consuming a server's SSE
    stream."""
    task = Task(id=TASK_ID, context_id=CONTEXT_ID, status=TaskStatus(state=TaskState.working))
    for i, chunk in enumerate(CHUNKS):
        artifact = Artifact(artifact_id="reply", parts=[_text_part(chunk)])
        update = TaskArtifactUpdateEvent(
            task_id=TASK_ID,
            context_id=CONTEXT_ID,
            artifact=artifact,
            append=i > 0,
            last_chunk=i == len(CHUNKS) - 1,
        )
        yield (task, update)
    final_status = TaskStatusUpdateEvent(
        task_id=TASK_ID,
        context_id=CONTEXT_ID,
        status=TaskStatus(state=TaskState.completed),
        final=True,
    )
    yield (task, final_status)
class FakeClient:
    def send_message(self, message: Message):
        return _fake_send_message(message)
async def main() -> None:
    agent_card = AgentCard(
        name="Repro Agent",
        description="Minimal agent card for reproduction purposes.",
        url="http://localhost:9999",
        version="1.0.0",
        capabilities=AgentCapabilities(streaming=True),
        default_input_modes=["text"],
        default_output_modes=["text"],
        skills=[],
    )
    message = Message(
        role=Role.user,
        parts=[_text_part("say hello")],
        message_id=str(uuid.uuid4()),
    )
    result = await StreamingHandler.execute(
        client=FakeClient(),
        message=message,
        new_messages=[],
        agent_card=agent_card,
        endpoint=agent_card.url,
    )
    actual = result["result"]
    print(f"Expected: {EXPECTED!r}")
    print(f"Actual:   {actual!r}")
    assert actual == EXPECTED, "BUG: chunks were joined with a space instead of concatenated"
if __name__ == "__main__":
    asyncio.run(main())
```

### Expected behavior

**Output:**
```
Expected: 'Hello, world!'
Actual:   'Hel lo,  wor ld !'
  ...
AssertionError: BUG: chunks were joined with a space instead of concatenated
```

### Screenshots/Code snippets

```python
# crewai/a2a/task_helpers.py, process_task_state()
response_text = " ".join(result_parts) if result_parts else ""
```

### Operating System

macOS Sonoma

### Python Version

3.11

### crewAI Version

1.15.21

### crewAI Tools Version

n/a

### Virtual Environment

Venv

### Evidence

see above

### Possible Solution

just reassemble with "".join instead of " ".join ?

### Additional context

no

parthiban-sivakumar:
## Related issue

Fixes #7473

## Summary

When an A2A agent streams its reply in small chunks, CrewAI joined the chunks with a space. So "Hello, world!" came back as "Hel lo,  wor ld !".

Each chunk comes as an artifact update with `append=True`. Now the streaming handler checks this flag. If `append` is true, the text is added to the same artifact's text with no space. Different artifacts are still joined with a space like before, so normal replies do not change.

The change is in one helper, `process_artifact_update` in `streaming/params.py`. It is used in the main stream loop and also in the reconnect path, so chunks that come after a reconnect are joined correctly too.

## Verification

I added `tests/a2a/test_streaming_artifact_text.py` with 4 tests:
- the chunks from the issue come back as "Hello, world!"
- two separate artifacts still have a space between them
- two artifacts streamed at the same time each keep their own text
- chunks after a reconnect continue the same text

Three of these tests fail on `main` and pass with this change. The script from the issue now prints `Hello, world!`. The full `lib/crewai/tests/` suite passes (5475 passed, 43 skipped), and ruff and mypy are clean.

## Additional context

This PR only fixes streaming. Polling and push notifications also join parts with a space, but there the finished artifact does not tell us which parts were streamed chunks, so I did not change them here.


parthiban-sivakumar:
@Vidit-Ostwal the 3 failing tests in this run are `tests/llms/azure/test_azure_responses.py::TestAzureResponsesProperties`. This PR only changes A2A streaming files, so I do not think they are related.

I ran them locally on this branch: the file alone passes (25 passed), the whole `tests/llms/azure` folder passes (108 passed), and CI's own split (`--splits 8 --group 5`) on Python 3.10 also passes. Could you please re-run the job?

VANDRANKI (commented):
Community review, not a merge-gate approval.

I traced `process_artifact_update` in `lib/crewai/src/crewai/a2a/updates/streaming/params.py` against all 5 new tests by hand, plus confirmed the final assembly point: `handler.py` line ~643 does `result=" ".join(result_parts) if result_parts else ""`. That's the key fact that makes the fix's design make sense: each entry in `result_parts` is one artifact's accumulated text, joined with a space against every *other* artifact at the very end, while within one artifact's entry, appended chunks are now concatenated with `"".join(texts)` (no separator). The old code did `result_parts.extend(part.root.text for part in artifact.parts if part.root.kind == "text")`, which put every chunk of every artifact update into its own slot in `result_parts`, so a server streaming one reply as `["Hel", "lo, ", "wor", "ld", "!"]` would come back as `"Hel lo,  wor ld !"` after the final space-join, since the join can't tell a mid-word chunk boundary from a real artifact boundary.

The new `artifact_positions: dict[str, int]` fixes that by remembering which slot in `result_parts` belongs to which `artifact_id`. I walked through each test:
- `test_appended_chunks_are_joined_without_spaces`: 5 chunks, same artifact, first `append=False` then rest `append=True`. Traces to a single `result_parts` entry built by concatenation, giving exactly `"Hello, world!"`.
- `test_interleaved_artifacts_append_to_their_own_text`: two different artifact IDs interleaved (`greeting`, `subject`, `greeting`, `subject`), each with its own append chunk. Traces to two independent slots that never cross-contaminate, joined with one space at the end: `"Hello World"`.
- `test_an_update_without_append_replaces_the_artifact_text`: a third update to the same artifact with `append=False` after two appends correctly hits the `elif update.append` / `else` branch's `else`, replacing the slot outright (`result_parts[position] = ...`) rather than appending, matching the docstring's claim that this mirrors `a2a.utils.append_artifact_to_task` replacing an artifact of the same ID.
- `test_chunks_after_a_reconnect_continue_the_same_text` is the one I checked most carefully, since it depends on state surviving an interruption: the diff threads a new `artifact_positions` parameter into `_try_recover_from_interruption` alongside `result_parts`, and both call sites in `execute()` pass the same live dict through (`artifact_positions=artifact_positions`). So a reconnect's resubscribed chunks still find the pre-disconnect artifact's position and append to it rather than starting a fresh, disconnected entry. I confirmed both `execute()` call sites of `_try_recover_from_interruption` were updated with this new argument, not just one.

I did not check every other caller of `process_status_update` (the sibling function in the same file) for a similar issue, this review is scoped to the artifact-text path the diff actually changes. Within that scope, the logic is correct and the test suite specifically targets the boundary cases (interleaving, replace-vs-append, reconnect) rather than just the straightforward case.

PR #7478, opened 2026-09-15
fix(gemini): keep fileData parts when formatting messages

### LINKED ISSUE #7476: [BUG] Native Gemini provider silently drops fileData content parts
### Description

The native Gemini provider silently drops `fileData` content parts in `_format_messages_for_gemini`.

`crewai_files` correctly formats an `ImageFile` backed by a `FileUrl` as a Gemini `fileData` block. The provider's message conversion then handles only `text` and `inlineData` dictionaries, so the file reference is omitted from the resulting `types.Content`.

This affects URI-based media inputs, including public HTTPS image URLs and Google Cloud Storage URIs. A mixed text/media message becomes a text-only request; a media-only message becomes a content object with no parts. The conversion raises no error for the unsupported block.

### Versions

- `crewai==1.15.20`
- `crewai-files==1.15.21`
- Python 3.12

### Minimal reproduction (no API call required)

```python
from crewai import LLM
from crewai_files import FileUrl, ImageFile, format_multimodal_content

uri = "https://storage.googleapis.com/generativeai-downloads/images/scones.jpg"

blocks = format_multimodal_content(
    {"image": ImageFile(source=FileUrl(url=uri))},
    "gemini",
)
assert any("fileData" in block for block in blocks)

llm = LLM(
    model="gemini/gemini-2.5-flash",
    api_key="test-key",  # Conversion only; no network request is made.
)

    {
        "role": "user",
        "content": [{"text": "Describe this image."}, *blocks],
    }
])

# Fails: the image reference was silently discarded.
assert any(part.file_data is not None for part in contents[0].parts)
```

The same issue can be reproduced with a manually supplied GCS block:

```python
    {
        "role": "user",
        "content": [{
            "fileData": {
                "fileUri": "gs://generativeai-downloads/images/scones.jpg",
                "mimeType": "image/jpeg",
            }
        }],
    }
])

# Currently an empty list of parts.
assert contents[0].parts
```

### Expected behavior

The Gemini provider should preserve `fileData` parts, including the URI, MIME type, and their position relative to text and inline-data parts. This should work for both synchronous and asynchronous model calls, which share the converter.

For example, convert the block using the supported Google Gen AI SDK API:

```python
types.Part.from_uri(
    file_uri=item["fileData"]["fileUri"],
    mime_type=item["fileData"]["mimeType"],
)
```

### Relevant implementation

- `GeminiCompletion._format_messages_for_gemini`: dictionary content parts currently have branches for `text` and `inlineData`, with no branch for `fileData`.
- `crewai_files.formatting.gemini.GeminiFormatter.format_block`: produces `fileData` for `UrlReference` and `FileReference`.

### References

- [Google Gen AI SDK: URI-based image input](https://googleapis.github.io/python-genai/)
- [Google Gemini API: file input methods](https://ai.google.dev/gemini-api/docs/file-input-methods)

parthiban-sivakumar:
## Related issue

Fixes #7476

## Summary

The native Gemini provider only handled `text` and `inlineData` parts in `_format_messages_for_gemini`. A `fileData` part was skipped with no error. So when a file is sent as a URL or as an uploaded file, Gemini got only the text, and a message with only a file became empty.

This PR adds a `fileData` branch. It uses `types.Part.from_uri` with the file URI and MIME type, and it keeps the order of the parts. Sync and async calls use the same function, so both are fixed.

## Verification

I added 2 tests in `tests/llms/google/test_google.py`:
- text, file and text parts are all kept, in the same order
- a message with only a file is not empty

Both tests fail on `main` and pass with this change. I also checked the real path: with `llm.call()` and an image URL in `files`, the request to Gemini now has the file part. Before the fix it had only the text. The full `lib/crewai/tests/` suite passes (5476 passed, 43 skipped), and ruff and mypy are clean.


Vidit-Ostwal:
Thanks for the contribution. This is a duplicate of #7479, which contains the selected fix for #7476 and is being merged. Closing this PR so the work tracks in one place.

PR #7500, opened 2026-09-16
feat(llm): llm_overlay — route agent roles to models for a block, without editing the crew

joaomdmoura:
## Related issue

None — maintainer change (part of the crewai-eval revamp: re-running a crew under other model configurations).

## Summary

A new module `crewai/llm_overlay.py`: a `ContextVar` `active` holding a `{role: model}` mapping, a `llm_overlay(mapping)` context manager that sets it for the block and always resets it, and `overlay_model_for(role)`. The two places an agent resolves its model from its role — `Agent.post_init_setup` (`agent/core.py`) and `LiteAgent.setup_llm` (`lite_agent.py`) — become `create_llm(overlay_model_for(self.role) or self.llm)`. When no overlay is active, or the role is not a key, nothing changes.

Why a construction-time context variable and not the existing `before_llm_call` hooks: the executor binds `llm=self.llm` before those hooks run, so a hook that swaps the executor's LLM misses the call it runs in; and `create_llm` is role-blind, so the hook belongs at the two role-aware sites. This is what lets a caller evaluate a flow under a different model per agent role when the flow builds its crews inside its methods.

## Verification

- [x] Tests added — `lib/crewai/tests/test_llm_overlay.py` (6 tests): an agent whose role is named constructs on the overlay's model and others keep their own; no leak outside the block; `LiteAgent` covered; a plain `threading.Thread` started inside the block does NOT see the overlay (contextvar semantics, documented in the module docstring with `contextvars.copy_context().run`).
- [x] Relevant tests and quality checks pass locally — `lib/crewai/tests/agents/`: 409 passed, 15 pre-existing skips; ruff, ruff format, mypy (strict) clean.

## Additional context

`create_llm("openai/gpt-4o").model` is `"gpt-4o"` (the native provider strips the prefix); the overlay passes the string through unchanged. Not exported from `crewai/__init__.py`, which exports classes only; import from `crewai.llm_overlay`. Verified end to end from crewai-evals: a flow that builds its crew inside a method ran its Analyst on `gpt-4o-mini` and, under the overlay, on `gpt-4.1-mini` (visible in the traces).

---

> [!NOTE]
> **Low Risk**
> Opt-in context manager with no behavior change when unused; only affects which model string is passed to create_llm at agent construction for matching roles.
> 
> **Overview**
> Adds **`crewai.llm_overlay`**, a context-scoped `{role: model}` map so callers can run the same crew-building code under different models without changing agent definitions. The **`llm_overlay(mapping)`** context manager stores the map in a **`ContextVar`** and always restores the previous value on exit (including on exceptions).
> 
> **`Agent`** and **`LiteAgent`** now resolve their LLM at init with **`create_llm(overlay_model_for(self.role) or self.llm)`**, so only agents whose **role** matches a key get the overlay model; others and all agents built outside the block are unchanged. **`function_calling_llm`** is not affected.
> 
> The module documents that the overlay follows the calling context, not new threads—callers must use **`contextvars.copy_context()`** when building agents off-thread. Tests cover role matching, no leakage, nested overlays, **`LiteAgent`**, and thread semantics.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9db7781388aa5025be92424a6624f71453f4a9d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Vidit-Ostwal (approved):
LGTM

PR #7501, opened 2026-09-16
fix(agents): carry from_cache on the native tool path's ToolUsageFinishedEvent

joaomdmoura:
## Related issue

None — bug found while measuring tool replay for crewai-eval.

## Summary

`CrewAgentExecutor._execute_single_native_tool_call` (native function calling) reads the tools handler's cache and skips the tool body on a hit — `from_cache` is computed and returned in its result dict — but the `ToolUsageFinishedEvent` it emits is built without `from_cache`, so the event defaults to `False`. Every consumer of the bus (the tracing listener, and so AMP's trace view) sees a replayed native tool call as a live one. The text-protocol path (`ToolUsage.on_tool_use_finished`) already carries the flag. One keyword argument fixes it.

Measured before the fix on a two-agent crew with a tool, cache pre-seeded: 2 cache hits, 0 tool executions, 0 events saying `from_cache`.

## Verification

- [x] Tests added — `lib/crewai/tests/agents/test_native_tool_from_cache.py`: a cached call yields `result["from_cache"] is True`, the tool body does not run, and the finished event carries `from_cache=True` with the cached output; a live call reports `False` and runs the tool once.
- [x] Relevant tests and quality checks pass locally — the two new tests pass; ruff clean on both files.

## Additional context

None.

---

> [!NOTE]
> **Low Risk**
> Single optional event field on an existing code path; behavior change is observability-only for cache hits.
> 
> **Overview**
> Native function-calling tool runs already set `from_cache` on the executor result when the tools handler cache hits, but **`ToolUsageFinishedEvent` was emitted without that field**, so it defaulted to `False` and tracing/UI treated cache replays as live executions.
> 
> This PR passes **`from_cache=from_cache`** into the finished event in `_execute_single_native_tool_call`, aligning the native path with the text-protocol behavior. Tests cover cache hits (no tool body run, event `from_cache=True`, cached output) and live calls (`from_cache=False`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 912b3208c64ff60f34488a7cbba41d940a1740f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Vidit-Ostwal (approved):
LGTM

PR #7511, opened 2026-09-16
fix: close sqlite connections in flow persistence and sqlite provider

### LINKED ISSUE #7510: [BUG] SQLiteFlowPersistence and SqliteProvider never close their SQLite connections (Windows file locks)
### Description

Follow-up to #7492 / #7493, which fixed `KickoffTaskOutputsSQLiteStorage`. The same pattern remains in the two other library-side SQLite backends:

- `crewai/flow/persistence/sqlite.py` — `SQLiteFlowPersistence.init_db`, `save_state`, `load_state`, `save_pending_feedback`, `load_pending_feedback`, `clear_pending_feedback`
- `crewai/state/provider/sqlite_provider.py` — `SqliteProvider.checkpoint`, `prune`, `from_checkpoint`

All nine use `with sqlite3.connect(...) as conn:` and never call `close()`. The `sqlite3.Connection` context manager only commits or rolls back, and since Python 3.11 the connection sits in a reference cycle (statement cache), so the OS file handle survives until a cyclic GC pass.

On Windows the open handle locks `flow_states.db` / the checkpoint database, so deleting, replacing or cleaning up the directory afterwards fails with `PermissionError: [WinError 32]`. In this repo's own suite on Windows, `test_flow_persistence.py`, `test_async_human_feedback.py`, `test_checkpoint.py` and `test_flow_persistence_factory.py` together show 30 failures on `main`; with this fix they show 3, all of them path-separator assertions unrelated to SQLite.

### Steps to Reproduce

```python
import os, tempfile
from crewai.flow.persistence.sqlite import SQLiteFlowPersistence

d = tempfile.mkdtemp()
db = os.path.join(d, "flows.db")
p = SQLiteFlowPersistence(db)
p.save_state("flow-1", "start", {"step": 1})
p.load_state("flow-1")
del p
os.remove(db)   # Windows: PermissionError [WinError 32]; Linux: succeeds
```

Same with `SqliteProvider().checkpoint(...)` followed by `from_checkpoint(...)`.

### Expected behavior

Each operation closes the connection it opened, so the database file is releasable immediately after use on every platform.

### Screenshots/Code snippets

```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\...\\Temp\\tmpabc\\flows.db'
```

### Operating System

Windows 11

### Python Version

3.13

### crewAI Version

main @ c0af9badb (1.15.21+)

### crewai-tools Version

same workspace commit

### Proposed fix

Same as #7493: wrap each connection in `contextlib.closing` while keeping the existing commit/rollback context manager. PR with lifecycle and failure-path regression tests follows. After this, the only remaining sites are in the CLI (`checkpoint_cli.py`, `task_outputs.py`), which I'd propose as one more small PR.

Disclosure per CONTRIBUTING: this report and the fix were prepared with the help of an AI coding assistant; please apply the `llm-generated` label.

SharoonSharif:
Fixes #7510. Follow-up to #7493.

**Problem.** `SQLiteFlowPersistence` (6 sites) and `SqliteProvider` (3 sites) use `with sqlite3.connect(...) as conn:`, which commits or rolls back but never closes. The connection survives in a reference cycle until a GC pass, so on Windows `flow_states.db` and checkpoint databases stay locked and later deletes/cleanups fail with `PermissionError (WinError 32)`.

**Fix.** Wrap each connection in `contextlib.closing`, keeping the existing commit/rollback semantics (`closing(sqlite3.connect(...)) as conn, conn`). No behaviour change other than closing on exit. The async paths already close via `aiosqlite`'s context manager. `test_checkpoint.py`'s own read-back connections are closed the same way, since they made its `TemporaryDirectory` cleanup fail on Windows.

**Test.** `lib/crewai/tests/test_sqlite_connection_lifecycle.py`: connection-tracking tests for both classes covering all nine sites, a failure-path test (an `INSERT` that raises inside `save_state` rolls back and still closes), and file-removal tests for the Windows symptom. On Windows, the related suites go from 30 failures on `main` to 3 (all path-separator assertions unrelated to SQLite).

**Verification (Windows 11, Python 3.13):**

```
uv run pytest lib/crewai/tests/test_sqlite_connection_lifecycle.py -q                     # 5 passed
uv run pytest lib/crewai/tests/test_flow_persistence.py lib/crewai/tests/test_async_human_feedback.py lib/crewai/tests/test_checkpoint.py lib/crewai/tests/test_flow_persistence_factory.py -q
uv run ruff check lib/crewai/src/crewai/flow/persistence/sqlite.py lib/crewai/src/crewai/state/provider/sqlite_provider.py
uv run ruff format --check lib/crewai/src/crewai/flow/persistence/sqlite.py lib/crewai/src/crewai/state/provider/sqlite_provider.py
uv run mypy --config-file pyproject.toml lib/crewai/src/crewai/flow/persistence/sqlite.py lib/crewai/src/crewai/state/provider/sqlite_provider.py
```

Remaining sites after this PR: `crewai_cli/checkpoint_cli.py` (5) and `crewai_cli/task_outputs.py` (1); happy to follow up.

Disclosure per CONTRIBUTING: prepared with the help of an AI coding assistant; please apply the `llm-generated` label.


VANDRANKI (commented):
Community review, not a merge-gate approval.

This fixes a real, well-known Python gotcha rather than a hypothetical one: `sqlite3.Connection` used as a context manager (`with sqlite3.connect(...) as conn:`) only manages the transaction (commit on success, rollback on exception), it never closes the underlying connection. The old code in both `flow/persistence/sqlite.py` and `state/provider/sqlite_provider.py` relied on `with sqlite3.connect(...) as conn:` everywhere, so every one of these operations leaked an open sqlite connection, relying on GC/`__del__` to eventually close it rather than closing it deterministically.

The fix, `with closing(sqlite3.connect(...)) as conn, conn:`, is a correct and slightly clever combination: `closing(...)` provides the actual `.close()` on exit, and reusing `conn` itself as a second context-manager item in the same `with` still gets you the original commit/rollback behavior. I checked the exit ordering matters here and it's right: `with` statement items exit in reverse order of entry, so the second item (`conn`, the transaction manager) exits first, i.e. commit-or-rollback happens before `closing(conn)` exits second and actually calls `.close()`. Closing before committing would silently lose the transaction; this doesn't do that.

The new `test_sqlite_connection_lifecycle.py` verifies this properly rather than just checking no exception is raised. `test_flow_persistence_closes_every_connection` monkeypatches `sqlite3.connect` to record every connection object handed out, exercises all 6 `SQLiteFlowPersistence` operations (`init_db` via construction, `save_state`, `load_state`, `save_pending_feedback`, `load_pending_feedback`, `clear_pending_feedback`), and then asserts each recorded connection raises `sqlite3.ProgrammingError: closed database` on further use, which is the only way to actually confirm `.close()` ran (the old buggy code would pass a weaker "no exception during normal use" test but fail this one). `test_flow_persistence_failed_write_rolls_back_and_closes` goes further and checks the failure path specifically: a custom `_FailingConnection` that raises on the INSERT statement confirms the connection still gets closed even when the transaction aborts, and that the failed write actually rolled back (`load_state` returns `None` afterward) rather than leaving a partial row.

The two `_db_is_removable_after_use` tests (deleting the file immediately after use, no GC pause) are a good practical proxy for "the file handle is really released," which matters on Windows where an open sqlite file handle blocks deletion.

I traced the changed call sites in both files and they're all mechanically the same substitution with no logic changes beyond the context-manager swap, so I don't see a way this introduces a new bug. This is a solid, correctly verified resource-leak fix.

Vidit-Ostwal (approved):
Approved: the closing-plus-transaction context manager preserves commit/rollback behavior and deterministically closes all nine affected SQLite connections. Lifecycle and rollback coverage are appropriately focused.

PR #7519, opened 2026-09-16
Support aliases as connection identifiers

vinibrsl:
App selectors only accepted UUID connection identifiers, so `apps=["github@private"]` failed validation.

Accept connection aliases and route them through Clipper like UUID identifiers. This supports private connections and future platform aliases without OSS releases.


Vidit-Ostwal (approved):
LGTM

PR #7525, opened 2026-09-17
fix(cli): close sqlite connections in checkpoint and task-outputs commands

### LINKED ISSUE #7524: [BUG] crewai-cli SQLite readers never close their connections (Windows file locks)
### Description

Follow-up to #7492 and #7510, which fixed the library-side SQLite backends. The same pattern remains in the third component that opens these databases: **`crewai-cli`**.

- `crewai_cli/checkpoint_cli.py` — `_list_sqlite`, `_info_sqlite_latest`, `_info_sqlite_id`, `_prune_sqlite`, and the `--dry-run` count inside `prune_checkpoints`
- `crewai_cli/task_outputs.py` — `load_task_outputs` (used by `crewai log-tasks-outputs`)

All six use `with sqlite3.connect(...) as conn:` and never call `close()`. The `sqlite3.Connection` context manager only commits or rolls back — it does not close the connection. Since Python 3.11 the connection additionally sits in a reference cycle via its statement cache, so the OS file handle survives until a cyclic GC pass.

On Windows the lingering handle locks the checkpoint database (and `latest_kickoff_task_outputs.db`), so deleting, replacing, or cleaning up the directory afterwards fails with `PermissionError: [WinError 32]`.

### Steps to Reproduce

Run any of the SQLite-backed checkpoint commands (or the task-outputs reader), then inspect the connections the command opened:

```python
import os, sqlite3, tempfile

import crewai_cli.task_outputs as task_outputs_module
from crewai_cli.task_outputs import load_task_outputs

d = tempfile.mkdtemp()
db = os.path.join(d, "latest_kickoff_task_outputs.db")
with sqlite3.connect(db) as conn:
    conn.execute(
        "CREATE TABLE latest_kickoff_task_outputs ("
        "task_id TEXT PRIMARY KEY, task_key TEXT, expected_output TEXT, "
        "output JSON, task_index INTEGER, inputs JSON, "
        "was_replayed BOOLEAN, timestamp DATETIME)"
    )
    conn.commit()

opened = []
real_connect = sqlite3.connect

def tracking_connect(*args, **kwargs):
    conn = real_connect(*args, **kwargs)
    opened.append(conn)
    return conn

with patch.object(task_outputs_module.sqlite3, "connect", tracking_connect):
    load_task_outputs(db)

# Expected on a closed connection:
for conn in opened:
    conn.execute("SELECT 1")   # sqlite3.ProgrammingError: Cannot operate on a closed database.
```

On `main` this raises **no** error: every connection is still open after the command returned. On Windows, deleting `db` at this point raises `PermissionError: [WinError 32]` instead.

### Expected behavior

Each command closes the connection it opened, so the database file is releasable immediately after use on every platform — the same guarantee #7493 and #7510 established for the library-side backends.

### Screenshots/Code snippets

`lib/cli/src/crewai_cli/checkpoint_cli.py`:

```python
def _list_sqlite(db_path: str) -> list[dict[str, Any]]:
    results = []
    with sqlite3.connect(db_path) as conn:   # commits/rolls back, never closes
        ...
```

`lib/cli/src/crewai_cli/task_outputs.py`:

```python
with sqlite3.connect(db_path) as conn:       # commits/rolls back, never closes
    ...
```

### Operating System

Windows 11 (the file-lock symptom is Windows-only; the leaked handle reproduces on every OS)

### Python Version

3.10, 3.11, 3.12

wangtaotaotao95:
## Related issue

Fixes #7524

## Summary

`with sqlite3.connect(...) as conn` only commits or rolls back — it never closes the connection. The connection then survives in a reference cycle (its statement cache, since Python 3.11) until a cyclic GC pass, so the OS file handle outlives the call.

`crewai-cli` still had six such sites, all in the SQLite-backed checkpoint/task-output readers:

- `crewai_cli/checkpoint_cli.py` — `_list_sqlite`, `_info_sqlite_latest`, `_info_sqlite_id`, `_prune_sqlite`, and the `--dry-run` count in `prune_checkpoints`
- `crewai_cli/task_outputs.py` — `load_task_outputs` (used by `crewai log-tasks-outputs`)

On Windows the lingering handle locks the checkpoint database / `latest_kickoff_task_outputs.db`, so any later delete, replace or temp-dir cleanup fails with `PermissionError: [WinError 32]`.

Each connection is now wrapped in `contextlib.closing`, using the pattern already established for the library-side backends in #7493 and #7510:

```python
with closing(sqlite3.connect(db_path)) as conn, conn:
```

The inner context manager keeps the existing commit/rollback semantics; the outer one closes the connection on the way out, including when the body raises. This is the third and last component in this bug class — after it, every `sqlite3.connect` in the repo is closed explicitly (the two library-side backends are covered by #7511).

## Verification

- Added `TestSqliteConnectionLifecycle` in `lib/crewai/tests/test_checkpoint_cli.py` (5 tests) and `lib/crewai/tests/test_task_outputs.py` (3 tests). The connection tests record every connection opened through the module under test and assert `sqlite3.ProgrammingError: Cannot operate on a closed database` when using it afterwards, so they fail on the old code and pass on the new one.
- Confirmed the new tests fail before the fix (`5 failed, 36 passed`) and pass after (`41 passed`).
- `uv run pytest tests/test_checkpoint.py tests/storage/ -q` → `69 passed`.
- `uv run ruff check` and `uv run ruff format --check` clean on both changed source files.

Note: the `os.remove`-style assertions in the tests are macOS/Linux-green either way; the leaked-handle assertion is the platform-independent signal, and the file-lock symptom itself only reproduces on Windows.

## Additional context

`lib/cli/tests/` is not exercised by the `tests.yml` workflow, which is why the regression tests live under `lib/crewai/tests/` alongside the existing `test_checkpoint_cli.py`.


Vidit-Ostwal (dismissed):
LGTM

Vidit-Ostwal (commented):
I agree with CodeRabbit: the file-release regression tests create their fixture databases with , which commits/rolls back but does not explicitly close the setup connection. On Windows, that fixture handle may retain the database lock before the production reader is exercised, so the deletion assertion does not reliably prove the intended behavior. Please wrap both fixture builders with  (or close explicitly in a finally block). The production connection changes otherwise look correct.

Vidit-Ostwal (commented):
I agree with CodeRabbit: the file-release regression tests create their fixture databases with a sqlite3 connection context manager, which commits or rolls back but does not explicitly close the setup connection. On Windows, that fixture handle may retain the database lock before the production reader is exercised, so the deletion assertion does not reliably prove the intended behavior. Please wrap both fixture builders with contextlib.closing around the connection context (or close explicitly in a finally block). The production connection changes otherwise look correct.

wangtaotaotao95:
Fair catch, and a slightly embarrassing one: the regression tests assert that the CLI releases its SQLite handles, while their own fixtures used the identical bare `with sqlite3.connect(...)`. Fixed in `0ed846d33`.

Wrapped the fixture builders in `contextlib.closing`, and with them the two read-side counts in the checkpoint suite that had the same shape — leaving those in place while asserting the pattern is wrong would have been inconsistent:

```python
with closing(sqlite3.connect(db_path)) as conn, conn:
    ...
```

So the only handles these tests can now observe are the ones the production code opens. `41 passed`.

On your point about the deletion assertion: you're right that it was not reliable, and it is also the weaker of the two signals I added — `unlink()` succeeding says nothing on Linux/macOS, where an open file can still be removed, so it only means something on Windows. The platform-independent assertion is the one in `test_load_task_outputs_closes_connection` and the `TestSqliteConnectionLifecycle` cases, which record every connection opened through the module under test and assert `sqlite3.ProgrammingError: Cannot operate on a closed database` when using them afterwards. That is what actually pins the behaviour; the `unlink` cases are the Windows-shaped symptom on top. Your fix removes the fixture handle that could have made the symptom fire for the wrong reason.

Also noticed you merged `main` into this branch while I was working — I rebased my follow-up on top of that commit rather than merging again, so the branch stays linear.

wangtaotaotao95:
Updated the branch to current main (0374c63129). The PR diff remains limited to the SQLite connection fix and its tests.

The refreshed matrix currently fails in unrelated Azure Responses tests under tests/llms/azure/test_azure_responses.py (mock_cls.call_args is None); the same base-only failure shape also appears on #7531 after updating that independent PR to the same main. Lint, type checks, CodeQL, and pip-audit pass. I am leaving the branch on current main rather than adding an out-of-scope Azure test change here.

@Vidit-Ostwal, the fixture-connection feedback is addressed on the current head; a re-review would be appreciated.

PR #7531, opened 2026-09-17
fix(tools): keep the output archive out of itself in FileCompressorTool

### LINKED ISSUE #7528: [BUG] FileCompressorTool includes its output ZIP in the archive
### Description

`FileCompressorTool` includes the output ZIP as an empty member when the archive is created inside the directory being compressed.

This is easy to hit with `input_path="."` and `output_path="bundle.zip"`. Opening the ZIP creates the output file before `os.walk()` enumerates the directory, so the new archive is written into itself.

### Steps to Reproduce

```python
from pathlib import Path
from tempfile import TemporaryDirectory
from zipfile import ZipFile

from crewai_tools import FileCompressorTool

with TemporaryDirectory(dir=".") as tmp:
    root = Path(tmp).resolve()
    (root / "payload.txt").write_text("hello", encoding="utf-8")

    FileCompressorTool()._run(
        input_path=str(root),
        output_path=str(root / "bundle.zip"),
    )

    with ZipFile(root / "bundle.zip") as archive:
        print(archive.namelist())
```

Current output:

```text
['bundle.zip', 'payload.txt']
```

### Expected behavior

The generated archive should contain `payload.txt`, but not `bundle.zip`. Extracting it should not create an empty archive that was absent from the source directory.

### Screenshots/Code snippets

The directory branch opens `output_path` first and then walks `input_path` without excluding that file:

```python
with zipfile.ZipFile(output_path, "w", zipfile.ZIP_DEFLATED) as zipf:
    for root, _, files in os.walk(input_path):
        for file in files:
            full_path = os.path.join(root, file)
            zipf.write(full_path, os.path.relpath(full_path, start=input_path))
```

### Operating System

macOS (15.5)

### Python Version

3.13

### crewAI Version

Current `main` at `7a01af27912c`; the bug is in `crewai-tools`.

### crewAI Tools Version

1.15.22

### Virtual Environment

Venv (`uv`)

### Evidence

A direct execution of the current `_compress_zip` implementation produces a zero-byte `bundle.zip` member. Existing compressor tests mock `_compress_zip`, so they do not inspect real archive membership.

I also searched open issues and pull requests for `FileCompressorTool`, archive self-inclusion, and ZIP output overlap, and did not find an existing fix.

### Possible Solution

While walking a directory, skip entries that identify the output archive. A filesystem identity check also covers a symlink alias to the output. Add a real-filesystem regression test for both a new output and `overwrite=True`.

### Additional context

AI-assisted report. I will open a focused fix with regression coverage and apply the required `llm-generated` label.

wangtaotaotao95:
## Related issue

Fixes #7528

## Summary

`_compress_zip` opened the output archive **before** walking `input_path`. When the output lands inside the source tree, the freshly created (still empty) archive is enumerated by `os.walk` and added to itself:

```python
FileCompressorTool()._run(input_path=".", output_path="bundle.zip")
# before: ['bundle.zip', 'payload.txt']
# after:  ['payload.txt']
```

Extracting the result produced an empty archive that never existed in the source directory.

The fix resolves the output path once and skips that entry during the walk:

```python
output_real_path = os.path.realpath(output_path)
...
full_path = os.path.join(root, file)
if os.path.realpath(full_path) == output_real_path:
    continue
```

`realpath` (rather than the arcname) is what makes this correct for a nested output: with `output_path="nested/bundle.zip"` the archive member is `nested/bundle.zip` while the on-disk path differs, so matching on the resolved file is what catches it.

**`_compress_tar` is deliberately untouched.** It looked like the same bug, so I checked before changing it — `TarFile.add()` already guards itself:

```python
# Skip if somebody tries to archive the archive...
if self.name is not None and os.path.abspath(name) == self.name:
    return
```

A `tar.gz` run of the same repro yields `['<dir>', '<dir>/payload.txt']` with no self-member, so there is nothing to fix there.

## Why this shipped: the test module never ran

`lib/crewai-tools/tests/tools/files_compressor_tool_test.py` imported the tool from the wrong path:

```python
from crewai_tools.tools.files_compressor_tool import FileCompressorTool
```

That package's `__init__.py` is **empty** (0 bytes, same on `main`), so the module raised `ImportError` at collection and **all 15 tests in it were silently not executing**. Reproduced on a clean `main`:

```
ERROR tests/tools/files_compressor_tool_test.py - ImportError while importing...
```

The import now uses the same path `crewai_tools/__init__.py` uses. This is the more important half of the fix — the file has been dead weight.

The second half of why it went unnoticed: every other test in the module mocks `zipfile.ZipFile`, so even once imported, the archive contents are never inspected. The three new tests drive the real `zipfile` path and assert on the resulting member list.

## Verification

- `uv run --directory lib/crewai-tools pytest tests/tools/files_compressor_tool_test.py -q` → `18 passed` (15 pre-existing + 3 new).
- The three new tests fail on unpatched `main` (`3 failed, 15 passed`) with the exact symptom from the issue:

```
E   AssertionError: assert ['bundle.zip', 'payload.txt'] == ['payload.txt']
E   AssertionError: assert 'nested/bundle.zip' not in ['payload.txt', 'nested/bundle.zip']
```

- Coverage added: the reported case, a nested output directory, and a positive control proving an unrelated `existing.zip` in the tree is still archived.
- `ruff check` clean on both changed files.

## Additional context

Worth a maintainer's glance: `crewai_tools.tools.<name>` packages generally have empty `__init__.py` files and are imported through the full module path. If other test modules use the short path, they will be silently dead the same way — I only audited this one.


Vidit-Ostwal (commented):
The existing realpath guard fixes the ordinary self-inclusion case, but it does not provide filesystem identity: hard-link aliases have different real paths while sharing the output inode. More importantly, checking identity after opening the archive is too late for overwrite=True, because opening a hard-linked output truncates the source file first. Please detect overlap before opening ZipFile (for example, by comparing device/inode or using samefile for existing paths), reject an output that aliases any file within the input tree, and add regression coverage for a normal in-tree overwrite plus the hard-link case. This also addresses the open CodeRabbit finding.

gaoanze888 (changes requested):
The ZIP fix itself looks right, but the new regressions still are not exercised by the normal suite. This file remains `files_compressor_tool_test.py`, while the root pytest config only collects `test_*.py`. Running `pytest lib/crewai-tools/tests/tools --collect-only` skips it; passing the file explicitly is why the reported 18 tests run. Could you rename it to `test_files_compressor_tool.py` (or move the cases into a collected test module) and confirm directory-level collection? The current green matrix does not cover this behavior.

wangtaotaotao95:
Both reviews addressed in `0eae97481`.

## `@gaoanze888` — the regressions were not actually collected

You're right, and I verified it rather than assuming: the root config sets `python_files = "test_*.py"`, so `pytest lib/crewai-tools/tests/tools --collect-only` never saw `files_compressor_tool_test.py`. My "18 passed" came entirely from naming the file explicitly, which is exactly the hole you described.

Renamed to `test_files_compressor_tool.py`. Directory-level collection now picks it up:

```
$ uv run pytest tests/tools --collect-only -q | grep -c files_compressor
21
```

One thing worth surfacing while I was in there: this is not specific to my file. **20 of the 33 modules in `lib/crewai-tools/tests/tools/` use the `*_test.py` suffix** and are therefore not collected by a directory-level run either — the same silent-dead-test shape that let the original bug ship. Fixing all of them is wider than this PR and would bury the fix, so I left them alone; flagging it here as a separate concern. There is also a hard collection error in `tests/tools/test_oxylabs_tools.py` that interrupts a bare `pytest tests/tools` run.

## `@Vidit-Ostwal` — identity, not paths, and before the archive is opened

Also correct, and the failure is worse than a missed exclusion — it is silent data loss. With the output hard-linked to a file inside the input tree, the previous guard could not see it (different path, same inode), and `ZipFile(output_path, "w")` truncated the shared inode before anything was read:

```text
before: payload.txt = 'important data'
after : payload.txt = 254 bytes, starts with b'PK\x03\x04'
result: Successfully compressed ...
```

The guard now runs before `ZipFile` touches the output and compares filesystem identity (`os.path.samefile`, i.e. device/inode) over the input tree. It distinguishes three cases deliberately:

- **output aliases an existing input file** (hard link, or a file compressed onto itself) → **rejected** with an error naming both paths;
- **new output inside the tree** → still excluded from the walk (the originally reported bug) rather than rejected, so `input_path=".", output_path="bundle.zip"` keeps working;
- **replacing an archive that already sits in the tree** → an ordinary overwrite, not a self-alias.

Rejections now surface as `Error: <reason>` instead of the generic "unexpected error occurred" wrapper, since it is a usage error rather than a compression failure.

Regression coverage added, both cases you asked for plus one:

| Test | Case |
|---|---|
| `test_zip_overwrites_an_archive_that_already_sits_in_input` | normal in-tree overwrite (must **not** be rejected) |
| `test_zip_output_hard_linked_to_an_input_file_is_rejected` | hard link — also asserts the source file is still intact afterwards |
| `test_zip_input_and_output_same_file_is_rejected` | a file compressed onto itself |

**Verification:** 21 passed with the fix; the 6 new tests fail on unpatched `main`, the hard-link one reporting `Successfully compressed` for a run that destroyed its input. This also covers the open CodeRabbit finding on line 126 — same root (identity vs path), fixed before the open rather than after.

wangtaotaotao95:
Correct, and the hole was real — the `realpath` comparison exempted too much. A symlink in the tree pointing at the output has a different path but the same `realpath`, so it was skipped before `samefile` ever ran, and opening the output truncated the symlink's target.

Confirmed against the revision you reviewed (`0eae97481`) with `data.zip` plus `alias.zip -> data.zip` and the output set to `alias.zip`:

```text
E   assert 'Successful' not in "Successfully compressed '.../test_zip_output_symlinked_from0' into '.../test_zip_output_symlinked_from0/data.zip'"
```

Note the message: it reports writing to `data.zip`. The run succeeded, the input file was truncated, and nothing surfaced.

Fixed in `fa4067e35` by exempting the output's own path **lexically** rather than by what it resolves to:

```python
output_abs_path = os.path.abspath(output_path)
...
if os.path.abspath(candidate) == output_abs_path:
    # Overwriting the archive that is already there is an ordinary overwrite.
    continue
if _aliases(candidate):     # os.path.samefile -> device/inode
    raise ValueError(...)
```

So a distinct symlink or hard-link alias in the tree now reaches the identity check and is rejected, while replacing an archive that sits at the output path is still an ordinary overwrite, and a brand-new output inside the tree is still excluded rather than rejected.

`test_zip_output_symlinked_from_input_is_rejected` covers it, including that the symlink's target survives the attempt; it skips itself where symlinks are unavailable rather than failing. `22 passed`, and directory-level collection still picks the module up.

gaoanze888 (changes requested):
The renamed test module is now collected, and the current 23 tests pass. I found one remaining alias case on `787001636ff4`, though: if the caller supplies an output symlink outside the input tree that points to a file inside it, `validate_file_path(output_path)` resolves the symlink before `_reject_output_aliasing_input()` runs. The guard then sees only the in-tree target path and exempts it as the output's own lexical path.

Repro: `source/payload.zip` plus `outside/bundle.zip -> source/payload.zip`, then `_run("source", "outside/bundle.zip", overwrite=True)` returns success and replaces `payload.zip`.

Could you preserve/check the caller's output identity before canonicalization, or otherwise reject this case before opening the archive, and add the external-symlink regression? The normal overwrite, hard-link, and in-tree symlink coverage otherwise looks good.

wangtaotaotao95:
Reproduced, and thanks for working the chain — this one is a direct consequence of the previous fix, which is worth stating plainly.

You're right about the mechanism: `validate_file_path` returns `os.path.realpath(path)`, so by the time `_reject_output_aliasing_input` runs, the caller's output has **already** been rewritten to its in-tree target. My "exempt the output's own path" comparison was therefore applied to the resolved path, and the in-tree target looked like the output the caller named. Confirmed on `787001636` with your repro:

```text
before: payload.zip members = ['inner.txt']
after : payload.zip members = []
result: Successfully compressed ...
```

Fixed in `480a22c62` by keeping both paths: the caller's spelling and the validated one.

```python
requested_output_path = output_path          # before validate_file_path resolves it
output_path = validate_file_path(output_path)
...
self._reject_output_aliasing_input(input_path, output_path, requested_output_path)
```

The exemption now compares against `requested_abs_path`, while the identity checks (`os.path.samefile`) still use the validated path — so the caller naming `outside/bundle.zip` no longer exempts `source/payload.zip`, and it reaches the `samefile` check and is rejected.

`test_zip_output_symlink_outside_input_pointing_into_it_is_rejected` covers it, asserting the in-tree file survives; it skips where symlinks are unavailable. All four alias shapes now have coverage and all pass:

| shape | expected |
|---|---|
| output named at an in-tree archive (ordinary overwrite) | allowed |
| hard link to an in-tree file | rejected |
| in-tree symlink to another in-tree file | rejected |
| **out-of-tree symlink into the tree** | rejected |

`24 passed`, `ruff check` clean. Fair warning that this is the third round on the same predicate — each tightening introduced the next edge, so if you can see a fourth shape that slips through, I would rather write it as a targeted test now than have you find it.

wangtaotaotao95:
Updated the branch to current main (0374c63129). The feature diff remains limited to FileCompressorTool and its tests.

The refreshed matrix currently fails in unrelated Azure Responses tests under tests/llms/azure/test_azure_responses.py (mock_cls.call_args is None); the same base-only failure shape also appears on #7525 after updating that independent PR to the same main. Lint, type checks, CodeQL, pip-audit, and CodeRabbit pass. I am leaving the branch on current main rather than adding an out-of-scope Azure test change here.

@gaoanze888, the requested alias cases and the later Windows normcase finding are addressed on the current head; a re-review would be appreciated.

gaoanze888 (approved):
Re-verified at exact head 6d86dfbca4d1.

The external-symlink data corruption I flagged is fixed. Counterexamples re-run:

- `outside/bundle.zip -> source/payload.zip` (zip): rejected with "Output ... is the same file as ... inside the input" and the source stays intact.
- `outside/bundle.tar.gz -> source/payload.tar.gz` (tar.gz): same rejection, source intact.

The mechanism is correct: `requested_output_path` is captured before `validate_file_path` resolves symlinks, and `_reject_output_aliasing_input` compares filesystem identity (`os.path.samefile`) over the input tree while exempting the caller's requested output lexically (not via `realpath`). It now runs for every format, not just zip.

Also confirmed the test file is now `test_files_compressor_tool.py` (collected by pytest) and the focused suite passes 25/25. Thanks for working through the chain.

wangtaotaotao95:
@coderabbitai review

wangtaotaotao95:
@coderabbitai review

wangtaotaotao95:
Addressed the latest review findings in `a687d20f`:

- Output opening now pins every parent directory with descriptor-relative `O_NOFOLLOW` opens before opening the final component, so a symlink or parent-directory substitution cannot redirect the later truncate/write.
- Added regression coverage for both final-component and parent-directory swaps.
- TAR tests now spy on `tarfile.open` and assert that every TAR format receives the exact file object yielded by `_open_validated_output`, while retaining archive-content checks.

Validation: 32 focused tests pass; Ruff and mypy pass.

@coderabbitai review

wangtaotaotao95:
Addressed the incremental review note in `HEAD`: parent-directory descriptors now use `O_PATH` when the platform provides it, with the existing `O_RDONLY` fallback elsewhere. This preserves `O_DIRECTORY | O_NOFOLLOW` while allowing traversal through execute-only directories. Revalidated 32 focused tests, Ruff, and mypy.

@coderabbitai review

wangtaotaotao95:
Branch updated again to current `main`. The refreshed CI failure is still outside this PR: the Python 3.10 core suite failed in three Azure Responses tests because the mocked client was never called (`test_responses_base_url_preserves_port`, `test_delegate_receives_optional_params`, and `test_responses_base_url_from_base_endpoint`), after 451 tests passed in that shard. This PR only changes `FileCompressorTool`, and the focused compressor suite passes locally: 32/32 tests. I do not have repository Actions admin rights to rerun the failed matrix; could a maintainer rerun it? Latest failing run: https://github.com/crewAIInc/crewAI/actions/runs/35945383715

PR #7544, opened 2026-09-17
fix(memory): keep lancedb scope filters inside path boundaries

### LINKED ISSUE #7543: [BUG] [BUG] LanceDBStorage scope-prefix filters cross path boundaries: `/app` matches `/apple`
### Description

`LanceDBStorage` treats memory scopes as hierarchical paths (`/crew/sales/agent`),
and the documented contract for a scoped view is "root path and below". But the
SQL predicates used for scope filtering on the read/query paths are naive
string prefixes:

- `search()` builds `scope LIKE '<prefix>%'`
- `_scan_rows()` builds `scope LIKE '<prefix>%'` — shared by `list_records()`,
  `get_scope_info()`, `list_categories()`, `count()`, and the
  categories/metadata branch of `delete()`

`LIKE '/app%'` matches not only `/app` and `/app/sub` but also the sibling scope
`/apple` (and `/appendix`, `/app2`, ...). Two consequences:

1. **Cross-scope leakage on reads.** `search`, `list_records`, `get_scope_info`,
   `count`, and `list_categories` for scope `/app` all include records stored
   under `/apple`. `Memory.recall(query, scope="/app")`, a `MemoryScope` rooted
   at `/app`, and a `MemorySlice` containing `/app` can therefore surface
   memories belonging to a different agent/crew scope — the exact isolation the
   scope mechanism exists to provide.
2. **Over-deletion on filtered deletes.** `delete(scope_prefix="/app",
   categories=[...])` scans `/apple` rows too and deletes matching records in
   the sibling scope.

The inconsistency is visible inside the same file: `reset()` already uses a
path-boundary-aware predicate (`scope >= '<prefix>' AND scope <
'<prefix>/\uffff'`), and the Qdrant edge backend filters on a `scope_ancestors`
payload field, so both correctly exclude `/apple` when operating on `/app`.
Only the LIKE-based paths are wrong.

A secondary inconsistency: `search()`/`_scan_rows()` never normalize a missing
leading slash (`scope_prefix="app"` produces `LIKE 'app%'`, which matches
nothing because every stored scope starts with `/`), while `delete()` and
`get_scope_info()` do normalize. The unescaped prefix is also interpolated raw
into the SQL string, so a scope containing `'` breaks the query and `%`/`_` act
as LIKE wildcards.

### Steps to Reproduce

```python
import tempfile

s = LanceDBStorage(path=tempfile.mkdtemp(), vector_dim=4, compact_every=0)
s.save([
    MemoryRecord(content="app",   scope="/app",     embedding=[0.0]*4),
    MemoryRecord(content="child", scope="/app/sub", embedding=[0.0]*4),
    MemoryRecord(content="sib",   scope="/apple",   embedding=[0.0]*4),
    MemoryRecord(content="root",  scope="/",        embedding=[0.0]*4),
])

print(sorted(r.scope for r, _ in s.search([0.0]*4, scope_prefix="/app")))
# ['/app', '/app/sub', '/apple']   <- expected ['/app', '/app/sub']
print(s.count("/app"))             # 3  <- expected 2
s.delete(scope_prefix="/app", categories=["mine"])
# also deletes the matching /apple record  <- expected: /apple untouched
```

The regression tests in
`lib/crewai/tests/memory/test_lancedb_scope_prefix.py` reproduce all of these
against a real LanceDB instance — no network or API keys required.

### Expected behavior

A scope filter for `/app` matches `/app` and its descendants (`/app/...`) only, never a sibling scope such as `/apple` — the same boundary semantics `LanceDBStorage.reset()` and the Qdrant backend already use.

### Screenshots/Code snippets

`lib/crewai/src/crewai/memory/storage/lancedb_storage.py` (upstream/main):

- `LanceDBStorage.search` — lines 389–392 (`scope LIKE '{like_val}'`)
- `LanceDBStorage._scan_rows` — lines 489–490 (`scope LIKE '{...}%'`), consumed
  by `list_records` (line 509), `get_scope_info` (line 519), `list_categories`
  (line 585), `count` (line 602 via `get_scope_info`), `list_scopes` (line 573),
  and the categories/metadata branch of `delete` (line 430).
- Note: `delete()`'s scope/older_than-only branch (line 456) has the same class
  of defect plus an extra `OR scope = '/'` clause; it is already covered by
  issue #7419 / PR #7472 and intentionally left out of this fix.

### Operating System

Ubuntu 22.04

### Python Version

3.12

### crewAI Version

1.15.22 (main @ 5c33fe4)

### crewAI Tools Version

1.15.22

### Virtual Environment

Venv

### Evidence

- Any scoped read leaks sibling-scope records: wrong results from
  `Memory.recall(scope=...)` (shallow and deep, via `scope_prefix`), inflated
  `get_scope_info`/`count` numbers, wrong `list_categories` counts.
- `MemoryScope`/`MemorySlice` views are documented as restricting visibility to
  "the root path and below"; they don't.
- Scoped category/metadata deletes remove records from sibling scopes.

### Possible Solution

Add one shared predicate builder that expresses "scope equals prefix or scope
is under prefix/":

```python
@staticmethod
def _scope_filter_sql(scope_prefix: str | None) -> str | None:
    if scope_prefix is None or not scope_prefix.strip("/"):
        return None
    if not prefix.startswith("/"):
        prefix = "/" + prefix
    prefix = prefix.replace("'", "''")
    return f"(scope = '{prefix}' OR starts_with(scope, '{prefix}/'))"
```

and use it in `search()` and `_scan_rows()` in place of the `LIKE 'prefix%'`
clauses. `starts_with` avoids LIKE wildcard interpretation of `%`/`_` in scope
names, and the literal is quote-escaped. This matches the semantics `reset()`
already implements via its range predicate and the Qdrant backend implements
via `scope_ancestors`. `delete()`'s dedicated scope/older_than branch is fixed
by open PR #7472.

### Additional context

Found and written with an AI coding agent (Devin, reviewed by Claude Code); please apply the `llm-generated` label. Related but separate: #7419 / #7472 cover the scope/older_than branch of `delete()`, which this report does not touch. A fix with regression tests is ready and will be linked here.

coderdailyone:
## Related issue

Fixes #7543

## Summary

`LanceDBStorage` filtered scopes with `scope LIKE '<prefix>%'` on the
read/query paths. Because the filter is a raw string prefix, a scoped lookup
for `/app` also matched the sibling scope `/apple` — and likewise `/appendix`,
`/app2`, etc. Concretely:

- `search(scope_prefix="/app")` returned `/apple` records, so
  `Memory.recall(scope=...)`, `MemoryScope` views, and `MemorySlice` views
  leaked memories across scope boundaries — the exact isolation the scope
  mechanism is for.
- `list_records`, `get_scope_info`, `list_categories`, and `count` (all via
  `_scan_rows`) included sibling-scope records, so `info("/app")` over-counted
  and reported sibling categories.
- `delete(scope_prefix="/app", categories=[...])` scanned sibling rows and
  deleted matching records in `/apple`.
- `search`/`_scan_rows` also skipped the leading-slash normalization that
  `delete`/`get_scope_info` apply, interpolated the prefix into SQL without
  escaping quotes, and let LIKE wildcards (`%`, `_`) inside scope names act as
  pattern characters.

The fix adds a shared `_scope_filter_sql()` helper that emits
`(scope = '<prefix>' OR starts_with(scope, '<prefix>/'))` — exact match plus
descendants only, with the prefix normalized and quote-escaped — and uses it in
`search()` and `_scan_rows()`. This matches the semantics `reset()` already
implements with its `scope >= 'p' AND scope < 'p/\uffff'` range predicate and
the Qdrant edge backend implements via its `scope_ancestors` payload field.
`list_scopes` output is unchanged (it post-filters children by
`startswith(prefix + '/')`).

`delete()`'s scope/older_than-only branch (which additionally contains an
`OR scope = '/'` clause) is not touched here — it is already addressed by open
issue #7419 / PR #7472. This change does fix the categories/metadata delete
branch because it routes through `_scan_rows`.

## Verification

New file `lib/crewai/tests/memory/test_lancedb_scope_prefix.py` (9 tests, real
LanceDB on tmp_path, no network/API keys). With the source change reverted and
the tests kept, these fail:

- `test_search_scope_prefix_excludes_sibling_scope` — `{'/app', '/app/sub', '/apple'}`
- `test_list_records_scope_prefix_excludes_sibling_scope` — same leak
- `test_scope_info_excludes_sibling_scope` — `record_count` 3 instead of 2
- `test_count_excludes_sibling_scope` — `count("/app")` 3 instead of 2
- `test_list_categories_excludes_sibling_scope` — `{'mine': 3}` instead of `{'mine': 2}`
- `test_scoped_category_delete_excludes_sibling_scope` — deleted 3 instead of 2
- `test_scope_prefix_without_leading_slash_is_normalized` — `"app"` matched nothing

The two controls (`test_root_scope_prefix_matches_everything`,
`test_list_scopes_still_returns_children_only`) pass before and after.

Local checks (Python 3.13):

- `uv run pytest lib/crewai/tests/memory/ -x -q` — 143 passed, 19 skipped
  (optional Qdrant edge tests).
- `uv run ruff check` / `ruff format --check` on the changed source file — clean
  (tests are outside the ruff include paths).
- `uv run mypy lib/crewai/src/crewai/memory/storage/lancedb_storage.py` — clean.

## Additional context

This change was written with an AI coding agent (Devin, reviewed by Claude
Code). Please apply the required `llm-generated` label; external contributors
cannot apply repository labels.

https://claude.ai/code/session_01CRrBenkCCr4pacMNwR4NuG


gaoanze888 (approved):
Reviewed at exact head eadd24052617.

The boundary fix is correct and well-tested. The old `scope LIKE 'prefix%'` both over-matched siblings (`/app` matching `/apple`) and under-specified the path boundary; `scope = 'prefix' OR starts_with(scope, 'prefix/')` pins it exactly. The new tests exercise search/list/get_scope_info/count/list_categories/delete against a real `LanceDBStorage`, plus no-leading-slash normalization and root, so `starts_with` support and semantics are verified.

One point worth naming: this also closes a latent SQL-injection vector. `normalize_scope_path` does not strip `'`, so a caller-supplied scope reached the old `f"scope LIKE '{like_val}'"` unescaped. The `replace("'", "''")` here is the actual defense, not just belt-and-suspenders.

Approving.

PR #7548, opened 2026-09-17
fix(tools): preserve directory listing paths

### LINKED ISSUE #7547: [BUG] DirectoryReadTool corrupts paths that repeat the base directory
### Description

`DirectoryReadTool` can report a path that does not exist when a descendant path repeats the validated base-directory string.

The tool walks the correct directory, but formats each result by calling `full_path.replace(directory, "")`. `str.replace()` removes every occurrence, not only the leading prefix. A repeated occurrence deeper in the path is therefore deleted as well.

### Steps to Reproduce

```python
from pathlib import Path
from tempfile import TemporaryDirectory

from crewai_tools import DirectoryReadTool

with TemporaryDirectory(dir=".") as tmp:
    base = Path(tmp).resolve()
    repeated = base / base.relative_to(base.anchor)
    repeated.mkdir(parents=True)
    target = repeated / "sentinel.txt"
    target.write_text("content", encoding="utf-8")

    result = DirectoryReadTool()._run(directory=str(base))
    print("actual:", target)
    print(result)
```

The actual file is under the repeated descendant path, but the reported path removes that repeated segment and points to a nonexistent file.

### Expected behavior

Every path returned by `DirectoryReadTool` should be the actual path found by `os.walk()` and should exist at the time it is listed.

### Screenshots/Code snippets

The current path formatting uses unrestricted replacement:

```python
os.path.join(root, filename).replace(directory, "")
```

### Operating System

macOS (15.5)

### Python Version

3.13

### crewAI Version

Current `main` at `5c33fe4c713a`.

### crewAI Tools Version

1.15.22

### Virtual Environment

Venv (`uv`)

### Evidence

On current `main`, the reproducer finds an existing `sentinel.txt` but returns a shorter path for which `Path.exists()` is false.

The directory has already been canonicalized by `validate_directory_path()`, and `os.walk()` already returns the correct root. The corruption happens only while reconstructing the output string.

Targeted issue and PR searches for `DirectoryReadTool`, incorrect/nonexistent paths, and the affected expression found no matching open work.

### Possible Solution

Return `os.path.join(root, filename)` directly. This preserves the existing absolute-path output and platform-native separators without performing a lossy string rewrite. Add a behavior test whose descendant repeats the canonical base path.

### Additional context

AI-assisted report and proposed fix. Please apply the required `llm-generated` label; external contributors cannot add repository labels directly.

gaoanze888:
## Related issue

Fixes #7547

## Summary

`DirectoryReadTool` walked the correct files but rebuilt each path with an unrestricted `str.replace(directory, "")`. If the base-directory string appeared again in a descendant path, that second occurrence was removed and the tool returned a plausible but nonexistent path.

The tool now returns the path already supplied by `os.walk()` via `os.path.join(root, filename)`. This keeps the existing absolute-path output while avoiding lossy string rewriting and using the host platform's path separator.

## Verification

- [x] Added a behavior test with a descendant path that repeats the canonical base path
- [x] The regression fails on unpatched `main` and passes with this change
- [x] Ruff check and format check pass for both changed files
- [x] `compileall` and `git diff --check` pass
- [x] Independent review found no correctness, compatibility, or security blocker

The full workspace environment could not be installed locally because `uv` repeatedly timed out downloading the large optional dependency set. CI provides the repository's Python-version test and type-check matrices.

## Additional context

AI-assisted contribution. Please apply the required `llm-generated` label; external contributors cannot add repository labels directly.


Vidit-Ostwal (approved):
LGTM

PR #7560, opened 2026-09-17
fix(llm): handle None, dict, and BaseModel in structured output validation (#7540)

### LINKED ISSUE #7540: [BUG] structured output validation crashes on None or dict content
### Description

`BaseLLM._validate_structured_output` calls `response.strip()` without checking the type. Native structured output can already be a dict, and some providers pass content=None. Both raise AttributeError before schema validation.

### Steps to Reproduce

```python

class Out(BaseModel):
    city: str

BaseLLM._validate_structured_output(None, Out)
BaseLLM._validate_structured_output({"city": "London"}, Out)
```

### Expected behavior

None is treated as empty. A dict is validated directly. Strings keep the current JSON parse path.

### Screenshots/Code snippets

`AttributeError: 'NoneType' object has no attribute 'strip'`

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

main

Ritiky23:
### Summary
Fixes #7540 by allowing `BaseLLM._validate_structured_output` to handle `dict`, `BaseModel`, and `None` inputs without raising an unhandled `AttributeError`.

### Problem
Previously, `BaseLLM._validate_structured_output` called `response.strip()` unconditionally. When providers or internal methods passed a native parsed `dict` (e.g., from tool calls or native structured output) or `None` content, an `AttributeError: 'NoneType'/'dict' object has no attribute 'strip'` was raised before schema validation could occur.

### Solution
- Handle `response is None` cleanly by raising `ValueError` when a `response_format` is expected, or returning `None` when `response_format is None`.
- Directly validate dictionary and existing `BaseModel` payloads using `response_format.model_validate(...)`.
- Retain JSON decoding and markdown JSON extraction regex for raw string outputs.
- Added comprehensive unit tests in `test_structured_output_validation.py`.

### Tests
- Ran `pytest lib/crewai/tests/llms/test_structured_output_validation.py` (9 passed).
- Ran `pytest lib/crewai/tests/llms/test_structured_output_multimodal.py` (2 passed).


gaoanze888 (approved):
Reviewed at exact head (fixes #7540).

The fix is correct and closes the reported AttributeError: `None` now raises a clear `ValueError` ("response is None"), `BaseModel` instances pass through or convert via `model_validate(model_dump())`, and `dict`/`list`/`tuple` go straight to `model_validate`. The string path is preserved and the `int|float|None -> float|None` narrowing is harmless (int remains assignable).

Tests cover the key cases (dict success/invalid, None with/without format, same-type passthrough, cross-type conversion, JSON string, embedded JSON, invalid string). The `logging.warning -> logger.warning` switch is a tidy improvement.

One minor non-blocking note: the trailing `not isinstance(response, str)` branch is a catch-all that partially overlaps the dict/list/tuple branch and will produce a generic validation failure for non-JSON-able types — fine defensively, just slightly redundant. Approving.

PR #7566, opened 2026-09-18
fix(guardrails): enforce result outcome invariant

### LINKED ISSUE #7565: [BUG] GuardrailResult accepts payloads for both outcomes
### Description

`GuardrailResult` does not reliably enforce the outcome invariant documented by its fields: a successful result may also carry an error, and a failed result may also carry a result payload.

The validator runs separately for `result` and `error`. Because Pydantic validates fields in declaration order, `error` is not available while `result` is validated, so `success=True, result=..., error=...` is accepted. The validator also uses truthiness, allowing contradictory falsey payloads such as `error=""` or `result=0`.

### Steps to Reproduce

```python
from crewai.utilities.guardrail import GuardrailResult

print(GuardrailResult(success=True, result="accepted", error="rejected"))
print(GuardrailResult(success=True, result="accepted", error=""))
print(GuardrailResult(success=False, result=0, error="rejected"))
```

All three currently construct successfully.

### Expected behavior

A successful `GuardrailResult` must not populate `error`, and a failed result must not populate `result`. Presence should be checked with `is not None`, so falsey values on the selected side remain valid.

The existing allowed states should remain valid, including:

```python
GuardrailResult(success=True, result="")
GuardrailResult(success=True, result=None)
GuardrailResult(success=False, error="")
GuardrailResult(success=False, error=None)
```

### Screenshots/Code snippets

The current field validator depends on `info.data` and truthiness:

```python
@field_validator("result", "error")
def validate_result_error_exclusivity(cls, v, info):
    ...
```

### Operating System

macOS (15.5)

### Python Version

3.13

### crewAI Version

Current `main` at `5c33fe4c713a`.

### crewAI Tools Version

1.15.22

### Virtual Environment

Venv (`uv`)

### Evidence

The validator has existed since `GuardrailResult` was introduced. Repository production paths construct results through `from_tuple()`, which already populates only the selected side, but `GuardrailResult` is a public model and its advertised validation invariant does not hold for direct construction or model validation.

Targeted searches for `GuardrailResult`, result/error exclusivity, and the validation message found no matching open issue or pull request. Existing LLM guardrail issues concern provider exceptions and retry behavior, not this model invariant.

### Possible Solution

Replace the order-dependent field validator with an `after` model validator that rejects only the field opposite to `success` when it is not `None`. Add regressions for ordinary and falsey contradictions plus `from_tuple()` compatibility.

### Additional context

AI-assisted report and proposed fix. Please apply the required `llm-generated` label; external contributors cannot add repository labels directly.

gaoanze888:
## Related issue

Fixes #7565

## Summary

`GuardrailResult` used a field validator to enforce the relationship between `success`, `result`, and `error`. That validator depended on field order and truthiness, so it accepted successful results carrying errors and falsey contradictory payloads.

The invariant now runs after the whole model is available:

- successful results reject any non-`None` error;
- failed results reject any non-`None` result;
- the selected side may still be falsey or `None`, preserving existing Agent, LiteAgent, and `from_tuple()` behavior.

## Verification

- [x] 10 focused invariant and factory cases pass
- [x] The same regression suite against unpatched `main` fails on three contradictory payloads
- [x] Ruff check and format check pass
- [x] `compileall` and `git diff --check` pass
- [x] Two independent reviews found no correctness or compatibility blocker

The full workspace environment could not be installed locally because `uv` repeatedly timed out downloading the large optional dependency set. CI provides the repository's Python-version test and type-check matrices.

## Additional context

AI-assisted contribution. Please apply the required `llm-generated` label; external contributors cannot add repository labels directly.


sattyamjjain:
Tested this independently at head ede0ed5.

On main, all three repros from #7565 construct without error. On this PR, all three raise ValidationError. The four allowed states from the issue still construct, and from_tuple works for (True, "x"), (False, "err"), (True, ""), and (False, "").

I also grepped main (6536114) for direct `GuardrailResult(...)` calls outside tests. There are none; everything goes through from_tuple, so the stricter check shouldn't break an internal path.

Caveat: I tested the class in isolation on pydantic 2.13.5, not the full crewAI suite.

PR #7572, opened 2026-09-18
fix(llm_overlay): a role and a key that differ only by surrounding whitespace match

joaomdmoura:
## Why

Roles declared in YAML often end in a newline (`role: >` folds to one). `llm_overlay` matched roles by exact text, so a mapping written for `Researcher` never reached an agent whose role is `Researcher\n`: the agent silently kept its declared `llm`. crewai-eval's mode 2 hit this on a production flow: two of its five agents (templated CrewBase roles from YAML) could only be reached by reproducing the newline inside the configuration key.

## What

- `llm_overlay(mapping)` sets a copy of the mapping with the whitespace around each role dropped; the caller's dict is not touched.
- `overlay_model_for(role)` strips `role` before the lookup; an empty or `None` role matches nothing. Case and the text in between are still matched exactly — no folding, no patterns.
- The module docstring, `llm_overlay`'s and `overlay_model_for`'s say so.

The other readers (`Agent.post_init_setup`, `Agent.interpolate_inputs`, `LiteAgent.setup_llm`) already go through `overlay_model_for`, so nothing else changes.

## Tests

`lib/crewai/tests/test_llm_overlay.py`: key `Researcher` matches `Researcher\n` and `  Researcher ` on the direct lookup and on an `Agent` build; a key carrying the newline matches a clean role; `researcher`, `Re searcher`, `Writer\n`, `""`, `" \n"` and `None` return `None`; the mapping passed in is unchanged after the block; a YAML-folded template role `Researcher for {repo}\n` matches after `interpolate_inputs`, which keeps the newline. 43 passed together with `test_create_llm_like.py`.

Companion on the library side: crewAIInc/crewai-evals#31 (the same rule wherever a role enters a trace or a configuration is matched).

---

> [!NOTE]
> **Low Risk**
> Localized matching change in overlay lookup with explicit ambiguity checks; no auth or data-path changes, though misconfigured duplicate keys now fail fast instead of picking an arbitrary model.
> 
> **Overview**
> **`llm_overlay` role matching now ignores surrounding whitespace** on both mapping keys and lookup roles, fixing cases where YAML-folded roles (e.g. trailing `\n`) never matched keys written for the clean text.
> 
> `llm_overlay()` stores a **normalized copy** of the caller’s dict (keys stripped; original dict unchanged). `overlay_model_for()` strips the role, treats `None`/blank as non-matches, and still requires exact case and inner text. **`_stripped()`** collapses whitespace-equivalent keys to one entry and **raises `ValueError`** if they map to different models so resolution never depends on dict order. Docs are updated to describe the behavior.
> 
> Agent/LiteAgent LLM resolution already calls `overlay_model_for`, so behavior changes are centralized. Tests cover symmetric key/role whitespace, negative cases, duplicate-key errors, non-mutation, and templated YAML roles after `interpolate_inputs`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db62e701f7f5bea67830b4bc240b610dd88a44d2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/crewai/src/crewai/llm_overlay.py
```diff
@@ -71,22 +73,38 @@ def llm_overlay(mapping: dict[str, str] | None) -> Iterator[None]:
...
+        return None
+    role = role.strip()
+    return mapping.get(role) if role else None
+
+
+def _stripped(mapping: dict[str, str] | None) -> dict[str, str] | None:
+    """A copy of ``mapping`` with the whitespace around each role dropped."""
+    if mapping is None:
+        return None
+    return {role.strip(): model for role, model in mapping.items()}
```
iris-clawd: Can we reject duplicate normalized roles instead of silently letting the later entry win? Otherwise, `{"Researcher": "a", " Researcher ": "b"}` selects a provider or model based on dictionary order.

lib/crewai/tests/test_llm_overlay.py
```diff
@@ -85,6 +85,52 @@ def test_nested_overlay_restores_the_outer_one() -> None:
         assert overlay_model_for("Researcher") == "openai/gpt-4o"
 
 
+def test_whitespace_around_a_role_or_a_key_is_ignored() -> None:
```
iris-clawd, line 88: Can we add a regression for duplicate normalized keys that pins the rejection contract? Without it, ambiguous configuration can regress to silently selecting the later model.

iris-clawd (dismissed):
The required checks are still pending. I could not verify the companion crewai-evals change.

The notes inline on llm_overlay.py and test_llm_overlay.py are what I'd want addressed before this merges.

Someone else should sign off on this too: normalizing overlay keys changes the matching contract: previously distinct keys such as `"Researcher"` and `" Researcher "` now collapse, with dictionary insertion order deciding which model wins. Worth a look even though the branch rules don't require them: pip-audit are failing.

lib/crewai/src/crewai/llm_overlay.py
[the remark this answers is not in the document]
joaomdmoura: Agreed — fixed in db62e701f. `_stripped` now refuses a mapping that names one role twice with different models: `{"Researcher": "a", " Researcher ": "b"}` raises `ValueError: llm_overlay: role 'Researcher' is mapped twice with different models ('a' and 'b'); give each role one model` before the overlay is set (nothing is left active). Two keys that collapse to one role with the SAME model are kept once — a harmless duplicate. So the model an agent runs on never depends on dictionary order. Test: `test_two_keys_that_are_one_role_with_different_models_are_refused` pins both cases.

joaomdmoura: Done in db62e701f: duplicates that normalise to one role are refused when their models differ (`ValueError` naming the role and both models, raised before the overlay is set), and collapsed when the model is the same. Pinned by `test_two_keys_that_are_one_role_with_different_models_are_refused`.

lib/crewai/tests/test_llm_overlay.py
joaomdmoura: Added in db62e701f: `test_two_keys_that_are_one_role_with_different_models_are_refused` — same-model duplicates collapse to one entry and resolve; different-model duplicates raise before anything is set, and `active.get()` is still `None` afterwards.

iris-clawd (approved):
Ambiguous mappings fail before changing the active context. I could not inspect the companion crewai-evals change.

Not holding this for it, but worth a ticket so it is not forgotten: This changes the public `llm_overlay` matching contract: surrounding whitespace is newly ignored, blank keys normalize to `""`, and conflicting normalized keys now raise `ValueError`. A match directly changes which model/provider an agent runs, affecting output and billing behavior.

Before merging: a human reviewer has requested changes.

PR #7573, opened 2026-09-18
feat(cli): prioritize popular platform integrations

Vidit-Ostwal:
## Summary

Prioritize commonly used CrewAI Platform integrations in the JSON crew wizard, surface Gmail and WhatsApp in Common tools, and rename the saved Platform credential to the Enterprise Action Auth Token.

## Commit

- `20ac775e4` — prioritize the Platform catalog and update CLI token persistence with legacy-variable fallback.

## Verification

- [x] Focused catalog, picker, token-validation, and integration-client tests
- [x] Pre-commit hooks: Ruff, Ruff format, and mypy


lib/cli/src/crewai_cli/create_json_crew.py
```diff
@@ -1178,11 +1180,11 @@ def create_json_crew(
     copy_assistant_instructions(folder_path)
 
     if platform_token:
-        os.environ["CREWAI_PLATFORM_INTEGRATION_TOKEN"] = platform_token
+        os.environ["CREWAI_ENTERPRISE_ACTION_AUTH_TOKEN"] = platform_token
         env_vars = load_env_vars(folder_path)
-        env_vars["CREWAI_PLATFORM_INTEGRATION_TOKEN"] = platform_token
+        env_vars["CREWAI_ENTERPRISE_ACTION_AUTH_TOKEN"] = platform_token
```
iris-clawd, line 1185: Can we update `crewai.context.get_platform_integration_token` to prefer this key and fall back to the legacy key too? Otherwise, generated projects authenticate Platform tools while existing skills and tracing silently miss the credential.

lib/cli/tests/test_create_crew.py
```diff
@@ -1040,7 +1042,7 @@ def test_json_create_saves_platform_token_to_env_file(tmp_path, monkeypatch):
     json_crew.create_json_crew("Platform Crew", skip_provider=True)
 
     env_file = tmp_path / "platform_crew" / ".env"
-    assert "CREWAI_PLATFORM_INTEGRATION_TOKEN=token" in env_file.read_text()
+    assert "CREWAI_ENTERPRISE_ACTION_AUTH_TOKEN=token" in env_file.read_text()
```
iris-clawd, line 1045: Can we add a regression that loads this generated environment through the core token resolver? This assertion confirms persistence, but not that current skills and tracing consumers receive the credential.

iris-clawd (changes requested):
I could not verify the exact AMP settings path or the new catalog entries across customer-managed deployments.

The notes inline on create_json_crew.py and test_create_crew.py are what I'd want addressed before this merges.

Someone else should sign off on this too: it changes the preferred credential/config key written into newly generated projects while retaining the old key only as a read fallback.

lib/cli/src/crewai_cli/create_json_crew.py
Vidit-Ostwal: Resolved in a21ca4fbc by retaining CREWAI_PLATFORM_INTEGRATION_TOKEN for CLI prompts, process environment, and generated .env files. The core resolver already consumes this existing key, so generated projects remain compatible with Skills, MCP, and tracing.

lib/cli/tests/test_create_crew.py
Vidit-Ostwal: Resolved in a21ca4fbc. Generated projects again persist CREWAI_PLATFORM_INTEGRATION_TOKEN, which is the key consumed by the core resolver; the existing persistence regression therefore verifies the compatible credential contract.

lib/cli/src/crewai_cli/create_json_crew.py
```diff
@@ -234,9 +234,9 @@
 _COMMON_TOOL_ORDER = [
     "SerperDevTool",
     "ScrapeWebsiteTool",
-    "DirectoryReadTool",
     "FileReadTool",
-    "FileWriterTool",
+    "platform:gmail",
+    "platform:whatsapp",
```
joaomdmoura, line 239: Are these the only ones that will be show? can we add Exa, msft and daytona as well?

PR #7593, opened 2026-09-18
ensure link is shown after finalizing traces

lorenzejay:
> [!NOTE]
> **Low Risk**
> Changes are mostly post-export console UX and validated optional URLs; grant/export behavior is extended with tests and does not alter core auth or execution logic.
> 
> **Overview**
> After traces are successfully exported, CrewAI now prints a one-time **Execution Traces** panel with the execution UUID and, when AMP provides it, a clickable viewer URL—instead of leaving users to hunt for the ID elsewhere.
> 
> Grant requests (initial and renewals) send **`include_trace_url: true`**, and optional **`trace_url`** values are validated to same-host http(s) URLs without credentials or control characters. **`GrantSpanExporter`** tracks export success/failure and exposes **`show_trace_summary()`**, which is invoked from authenticated shutdown (after **`session.shutdown()`**) and from ephemeral upload completion; output is skipped in TUI mode, when tracing messages are suppressed, on export failure, or before finalization.
> 
> **`pyproject.toml` / `uv.lock`** also pin **`anyio>=4.14.2`** for known security advisories.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a82066a0f3378c5850c3016e88e27a7492485c67. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #7597, opened 2026-09-18
feat(tracing): task spans say the declared output format and what came out, agent spans carry the prompt and answer, tool spans say whether the cache answered

joaomdmoura:
## What was missing

A reader of a run's OTel spans (the trace viewer, or anything else that consumes them) could see:

- a task's raw output, but not the format it declared (`output_json` / `output_pydantic`), nor whether a Pydantic object or a JSON dict actually came out of it;
- an agent's goal, backstory, model and tools, but not the prompt the agent was handed or the answer it gave;
- a tool's arguments and result, but not whether the tool ran or the cache answered.

Each had to be reconstructed from other spans, or could not be known at all. The legacy `trace_events` payload carried the agent prompt and output; the OTel spans did not.

## What the spans now carry

**`execute task`**
- `crewai.task.output_format`: `json`, `pydantic` or `raw`. Read from the declaration on start and on failure, from the `TaskOutput` on completion.
- `crewai.task.output_pydantic_produced`, `crewai.task.output_json_produced` (completion only): what the run actually yielded, so a declared-pydantic task that fell back to raw text is told apart from one that parsed.

**`execute agent`**
- `gen_ai.input.messages`: the task prompt. `gen_ai.output.messages`: the agent's answer. Same spec shape the task span already uses for its own text, through the existing per-attribute byte cap (`CREWAI_OTEL_MAX_ATTR_BYTES`, default 32 KiB); over the cap the payload is bounded and `<attr>.truncated` / `<attr>.original_size_bytes` are set, so a cut is never mistaken for a short text.

**`call tool`**
- `crewai.tool.from_cache`: whether the result was served from the tool cache. New `crewai_tool()` builder in `semantic_conventions.py` next to `crewai_tool_failure()`.

`handle_task_completed` now reads `event.output` (required on `TaskCompletedEvent`) instead of `task.output`, which removed the `if task.output else None` guard.

## Content and redaction

The agent span's prompt and answer are the same text the spans beneath it already export: the `call llm` span carries the whole prompt array (`input_messages=event.messages`, `handlers.py:2033`, unconditional) and the `execute task` span carries the answer as `crewai.task.output` / `gen_ai.output.messages`. No content-capture switch exists in crewAI's tracing for any of it; the only knob is the size cap. Redaction is the caller's by design (`TraceSession` docstring), and the hosted caller's redactor exempts attributes by key name and runs everything else through presidio — so the text is carried under `gen_ai.input.messages` / `gen_ai.output.messages` only, the keys the LLM span already uses, and gets the same pass. `test_the_agent_text_travels_under_the_standard_keys_only` pins that no other attribute carries it. An agent-only opt-in would protect nothing while making every consumer depend on a flag; a content switch worth having covers every message-carrying attribute at once and is its own change.

## Tests

`lib/crewai/tests/telemetry/test_span_task_agent_tool_attributes.py`, 10 cases: pydantic / json / raw task → format and produced flags; declared-pydantic task that produced only raw text; failed task still carries the declared format; agent span carries a 22 KB prompt and the answer verbatim, unmarked; under a 2 KiB cap the prompt is bounded and marked while the short answer is untouched; tool span `from_cache` true / false.

```
pytest lib/crewai/tests/telemetry lib/crewai/tests/tracing lib/crewai/tests/events lib/crewai/tests/utilities/test_events.py
5 failed, 780 passed, 2 skipped
```

The 5 failures are all `test_session_trace_export.py` viewer-link print counts and fail identically on clean `main` (checked with this branch's source files replaced by main's); they come from #7593, not from this change.


lib/crewai/src/crewai/telemetry/tracing/handlers.py
```diff
@@ -1090,6 +1115,7 @@ def handle_agent_execution_started(
             tool_definitions=agent.tools,
             system_instructions=agent.backstory,
         ),
+        **semantic_conventions.gen_ai_io(input_value=event.task_prompt),
```
iris-clawd, line 1118: Can we require explicit content-capture opt-in before adding the agent prompt and answer attributes? Both may contain credentials or customer data, and configured exporters otherwise receive them automatically.

lib/crewai/tests/telemetry/test_span_task_agent_tool_attributes.py
```diff
@@ -0,0 +1,281 @@
...
+        ctx,
+        agent,
+        AgentExecutionCompletedEvent(
+            agent=agent, task=None, output=output, started_event_id=started.event_id
+        ),
+    )
+    return _only_span(exporter, "execute agent")
+
+
+def test_execute_agent_span_carries_the_exact_prompt_and_answer(pipeline) -> None:
```
iris-clawd, line 224: Can we add a regression proving the default or disabled path omits both agent-content attributes? Without it, content capture could become unconditional again without a failing test.

iris-clawd (dismissed):
I could not verify private hosted or Enterprise redaction behavior.

The notes inline on handlers.py and test_span_task_agent_tool_attributes.py are what I'd want addressed before this merges.

Someone else should sign off on this too: agent prompts and answers may contain credentials, personal data, or proprietary customer content.

lib/crewai/src/crewai/telemetry/tracing/handlers.py
joaomdmoura: Looked into this before changing anything. Three facts:

1. **No new class of content leaves the process.** The `call llm` span under the agent already exports the whole prompt array — `input_messages=event.messages`, `handlers.py:2033`, unconditional — and the agent's task prompt is inside its first call's user message. The answer is already on the `execute task` span as `crewai.task.output` and `gen_ai.output.messages`. crewAI's tracing has no content-capture switch for any of that today; the only knob is the size cap `CREWAI_OTEL_MAX_ATTR_BYTES`.

2. **Redaction is the caller's by design** — `TraceSession`'s docstring: "Exporters and redaction processors are supplied by the standalone or hosted caller." The hosted caller's redactor (`PIIRedactor._redact_mapping`, crewai-enterprise) exempts attributes by key — a protected-key set, the `gen_ai.usage.` / `gen_ai.request.` / `crewai.principal.` / `crewai.execution.` / `enduser.` prefixes, and last segments like `id` / `name` / `role` / `model` — and runs everything else through presidio. `gen_ai.input.messages` / `gen_ai.output.messages` are not exempt, so on the agent span they get exactly the pass they get on the LLM span. That is why this PR carries the text under those two keys only; an earlier draft also had `crewai.agent.task_prompt` / `crewai.agent.output` copies, and I dropped them.

3. **An opt-in for the agent span alone would protect nothing** — the same text keeps leaving on the spans beneath it — while making every consumer of the spans depend on a flag. A content switch that means something has to cover every message-carrying attribute (LLM messages, task output, system instructions, tool arguments and results) at once. That is its own change, and I'd rather not fold it into this one.

Kept as is. 7aeb58dc9 adds the test described in the other thread.

lib/crewai/tests/telemetry/test_span_task_agent_tool_attributes.py
joaomdmoura: There is no disabled path to test — see the other thread for why an agent-only switch is not added. What *can* regress is the shape: a future change could put the text under a new `crewai.agent.*` key, which a redactor selecting by key name would not cover.

7aeb58dc9 pins that: `test_the_agent_text_travels_under_the_standard_keys_only` asserts the prompt and the answer appear under `gen_ai.input.messages` / `gen_ai.output.messages` and under no other attribute of the span.

iris-clawd (approved):
I could not inspect the private hosted redactor, but these values already leave through the same standard GenAI keys. The shared shaping and truncation path keeps the new copies bounded.

Not holding this for it, but worth a ticket so it is not forgotten: `lib/crewai/src/crewai/telemetry/tracing/handlers.py` now unconditionally places agent prompts and answers on another exported span. The claim that hosted/Enterprise redaction treats these attributes safely cannot be verified from this public repository; the prior review explicitly could not verify that private integration behavior.

PR #7613, opened 2026-09-19
fix(bedrock): treat throttling as a retryable rate limit, not a context-window error

### LINKED ISSUE #7377: Classify Bedrock throttles as retryable rate limits with exponential backoff
## Problem

AWS Bedrock throttling responses such as `ThrottlingException` and “Too many tokens, please wait before trying again” are currently being mapped as context-window errors. This sends execution into context recovery instead of retrying the request.

## Expected behavior

Treat Bedrock throttling as a retryable rate-limit error and retry with exponential backoff, consistent with the provider's throttling semantics.

The retry behavior should:

- Detect the relevant Bedrock throttling responses.
- Avoid triggering context-length recovery for throttling.
- Apply bounded exponential backoff between attempts.
- Preserve the existing behavior for genuine context-window errors.
- Include unit tests covering throttling and non-throttling error classification.

Related Linear ticket: [OSS-169](https://linear.app/crewai/issue/OSS-169/classify-bedrock-throttles-as-retryable-rate-limits)

This issue is intentionally separate from guarding the context-recovery path when `llm.acall()` is unavailable ([OSS-170](https://linear.app/crewai/issue/OSS-170/prevent-context-recovery-from-fatally-failing-bedrock-flows)).

Debasishhh:
## Summary

Closes #7377.

`ThrottlingException` messages from Bedrock (e.g. *"Too many tokens, please wait before trying again"*) were matching the generic `"too many tokens"` entry in `CONTEXT_LIMIT_ERRORS` (`crewai/utilities/exceptions/context_window_exceeding_exception.py`), so a throttled request was routed into context-window recovery (summarize or `SystemExit`) instead of being retried — exactly as described in the issue.

Root cause, traced in `lib/crewai/src/crewai/llms/providers/bedrock/completion.py`:
1. `ThrottlingException` is caught and re-raised as a plain `RuntimeError` embedding AWS's raw message.
2. The outer `call()`/`acall()` handler runs every caught exception's `str()` through `is_context_length_exceeded()`.
3. AWS's throttle message text contains "too many tokens", so it false-positives as a context-length error.

Separately, the only retry that ever happened was boto3's own adaptive retry (`max_attempts=3`, already configured on the client) — once that's exhausted, the request just failed (misclassified, on top of that). No application-level retry existed for any provider in this codebase (OpenAI etc. rely entirely on their SDK's own retry, same as boto3 here).

## Changes

- Added `BedrockThrottlingError` (a `RuntimeError` subclass) so a throttled request is never classified by message content alone.
- Added `_call_with_throttle_retry` / `_acall_with_throttle_retry`, wrapping all 4 boto3 call sites (`converse`, `converse_stream`, sync + async) with an application-level retry using exponential backoff (1s, 2s, 4s) for throttling that outlasts boto3's own adaptive retries.
- `call()`/`acall()` now skip the context-length check for `BedrockThrottlingError` instances, so it can no longer be swept into context-window handling.
- `_handle_client_error` now returns `(error_code, message)` instead of just `message`, so the streaming call sites can also raise the right exception type.

## Test plan

- Added `test_bedrock_throttling_is_not_misclassified_as_context_length_error` — reproduces the exact issue scenario (AWS's real throttle message text) and asserts it does **not** surface as `LLMContextLengthExceededError`.
- Added `test_bedrock_throttling_retries_then_succeeds` — asserts a request that's throttled twice then succeeds returns correctly, with `time.sleep` called with the expected backoff (1.0s, 2.0s).
- Updated `test_bedrock_client_error_handling`'s throttling case to assert `BedrockThrottlingError` and that all 4 attempts (1 + 3 retries) happen; `time.sleep` is mocked so the test stays fast.
- `uv run pytest lib/crewai/tests/llms/bedrock/ -q` → 38 passed (1 unrelated pre-existing failure locally due to a missing `litellm` extra, confirmed present on `main` without this change too).
- `uv run ruff check` / `ruff format --check` clean on both changed files.


Debasishhh:
Good catch — the previous version only guarded `BedrockCompletion.call`/`acall`'s own exception handling. The `BedrockThrottlingError` still re-propagated up to `CrewAgentExecutor`/`LiteAgent`/the experimental executor, which each independently call `is_context_length_exceeded` on whatever they catch, so it was still misclassified one layer up in the actual agent loop.

Fixed by introducing a shared `LLMRateLimitError` in `base_llm.py` and checking it inside `is_context_length_exceeded` itself (the one function every executor calls). `BedrockThrottlingError` now subclasses it, so this holds regardless of which layer catches the exception.

Also added the requested non-VCR mocked async tests for `_ahandle_converse` and `_ahandle_streaming_converse` (retry count + backoff timing), plus unit tests on `is_context_length_exceeded` directly.

PR #7612, opened 2026-09-19
fix(knowledge): chunk CSV/JSON file text instead of the content dict repr

Lesereingrape:
## Related issue

Fixes #7611

## Summary

`CSVKnowledgeSource.add()` / `aadd()` and `JSONKnowledgeSource.add()` / `aadd()` chunked `str(self.content)` — the Python repr of the whole `dict[Path, str]` — instead of each file's text, so every chunk handed to the embedder carried `PosixPath('...')` wrappers, quotes and `\\n` literals, and a multi-file source had all of its files merged into one string.

Measured on `3831e8b` (absolute paths shortened here):

```python
CSVKnowledgeSource(file_paths=[d / "data.csv"]).chunks    # d/data.csv = "Name,Age\nBrandon,30\nAlice,25\n"
# ["{WindowsPath('.../data.csv'): 'Name Age\\nBrandon 30\\nAlice 25\\n'}"]
JSONKnowledgeSource(file_paths=[d / "data.json"]).chunks  # {"name": "Brandon", "age": 30}
# ["{WindowsPath('.../data.json'): 'name: Brandon\\nage: 30\\n'}"]
CSVKnowledgeSource(file_paths=[a, b]).chunks              # a="Name\nBrandon\n", b="City\nChicago\n"
# ["{WindowsPath('.../multi_a.csv'): 'Name\\nBrandon\\n', WindowsPath('.../multi_b.csv'): 'City\\nChicago\\n'}"]
```

After the change:

```python
# ['Name Age\nBrandon 30\nAlice 25\n']
# ['name: Brandon\nage: 30\n']
# ['Name\nBrandon\n', 'City\nChicago\n']
```

`BaseFileKnowledgeSource` declares `content: dict[Path, str] = Field(init=False, default_factory=dict)` and its `model_post_init` always assigns `self.content = self.load_content()`, so for these two sources the `isinstance(self.content, dict)` branch is not a corner case — it is the only branch that ever runs (`content` cannot be passed to the constructor). That makes the `else` arm dead and the repr the permanent behavior.

The fix drops the repr and iterates the mapping, which is exactly what the three sibling sources already do — `text_file_knowledge_source.py:26`/`:33`, `pdf_knowledge_source.py:46`/`:53`, `excel_knowledge_source.py:151`/`:165` — so CSV/JSON now match the per-file chunking the rest of the package has. Same four-line shape in both `add()` and `aadd()` of each class; no change to any signature, field, or stored format.

## Verification

[...]

Issue #7611, opened 2026-09-19
CSV and JSON knowledge sources embed the repr of the content dict instead of the file text


Lesereingrape:
## Description

`CSVKnowledgeSource` and `JSONKnowledgeSource` chunk `str(self.content)` — the Python **repr of the whole `dict[Path, str]`** — instead of each file's text. Every chunk that gets embedded and stored therefore carries the `PosixPath('...')` wrapper, quotes and escaped newlines, and a multi-file source has all of its files merged into one string.

```python
# lib/crewai/src/crewai/knowledge/source/csv_knowledge_source.py:30  (same in aadd() at :37)
# lib/crewai/src/crewai/knowledge/source/json_knowledge_source.py:42 (same in aadd() at :49)
content_str = (
    str(self.content) if isinstance(self.content, dict) else self.content
)
new_chunks = self._chunk_text(content_str)
```

`BaseFileKnowledgeSource` declares `content: dict[Path, str] = Field(init=False, default_factory=dict)` and its `model_post_init` runs `self.content = self.load_content()` unconditionally, so for these two sources the `isinstance(..., dict)` branch is not a corner case — it is the only branch that ever runs (`content` cannot even be passed to `__init__`).

Every other file knowledge source in the same package already iterates the mapping instead: `text_file_knowledge_source.py:26`/`:33`, `pdf_knowledge_source.py:46`/`:53`, `excel_knowledge_source.py:151`/`:165`. CSV and JSON are the only two outliers, and after the fix `str(self.content)` no longer appears anywhere under `lib/crewai/src/crewai/knowledge/`.

## Steps to Reproduce

```python
from pathlib import Path
from unittest.mock import MagicMock

from crewai.knowledge.source.csv_knowledge_source import CSVKnowledgeSource

path = Path("knowledge/data.csv")
path.parent.mkdir(exist_ok=True)
path.write_text("Name,Age\nBrandon,30\nAlice,25\n", encoding="utf-8")

source = CSVKnowledgeSource(file_paths=[path])
source.storage = MagicMock()
source.add()
print(source.chunks)
```

## Actual behavior (measured on `main` @ `3831e8b`)

```
["{WindowsPath('C:/.../knowledge/data.csv'): 'Name Age\\nBrandon 30\\nAlice 25\\n'}"]
```

With `{"name": "Brandon", "age": 30}` through `JSONKnowledgeSource`:

```
[...]

PR #7615, opened 2026-09-19
fix(tools): resolve a coworker that a model spelled as a quoted list

### LINKED ISSUE #7614: Delegating to a coworker fails when the model emits a Python-repr list: coworker="['researcher']"
## Description

The delegation tools tolerate a coworker that a weak model emitted as a *list* instead of a string — `_get_coworker()` unwraps `[...]` and takes the first entry. But the unwrap keeps the entry's quotes, and role matching only strips **double** quotes, so the Python-repr spelling (`['researcher']`, what `str(list)` and most small models produce) never matches, while the JSON spelling (`["researcher"]`) does.

The failure is self-contradicting: the tool refuses the role and then prints that same role as an available option.

```
coworker="['Senior Researcher']"
-> Error executing tool. coworker mentioned not found, it must be one of the following options:
   - senior researcher
   - analyst
```

## Steps to Reproduce

```python
from crewai.tools.agent_tools.agent_tools import AgentTools

senior = Agent(role="Senior Researcher", goal="g", backstory="b", allow_delegation=False)
analyst = Agent(role="Analyst", goal="g", backstory="b", allow_delegation=False)
delegate = AgentTools(agents=[senior, analyst]).tools()[0]

with patch.object(Agent, "execute_task", lambda self, task, context=None, tools=None: self.role):
    for coworker in ["Senior Researcher", '["Senior Researcher"]', "['Senior Researcher']"]:
        print(repr(coworker), "->", delegate.run(coworker=coworker, task="t", context="c")[:60])
```

## Actual behavior (measured on `main` @ `3831e8b`)

| `coworker` as the model emits it | value used for role matching | outcome |
| --- | --- | --- |
| `Senior Researcher` | `'senior researcher'` | delegates |
| `[Senior Researcher]` | `'senior researcher'` | delegates |
| `["Senior Researcher"]` | `'senior researcher'` | delegates |
| `['Senior Researcher']` | `"'senior researcher'"` | **"coworker mentioned not found"** |
| `['Senior Researcher', 'Analyst']` | `"'senior researcher'"` | **"coworker mentioned not found"** |
| `'Senior Researcher'` (quoted scalar) | `"'senior researcher'"` | **"coworker mentioned not found"** |

## Expected behavior

All six spellings resolve to the `Senior Researcher` agent. A quoted name is the same defect as a padded name, and a padded name is already tolerated.

## Root cause

`_get_coworker()` (`tools/agent_tools/base_agent_tools.py:37`) hands the unwrapped entry straight to matching, and `sanitize_agent_name()` (`:20`) normalizes whitespace, case, and `"` only:

```python
normalized = " ".join(name.split())
return normalized.replace('"', "").casefold()
```

Its own docstring says the value comes back with "quotes removed", and the comment at `:65` explains why the tolerance exists at all ("less-powerful LLM's have difficulty producing valid JSON"). Single quotes are the other half of that JSON-shaped problem and are the natural output of `str(list)` / a Python-repr argument, so the branch misses its main case.

`AskQuestionTool` shares `_get_coworker()`, so asking a coworker a question fails the same way.

## Possible Solution

Strip the wrapping quotes where the malformed value is already being normalized, in `_get_coworker()`, so both the list and the scalar form are handled in one place and the agents' own role names are never touched:

```python
if coworker:
    is_list = coworker.startswith("[") and coworker.endswith("]")
    if is_list:
        coworker = coworker[1:-1].split(",")[0]
    coworker = coworker.strip().strip("'\"")
```

I would not widen `sanitize_agent_name()` to drop `'` instead: it is applied to the agents' real role names as well, so an apostrophe in a role (`Data Steward's Office`) would stop matching. `_get_coworker()` only ever sees a model-produced argument.

## Additional context

Nothing under `lib/crewai/tests/` asserts the current single-quote behavior; the two existing list tests use the unquoted `coworker="[researcher]"` form, which is why the gap is invisible.

Authored with an AI coding assistant. `.github/CONTRIBUTING.md` requires the `llm-generated` label for agent-authored contributions and external contributors cannot apply labels here — maintainers, please add it.

## Operating System

Windows 11 (nothing OS specific)

## Python Version

3.13

## crewAI Version

1.15.22 (`main` @ `3831e8b`)

## crewAI Tools Version

## Virtual Environment

Venv

Lesereingrape:
## Related issue

Fixes #7614

## Summary

`_get_coworker()` already knows that a weak model may hand over a list where a string was expected, and unwraps `[...]` to its first entry — but it leaves that entry's quotes in place, and role matching (`sanitize_agent_name()`, same file) strips `"` only. So the JSON spelling `["researcher"]` delegates fine while the Python-repr spelling `['researcher']` — what `str(list)` produces — comes back as "coworker mentioned not found" while listing that very role as available:

```python
coworker='["Senior Researcher"]'             -> delegates
coworker="['Senior Researcher']"             -> "coworker mentioned not found… - senior researcher"
coworker="['Senior Researcher', 'Analyst']"  -> same
coworker="'Senior Researcher'"               -> same
```

The unwrap now strips the wrapping quotes (and surrounding whitespace) once, after it, which covers the list form and the quoted-scalar form together for both `DelegateWorkTool` and `AskQuestionTool`, since they share this method.

Deliberately *not* done: teaching `sanitize_agent_name()` to drop `'` too. Its own docstring ("quotes removed") invites that reading, but the function is applied to the agents' real role names as well as to the model's argument, so a role containing an apostrophe would stop matching. `_get_coworker()` only ever sees the incoming argument.

Behavior is unchanged for every spelling that works today: `.strip()` and `.strip("'\"")` are no-ops on a bare name, on `[researcher]`, and on `["researcher"]` (the double quotes were already removed later, and the two paths now agree).

## Verification

- New `test_delegate_work_resolves_coworker_spellings`, parametrized over the five spellings, with `Agent.execute_task` stubbed so no LLM or cassette is involved. On `3831e8b`: **2 failed, 3 passed** — the two single-quote spellings fail, and the three controls (`researcher`, `[researcher]`, `["researcher"]`) pass both before and after, which is the no-regression evidence. On this branch: 5 passed.
- `uv run pytest lib/crewai/tests/tools/agent_tools/test_agent_tools.py -q` → **13 passed** (the 8 pre-existing VCR-replay tests included).
- Blast radius: `lib/crewai/tests/tools` + `lib/crewai/tests/agents` run on the base and on this branch, fail-sets diffed → **exactly the 2 new cases flip, 0 regressions**. 32 failures remain identical on both sides in this environment (`crewai_files` is not installed here, plus the platform/a2a suites that need it); CI installs the full workspace.
- `uv run ruff check` / `ruff format --check` → clean on the changed source. `uv run mypy` → no error in `base_agent_tools.py` (the 10 reported errors are the pre-existing `no-any-unimported` findings in `crew_docling_source.py`).

## Additional context

The comment above the matching code spells out why the tolerance exists — "less-powerful LLM's have difficulty producing valid JSON" — so a single-quoted argument is squarely inside the case this branch was written for, and it is the half that was missed. The user-visible symptom is the loop where the tool refuses a role and then prints it as an allowed option, which a model usually retries by passing the bare name; so it costs a wasted tool call rather than breaking the crew outright.

Follow-up this PR does not attempt: `coworker[1:-1].split(",")[0]` silently drops every entry after the first, so `['Researcher', 'Analyst']` never considers `Analyst`. Splitting a quoted name on `,` is also wrong for a role that contains a comma. Both are about *which* coworker a multi-entry list should pick, i.e. a design call rather than a missed guard, so they belong in their own issue if maintainers want them addressed.

Authored with an AI coding assistant. `.github/CONTRIBUTING.md` requires the `llm-generated` label for agent-authored contributions and external contributors cannot apply it here — maintainers, please add it.


Lesereingrape:
Heads-up for maintainers so this does not get reviewed twice: #7620 (opened a few hours after this one, same issue #7614) fixes the same defect in `sanitize_agent_name` so both quote styles are stripped on both sides of the comparison, while this commit strips wrapping quotes at the extraction point in `_get_coworker`.

They are equivalent for every spelling in the issue. The strip here sits in the `if coworker:` branch, not inside `if is_list:`, so a plain quoted scalar (`coworker="'Senior Researcher'"`) resolves too, and `.strip("'\"")` only removes *wrapping* quotes, so a role whose own name contains an apostrophe is unaffected.

Either one closes #7614 — merging one and closing the other is fine by me. If the sanitizer route is preferred, the parametrized spelling table in this PR's test (`researcher`, `[researcher]`, `["researcher"]`, `['researcher']`, `['researcher', 'writer']`) is copy-pasteable into #7620 as extra coverage.

Lesereingrape:

unrelated to this diff.

PR #7617, opened 2026-09-19
fix(knowledge): reject chunk settings that cannot produce chunks

### LINKED ISSUE #7616: Knowledge sources accept chunk settings that silently store nothing or drop characters
## Description

Every knowledge source accepts `chunk_size` / `chunk_overlap` combinations that cannot produce a valid chunking window. `_chunk_text()` then either stores **zero** documents or **silently drops characters** from the text it embeds, and neither case reports an error to the caller.

The chunking step is derived from the pair, and neither field is validated against the other (or against zero):

```python
# lib/crewai/src/crewai/knowledge/source/base_knowledge_source.py:19
    chunk_size: int = 4000
    chunk_overlap: int = 200
```

```python
# lib/crewai/src/crewai/knowledge/source/base_knowledge_source.py:43
    def _chunk_text(self, text: str) -> list[str]:
        """Utility method to split text into chunks."""
        return [
            text[i : i + self.chunk_size]
            for i in range(0, len(text), self.chunk_size - self.chunk_overlap)
        ]
```

`chunk_size - chunk_overlap` is the `range()` step, so it is the only thing deciding what reaches storage. Three shapes are reachable through the public constructor:

| `chunk_size` | `chunk_overlap` | step | what happens |
| --- | --- | --- | --- |
| 100 | 200 | `-100` | `range(0, len(text), -100)` is empty → 0 chunks → `storage.save([])` |
| 100 | 100 | `0` | `add()` raises `ValueError: range() arg 3 must not be zero` |
| 10 | `-1` | `11` | window advances by more than `chunk_size` → characters between chunks are never stored |

The negative-overlap row is the one that worries me most: the run reports success, the knowledge base is written, and part of every document is simply unretrievable. The `overlap > size` row is the more likely typo (`chunk_overlap=4000, chunk_size=2000`), and it produces an *empty* knowledge base that still looks like a successful ingest.

## Steps to Reproduce

No storage backend and no LLM needed — a stub storage object records what `add()` hands over.

```python
from unittest.mock import MagicMock

from crewai.knowledge.source.string_knowledge_source import StringKnowledgeSource

ALPHABET = "0123456789ABCDEFGHIJ"


def case(chunk_size: int, chunk_overlap: int, text: str = ALPHABET) -> None:
    source = StringKnowledgeSource(
        content=text, chunk_size=chunk_size, chunk_overlap=chunk_overlap
    )
    storage = MagicMock()
    source.storage = storage
    try:
        source.add()
    except Exception as exc:  # noqa: BLE001
        print(f"({chunk_size}, {chunk_overlap}) add() raised {type(exc).__name__}: {exc}")
        return
    saved = storage.save.call_args[0][0]
    lost = "".join(sorted(set(text) - set("".join(saved))))
    print(f"({chunk_size}, {chunk_overlap}) saved {len(saved)} chunk(s) {saved!r} chars_lost={lost!r}")


case(100, 200, "x" * 5000)  # 5000 characters of content, nothing stored
case(100, 100)
case(10, 11)
case(10, -1)
case(0, 0)
case(-5, 0)
```

## Actual behavior (measured, `main` @ `3831e8b`, Python 3.13 / Windows)

```
(100, 200) saved 0 chunk(s) [] chars_lost='x'
(100, 100) add() raised ValueError: range() arg 3 must not be zero
(10, 11) saved 0 chunk(s) [] chars_lost='0123456789ABCDEFGHIJ'
(10, -1) saved 2 chunk(s) ['0123456789', 'BCDEFGHIJ'] chars_lost='A'
(0, 0) add() raised ValueError: range() arg 3 must not be zero
(-5, 0) saved 0 chunk(s) [] chars_lost='0123456789ABCDEFGHIJ'
```

All six constructions above succeed. The first, third and sixth store *less than what was given*, with no signal at all.

Note the `(10, -1)` output: `'0123456789'` + `'BCDEFGHIJ'` — the character at index `10` is in neither chunk, so it never enters the embedding input.

## Expected behavior

A source that cannot chunk its content should fail where the mistake is made — at construction, with a message naming both numbers — instead of writing an empty or partial knowledge base, or raising a raw `range()` complaint from deep inside `add()`.

## Cause

`BaseKnowledgeSource` declares `chunk_size` / `chunk_overlap` as plain `int` fields, so pydantic only checks the type. The relation that actually matters (`0 <= chunk_overlap < chunk_size`, `chunk_size > 0`) is implicit in the `range()` step at `:47` and is never asserted.

`_chunk_text()` is also duplicated verbatim in the six subclasses that carry their own copy (`csv_knowledge_source.py:50`, `excel_knowledge_source.py:180`, `json_knowledge_source.py:62`, `pdf_knowledge_source.py:62`, `string_knowledge_source.py:40`, `text_file_knowledge_source.py:42`), so a per-subclass fix would have to be repeated. All eight sources inherit the field declarations from `BaseKnowledgeSource`, which is where a single guard belongs.

## Possible solution

Validate the pair on the model, where both fields are visible, so the check no longer depends on when `add()` happens to run:

```python
def _validate_chunk_settings(self) -> Self:
    """_chunk_text steps by chunk_size - chunk_overlap: zero makes range() raise, negative makes it yield no chunk."""
    if self.chunk_size <= 0:
        raise ValueError("chunk_size must be a positive integer")
    if self.chunk_overlap < 0:
        raise ValueError("chunk_overlap must be a non-negative integer")
    if self.chunk_overlap >= self.chunk_size:
        raise ValueError(
            f"chunk_overlap ({self.chunk_overlap}) must be smaller than "
            f"chunk_size ({self.chunk_size})"
        )
    return self
```

This is a narrowing, not a behavior change: every pair that produced correct chunks before still constructs, and only pairs that could not chunk at all are now rejected up front. The `chunk_overlap >= chunk_size` branch subsumes the `step == 0` and `step < 0` cases; the `chunk_size <= 0` and `chunk_overlap < 0` branches exist so the message names the bad field instead of blaming the other one.

Deduplicating the seven `_chunk_text()` copies looks like a separate, mechanical cleanup — I left it out to keep this change reviewable.

## Environment

- crewAI: `1.15.22` (`main` @ `3831e8b`)
- Python 3.13, Windows 11 (nothing OS specific)
- Verified with `lib/crewai/tests/knowledge/`: 6 new failing-before / passing-after cases plus controls; the only failures in that directory are two pre-existing `test_docling_*` `ImportError`s (`docling` not installed locally) that reproduce identically on `main`.

Authored with an AI coding assistant. `.github/CONTRIBUTING.md` requires the `llm-generated` label for agent-authored contributions; external contributors cannot apply labels, so maintainers please add it.

Lesereingrape:
## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Other

## Summary

Knowledge sources accepted `chunk_size` / `chunk_overlap` pairs that cannot produce a valid chunking window, and then either stored **zero** documents or **silently dropped characters** from the text they embedded — with no error reaching the caller. `BaseKnowledgeSource` now validates the pair once, on the model, so all eight sources reject it where the mistake is made.

Closes #7616

## Detail

`_chunk_text()` uses `chunk_size - chunk_overlap` as the `range()` step, so that difference is the only thing deciding what reaches storage; the two fields were plain `int`s with no relation asserted. Three shapes were reachable through the public constructor:

| `chunk_size` | `chunk_overlap` | step | before | after |
| --- | --- | --- | --- | --- |
| 100 | 200 | `-100` | `range()` empty → `storage.save([])`, ingest reports success | `ValidationError: chunk_overlap (200) must be smaller than chunk_size (100)` |
| 100 | 100 | `0` | `ValueError: range() arg 3 must not be zero` from inside `add()` | same `ValidationError`, raised at construction |
| 10 | `-1` | `11` | saved `['0123456789', 'BCDEFGHIJ']` — the char at index 10 is in no chunk | `ValidationError: chunk_overlap must be a non-negative integer` |
| 0 / -5 | – | – | `range()` step 0 / empty → crash or empty KB | `ValidationError: chunk_size must be a positive integer` |

Measured on `main` @ `3831e8b` with a stub storage object; reproduction script and raw output are in the linked issue.

The guard sits on `BaseKnowledgeSource` rather than in each source because `_chunk_text()` is duplicated verbatim in six subclasses (`csv`, `excel`, `json`, `pdf`, `string`, `text_file`) while all eight inherit the field declarations. Deduplicating those copies is a separate mechanical cleanup and is deliberately not part of this diff.

## Interaction / alternatives considered

`lib/crewai-files/src/crewai_files/processing/transformers.py` handles the same class of mistake by **clamping** (`start_pos = max(start_pos + 1, end_pos - overlap_chars)`, with a test named `test_chunk_overlap_larger_than_max_chars`). I chose to raise instead, because the clamping path changes which characters end up in each chunk — for the knowledge sources that means re-embedding content the user did not configure, and it keeps the misconfiguration invisible. If maintainers would rather match `crewai-files` and clamp, say so and I will convert this to a `step = max(1, ...)` form plus the corresponding tests.

The one intentional behavior change: a source that today constructs and silently stores nothing now raises at construction. Anything that produced correct chunks before still constructs unchanged — that is what the control cases below pin down.

## Evidence before / after

```
# base (main @ 3831e8b) — new tests, source reverted, same test file
6 failed, 3 passed, 18 deselected          # the 3 passing are the accepted-settings controls

# branch
9 passed, 18 deselected
```

```
# full knowledge directory, branch
2 failed, 61 passed    # both failures are pre-existing: test_docling_source and
                       # test_multiple_docling_sources raise ImportError because
                       # `docling` is not installed in my venv; they fail
                       # identically on main with the source file reverted
```

## Test coverage

5 new test functions / 9 parametrized cases in `lib/crewai/tests/knowledge/test_knowledge.py`:

- `test_overlap_above_chunk_size_is_rejected_at_construction` — asserts the exact message, so the numbers cannot be transposed
- `test_overlap_matching_chunk_size_is_rejected_at_construction` — the old `range() arg 3 must not be zero` case
- `test_non_positive_chunk_settings_are_rejected_at_construction[(0,0),(-5,0),(100,-1)]`
- `test_file_sources_inherit_the_chunk_settings_guard` — a file source, to show the base-class guard is reached by subclasses
- `test_accepted_chunk_settings_still_produce_chunks[(1000,200),(100,99),(100,0)]` — controls that pass on base **and** on the branch, so the guard is not over-tight

All run offline (stub storage, no embedding provider, no cassette).

## Validation

- [x] I ran `ruff check` and `ruff format --check` on the changed source file — clean
- [x] I ran `mypy` on the changed source file — `Success: no issues found in 1 source file`
- [x] I ran the affected test module and the whole `lib/crewai/tests/knowledge` directory, on base and on the branch
- [x] I re-checked open PRs against the two paths I touch: `base_knowledge_source.py` has no other open PR; the only other hits on `test_knowledge.py` are my own #7610 and #7612
- [x] New and existing tests pass locally with my changes (the two `docling` import failures are environment-only and reproduce on `main`)

## How to verify

```bash
uv run pytest lib/crewai/tests/knowledge/test_knowledge.py -k chunk -q
```

or confirm the behavior directly:

```python
from crewai.knowledge.source.string_knowledge_source import StringKnowledgeSource

StringKnowledgeSource(content="x" * 5000, chunk_size=100, chunk_overlap=200)
# ValidationError: chunk_overlap (200) must be smaller than chunk_size (100)
```

---

Disclosure: **AI assistance was used to investigate, implement and test this change**, per `.github/CONTRIBUTING.md` — maintainers, could you add the `llm-generated` label? Outside collaborators cannot apply labels, and the PR template's "AI generated code" checkbox is not reachable from the `gh` CLI, so this note is the disclosure. Every number quoted above comes from a measured run against `main` @ `3831e8b` and against this branch, using the commands in the Validation section.


Lesereingrape:

unrelated to this diff.

PR #7639, opened 2026-09-19
fix(llms): give gemini -latest aliases full tool capabilities

### LINKED ISSUE #7636: [BUG] Gemini native provider: version-less aliases (gemini-flash-latest, gemini-flash-lite-latest, gemini-pro-latest) lose tool support — tools are silently dropped from requests
### Description

The native Gemini provider derives `supports_tools`, `is_gemini_2_0`, and the automatic `thinking_config` from a version regex — `re.search(r"gemini-(\d+(?:\.\d+)?)", model.lower())` in `GeminiCompletion._normalize_gemini_fields` (`lib/crewai/src/crewai/llms/providers/gemini/completion.py:99`). The version-less aliases that crewAI lists in its own `GEMINI_MODELS` table (`gemini-flash-latest`, `gemini-flash-lite-latest`, `gemini-pro-latest`; `lib/crewai/src/crewai/llms/constants.py:310-312`) contain no digit after `gemini-`, so the regex never matches and they are treated as pre-1.5 models.

Two user-visible consequences:

1. `_prepare_generation_config` only attaches tools `if tools and self.supports_tools` (line 490), so `LLM(model="gemini/gemini-flash-latest").call(messages, tools=[...])` sends a request with **no tools at all** — the model can never call them.
2. `supports_function_calling()` returns False, so an Agent on one of these aliases silently downgrades from native function calling to the ReAct text protocol.

Structured output also takes the legacy `response_schema` path and thinking is not auto-enabled, unlike the numbered equivalents (e.g. `gemini-2.5-flash`).

### Steps to Reproduce

1. `pip install "crewai[google-genai]"` (no API key needed).
2. Run:

```python
g = GeminiCompletion(model="gemini-flash-latest")
print(g.supports_tools, g.supports_function_calling())
cfg = g._prepare_generation_config(None, [{"type": "function", "function": {"name": "lookup", "description": "d", "parameters": {"type": "object", "properties": {}}}}], None)
print(getattr(cfg, "tools", None))
```

3. Compare with `GeminiCompletion(model="gemini-2.5-flash")`.

### Expected behavior

The `-latest` aliases resolve to current 2.5-generation models, so `supports_tools` / `supports_function_calling()` are True and the request config contains the tools, exactly as for `gemini-2.5-flash`.

### Screenshots/Code snippets

```
model                                 supports_tools  supports_fc  tools_sent  thinking_auto
gemini-2.5-flash                      True            True         True        True
gemini-2.0-flash                      True            True         True        False
gemini-flash-latest                   False           False        False       False
gemini-flash-lite-latest              False           False        False       False
gemini-pro-latest                     False           False        False       False
```

### Operating System

Windows 11 (reproduces anywhere; capability detection is offline)

### Python Version

3.12 (verified on 3.13.12 as well)

### crewAI Version

1.15.22 (also reproduced on `main` @ `3831e8b`)

### crewAI Tools Version

n/a (not required)

### Virtual Environment

Venv (uv)

### Evidence

Script output above: the three aliases report `supports_tools=False` and omit tools from the `GenerateContentConfig`, while `gemini-2.5-flash` includes them. `llm.py:474` documents "latest" versions as intended-supported; `constants.py:310-312` lists the aliases.

### Possible Solution

When the version regex misses, consult an explicit alias map for the version-less aliases (e.g. `{"gemini-flash-latest": 2.5, "gemini-flash-lite-latest": 2.5, "gemini-pro-latest": 2.5}`) rather than loosening the regex — a "first number anywhere" regex would mis-detect `gemini-gemma-2-27b-it` (gemma has no function calling) as 2.0. Same class of fix as merged PR #4009 for Anthropic. Happy to open a PR with this plus regression tests.

### Additional context

Also affects `gemini-robotics-er-1.5-preview` and the `gemini-live-*` ids in the table (out of scope for the minimal fix; can be handled separately). Verified no existing issue/PR covers this (searched supports_tools, alias names, thinking config, and all open PR file lists touching the provider on 2026-09-20). This issue was found and written with AI assistance; please apply the `llm-generated` label per the contribution policy.

Madin-H23:
## Related issue

Fixes #7636

## Summary

The version-less aliases listed in `GEMINI_MODELS` (`gemini-flash-latest`, `gemini-flash-lite-latest`, `gemini-pro-latest`) never match the `gemini-<version>` regex in `_normalize_gemini_fields`, so they fell back to pre-1.5 defaults: `supports_tools=False` dropped tools from `GenerateContentConfig` entirely, `supports_function_calling()` returned `False` (silently downgrading agents to the ReAct protocol), structured output took the legacy `response_schema` path, and thinking was not auto-enabled.

Following the approach sketched in the issue (same shape as the Anthropic fix in #4009): a new `GEMINI_LATEST_ALIASES` map in `llms/constants.py` (next to `GEMINI_MODELS`) resolves the three aliases to the 2.5 generation when the version regex misses. The regex itself is untouched, so ids like `gemini-gemma-2-27b-it` — which carry a version but no function calling — keep the pre-1.5 defaults.

## Verification

- `pytest lib/crewai/tests/llms/google/test_google.py -n0` → all previously-passing tests still pass (the 9 pre-existing cassette-based failures reproduce identically on unpatched `main`); the new tests add 4 passes:
  - parametrized over all three aliases: `supports_tools` is `True` and `supports_function_calling()` is `True`
  - `gemini-gemma-2-27b-it` guard: still `supports_tools == False` — pins that the alias map doesn't loosen the regex
  - red-check: the alias tests fail against unpatched `main` (verified via `git stash`)
- `ruff check` / `ruff format --check` clean on all three files.

## Additional context

Out of scope per the issue: `gemini-robotics-er-1.5-preview` and the `gemini-live-*` ids.


Madin-H23:
CodeRabbit's out-of-scope warning was correct: this branch was cut from the working branch of #7587 and carried that commit. The branch is now rebased onto current `main` — the diff contains only the Gemini alias fix (3 files) plus the docstring for `_normalize_gemini_fields` (addressing the docstring-coverage warning). Tests re-run after the rebase: 48 passed, the 9 failures are the pre-existing cassette-based ones that reproduce identically on unpatched `main`.

PR #7648, opened 2026-09-20
feat(tracing): record the last traced run for crewai eval instead of printing it

joaomdmoura:
## Related issue

Groundwork for `crewai eval` (a following PR): the CLI evaluates the last traced run without the user pasting an id.

## Summary

- After a traced run nothing about tracing is printed any more. The "Execution trace ID / View traces" panel exposed an internal id for no purpose a user has.
- Once the run's spans have reached Wharf, crewAI writes `.crewai/last_run.json` in the project — the execution id, when the run started and ended, whether it was traced anonymously or under an account, and the AMP base url. Written atomically; a write failure never fails the run; only a run whose every export succeeded is recorded; silent in the TUI and under message suppression too; off under the test suite; inert inside a deployment (the platform binds the execution before crewAI's own tracing starts, so this path never runs there).
- `GrantSpanExporter.record_export` replaces `show_trace_summary` at both finish points (authenticated shutdown, anonymous share). The grant request is unchanged (`include_trace_url` stays; AMP may use it later).
- The crew, flow and json_crew scaffolds ignore `.crewai/`, as the declarative flow scaffold already did.

## Verification

- [x] Tests: `lib/crewai/tests/telemetry` and `lib/crewai/tests/tracing` (the four panel tests rewritten around "nothing printed, run recorded"; new `test_last_run.py`), `lib/cli/tests/test_scaffold_gitignore.py` — 595 passed locally. The two failures in `test_create_crew.py` (`crewai[tools]>=1.15.22` version specifier) fail identically on clean `main`.
- [x] `ruff check`, `ruff format --check`, `mypy` clean on the changed files.

## Additional context

`crewai eval` (next PR) reads `.crewai/last_run.json` by default and takes `--run <execution id>` for another run.

---

> [!NOTE]
> **Low Risk**
> Changes are additive local filesystem metadata and UX around trace export; failures never fail runs, with broad test coverage for concurrency and partial exports.
> 
> **Overview**
> Replaces the post-run **execution trace ID panel** with a silent **`.crewai/last_run.json`** record so upcoming **`crewai eval`** can target the last fully exported run without copying an id.
> 
> After spans reach Wharf, **`GrantSpanExporter.record_export`** (replacing **`show_trace_summary`**) writes execution id, tier, start/end times, and AMP base URL atomically with optional **`flock`**; only **complete** exports are recorded (partial or truncated ephemeral traces are skipped). Console output is reduced to a single optional **“View traces:”** link when AMP provides a URL; TUI/message suppression still hides the link but **recording continues**. Recording is disabled under **`CREWAI_TESTING`**. CLI scaffolds (**crew**, **flow**, **json_crew**) now gitignore **`.crewai/`** (aligned with declarative flow).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2d93e855f0cb1fca423573395e0656fcc0e8c91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,97 @@
...
+        "started_at": _iso(started_at_ns),
+        "finished_at": _iso(finished_at_ns),
+        "recorded_at": datetime.now(timezone.utc).isoformat(timespec="milliseconds"),
+        "amp_base_url": amp_base_url,
+    }
+    path: Path | None = None
+    try:
+        path = last_run_path()
+        path.parent.mkdir(parents=True, exist_ok=True)
+        temporary = path.with_name(path.name + ".tmp")
```
iris-clawd: Can we use a unique temporary file in this directory before `os.replace`? Concurrent crews share this path. One process can replace the temporary file, causing the other to silently lose its successful run record.

iris-clawd (changes requested):
I could not verify private host versions still deployed in the wild.

The note inline on last_run.py is what I'd want addressed before this merges.

Someone else should sign off on this too: the new `.crewai/last_run.json` shape is a stored-data/serialization contract intended for `crewai eval`, including execution identity, tracing tier, timestamps, and AMP endpoint.

lib/crewai/src/crewai/telemetry/tracing/ephemeral.py
```diff
@@ -143,7 +143,7 @@ def share(self, execution_uuid: str, *, first_time: bool = False) -> None:
                     logger.warning("Ephemeral trace export failed; buffer discarded")
             finally:
                 exporter.shutdown()
-            exporter.show_trace_summary()
+            exporter.record_export()
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in c7e50bdaa: `share()` calls `record_export()` only when the buffer dropped nothing — a truncated trace is still uploaded but not recorded, since the grader could not read that run whole. Test: a buffer capped at one span shares what survived and writes no record.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,92 @@
...
+    }
+    path = last_run_path()
+    try:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        temporary = path.with_name(path.name + ".tmp")
+        temporary.write_text(json.dumps(record, indent=2) + "\n", encoding="utf-8")
+        os.replace(temporary, path)
+    except OSError as error:
+        logger.debug("Could not record the last run in %s: %s", path, error)
+        return None
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in c7e50bdaa: `last_run_path()` is inside the guard (a vanished cwd fails it too), same debug log, same None.

docs/edge/en/observability/tracing.mdx
```diff
@@ -154,6 +154,8 @@ Traces are uploaded only after a successful authenticated export or an explicitl
 For traces associated with your account, open the [Traces tab in the CrewAI AMP dashboard](https://app.crewai.com/crewai_plus/trace_batches) to view agent interactions, tool usage, and LLM calls.
 ![CrewAI Tracing Interface](/images/view-traces.png)
 
+After a run whose spans reached CrewAI AMP, crewAI writes `.crewai/last_run.json` in the project directory: the run's execution id, whether it was traced anonymously or under your account, when it started and finished, and the AMP it was sent to. `crewai eval` reads it to evaluate the last run without you copying an id. The scaffold's `.gitignore` covers `.crewai/`; only a run exported whole is recorded, and nothing is written under `CREWAI_TESTING=true`.
```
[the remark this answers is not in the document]
joaomdmoura: Done in 97eceda88: the paragraph in en, ar, ko and pt-BR now says nothing is recorded inside a deployment, where the platform owns the trace.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
joaomdmoura: Done in 97eceda88: each write goes through `tempfile.mkstemp` in the same directory (`.last_run.json.<random>.tmp`), then `os.replace`; a failed write removes its own temporary file. Tests: two writers use two names and leave only `last_run.json`; a failed replace leaves nothing behind.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,109 @@
...
+        path.parent.mkdir(parents=True, exist_ok=True)
+        # A temporary file of its own per writer: two crews finishing together in one
+        # project must not write through the same name, or one record is lost.
+        descriptor, temporary = tempfile.mkstemp(
+            prefix=f".{path.name}.", suffix=".tmp", dir=path.parent
+        )
+        try:
+            with os.fdopen(descriptor, "w", encoding="utf-8") as handle:
+                handle.write(json.dumps(record, indent=2) + "\n")
+            os.replace(temporary, path)
```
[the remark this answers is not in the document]
joaomdmoura: Done in ae2bf77b5, without a lock: `record_last_run` compares `finished_at` (`recorded_at` when absent) with the record already there and keeps the later-finished run, so the older writer arriving last leaves the newer run recorded — the interleaving test forces exactly that order. The same run recorded again and a record without a comparable time never block the run just finished. I left out a cross-process lock on purpose: the file is a convenience pointer for `crewai eval`, a lock file in every project plus platform-specific locking is more machinery than it protects, and what remains is a microsecond window between the read and the replace that only two runs finishing in the same instant could hit — either is then a fair "last run".

lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,133 @@
...
+    or one without a comparable time, never wins over the run just finished."""
+    if not existing or existing.get("execution_id") == record["execution_id"]:
+        return True
+    ours = record["finished_at"] or record["recorded_at"]
+    theirs = existing.get("finished_at") or existing.get("recorded_at")
+    if not isinstance(theirs, str) or not isinstance(ours, str):
+        return True
+    return (
+        ours >= theirs
+    )  # both ISO 8601 in UTC with the same precision: text order is time order
```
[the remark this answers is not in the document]
joaomdmoura: Done in 0f14e159c, the cheaper half of the ask: a tie keeps the record already there (`>` instead of `>=`), so an older execution can no longer replace a newer one that finished in the same millisecond. I did not add an unrounded `finished_at_ns` to the file: the record already orders runs as finely as it stores them, and a sub-millisecond distinction between two runs finishing in the same instant is not one `crewai eval` needs — either is a fair "last run". Test extended with the same-millisecond case.

joaomdmoura:
On the outside-diff note about `grants.py` (a deployment run reaching `record_export` through the platform integration token): fixed in 0f14e159c. `recording_enabled()` is now false whenever the platform's integration token is present, so a container that did not have the host bind the trace still writes nothing — the docs' promise holds structurally instead of by the usual code path. Test added.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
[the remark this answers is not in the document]
joaomdmoura: Correction to my last reply: the strict tie-break I pushed for this was wrong, and CI caught it. Two records written in the same millisecond with no `finished_at` fell back to `recorded_at`, compared equal, and the second write was dropped — so a plain sequence of two runs could leave the older one recorded. Locally the writes landed in different milliseconds and the suite passed; that was luck, not correctness.

ceb4cd730 narrows the rule instead: a write stands unless the record already there is a DIFFERENT run with a later `finished_at`. No completion time on either side, the same run again, or a tie all leave the write standing, so "the last run recorded" is what a reader gets. The case this exists for is unchanged — the older-finished writer arriving last still does not replace the newer one — and there is no sub-millisecond field to add, since the tie no longer decides anything.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,144 @@
...
+    """Off under the test suite, so kickoff-level tests leave no file behind,
+    and off inside a deployment — the platform's integration token marks one —
+    so a run there never writes a file the platform never reads. (A deployment
+    normally never gets here at all: the host binds the trace before crewAI
+    would start its own; this is the guard for a container that did not.)"""
+    from crewai.context import get_platform_integration_token
+
+    if os.environ.get("CREWAI_TESTING", "").lower() == "true":
+        return False
+    return get_platform_integration_token() is None
```
iris-clawd: Can we use a deployment-specific signal here? Local JSON crews persist this token in `.env`, which `crew_base.py:130` loads, so successful local traces produce neither this record nor the removed identifier.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,144 @@
...
+        path.parent.mkdir(parents=True, exist_ok=True)
+        # A temporary file of its own per writer: two crews finishing together in one
+        # project must not write through the same name, or one record is lost.
+        descriptor, temporary = tempfile.mkstemp(
+            prefix=f".{path.name}.", suffix=".tmp", dir=path.parent
+        )
+        try:
+            with os.fdopen(descriptor, "w", encoding="utf-8") as handle:
+                handle.write(json.dumps(record, indent=2) + "\n")
+            if _keep(record, read_last_run(path.parent.parent)):
```
iris-clawd: Can we serialize this comparison and replacement across processes? An older run can read first, pause while a newer run replaces the file, then overwrite it and become the wrong eval target.

lib/crewai/tests/telemetry/test_last_run.py
```diff
@@ -0,0 +1,124 @@
...
+    assert [child.name for child in (project / ".crewai").iterdir()] == ["last_run.json"]
+
+
+def test_a_failed_write_leaves_no_temporary_file(project, monkeypatch):
+    monkeypatch.setattr(last_run.os, "replace", lambda src, dst: (_ for _ in ()).throw(OSError("disk full")))
+    assert last_run.record_last_run(execution_id="x", tier=None, started_at_ns=None, finished_at_ns=None, amp_base_url=None) is None
+    assert list((project / ".crewai").iterdir()) == []
+
+
+def test_the_run_that_finished_last_stays_recorded_whichever_writer_comes_last(project):
```
iris-clawd, line 115: Can we cover the actual stale-read interleaving? Pause the older writer after its read, let the newer writer replace, then resume it. This sequential test passes while that bug remains.

lib/crewai/src/crewai/telemetry/tracing/ephemeral.py
```diff
@@ -143,7 +143,10 @@ def share(self, execution_uuid: str, *, first_time: bool = False) -> None:
                     logger.warning("Ephemeral trace export failed; buffer discarded")
             finally:
                 exporter.shutdown()
-            exporter.show_trace_summary()
+            if (
+                not self._dropped
```
iris-clawd, line 147: Good catch. Retained spans still upload, while a truncated anonymous trace cannot become the eval target.

iris-clawd (dismissed):
I could not verify the deployment signal used by private Factory, on-prem, and Enterprise host versions.

The notes inline on last_run.py are what I'd want addressed before this merges. The other inline notes are small.

Someone else should sign off on this too: the new `.crewai/last_run.json` format is a serialization/stored-data contract for another CLI command.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
joaomdmoura: You're right, and it is worse than a missed signal — that guard was a regression, so c93892090 removes it.

`CREWAI_PLATFORM_INTEGRATION_TOKEN` is a credential a developer is told to get for platform tools, and `crewai create crew` writes it into the project's own `.env` (create_json_crew.py:1183, pinned by test_create_crew.py:1043), which `crew_base.py:130` and `run_crew.py:313-317` load. So the guard read "this developer uses platform tools", and those users would have got no record at all. It is also wrong in the other direction: the enterprise API sets that token only when an action token exists (crew_run_task.py:105-108), so a deployment without integrations does not have it.

I did not swap in a real deployment marker either, though `CREWAI_PLUS_ID` and `CREWAI_DEPLOYMENT_INSTANCE_ID` are provisioner-injected and on the platform's restricted list. A deployment cannot reach this code: the host sets `crew.tracing = False` / `flow.tracing = False` at all five kickoff sites, so `_start_tracing` returns before `GrantSpanExporter` exists; and even past that, `should_enable_tracing` is False in a container so there is no authenticated exporter, and there is no TTY so the ephemeral path never gets consent. A fourth check that is never the operative one is not worth the local breakage.

The module docstring claimed the platform "binds the execution before crewAI would start its own tracing". It does not — enterprise's `telemetry_session` sets its own contextvar, not crewAI's. The docstring and the four translated docs now say the real reason. Test added: a project carrying the platform token is still recorded.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
joaomdmoura: Done in c93892090. Reading the record, deciding and replacing it are now one step under a lock file beside the record (`flock`), so the interleaving you describe cannot happen: the older writer's comparison runs against whatever the newer writer left, not against what it read before pausing. Where the platform has no `flock` (Windows) the write goes ahead unserialised and says so in the docstring — the record is a convenience pointer for `crewai eval`, never worth failing a run over.

lib/crewai/tests/telemetry/test_last_run.py
joaomdmoura: Added in c93892090. Three writers race on one project with a widened window inside the read-decide-replace region; the test asserts no two are ever inside it and that the latest-finished run is what remains. I checked it actually catches the bug: with the lock disabled it fails, with it it passes.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,143 @@
...
+    """Off under the test suite, so kickoff-level tests leave no file behind,
+    and off inside a deployment — the platform's integration token marks one —
+    so a run there never writes a file the platform never reads. (A deployment
+    normally never gets here at all: the host binds the trace before crewAI
+    would start its own; this is the guard for a container that did not.)"""
+    from crewai.context import get_platform_integration_token
+
+    if os.environ.get("CREWAI_TESTING", "").lower() == "true":
+        return False
+    return get_platform_integration_token() is None
```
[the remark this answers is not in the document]
joaomdmoura: Correct, and it was my regression — removed in c93892090. `crewai create crew` writes that token into the project's own `.env` (create_json_crew.py:1183) and `crew_base.py:130` loads it, so the guard read "this developer uses platform tools". I did not swap in a real deployment marker either: a deployment cannot reach this code, because the host sets `crew.tracing = False` / `flow.tracing = False` at all five kickoff sites, so `_start_tracing` returns before the exporter exists. The regression test is in a2f455e7f and now exercises the real guard rather than the fixture's stub.

lib/crewai/src/crewai/telemetry/tracing/last_run.py
```diff
@@ -0,0 +1,179 @@
...
+        yield
+        return
+    try:
+        handle = open(directory / LOCK_FILE, "a")
+    except OSError:
+        yield
+        return
+    try:
+        fcntl.flock(handle, fcntl.LOCK_EX)
+        yield
```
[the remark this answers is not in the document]
joaomdmoura: Real, and fixed in a2f455e7f. A refused `flock` — some network mounts — escaped `_exclusive`, the outer handler deleted the temporary file, and the run was not recorded at all. It is now logged and the write goes ahead unserialised, which is what already happened when the lock file could not be opened or `fcntl` is absent. Tests cover both an `OSError` and a `PermissionError` from `flock`.

lib/crewai/tests/telemetry/test_last_run.py
```diff
@@ -0,0 +1,193 @@
...
+    """`crewai create crew` writes CREWAI_PLATFORM_INTEGRATION_TOKEN into the project's
+    own .env, so it says "this developer uses platform tools", never "this is a
+    deployment". Reading it as a deployment marker would leave those users with no
+    run for `crewai eval` to find."""
+    monkeypatch.setenv("CREWAI_PLATFORM_INTEGRATION_TOKEN", "a token from the project's .env")
+
+    path = last_run.record_last_run(execution_id="local", tier="authenticated", started_at_ns=None, finished_at_ns=None, amp_base_url=None)
+
+    assert path is not None
+    assert last_run.read_last_run(project)["execution_id"] == "local"
```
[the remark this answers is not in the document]
joaomdmoura: Right — the test proved nothing. Fixed in a2f455e7f: it stubs only `project_dir`, clears `CREWAI_TESTING`, and asserts `recording_enabled()` is True with the platform token set, so the real guard is what runs.

lib/crewai/src/crewai/telemetry/tracing/grants.py
```diff
@@ -172,18 +166,24 @@ def create(self, execution_uuid: str) -> TraceGrant:
...
+    (``last_run``) so ``crewai eval`` can find the run without an id being
+    shown to anyone.
+    """
 
     def __init__(self, client: TraceGrantClient, grant: TraceGrant):
         self._client = client
         self._grant = grant
         self._lock = Lock()
         self._delegate = self._exporter(grant)
-        self._trace_url = grant.trace_url
```
lorenzejay, line 182: we need this url i believe to display the user to see the traces

lorenzejay (changes requested):

lib/crewai/src/crewai/telemetry/tracing/grants.py
```diff
@@ -250,32 +268,27 @@ def _export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
...
-            content.append("Traces exported\n", style="green bold")
-            content.append("Execution trace ID: ", style="white")
-            content.append(self._grant.execution_uuid, style="green")
-            if self._trace_url:
-                content.append("\n\nView traces:\n", style="white bold")
-                content.append(
-                    self._trace_url,
-                    style=Style(color="cyan", underline=True, link=self._trace_url),
-                )
-            ConsoleFormatter(verbose=True).print_panel(
```
lorenzejay, line 276: yeah we need this part ^

lorenzejay (changes requested):

lib/crewai/src/crewai/telemetry/tracing/grants.py
joaomdmoura: Agreed — restored in a2d93e855. `self._trace_url` is back on the exporter, and a renewed grant still upgrades it when the first one came without a URL.

lib/crewai/src/crewai/telemetry/tracing/grants.py
joaomdmoura: Done in a2d93e855, in a lighter form than the panel that was there.

One line, `View traces: <url>`, printed once by `record_export` — the same moment the run is recorded — and only for a run exported whole, since a partial one is not worth pointing at. Still silent when AMP granted no viewer URL, under the TUI, and wherever tracing messages are suppressed.

What stays gone is the part João asked to remove: the execution id and the `Traces exported` panel around it. `crewai eval` reads the id from `.crewai/last_run.json`, so nobody needs to copy it. And there is no double link on a first run — the legacy first-run panel is skipped once a kickoff owns the session (trace_listener.py:216-219).

Checked on a real crew run against a local stand-in AMP: stdout is exactly that one line, and the id appears nowhere.

lorenzejay (approved):

PR #4541, opened 2026-02-20
fix: prevent writer starvation in RWLock

Waqar53:
Add _waiting_writers counter so new readers block when writers are waiting for the lock. Without this, a continuous stream of readers can indefinitely starve writers.

Changes:
- Add _waiting_writers field initialized to 0
- r_acquire now blocks when _waiting_writers > 0
- w_acquire increments _waiting_writers before waiting, decrements in finally block after acquiring
- Add test_writer_not_starved_by_continuous_readers test

This is standard practice for production RWLock implementations and does not change behavior when there are no waiting writers.

---

> [!NOTE]
> **Medium Risk**
> Changes core concurrency/locking behavior and may affect throughput/latency or expose subtle deadlocks under edge scheduling, though the change is small and covered by a new regression test.
> 
> **Overview**
> Prevents writer starvation in `RWLock` by tracking `_waiting_writers` and blocking new readers in `r_acquire()` whenever a writer is queued.
> 
> Updates `w_acquire()` to increment/decrement the waiting-writer counter around the wait loop, and adds a regression test (`test_writer_not_starved_by_continuous_readers`) that simulates continuous readers and asserts a writer acquires the lock within a timeout.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 43706a59ad40dfc3ff0790c1a5907c7feb7fac30. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #7649, opened 2026-09-20
feat(cli): crewai eval evaluates the last traced run through AMP

joaomdmoura:
## Related issue

Pairs with #7648 (crewAI records the last traced run in `.crewai/last_run.json`). Independent branches; this one only reads the file.

## Summary

- New command `crewai eval`: evaluates the project's last traced run through AMP, or `--run EXECUTION_ID` for another run. It asks AMP (`POST /crewai_plus/api/v1/tracing/evaluations`) with the saved `crewai login` when there is one, prints and opens the URL AMP answers with, waits for the verdict and prints it (goal gate and the four grades). Exit 1 only when the evaluation itself failed.
- No traced run recorded: it offers to turn tracing on for the project (`CREWAI_TRACING_ENABLED=true` in `.env`, written with `set_key` so nothing else in the file moves) and run the crew now; without a terminal, or declined, it prints the three steps.
- AMP's refusals are printed in AMP's own words: a run that needs an account (401 `account_required`), a refused credential, a run AMP does not hold (404), rate limiting (429 with `Retry-After`), any other status with its message.
- The two AMP calls live on the CLI's `PlusAPI` subclass, so no crewai-core release is needed.
- Who may evaluate what is AMP's decision (crewai-plus #4541 and the evaluations endpoint that follows): an anonymous run once without an account, then it needs one; a run traced while logged in for that organization's members; a deployment execution for members who may see its traces.

## Verification

- [x] `lib/cli/tests/test_eval_crew.py` (happy path, `--run`, anonymous caller, failed evaluation, every refusal, the offer to run with tracing on, the decline, a run that leaves no record, the CLI mapping and help, the record reader), `test_plus_api.py` (the two calls), `test_cli.py` — 87 passed. AMP is a scripted double; nothing reaches the network.
- [x] `ruff check`, `ruff format --check`, `mypy` (strict, lib/cli) clean.

## Additional context

- The AMP side of `/tracing/evaluations` is the next PR on crewai-plus; until it is deployed the command gets AMP's 404 and says so.
- Docs (`docs/edge/en`) for the command follow once the whole flow is live, so the page describes what actually works.
- `crewai test` (the old "Test Crew Performance") is deprecated later; `crewai eval` sits beside it for now.

---

> [!NOTE]
> **Medium Risk**
> New CLI surface with AMP HTTP calls and credential routing rules; mistakes could leak tokens or mis-handle auth, but behavior is heavily tested and defaults to anonymous access when unsure.
> 
> **Overview**
> Adds **`crewai eval`** to evaluate a traced crew run through CrewAI AMP: it reads `.crewai/last_run.json` (or **`--run EXECUTION_ID`**), starts an evaluation via AMP, opens the report URL, polls until finished, and prints the goal gate plus grades. Exit **1** when the evaluation fails, not when the gate fails.
> 
> When no run is recorded, the CLI can prompt to set **`CREWAI_TRACING_ENABLED=true`** in `.env`, run the crew, then evaluate; non-interactive paths get step-by-step instructions instead.
> 
> **`PlusAPI`** gains **`create_evaluation`** / **`get_evaluation`** on the tracing evaluations API (CLI-only, no crewai-core bump). Auth is guarded: the saved login is sent only to trusted AMP origins over HTTPS (or loopback HTTP); other targets are called anonymously, with clear errors for AMP refusals and broken credential storage.
> 
> Broad **`test_eval_crew.py`** coverage includes happy path, security, polling, and CLI wiring; **`test_plus_api.py`** covers the new API methods.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f8be9e538e24b4c478efe60536a9e42a4c5a25e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -0,0 +1,259 @@
...
+    get_or_create_project_id()
+    record = read_last_run() or {}
+    execution_id = run_id or record.get("execution_id")
+    if execution_id is None:
+        execution_id = _run_now_or_explain()
+        record = read_last_run() or {}
+
+    # The record names the AMP the run was traced to; a run named by hand goes to the configured AMP.
+    amp_base_url = None if run_id else record.get("amp_base_url")
+    client = PlusAPI(api_key=saved_login(), base_url=amp_base_url or None)
```
iris-clawd: Can we validate the recorded AMP origin before pairing it with `saved_login()`? Core sends the bearer token to this base URL, so a modified `.crewai/last_run.json` can disclose it. Please derive trust from configured AMP origins or withhold credentials pending explicit confirmation.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -0,0 +1,259 @@
...
+        )
+    steps = (
+        "No traced run is recorded in this project. Turn tracing on and run the crew, "
+        f"then come back:\n  1. add {TRACING_ENV_VAR}=true to .env\n  2. crewai run\n  3. crewai eval"
+    )
+    if is_dmn_mode_enabled() or not sys.stdin.isatty():
+        console.print(steps, style="yellow")
+        raise SystemExit(1)
+    if not click.confirm(
+        "No traced run is recorded in this project. Turn tracing on and run the crew now?",
```
iris-clawd: Can this prompt state that tracing remains enabled and require an explicit yes? It permanently writes the override and then runs project code, while the current default accepts both effects on Enter.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -0,0 +1,259 @@
...
+                if misses >= POLL_RETRIES:
+                    _refused(response, subject)
+                time.sleep(POLL_SECONDS)
+                continue
+            if response.status_code != 200:
+                _refused(response, subject)
+            misses = 0
+            payload = _payload(response) or {}
+            status = payload.get("status")
+            if status in FINISHED:
```
iris-clawd, line 289: Can we validate the completed payload before returning it? A missing verdict currently prints `INCONCLUSIVE` and exits 0, while non-dict verdict or grades values raise `AttributeError`. Please fail with a protocol error instead.

lib/cli/tests/test_eval_crew.py
```diff
@@ -0,0 +1,337 @@
...
+def test_the_last_run_is_evaluated_the_url_opened_and_the_verdict_printed(project, monkeypatch, capsys):
+    directory, opened = project
+    record_last_run(directory)
+    amp = install(monkeypatch, FakeAMP(statuses=[httpx.Response(200, json={"id": "ev-1", "status": "running"}), done()]))
+
+    eval_module.eval_crew()
+
+    out = capsys.readouterr().out
+    assert amp.api_key == "login-token"
+    assert amp.base_url == "https://amp.test"  # the AMP the run was traced to, off the record
```
iris-clawd: Please add regressions for an untrusted recorded AMP origin with a saved token and for `done` payloads with missing or non-dict verdict/grades. Without them, the credential disclosure and malformed-response failures can return unnoticed.

iris-clawd (changes requested):
I could not verify the paired AMP endpoint or the last-run writer here. The retry and interruption handling look right.

The notes inline on eval_crew.py are what I'd want addressed before this merges. The other inline notes are small.

Someone else should sign off on this too: `lib/cli/src/crewai_cli/cli.py` adds the public `crewai eval --run EXECUTION_ID` contract.

lib/cli/src/crewai_cli/eval_crew.py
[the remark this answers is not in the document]
joaomdmoura: Fixed in 3f21d108d: the client is built from the configured AMP only (`CREWAI_PLUS_URL`, the saved settings, app.crewai.com) — the record's `amp_base_url` is never handed to `PlusAPI`. The project's `.env` is loaded first, as `crewai run` loads it, so the configured AMP is the one the run was traced to; a record naming another address gets a one-line note and no credential goes there. Regression: a record pointing at a foreign origin with a saved token.

joaomdmoura: Done in 3f21d108d — trust comes from the configured AMP origins only; the recorded address is never used for a request. Loading the project's `.env` first (as `crewai run` does) is what closes the divergence the earlier review found, without a confirmation prompt. A record naming a different address prints a note and the credential stays with the configured AMP. Test added.

lib/cli/src/crewai_cli/eval_crew.py
joaomdmoura: Done in 3f21d108d: the prompt now reads 'Turn tracing on (CREWAI_TRACING_ENABLED=true stays in .env) and run the crew now?' with `default=False` — Enter declines and prints the steps; only an explicit yes writes and runs. The test pins both the text and the default.

lib/cli/src/crewai_cli/eval_crew.py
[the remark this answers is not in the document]
joaomdmoura: Fixed in 3f21d108d, a step further than the suggestion: `_well_formed_verdict` requires a dict with a string `gate` and a `grades` dict whose values are ints or null; anything else on a `done` answer is 'AMP answered done without a verdict (protocol error)', exit 1. Seven malformed shapes are tested.

joaomdmoura: Done in 3f21d108d — `_well_formed_verdict` validates the `done` payload before `_wait` returns it (gate a string, grades a dict of ints/null); a missing or malformed verdict is a protocol error, exit 1, and `_print_verdict` no longer touches an unvalidated shape. Seven malformed shapes tested.

lib/cli/tests/test_eval_crew.py
joaomdmoura: Added in 3f21d108d: an untrusted recorded origin with a saved token (the token goes to the configured AMP, a note is printed), the same-AMP case through the project's `.env`, and seven `done` payloads with a missing or malformed verdict/grades. 59 tests in the two files.

lib/cli/src/crewai_cli/eval_crew.py
joaomdmoura: Owner's decision (João, today): the confirm is y/n with Y as the default — d849f9d7d puts it back. The prompt keeps the explicit wording: tracing stays on in `.env` and the crew runs now, so Enter accepts something the line just described.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -0,0 +1,291 @@
...
+        "No traced run is recorded in this project. Turn tracing on and run the crew, "
+        f"then come back:\n  1. add {TRACING_ENV_VAR}=true to .env\n  2. crewai run\n  3. crewai eval"
+    )
+    if is_dmn_mode_enabled() or not sys.stdin.isatty():
+        console.print(steps, style="yellow")
+        raise SystemExit(1)
+    if not click.confirm(
+        f"No traced run is recorded in this project. Turn tracing on ({TRACING_ENV_VAR}=true "
+        "stays in .env) and run the crew now?",
+        default=True,  # João, 2026-09-20: y/n with Y as the default — the prompt says what Enter does
```
iris-clawd, line 207: Can we require an explicit yes instead of accepting Enter? This permanently enables tracing and immediately executes project code, so the safe default should leave both effects off.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -0,0 +1,291 @@
...
+        return None
+    loaded["execution_id"] = str(loaded["execution_id"])
+    return loaded
+
+
+def saved_login() -> str | None:
+    """The `crewai login` token, or None: AMP then treats the caller as anonymous."""
+    try:
+        return get_auth_token()
+    except Exception:
```
iris-clawd: Can we catch only `AuthError` and surface unexpected token-store failures? Swallowing every exception silently turns an authenticated evaluation into anonymous access, potentially consuming the anonymous allowance or producing a misleading refusal.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -0,0 +1,291 @@
...
+    _print_verdict(finished, url)
+    if finished.get("status") != "done":
+        raise SystemExit(1)
+
+
+def _load_project_env() -> None:
+    """The project's .env, as `crewai run` loads it — so CREWAI_PLUS_URL here is the one the run used."""
+    env_file = Path.cwd() / ".env"
+    if env_file.is_file():
+        load_dotenv(env_file, override=True)
```
iris-clawd, line 153: Can we avoid pairing a project-controlled `CREWAI_PLUS_URL` with the saved global login token without explicit trust? Running this in an untrusted project sends that bearer token to its chosen host. Can we use trusted saved settings or withhold credentials for project overrides?

iris-clawd (dismissed):
I could not verify the paired AMP endpoint, last-run writer, or self-hosted endpoint availability. The recorded-origin fix and malformed-response handling look solid.

The notes inline on eval_crew.py are what I'd want addressed before this merges.

Someone else should sign off on this too: introduces public `crewai eval --run EXECUTION_ID` behavior and persists `CREWAI_TRACING_ENABLED=true` to project `.env`.

lib/cli/src/crewai_cli/eval_crew.py
joaomdmoura: Done in 0f6e499df. Only `AuthError` reads as anonymous now, which matches `tracing_credential()` in the library (grants.py:51-54) and the two other downgrade sites. Anything else — a rotated Fernet key, a half-written store, a directory a past `sudo crewai login` left owned by root — is reported with its class and message and a pointer to `crewai login`, rather than silently spending the run's one anonymous read and then refusing a user who believes they are logged in. Tests cover both bands.

lib/cli/src/crewai_cli/eval_crew.py
joaomdmoura: Fixed in 0f6e499df, keeping the self-hosted case working. The request still follows the project's `CREWAI_PLUS_URL`, because that is how a self-hosted project is wired and the run really was traced there. The credential no longer does: it goes only to an origin this machine is logged in to — `crewai enterprise configure`'s saved `enterprise_base_url`, an address already exported in this shell (snapshotted before `.env` is loaded, since that load uses `override=True`), or app.crewai.com. Anywhere else the run is read anonymously and the command says so, naming `crewai enterprise configure`.

That keeps the supported factory path a no-op, since `crewai enterprise configure <url>` is what makes a self-hosted login work at all and it writes that URL into settings in the same step.

Worth saying plainly: this closes the smaller hole. `crewai run` sends the same saved login to the same `.env`-chosen URL (run_crew.py:313-317 → grants.py:114 with `base_url=None`), and the same `.env` can flip `CREWAI_TRACING_ENABLED` — so a hostile project already exfiltrates the token on a plain run, before `crewai eval` exists. I did not widen this PR to that; it deserves its own.

lib/cli/src/crewai_cli/eval_crew.py
joaomdmoura: Keeping Enter as yes, on the owner's call today: y/n with Y as the default. The prompt names both effects before Enter accepts them — that tracing stays on in `.env` and that the crew runs now — and declining prints the three manual steps. I had it your way earlier in the day and reverted it when João asked.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -0,0 +1,346 @@
...
+        Settings().enterprise_base_url,
+        DEFAULT_CREWAI_ENTERPRISE_URL,
+    )
+    return {origin for origin in map(_origin, candidates) if origin}
+
+
+def _origin(url: str | None) -> str | None:
+    parsed = urlparse(str(url or ""))
+    return (
+        f"{parsed.scheme}://{parsed.netloc}".lower()
```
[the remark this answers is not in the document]
joaomdmoura: Right — fixed in 0185dc7a2. The credential now needs both: an origin this machine is logged in to AND an encrypted connection. `_encrypted()` allows HTTPS, or plain HTTP to this machine (localhost, its subdomains, loopback addresses), which is the same rule `TraceGrantClient` already applies to collector grants (grants.py:91-106) — so local development against a dev AMP keeps working and `http://amp.internal` does not. Anywhere else the run is read anonymously and the line says which of the two reasons applied. Tests cover both paths plus the rule over nine origins.

joaomdmoura:
On the outside-diff note about the grade domain: fixed in 0185dc7a2. `isinstance(grade, int)` accepted `True` (a bool is an int in Python) and out-of-range values like 6, both of which would have printed as real grades and let the command exit 0. A grade is now an exact `int` in 1..5, or null, and the malformed-verdict cases gained `True`, 6, 0 and 4.5.

lorenzejay (approved):

PR #7654, opened 2026-09-20
fix(tools): close S3 response body in S3ReaderTool

### LINKED ISSUE #7646: [BUG] S3ReaderTool leaves response bodies open
## Description

`S3ReaderTool` reads the `StreamingBody` returned by `boto3.client("s3").get_object()` but never closes it. The body owns an HTTP response/connection; leaving it open makes cleanup dependent on garbage collection and can retain pooled connections across repeated tool calls.

The leak is also present on failure paths. For example, if reading succeeds but UTF-8 decoding raises, the method exits without releasing the body.

> This report was prepared with AI assistance. I cannot apply repository labels as an external contributor; please add the required `llm-generated` label.

## Steps to reproduce

1. Mock the S3 client so `get_object()` returns a closeable body.
2. Call `S3ReaderTool._run("s3://bucket/key.txt")`.
3. Observe that `body.read()` is called but `body.close()` is not.
4. Repeat with invalid UTF-8 or a body whose `read()` raises; cleanup is still skipped.

## Expected behavior

Once `get_object()` succeeds, its response body should be closed exactly once whether reading and decoding succeed or raise. Existing result and exception behavior should otherwise remain unchanged.

## Screenshots/Code snippets

Current path:

```python
response = s3.get_object(Bucket=bucket_name, Key=object_key)
result = response["Body"].read().decode("utf-8")
return result
```

## Operating System

macOS Sonoma

## Python Version

3.13 (supported by the repository; not listed in the issue template)

## crewAI Version

Current `main` (`0374c631297f`)

## crewAI Tools Version

Current `main` (`0374c631297f`)

## Virtual Environment

Venv (`uv`)

## Evidence

Deterministic offline tests with a mocked S3 response show zero `close()` calls on success, decode failure, and read failure. Wrapping read/decode in `try/finally` makes all three paths close the body exactly once.

## Possible Solution

Store `response["Body"]` in a local variable and close it in `finally` around `read().decode("utf-8")`.

## Additional context

No matching open issue or pull request was found by searching for `S3ReaderTool`, `StreamingBody`, and response-body cleanup. The change does not alter existing `ClientError` formatting or decoding behavior.

FanouZeng-TT:
## Related issue

Fixes #7646

## Summary

- close the S3 `get_object()` response body in `S3ReaderTool` so pooled HTTP connections are released deterministically
- wrap read/decode in `try/finally` so the body is closed exactly once on success and on every failure path
- add offline regression tests with a mocked S3 client covering success, decode failure, and read failure

The body previously stayed open on every path, leaving cleanup to garbage collection and retaining pooled connections across repeated tool calls.

## Verification

- `uv run --frozen pytest lib/crewai-tools/tests/tools/s3_reader_tool_test.py -q` — red baseline before the fix: 3 failed (success, decode failure, read failure), 1 passed (ClientError behavior baseline); after the fix: 4 passed
- `ruff check` on both changed files — passed
- Tests are fully offline: a fake `boto3`/`botocore` module pair is injected via `monkeypatch.setitem(sys.modules, ...)`, so neither the SDK nor network access is required

## Additional context

This PR was prepared with AI assistance; please apply the `llm-generated` label (external contributors cannot add repository labels).


FanouZeng-TT:
This PR was prepared with AI assistance. Per the contributing requirements it needs the `llm-generated` label — external contributors cannot add repository labels, so could a maintainer apply it? Thanks!

gaoanze888:
Thanks for the fix. #7647 already addresses this same issue (#7646) and additionally keeps a failing `close()` from masking the original result or read/decode exception, which a bare `finally: body.close()` would do. Suggest closing this in favor of #7647 to avoid duplicate maintenance.

FanouZeng-TT:
Agreed — thanks for catching it, and sorry for the duplicate. #7647 predates this one and is the stronger fix: wrapping the `get_object()` body in `try/finally` also handles the case you point out, where a failing `close()` would mask the original exception if `close()` were all that ran. Closing this in favor of #7647.

PR #7663, opened 2026-09-20
fix(tools): keep selenium driver reusable

### LINKED ISSUE #7662: [BUG] SeleniumScrapingTool fails on every call after the first (driver closed per run)
## Description

`SeleniumScrapingTool` closes the browser window after every single call: `_run` ends with `finally: self.driver.close()`. A real chromedriver closes the only window, so the next call to the same tool instance fails with `no such window: target window already closed`. The tool is effectively one-shot, while the documented usage pattern attaches one instance to an agent and lets the crew call it repeatedly.

The class's own cleanup API has the same problem in reverse: `close()` also calls `driver.close()`, which closes a window but leaves the browser and chromedriver processes running.

> This report was prepared with AI assistance. I cannot apply repository labels as an external contributor; please add the required `llm-generated` label.

## Steps to reproduce

1. Create `SeleniumScrapingTool()` and call `_run(website_url=...)` — succeeds.
2. Call the same instance again.
3. The second call returns `Error scraping website: no such window: target window already closed`, because the first call closed the only window.

## Expected behavior

A tool instance should stay usable across calls, matching the documented agent-attached usage. Cleanup should be explicit: `close()` ends the session (`quit()`), terminating browser and driver processes.

## Screenshots/Code snippets

```python
finally:
    if self.driver is not None:
        self.driver.close()  # closes the only window -> next call fails
...
def close(self) -> None:
    if self.driver is not None:
        self.driver.close()  # window closed, process still running
```

## Operating System

macOS Sonoma

## Python Version

3.13 (supported by the repository; not listed in the issue template)

## crewAI Version

Current `main` (`0374c631297f`)

## crewAI Tools Version

Current `main` (`0374c631297f`)

## Virtual Environment

Venv (`uv`)

## Evidence

A deterministic fake driver that models chromedriver window semantics reproduces the failure on current `main`: the second `_run` returns the "no such window" error. After removing the per-run close and switching `close()` to `quit()`, both calls succeed and `close()` records a `quit`.

## Possible Solution

Delete the `finally` block in `_run` (the driver's `get()` reuses the same window, so no window accumulation) and make `close()` call `driver.quit()`.

## Additional context

No open pull request touches this tool; a related closed PR (#6870) was closed for lacking an issue, and open issue #7380 (a vendor proposing a different scraping tool) independently describes the session-survival pain point.

gaoanze888:
## Related issue

Fixes #7662

## Summary

`SeleniumScrapingTool` closed the browser window after every `_run`, so any second call to the same instance failed with `no such window: target window already closed` — contradicting the documented single-instance agent usage. `close()` also only closed a window, leaving the browser and chromedriver processes alive.

Changes:

- `_run` no longer closes the driver; `driver.get()` reuses the same window across calls.
- `close()` now calls `driver.quit()`, ending the session and terminating the process.
- Existing tests that asserted the per-run close now assert the driver is left untouched; two new regression tests model real chromedriver window semantics.

Note for one-shot script users: cleanup is now explicit via `close()` instead of implicit per-call window closing.

> This PR was prepared with AI assistance. I cannot apply repository labels as an external contributor; please add the required `llm-generated` label.

## Verification

```text
PYTHONPATH=lib/crewai-tools/src:lib/crewai/src:lib/crewai-core/src \
  ../crewAI-contrib/.venv/bin/pytest \
  lib/crewai-tools/tests/tools/selenium_scraping_tool_test.py \
  -q -n 0 -o addopts=''
# 10 passed

../crewAI-contrib/.venv/bin/ruff check <changed files>
../crewAI-contrib/.venv/bin/ruff format --check <changed file>
../crewAI-contrib/.venv/bin/python -m compileall -q <changed file>
git diff --check
```

Mutation check: restoring the per-run `finally: driver.close()` makes 7 tests fail (the 5 flipped assertions plus both new regression tests).

`mypy` reports no errors in the changed file; repo-wide mypy count is identical to baseline.

## Additional context

This changes observable behavior only in the failure mode that made the tool one-shot; success-path outputs are unchanged.


Vidit-Ostwal (approved):
LGTM

PR #7672, opened 2026-09-21
fix(crew): preserve replay data after async for-each runs

### LINKED ISSUE #6704: [BUG] kickoff_for_each_async / akickoff_for_each wipe replay data on every run

wangtaotaotao95:
## Related issue

Fixes #6704

## Summary

- remove the trailing replay-storage reset shared by kickoff_for_each_async and akickoff_for_each
- retain the existing concurrent behavior without promising deterministic last-writer ordering
- exercise both public entrypoints against the real SQLite-backed TaskOutputStorageHandler
- preserve Deep Santoshwar authorship by cherry-picking the original #6705 commit and adding the missing native-async coverage separately

This refresh supersedes stale PR #6705 while retaining its original commit and credit.

## Verification

Commands:

- uv run --extra litellm pytest lib/crewai/tests/test_crew.py -q (137 passed, 1 skipped)
- uv run ruff check lib/crewai/src/crewai/crews/utils.py lib/crewai/tests/test_crew.py
- uv run ruff format --check lib/crewai/src/crewai/crews/utils.py lib/crewai/tests/test_crew.py
- uv run mypy lib/crewai/src/crewai/crews/utils.py
- git diff --check

## Additional context

The copies execute concurrently and share the replay store, so the surviving copy is intentionally unspecified. This change only fixes the unconditional wipe-to-empty after all copies complete.


wangtaotaotao95:
Per .github/CONTRIBUTING.md, this PR requires the llm-generated label. I attempted to add it, but GitHub denied AddLabelsToLabelable for my account. Maintainers, please apply the label.

wangtaotaotao95:
The latest `Run Tests` workflow is red because GitHub canceled the only reported failing shard after its test process had completed successfully (`716 passed, 5 skipped`). Fail-fast then canceled the remaining matrix jobs. I cannot rerun the upstream workflow without repository admin permission; could a maintainer please rerun failed jobs? The relevant run is https://github.com/crewAIInc/crewAI/actions/runs/35559196017.

wangtaotaotao95:
Branch updated to current `main`. The refreshed CI failure is unrelated to this PR: `tests/telemetry/test_flow_telemetry.py::test_infrastructure_flows_do_not_pollute_outcome_signals` failed under Python 3.13 after 717 tests passed in that shard. I do not have repository Actions admin rights to rerun the failed matrix; could a maintainer rerun it?

PR #7675, opened 2026-09-21
fix(openai): collapse multimodal system instructions

### LINKED ISSUE #7674: [BUG] OpenAI Responses instructions stringify multimodal system content as Python repr
### Description

OpenAICompletion._prepare_responses_params converts non-string system message content with str(content). LLMMessage.content may be a list of multimodal content parts, so the Responses API instructions receive a Python repr such as [{"type": "text", ...}] instead of the actual text.

This violates the repository message-content contract and leaks Python syntax plus image metadata into the instruction string.

### Reproduction

Create an OpenAICompletion with api set to responses, then call _prepare_responses_params with a system message whose content contains one text part and one image_url part.

Current result:

    params["instructions"] == "[{'type': 'text', 'text': 'Follow the receipt policy.'}, {'type': 'image_url', ...}]"

### Expected behavior

System content should be collapsed with the shared message_content_text helper. The example should produce:

    Follow the receipt policy.

Image-only system content should use the existing [multimodal content] fallback rather than a Python repr.

### Evidence

A deterministic unit test on current main (0374c631) fails because the instruction string contains the full list repr. No network request or credentials are required.

### Proposed fix

Use message_content_text(message) when extracting system instructions in _prepare_responses_params, matching the other CrewAI paths that already honor multimodal LLMMessage content.

This report and the proposed regression test were prepared with AI assistance. Please add the required llm-generated label.

wangtaotaotao95:
## Related issue

Fixes #7674

## Summary

OpenAI Responses requests previously converted non-string system content with `str(content)`. For multimodal content arrays, that placed a Python list/dict representation—including image metadata—into `instructions`.

This change reuses the shared `message_content_text` helper so only textual content parts become Responses instructions. It also adds a regression test proving image content is excluded rather than stringified.

## Verification

Commands run:

- `uv run --extra litellm pytest lib/crewai/tests/llms/openai/test_openai.py -q` — 108 passed, 1 skipped
- `uv run ruff check ...` — passed
- `uv run ruff format --check ...` — passed

## Additional context

The patch and tests were prepared with AI assistance and manually reviewed and verified.


wangtaotaotao95:
This contribution was prepared with AI assistance. I attempted to add the required `llm-generated` label, but fork contributors do not have permission to apply repository labels. Could a maintainer please add it?

PR #7687, opened 2026-09-21
fix(memory): handle Windows OSError in QdrantEdgeStorage orphaned-shard cleanup

### LINKED ISSUE #7684: [BUG] QdrantEdgeStorage crashes on Windows during orphaned-shard cleanup (os.kill signal 0 unsupported)
### Description

QdrantEdgeStorage._cleanup_orphaned_shards() (lib/crewai/src/crewai/memory/storage/qdrant_edge_storage.py)
uses os.kill(pid, 0) — the POSIX idiom for checking whether a process is alive without
sending a real signal — and only catches ProcessLookupError (dead process) and
PermissionError (alive, inaccessible).

On Windows, sending signal 0 isn't supported the same way, and Python's os.kill raises a
generic OSError instead of ProcessLookupError. That isn't caught, so it propagates and
crashes QdrantEdgeStorage.__init__ for any Windows user whose storage directory contains a
leftover worker-<pid> shard from a previous process (an ordinary situation after a crash or
force-kill).

Actual behavior: QdrantEdgeStorage initialization crashes with an unhandled OSError
(see traceback under "Evidence / logs / error messages" below) instead of silently
skipping or cleaning up the orphaned shard.

### Steps to Reproduce

Run the existing test on Windows:

\`\`\`
pytest lib/crewai/tests/memory/test_qdrant_edge_storage.py::test_orphaned_shard_cleanup -v
\`\`\`

### Expected behavior

Orphaned-shard cleanup should detect that the owning process is no longer running
(on Windows as well as POSIX) and clean up the shard without raising, matching the
behavior already verified by the existing test test_orphaned_shard_cleanup.

### Screenshots/Code snippets

Not applicable — see the traceback under "Evidence / logs / error messages" below.

### Operating System

Windows 11

### Python Version

3.12

### crewAI Version

1.15.22

### crewAI Tools Version

1.15.22

### Virtual Environment

Venv

### Evidence

Reproduced via the existing test suite on Windows:

    uv run pytest lib/crewai/tests/memory/test_qdrant_edge_storage.py::test_orphaned_shard_cleanup -v

Traceback:

```
lib\crewai\tests\memory\test_qdrant_edge_storage.py:284: in test_orphaned_shard_cleanup
    s2 = _make_storage(str(base))
         ^^^^^^^^^^^^^^^^^^^^^^^^
lib\crewai\tests\memory\test_qdrant_edge_storage.py:27: in _make_storage
    return QdrantEdgeStorage(path=path, vector_dim=vector_dim)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
lib\crewai\src\crewai\memory\storage\qdrant_edge_storage.py:143: in __init__
    self._cleanup_orphaned_shards()
lib\crewai\src\crewai\memory\storage\qdrant_edge_storage.py:822: in _cleanup_orphaned_shards
    os.kill(pid, 0)
E   OSError: [WinError 11] An attempt was made to load a program with an incorrect format
```

The test simulates an orphaned shard using a fake, guaranteed-dead PID, and expects
`_cleanup_orphaned_shards` to detect and clean it up. On Windows, `os.kill(pid, 0)`
raises a generic `OSError` rather than `ProcessLookupError`, so it's uncaught and
crashes storage initialization.

### Possible Solution

Yes — os.kill(pid, 0) needs a cross-platform liveness check. On Windows this can be done via ctypes.OpenProcess/CloseHandle instead of relying on POSIX signal semantics, following the same sys.platform != "win32" pattern already used in lancedb_storage.py in this codebase. I have a fix ready and will open a PR referencing this issue.

### Additional context

- Actual Python version tested: 3.13.6 (not listed in the dropdown above; closest
  available option, 3.12, was selected instead).
- This is invisible in CI: .github/workflows/tests.yml runs tests on ubuntu-latest
  only, with no Windows job in the matrix, so this code path has never been
  exercised on Windows in CI.
- Impact: any Windows user running QdrantEdgeStorage will hit this crash on
  initialization if their storage directory contains a leftover worker-<pid>
  shard directory from a previous process (e.g. after a crash or force-kill) —
  an ordinary, not-edge-case situation.

josephkamau32:
Fixes #7684

## What
`QdrantEdgeStorage._cleanup_orphaned_shards()` used `os.kill(pid, 0)` — the POSIX idiom
for probing process liveness without sending a real signal — catching only
`ProcessLookupError` and `PermissionError`. On Windows, signal `0` isn't supported the
same way and raises a generic `OSError` instead, which wasn't caught, crashing
`QdrantEdgeStorage.__init__` for any Windows user with a leftover `worker-<pid>` shard
from a previous process.

## Fix
Added a small cross-platform `_pid_is_alive()` helper that uses
`ctypes.OpenProcess`/`CloseHandle` on Windows and the existing `os.kill(pid, 0)` logic
on POSIX, following the `sys.platform != "win32"` convention already used elsewhere in
this package (`lancedb_storage.py`).

## Testing
```
uv run pytest lib/crewai/tests/memory/test_qdrant_edge_storage.py -n 0 -v
```
`test_orphaned_shard_cleanup` — the test that reproduces this bug — now passes on
Windows. All other tests in the file are unaffected (18 passed, the one pre-existing
unrelated failure, `test_memory_with_qdrant_edge`, is a separate Windows/pytest-recording
network-guard issue unrelated to this change).

```
uv run mypy lib/crewai/src/crewai/memory/storage/qdrant_edge_storage.py
Success: no issues found in 1 source file

uv run ruff check lib/crewai/src/crewai/memory/storage/qdrant_edge_storage.py
All checks passed!
```

No test assertions were changed — the existing test already encoded the correct
expected behavior; this fix makes it pass on Windows without touching the test itself.


josephkamau32:
Pushed a follow-up commit (c0e0c4f) addressing CodeRabbit's review:

- The Windows liveness check now distinguishes ERROR_ACCESS_DENIED (process
  is alive, just inaccessible) from ERROR_INVALID_PARAMETER (no such
  process), using ctypes.get_last_error() rather than treating every
  OpenProcess failure as "dead."
- Added three regression tests (test_windows_access_denied_pid_is_treated_as_alive,
  test_windows_invalid_parameter_pid_is_treated_as_dead,
  test_windows_live_process_returns_true_and_closes_handle) that mock the
  Win32 API directly so they run on Linux CI too, not just on Windows.

CodeRabbit's re-review confirms no actionable comments remain.

PR #7644, opened 2026-09-20
chore(deps): widen crewai-cli uv pin

FanouZeng-TT:
## Related issue

Fixes #7633

## Summary

- align `crewai-cli`'s runtime uv requirement with the workspace's existing `uv>=0.11.15,<1` constraint
- remove the `<0.12` ceiling imposed by `uv~=0.11.6`
- keep the workspace lock metadata in sync

The previous CLI requirement allowed uv versions below the repository's GHSA-4gg8-gxpx-9rph security floor and capped every consumer environment below uv 0.12. Reusing the workspace constraint preserves uv as an installed CLI prerequisite while applying the security floor and compatibility range already selected by this repository.

## Verification

- [ ] Tests added or updated for the changed behavior — not applicable; this is dependency metadata only, verified through lock resolution and built wheel metadata
- [x] Relevant tests and quality checks pass locally

- `uv lock --check` — passed
- `uv run pre-commit run --files lib/cli/pyproject.toml uv.lock` — `uv-lock` passed; Python-only hooks skipped these metadata files
- `uv build --package crewai-cli --out-dir /tmp/crewai-cli-dist` — passed
- built wheel metadata contains `Requires-Dist: uv<1,>=0.11.15`
- `uv export --package crewai-cli --no-dev --no-emit-project --no-annotate --no-header` resolves `uv==0.11.19` under the repository's pinned package cutoff

## Additional context

No functional Python code changed. The generated `uv.lock` diff only updates the `crewai-cli` requirement; it does not upgrade unrelated packages or change the lockfile cutoff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7700, opened 2026-09-22
fix(cli): widen the uv runtime pin to >=0.11.15,<1

### LINKED ISSUE #7633: [BUG] crewai-cli's runtime pin uv~=0.11.6 installs a second, capped uv into every consumer .venv
### Description

`crewai-cli` lists `uv~=0.11.6` in its runtime `dependencies` (`lib/cli/pyproject.toml`, line 24; the 1.15.22 wheel on PyPI carries the same requirement). `crewai` depends on `crewai-cli==1.15.22`, so `uv add crewai` in any project resolves the `uv` PyPI package and drops a second uv binary into that project's `.venv/bin`.

The CLI never imports `uv` as a library. There is no `import uv` / `from uv` anywhere under `lib/cli/src`; the only uses are subprocess argv:

- `install_crew.py:52` — `["uv", "sync"]`
- `run_crew.py:445` — `["uv", "run", "--no-sync", "python", "-c", ...]`
- `run_crew.py:764` — `["uv", "run", script_name]`

Those resolve `uv` through `PATH`. With the venv activated (or `.venv/bin` on `PATH` via direnv and similar), `PATH` finds the venv copy first, so both the CLI's subprocess calls and the user's own `uv ...` commands silently run the pinned uv instead of the one they installed.

`~=0.11.6` means `>=0.11.6,<0.12`. The newest release it can pick is 0.11.33; the current uv release is 0.12.17, and the gap only grows from here. Every project that depends on crewai inherits the cap.

### Steps to Reproduce

1. `mkdir repro && cd repro && uv init --python 3.12`
2. `uv add crewai`
3. `ls -la .venv/bin/uv`
4. `.venv/bin/uv --version`
5. `uv --version` (outside the venv, then again with it activated)

### Expected behavior

No `uv` binary inside `.venv/bin`. `uv --version` reports the same uv whether or not the venv is activated: the one the user installed.

### Screenshots/Code snippets

```console
$ ls -la .venv/bin/uv
-rwxr-xr-x 1 davidlong wheel 39856400 Sep 18 21:19 .venv/bin/uv
$ .venv/bin/uv --version
uv 0.11.33 (fece32fc5 2026-07-28 aarch64-apple-darwin)
$ grep -n -A1 '^name = "uv"$' uv.lock
3305:name = "uv"
3306-version = "0.11.33"
```

Outside the venv, `uv --version` prints whatever you installed (0.12.17 is current). With the venv activated it prints 0.11.33.

### Operating System

macOS Sonoma

### Python Version

3.12

### crewAI Version

1.15.22

### crewAI Tools Version

Not installed (not involved).

### Virtual Environment

Venv (uv-managed `.venv`)

### Evidence

See the snippets above. On `main`:

- `lib/cli/pyproject.toml:24` — `"uv~=0.11.6"`
- `pyproject.toml:276` — `"uv>=0.11.15,<1"` (the workspace's own requirement, set in #5992)

### Possible Solution

At minimum, widen `crewai-cli`'s pin to match the workspace: `uv>=0.11.15,<1`. That removes the `<0.12` cap and adopts the same floor the repo already chose for itself in #5992 (which cites GHSA-4gg8-gxpx-9rph as the reason for 0.11.15). Today the shipped package still allows 0.11.6 through 0.11.14, so it is looser on that advisory than the repo's own tooling.

The stronger option is to drop `uv` from `dependencies` entirely and document it as a prerequisite. The CLI already assumes `uv` is on `PATH` for `uv sync` / `uv run`, and a pip-installed uv inside a venv is rarely what anyone wants. That changes the install story, so I'd treat it as a follow-up.

I'm happy to open the PR for either. The PR template asks first-time contributors to link an open issue, hence filing this first.

### Additional context

None beyond the above.

david-long1:
## Related issue

Fixes #7633

## Summary

`crewai-cli` pins `uv~=0.11.6` (`>=0.11.6,<0.12`) as a runtime dependency. Because `crewai` depends on `crewai-cli`, every project that adds `crewai` inherits that cap: the `uv` PyPI package lands in its `.venv/bin`, and the newest release it can ever pick is the last 0.11.x. The pin is also looser than the workspace's own `uv>=0.11.15,<1` (`pyproject.toml`, set in #5992 for GHSA-4gg8-gxpx-9rph), so the shipped package still admits 0.11.6–0.11.14.

This widens the pin to `>=0.11.15,<1`, the same requirement the workspace already uses for itself. `uv.lock` changes only the `requires-dist` specifier for `crewai-cli`; the resolved `uv` version is unchanged.

## Verification

- [ ] Tests added or updated for the changed behavior — metadata-only change, no code path to test
- [x] `uv lock --check` passes with uv 0.11.3 (the version CI pins in `astral-sh/setup-uv`)
- [x] `uv build --wheel lib/cli` and inspected `METADATA`: `Requires-Dist: uv<1,>=0.11.15`

## Additional context

Dropping `uv` from `dependencies` entirely (the CLI only calls it via `subprocess`, never imports it) would stop the second binary from appearing in consumer venvs at all, but it changes the install story for anyone relying on the venv copy, so it's left for a follow-up as noted in the issue.


david-long1:
Closing — #7644 was opened first with the same change and already references #7633. Backing that one instead.

PR #7701, opened 2026-09-22
fix(cli): print whatever areas the evaluation graded

joaomdmoura:
`crewai eval` prints the verdict AMP returns. It builds that line from a fixed tuple of area names:

```python
for area in ("goal", "quality", "process", "cost")
```

The areas belong to the evaluation, not to the client. A fixed list does two wrong things as soon as the evaluator's vocabulary moves ahead of an installed CLI: it **drops every area it has not heard of**, and it **invents "not measured" for ones that no longer exist**. The user sees one real grade and three phantom blanks, with nothing to say that anything was dropped.

It now prints the areas it was sent, in the order they arrived.

The verdict is still validated exactly as before — `_well_formed_verdict` / `_a_grade`, a grade being an `int` in 1..5 or `null` — so a malformed payload is still a protocol error rather than something printed.

### Why now

crewAIInc/crewai-evals#51 renames the graded areas to CrewAI's own four nouns (goal, tasks, agents, tools) and drops cost from the scorecard, because a 1–5 grade needs a budget to fall short of and a recorded run declares none. This change lands **first** so an installed CLI keeps working through that rollout instead of after it.

One new test covers a payload whose areas the CLI has never seen.

---

> [!NOTE]
> **Low Risk**
> CLI-only display and response validation for `crewai eval`; no auth, persistence, or API contract changes beyond stricter handling of malformed AMP payloads.
> 
> **Overview**
> **`crewai eval`** no longer hard-codes verdict areas (`goal`, `quality`, `process`, `cost`). It prints whatever keys AMP returns in `grades`, in server order, including empty grade maps (gate only, no trailing `·`).
> 
> AMP-sourced strings (areas, URLs, errors, refusal messages) are rendered with **`rich.text.Text`** instead of Rich markup in f-strings, so bracket syntax in payloads cannot restyle output or spoof clickable “Follow it at” links.
> 
> **`_start_evaluation`** now requires a non-empty string **`id`** (protocol error otherwise). A non-string **`url`** triggers a yellow warning, clears the link, and the run still polls for the verdict. **`_fail`** and verdict printing follow the same literal-text pattern.
> 
> Tests cover dynamic area names, markup in names, hostile URL markup, malformed `url`/`id`, empty grades, and literal AMP refusal text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b33f4b4605d4a927007d2fd34102f77bfa0f28e0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -324,12 +324,14 @@ def _print_verdict(finished: dict[str, Any], url: str | None) -> None:
...
     grades = verdict.get("grades") or {}
     parts = [
-        f"{area} {grades[area]}/5"
-        if grades.get(area) is not None
-        else f"{area} not measured"
-        for area in ("goal", "quality", "process", "cost")
+        f"{area} {grade}/5" if grade is not None else f"{area} not measured"
+        for area, grade in grades.items()
     ]
     console.print(f"Goal gate: [{style}]{gate}[/{style}] · " + " · ".join(parts))
```
iris-clawd: Can we add the separator only when `parts` is non-empty? `_well_formed_verdict` accepts `{}`, so an evaluator can currently show users the malformed line `Goal gate: PASSED · `.

lib/cli/tests/test_eval_crew.py
```diff
@@ -95,6 +95,24 @@ def test_the_last_run_is_evaluated_the_url_opened_and_the_verdict_printed(projec
...
+    # evaluation's vocabulary moves ahead of an installed CLI.
+    directory, _ = project
+    record_last_run(directory)
+    graded = done(grades={"goal": 5, "tasks": 3, "agents": 4, "tools": None})
+    install(monkeypatch, FakeAMP(statuses=[httpx.Response(200, json={"id": "ev-1", "status": "running"}), graded]))
+
+    eval_module.eval_crew()
+
+    out = capsys.readouterr().out
+    assert "goal 5/5" in out and "tasks 3/5" in out and "agents 4/5" in out
```
iris-clawd: Can we assert the complete grade segment in order? The current assertions still pass if this path later reorders AMP’s response, leaving the stated ordering behavior unprotected.

iris-clawd (dismissed):
Tests were not run because required CI is red. I could not verify the evaluator rollout order.

The note inline on eval_crew.py is what I'd want addressed before this merges. The other inline notes are small.

lib/cli/src/crewai_cli/eval_crew.py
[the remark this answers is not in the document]
joaomdmoura: Good catch, and it is a bug this PR introduced. Fixed in 189e07cc1.

`_well_formed_verdict` accepts an empty `grades`, and the fixed list of areas was hiding it: there was always something after the separator. Printing what arrives exposed it, so an evaluation that graded nothing produced `Goal gate: PASSED · ` with a dangling separator.

The separator is added only when there is something to separate now, and there is a test for the empty case asserting `"Goal gate: PASSED ·"` does **not** appear.

joaomdmoura: Fixed in 189e07cc1 — same finding as CodeRabbit's on this line, and you are both right that this PR is what exposed it.

`_well_formed_verdict` accepts `{}`, and the fixed list of areas meant there was always something after the separator. The separator is now added only when `parts` is non-empty, with a test asserting that an evaluation which graded nothing prints the gate alone and that `"Goal gate: PASSED ·"` never appears.

On the red CI: the only genuine failure in that run is `tests/telemetry/test_telemetry.py::test_telemetry_fails_due_connect_timeout`, asserting `logger_mock.call_count == export_mock.call_count` and getting `2 == 1` in one of eight parallel shards. It is in a file this PR does not touch, it passes locally on this branch, and every other shard is green. The area this PR does change is 58 tests, all passing.

lib/cli/tests/test_eval_crew.py
joaomdmoura: Fixed in 189e07cc1 — you are right that the assertions did not protect the behaviour the PR claims.

Checking each segment on its own would have passed even if this path sorted the areas or printed its own list. It asserts the whole line in order now:

```
Goal gate: PASSED · goal 5/5 · tasks 3/5 · agents 4/5 · tools not measured
```

plus the negative that `quality` and `process` appear nowhere, so a reintroduced fixed list fails the test rather than passing quietly.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -324,14 +324,22 @@ def _print_verdict(finished: dict[str, Any], url: str | None) -> None:
...
+        for area, grade in grades.items()
     ]
-    console.print(f"Goal gate: [{style}]{gate}[/{style}] · " + " · ".join(parts))
+    # `_well_formed_verdict` accepts an empty `grades`, and a fixed list of
+    # areas used to hide that: there was always something after the separator.
+    # An evaluation that graded nothing prints the gate alone.
+    line = f"Goal gate: [{style}]{gate}[/{style}]"
+    if parts:
+        line += " · " + " · ".join(parts)
+    console.print(line)
```
iris-clawd, line 368: Can we render AMP-provided area names as literal text? Rich parses this string as markup, so a name such as `[red]tasks[/red]` can restyle the verdict. Escaping each area or composing a `Text` object would preserve the output.

lib/cli/tests/test_eval_crew.py
```diff
@@ -95,6 +95,40 @@ def test_the_last_run_is_evaluated_the_url_opened_and_the_verdict_printed(projec
...
 
 
+def test_the_verdict_prints_whatever_areas_the_evaluation_graded(project, monkeypatch, capsys):
+    # The areas are the evaluator's to name. A client printing its own list
+    # would drop the ones it had not heard of and invent "not measured" for
+    # ones that no longer exist — which is what happens the moment the
+    # evaluation's vocabulary moves ahead of an installed CLI.
+    directory, _ = project
+    record_last_run(directory)
+    graded = done(grades={"goal": 5, "tasks": 3, "agents": 4, "tools": None})
```
iris-clawd, line 105: Can we include a bracketed area name and assert that its brackets print literally? That case currently passes through Rich markup parsing instead of preserving AMP’s name.

iris-clawd (dismissed):
I could not verify the separate evaluator rollout order.

The note inline on eval_crew.py is what I'd want addressed before this merges. The other inline notes are small.

lib/cli/src/crewai_cli/eval_crew.py
joaomdmoura: Real bug, and one this PR introduced. Fixed in f0eb87607.

`console = Console()` with markup on, so the verdict was being parsed. Verified against rich 14.3.2: `[red]tasks[/red]` restyles the line, `bad [/bold] tag` raises `MarkupError`, `[notice] x` silently deletes the bracket, and `[link=http://evil]label[/link]` emits a hyperlink whose label and target disagree.

The line is a `Text` now. The gate keeps its colour through `append(style=)` and every value from the payload is appended literally — area names, the gate itself, the error, the URL.

**Reviewing the rest of the file for the same defect found four more, all pre-dating this PR:**

- `Follow it at [cyan underline]{url}[/cyan underline]` — the worst of them, because that line invites a click. A `url` of `[link=http://attacker/]https://app.crewai.com/e/ev-1[/link]` prints a trustworthy label over a hostile target.
- the same URL in the Ctrl-C hint, where an unbalanced tag raises *inside* the interrupt handler and replaces the hint with a traceback.
- the anonymous-read notice, whose address comes from the project's `.env`. That is the one line on that path telling a user their credential is not being sent.
- the AMP-mismatch warning and the execution id, both read from `.crewai/last_run.json`.
- `_fail`, whose message is usually AMP's own sentence.

Same defect, same payload, same file, so they are fixed with it. No caller relied on markup in any of them.

**Beyond this file there is a lot more of it, and it needs its own change rather than this one.** I swept `lib/cli/src/crewai_cli/` for the same pattern. Every `Console` in the package is constructed bare, and there is no escaping convention anywhere in `lib/` — no `markup=False`, no `rich.markup.escape`, no `Text.from_markup`. The widest one is `command.py:49-80`, the shared Plus API error path, which prints the raw HTTP body and is reached by every authenticated command. The one I would fix first, though, is in `crewai-core`: `auth/oauth2.py:123-124` prints the device-authorization `verification_uri` straight from the response, on the line that tells a user where to navigate to log in.

Happy to open that as a separate PR. It should not ride along here.

lib/cli/tests/test_eval_crew.py
joaomdmoura: Added in f0eb87607, and it fails without the fix — I checked, rather than trusting it.

The area test names two areas `[red]tasks[/red]` and `[bold]goal` and asserts both keep their brackets. Without the change the first renders as a restyled `tasks`, so the assertion fails on exactly the behaviour you are asking to protect.

Two more of the same kind, since the sweep of this file turned up worse instances than the area names:

- a `url` of `[link=http://attacker.test/]https://app.crewai.com/e/ev-1[/link]` is printed rather than followed, covering the line that invites a click.
- a refusal message containing `[id]` keeps it, covering `_fail`.

joaomdmoura:
On the red `tests` check, so it does not block review: it is not this PR.

This branch changes **zero files** outside `lib/cli/` — `git diff --stat origin/main...HEAD -- lib/crewai/` is empty — and the failures are in `lib/crewai`:

- the first run failed on `tests/telemetry/test_telemetry.py::test_telemetry_fails_due_connect_timeout` (`assert 2 == 1` on a retry counter, in one of eight parallel shards). It passes locally on this branch.
- this run failed on 25 tests in `lib/crewai/tests/llms/azure/test_azure_responses.py` with `TypeError: 'NoneType' object is not subscriptable`. They fail locally on this branch too, and since the branch touches nothing in that package, they fail on `main` for the same reason.

Two different unrelated failures across two runs, both in packages this PR does not touch.

The area it does change is green: 61 tests in `lib/cli/tests/test_eval_crew.py`, plus `ruff check` and `ruff format --check` over all 940 files.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -63,14 +64,19 @@ def eval_crew(run_id: str | None = None) -> None:
...
     started = _start_evaluation(client, execution_id)
     url = started.get("url")
-    console.print(f"Evaluating run [bold]{execution_id}[/bold]")
+    console.print(Text("Evaluating run ").append(execution_id, style="bold"))
     if url:
-        console.print(f"Follow it at [cyan underline]{url}[/cyan underline]")
+        # Appended, never interpolated: this line invites a click, so a `url`
+        # carrying `[link=…]` would print a trustworthy label over a hostile
+        # target. The style belongs to the span, not to the string.
+        console.print(Text("Follow it at ").append(url, style="cyan underline"))
```
iris-clawd, line 79: Can we reject a non-string `url` as a protocol error before this append? A truthy list or object from AMP currently raises `TypeError`, giving the user a traceback instead of an actionable failure. Please add that malformed-response regression.

iris-clawd (dismissed):
I did not run the tests because required CI has no successful test run. I could not verify the separate evaluator rollout order.

The note inline on eval_crew.py is what I'd want addressed before this merges.

Someone else should sign off on this too: it changes the CLI’s observable output contract from a fixed scorecard to server-defined area names and ordering; scripts or users may depend on the previous output.

lib/cli/src/crewai_cli/eval_crew.py
joaomdmoura: Fixed in b33f4b460, and you are right that this PR caused it. Composing the line means appending the url rather than interpolating it, and `Text.append` takes only a string, so a malformed response went from printing oddly to `TypeError: Only str or Text can be appended to Text`.

I did not make it a protocol error, though, and the reason is that the two fields are not the same kind of thing:

- **the id** is what every later call is made with, so a missing or non-string one has nothing to carry on with. That stays the protocol error it already was, and it now checks the type rather than just truthiness.
- **the url** is only ever shown and opened. A malformed one costs the link and says so in one yellow line, and the run continues — the evaluation is already running on AMP's side, and its verdict is what the user came for. Exiting 1 there would throw away the answer over a cosmetic field.

Two regressions, both failing without the change: a url of `["not", "a", "string"]` prints the notice, still reaches the verdict, and hands nothing to a browser; a non-string id exits with "without an evaluation id". The first reproduces exactly the TypeError you described.

Happy to switch the url to a hard failure if you would rather the protocol be strict there too — it is one line — but I would rather not lose a verdict over a field we only print.

lib/cli/src/crewai_cli/eval_crew.py
```diff
@@ -63,14 +64,17 @@ def eval_crew(run_id: str | None = None) -> None:
...
```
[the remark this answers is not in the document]
joaomdmoura: Fixed in b33f4b460 — same finding as the one below, and correct.

`Text.append` takes only a string, so a truthy non-string `url` raised instead of printing. The id and the url are now handled apart: the id must be a string or it is the protocol error it already was, while a malformed url costs the link, says so, and lets the run finish, since the evaluation is already running and the verdict is the point of the command.

Two regressions cover both, and the url one reproduces the exact `TypeError` without the fix.

alex-clawd (approved):
Approving. This is the right ownership boundary: the evaluator owns the scorecard vocabulary, and the CLI renders the protocol it received. A fixed tuple was not compatibility; it silently dropped new areas and invented blanks for removed ones.

**The dynamic rendering preserves validation rather than weakening it.** `_well_formed_verdict` still requires a dict, a string gate, a grades dict, and every grade to be an exact `int` in 1..5 or `null`; `True` is explicitly rejected despite being an `int` subclass. Only the list of keys stopped being client-owned. Empty grades now print the gate alone without a dangling separator, which is the honest output.

**Switching wire-sourced strings to `rich.text.Text` is the subtle fix that makes dynamic keys safe.** Once area names come from AMP, an area such as `[red]tasks[/red]` would otherwise be interpreted as markup. The same was true for refusal messages and URLs; interpolating a hostile `[link=…]` into the “Follow it at” line could display one trusted-looking label over a different target. Appending the URL as a styled Text span keeps the content literal and the style local.

**Protocol handling improved in the right direction:** a missing or non-string evaluation id is a hard protocol error because every later call depends on it, while a malformed URL only removes the convenience link and polling continues because the evaluation already started. Different severity for different fields, rather than rejecting the whole response uniformly.

I ran the targeted CLI suite locally: **63 passed**. All repository CI is green across Python 3.10–3.13, lint, type checking, CodeQL and pip-audit.

One non-blocking note: `_open(url)` still trusts AMP to provide a browser-safe URL; this PR prevents Rich-link spoofing but does not restrict scheme or host. That is pre-existing and AMP is the trusted source, so not part of this change.

PR #7707, opened 2026-09-22
chore(cli): move `crewai eval` under experimental

joaomdmoura:
Nothing about the command changes. It is still `crewai eval`, still listed in `--help`, still takes `--run`, still prints the same lines. Only the module moved:

| before | after |
|---|---|
| `crewai_cli/eval_crew.py` | `crewai_cli/experimental/eval_crew.py` |
| `tests/test_eval_crew.py` | `tests/experimental/test_eval_crew.py` |

The package says what the command's location could not: its output and its options may change between releases without the deprecation cycle a settled command gets.

That is honest about where it stands. Most of what `crewai eval` depends on lives outside this repository — an AMP endpoint and the evaluator behind it — and both are new enough that the report's shape is still moving. The command already absorbs some of that on its own, since crewAIInc/crewAI#7701 made it print whatever the evaluation graded rather than a list of its own, but the promise it makes to a user should match the promise we can keep.

`crewai/experimental/` plays the same role on the library side, so this follows the convention the repository already has rather than inventing one. The CLI had no experimental package before; it does now, and `crewai eval` is its first tenant.

**Not moved, deliberately:** `crewai/telemetry/tracing/last_run.py` (from crewAIInc/crewAI#7648). It is written by the tracing pipeline on every traced run, not by this command — `grants.py` calls it. Moving it would leave settled tracing code importing an experimental module, which is backwards.

63 tests in the moved file pass, `crewai --help` still lists the command, `ruff check` and `ruff format --check` are clean over all 941 files.

---

> [!NOTE]
> **Low Risk**
> Refactor-only module relocation with unchanged user-facing CLI wiring; no auth, tracing, or eval logic changes in the shown diff.
> 
> **Overview**
> Introduces **`crewai_cli/experimental/`** as the home for CLI commands whose behavior, output, and flags may change without a deprecation cycle. **`crewai eval`** is the first command there: implementation and tests move from top-level `eval_crew` into `experimental/eval_crew` and `tests/experimental/`, while **`cli.py`** still lazy-loads the same entrypoint so users keep **`crewai eval`**, **`--help`**, and **`--run`** unchanged.
> 
> The new package docstring states that experimental commands remain fully listed and invocable like settled commands; only the stability promise differs, aligned with the main library’s `crewai/experimental/` convention and the AMP-backed evaluator’s still-evolving report shape.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a97385b521f97f3a7961433169ad07f1ef333038. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


alex-clawd (approved):
Approving. This is genuinely a relocation, not a behavioral refactor.

`git diff --name-status -M` confirms the implementation is a **100% rename** from `crewai_cli/eval_crew.py` to `crewai_cli/experimental/eval_crew.py`. The test file is a 99% rename; its only substantive change is importing the implementation from the new package. `cli.py` lazy-loads the same `eval_crew` function, so the user-facing command, `--run`, help text, and import cost remain unchanged.

**No stale source imports remain.** The only grep hit for `crewai_cli.eval_crew` is a compiled `__pycache__` artifact in my checkout, not tracked code.

**The boundary is explained instead of merely encoded.** The package docstring says what experimental means here: fully listed and invocable, but behavior, output, and options may move without the settled command deprecation cycle. That is a more useful contract than hiding the command or adding an “experimental” prefix users have to type.

**Keeping `last_run.py` out of the experimental package is correct.** It belongs to the tracing pipeline and is written on every traced run through `grants.py`; moving it just because `crewai eval` consumes it would make settled tracing code depend on an experimental namespace, reversing the dependency.

I ran the moved test file locally: **63 passed**. Repository CI is green across Python 3.10–3.13, lint, formatting, typing, CodeQL, and pip-audit.

One release note worth preserving: moving a Python module is a breaking import for anyone who imported `crewai_cli.eval_crew` directly, even though it was not a public command contract. The PR intentionally does not leave a compatibility shim because the point is to label the surface experimental; that is reasonable, but it should be treated as internal-import breakage rather than literally “nothing changes.”

